PR #8, opened 2023-02-24
Fix translation of atomic_add

thedataking:
The initial transpile used a version of dav1d which did not handle C11 atomics correctly. This change fixes translations of atomic_add.


randomPoison (approved):

kkysen:
> The initial transpile used a version of dav1d which did not handle C11 atomics correctly.

Do you mean `c2rust`, not `dav1d`?

PR #15, opened 2023-03-08
Setup basic CI

### LINKED ISSUE #4: Setup CI
Setup basic CI for the project such that we can verify that any changes put in a PR still build and pass tests (if we have tests setup). We probably just want to use GitHub Actions for this.

thedataking:
Closes #4.

Notes:
- covers x86_64 on ubuntu-latest. 
- caches rust toolchain and crates
- all tests are run (not just those that exercise the transpiled rust). I ran into some problem running the transpiled subset; not sure this is worth chasing down but it does increase the time to complete the test.
- uploads the test log as build artefact.


randomPoison (approved):
🚀

Issue #23, opened 2023-03-17
Clean up control flow where c2rust emitted state machines


thedataking:
Grep for `match current_block_` to find instances of this pattern. One can look at the original C code to determine the ideal mapping into structured control flow in Rust. There seem to be 19 instances across 8 files.

randomPoison:
The following files/functions still have state machine control flow that needs to be cleaned up:

* [ ] `cdef_apply.rs` - `rav1d_cdef_brow`
* [ ] `recon.rs` - `decode_coefs`
* [ ] `thread_task.rs` - `reset_task_cur`
* [ ] `thread_task.rs` - `insert_tasks`
* [ ] `thread_task.rs` - `check_tile`

The following ALSO have state machine control flow, but are lower priority:

* [ ] `seek_stress.rs` - `main_0`
* [ ] `ivf.rs` - `ivf_seek`
* [ ] `y4m2.rs` - `y4m2_write`
* [ ] `yuv.rs` - `yuv_write`

kkysen:
The latter ones are not in the `rav1d` library.  They have pretty much zero priority.

PR #24, opened 2023-03-18
Support 32-bit builds

thedataking:


src/cdef_apply_tmpl_16.rs
```diff
@@ -2179,20 +2177,20 @@ pub unsafe extern "C" fn dav1d_cdef_brow_16bpc(
         ptrs[0 as libc::c_int
             as usize] = (ptrs[0 as libc::c_int as usize])
             .offset(
-                (8 as libc::c_int as libc::c_long
+                (8
```
ahomescu, line 2180: Re-run `cargo fmt`?

tools/input/annexb.rs
```diff
@@ -373,6 +357,8 @@ unsafe extern "C" fn annexb_read(
         dav1d_data_unref(data);
         return -(1 as libc::c_int);
     }
+    // println!("ptr[0]: {:#04x}, ptr[1]: {:#04x}, ptr[2]: {:#04x}, len: {:?}",
```
ahomescu, line 360: Remove these?

src/recon_tmpl_16.rs
```diff
@@ -3878,6 +3560,8 @@ unsafe extern "C" fn decode_coefs(
         }
     } else {
         dc_sign_ctx = get_dc_sign_ctx(tx as libc::c_int, a, l) as libc::c_int;
+        println!("xcheck:tx:{}:dc_sign_ctx:{}:*a:{}:*l:{}",
```
ahomescu, line 3563: Remove this?

src/recon_tmpl_8.rs
```diff
@@ -3833,6 +3827,8 @@ unsafe extern "C" fn decode_coefs(
         }
     } else {
         dc_sign_ctx = get_dc_sign_ctx(tx as libc::c_int, a, l) as libc::c_int;
+        println!("xcheck:tx:{}:dc_sign_ctx:{}:*a:{}:*l:{}",
```
ahomescu, line 3830: Remove?

tools/input/annexb.rs
```diff
@@ -160,6 +145,8 @@ unsafe extern "C" fn parse_obu_header(
     if *buf as libc::c_int & 0x80 as libc::c_int != 0 {
         return -(1 as libc::c_int);
     }
+    // let buf_slice = std::slice::from_raw_parts(buf, buf_size as size_t);
```
ahomescu, line 148: Remove?

tools/input/annexb.rs
thedataking: Derp, fixed.

src/cdef_tmpl_16.rs
```diff
@@ -7,7 +7,7 @@ extern "C" {
     static dav1d_cdef_directions: [[int8_t; 2]; 12];
 }
 
-#[cfg(feature = "asm")]
+// #[cfg(feature = "asm")]
```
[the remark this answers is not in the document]
thedataking: I should have been clearer. I made this commit quickly as a workaround; don't have a strong opinon on what the best resolution is. That said, it might be worthwhile to stay close to the C code for the moment. AFAICT, the C code unconditionally [defines](https://github.com/memorysafety/rav1d/blob/d35c4ed2edd48026508cd77cf35f674ae75c44ac/src/cpu.h) and [uses](https://github.com/memorysafety/rav1d/blob/d35c4ed2edd48026508cd77cf35f674ae75c44ac/src/cpu.h#L54) these variables. .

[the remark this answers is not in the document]
thedataking: I think I was able to do what you suggest with some more compile-time guards.

randomPoison (approved):

include/stdint.rs
kkysen, line 1: Should we also fix the other integer types now?

thedataking: Not sure what you mean. If you mean, get rid of `__int64_t` and similar types in `include/stdint.rs` that doesn't do much since those type aliases aren't used and the `include/*rs` files should eventually become dead code. If you mean change all the libc integer types to Rust types, I think we want to do that towards the end of the project. 

IMHO, we'd be better off getting rid of duplicate functions from header files, cleaning up the control flow (#23), or getting rid of templates (#25) if you don't want to do the previous items twice. Not sure what Nicole thinks is best.

PR #28, opened 2023-03-22
Setup aarch64 asm

randomPoison:
* Update `build.rs` to build and link assembly for aarch64.
* Add asm calls to `msac.rs`.


thedataking:
You probably also need to update `dav1d_init_cpu` which has an aarch64/arm specific code path.

randomPoison:
I was gonna do the `david_init_cpu` stuff for arm in a separate PR. @thedataking let me know if you think it makes more sense to do as part of this one.

PR #29, opened 2023-03-23
Deduplicate fixed-width (`stdint.h`) and standard (`stddef.h`) integer types

kkysen:
This solely deduplicates the typedefs of the cross-platform int types, but doesn't actually change any of them.  Thus, this should be as compatible as the current code with future changes for portability (e.x. #24).


thedataking:
Looks good to me. These type aliases should eventually be replaced with Rust types but this looks good for now.

Tiny title nitpick: "Deduplicate standard type definitions and fixed width integer types" would a more accurate summary of your change. stddef.h contains the standard type definitions and stdint.h defines the C99 fixed width integer types.

kkysen:
I'm planning on actually moving these new modules to `include/` rather than `src/`, so it'd be `crate::include::{stddef, stdint}`.  This matches the deduplication I'm doing of the headers in `include/`, which also don't currently have Rust modules as they're headers, but which I'm creating as I deduplicate the types in those headers.

randomPoison:
Two requests:

* Could you change the imports to use globs to cut down on verbosity? i.e. one `use crate::src::stdint::*;` instead of having a separate line for each type?
* Could you move the imports to the top of the file, where `use` statements would normally go? For some of the files I've already started putting the `use` statements there.

thedataking:
> I'm planning on actually moving these new modules to `include/` rather than `src/`

Keeping a correspondence between the C and Rust code sounds useful. Keep in mind that we'll want to remove all traces of C headers at some point during the migration.

randomPoison (approved):

Issue #26, opened 2023-03-22
Deduplicate code


kkysen:
The transpiled code contains a ton of duplicated code, which can be split into a few categories:

* [x] Deduplicate `rav1d` function prototype declarations (`extern` imports) into proper `use` imports.
* [x] Deduplicate non-`rav1d` (e.x. `libc`, etc.) function prototype declarations (`extern` imports) into a single `extern` block (perhaps one per shared library).
* [x] Deduplicate `rav1d` types from headers.
* [x] Deduplicate non-`rav1d` types from headers into a single, new module.  These seem to mostly be std types.
* [x] Deduplicate `rav1d` constants from headers.
* [x] Deduplicate non-`rav1d` constants from headers.
* [x] #9: Deduplicate inline functions from headers.
* [ ] Deduplicate expanded macros (if possible, rewrite them as functions in C first).
* [ ] #25: Deduplicate templates files.

kkysen:
Does this make sense for duplicated code, and are there any categories I'm missing?  If they look good, I'll turn them each into separate issues and start working on each of them.

Also, what order should I work on them?  Does it matter?  I'm inclined to do them in roughly the order listed above, but not for any particular reason I think.  The exception is newtypes, which I think need to be done before functions, as otherwise those will be considered different types.  Another consideration is if types and function definitions are deduplicated before function prototype declarations, should their imports be turned into more `extern` imports or `use` imports?

randomPoison:
This is effectively done except for #25, which is in review. I'm closing this one now that the remaining work is tracked elsewhere.

PR #30, opened 2023-03-25
Initial portability fixes

thedataking:
The following changes are necessary but not sufficient to run on platforms other than Linux. I have a few more fixes in the pipeline but I'd like to land these first to not unduly interfere with #26.

### DIFF
--- a/c2rust-lib.rs
+++ b/c2rust-lib.rs
@@ -112,3 +112,20 @@ pub mod y4m2;
 pub mod yuv;
 } // mod output
 } // mod tools
+
+// NOTE: temporary code to support Linux and macOS, should be removed eventually
+cfg_if::cfg_if! {
+    if #[cfg(target_os = "linux")] {
+        extern "C" {
+            pub static mut stdout: *mut libc::FILE;
+            pub static mut stderr: *mut libc::FILE;
+        }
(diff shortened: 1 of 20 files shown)


src/picture.rs
```diff
@@ -1,13 +1,12 @@
 use ::libc;
 use crate::src::cdf::CdfContext;
 use crate::src::msac::MsacContext;
+use crate::{stderr,errno_location};
```
kkysen, line 4: In some of the `use`s, there are spaces missing after the commas.  It would be fixed once we do a `cargo fmt` later on once we're reading to do that, though.

c2rust-lib.rs
```diff
@@ -112,3 +112,28 @@ pub mod y4m2;
...
+            pub static mut stdout: *mut libc::FILE;
+            #[link_name = "__stderrp"]
+            pub static mut stderr: *mut libc::FILE;
+        }
+
+        unsafe fn errno_location() -> *mut libc::c_int {
+            libc::__error()
+        }
+    }
+}
```
kkysen, line 139: This could go in its own module I think.

kkysen (approved):
LGTM in general.  I don't think it will cause any conflicts with the deduplication work I'm doing, as I haven't touched these parts yet.

PR #31, opened 2023-03-27
Get CPU flags for arm targets

randomPoison:
Translate the logic for [`dav1d_get_cpu_flags_arm`](https://github.com/memorysafety/rav1d/blob/39968db983ce4211e48ca173e8d5d189b3b666c6/src/arm/cpu.c#L83). I tried to replicate the original logic using Rust's conditional compilation mechanisms, but the translations isn't one-to-one.

### DIFF
--- a/src/arm/cpu.rs
+++ b/src/arm/cpu.rs
@@ -21,7 +21,7 @@ pub unsafe extern "C" fn dav1d_get_cpu_flags_arm() -> libc::c_uint {
             todo!("Android is not yet supported")
         } else {
             if (libc::getauxval(libc::AT_HWCAP) & NEON_HWCAP) != 0 {
-                flags |= DAV1D_ARM_CPU_FLAG_NEON
+                flags |= DAV1D_ARM_CPU_FLAG_NEON;
             }
         }
     }
--- a/src/arm/cpu.rs
+++ b/src/arm/cpu.rs
@@ -17,6 +17,8 @@ pub unsafe extern "C" fn dav1d_get_cpu_flags_arm() -> libc::c_uint {
             target_os = "macos"
         ))] {
             flags |= DAV1D_ARM_CPU_FLAG_NEON;
+        } else if #[cfg(target_os = "android")] {
+            todo!("Android is not yet supported")
         } else {
             if (libc::getauxval(libc::AT_HWCAP) & NEON_HWCAP) != 0 {
                 flags |= DAV1D_ARM_CPU_FLAG_NEON
--- a/src/arm/cpu.rs
+++ b/src/arm/cpu.rs
@@ -1,0 +1,28 @@
+use cfg_if::cfg_if;
+use libc;
+
+pub const DAV1D_ARM_CPU_FLAG_NEON: libc::c_uint = 1 << 0;
+pub const NEON_HWCAP: libc::c_ulong = 1 << 12;
+
+#[no_mangle]
+#[cold]
+pub unsafe extern "C" fn dav1d_get_cpu_flags_arm() -> libc::c_uint {
+    let mut flags = 0;
+
+    // TODO: Support Android by parsing `/proc/cpuinfo` the way the original C does.
+    cfg_if! {
+        if #[cfg(any(
+            target_arch = "aarch64",
+            target_os = "windows",
+            target_os = "macos"
+        ))] {
+            flags |= DAV1D_ARM_CPU_FLAG_NEON;
+        } else {
+            if (libc::getauxval(libc::AT_HWCAP) & NEON_HWCAP) != 0 {
+                flags |= DAV1D_ARM_CPU_FLAG_NEON
+            }
+        }
+    }
+
+    flags
+}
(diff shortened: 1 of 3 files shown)


src/arm/cpu.rs
```diff
@@ -0,0 +1,28 @@
...
+
+pub const DAV1D_ARM_CPU_FLAG_NEON: libc::c_uint = 1 << 0;
+pub const NEON_HWCAP: libc::c_ulong = 1 << 12;
+
+#[no_mangle]
+#[cold]
+pub unsafe extern "C" fn dav1d_get_cpu_flags_arm() -> libc::c_uint {
+    let mut flags = 0;
+
+    // TODO: Support Android by parsing `/proc/cpuinfo` the way the original C does.
```
thedataking, line 12: Couldn't you insert the android case and have it panic? That way, we're sure this cannot silently fail on Android.

randomPoison: Good call, done.

src/arm/cpu.rs
```diff
@@ -0,0 +1,30 @@
...
+pub const NEON_HWCAP: libc::c_ulong = 1 << 12;
+
+#[no_mangle]
+#[cold]
+pub unsafe extern "C" fn dav1d_get_cpu_flags_arm() -> libc::c_uint {
+    let mut flags = 0;
+
+    cfg_if! {
+        if #[cfg(any(
+            target_arch = "aarch64",
```
ahomescu, line 14: This is `aarch64 || windows || macos`. Do you want something like `aarch64 && (windows || macos)`?

src/arm/cpu.rs
```diff
@@ -0,0 +1,30 @@
...
+            target_os = "windows",
+            target_os = "macos"
+        ))] {
+            flags |= DAV1D_ARM_CPU_FLAG_NEON;
+        } else if #[cfg(target_os = "android")] {
+            // TODO: Support Android by parsing `/proc/cpuinfo` the way the original C does.
+            todo!("Android is not yet supported")
+        } else {
+            if (libc::getauxval(libc::AT_HWCAP) & NEON_HWCAP) != 0 {
+                flags |= DAV1D_ARM_CPU_FLAG_NEON
```
ahomescu, line 24:
```suggestion
                flags |= DAV1D_ARM_CPU_FLAG_NEON;
```

src/arm/cpu.rs
```diff
@@ -0,0 +1,30 @@
+use cfg_if::cfg_if;
+use libc;
+
+pub const DAV1D_ARM_CPU_FLAG_NEON: libc::c_uint = 1 << 0;
```
ahomescu, line 4: Where is this used? Do those places import it from here?

src/arm/cpu.rs
```diff
@@ -0,0 +1,30 @@
...
+        if #[cfg(any(
+            target_arch = "aarch64",
+            target_os = "windows",
+            target_os = "macos"
+        ))] {
+            flags |= DAV1D_ARM_CPU_FLAG_NEON;
+        } else if #[cfg(target_os = "android")] {
+            // TODO: Support Android by parsing `/proc/cpuinfo` the way the original C does.
+            todo!("Android is not yet supported")
+        } else {
```
thedataking, line 22: nit: In the C version, 
- this path is guarded on `ARCH_ARM` which I think means 32-bit arm (so `target_arch  = "arm"`), and
- it comes before the android path

Would it be better to guard it on 32-bit arm like the original code and put the test before android? I'm not 100% convinced this even matters, so you decide what makes sense here.

thedataking (approved):

src/arm/cpu.rs
randomPoison: The original C was

```c
#if defined(__ARM_NEON) || defined(__APPLE__) || defined(_WIN32) || ARCH_AARCH64
```

So that's what I was trying to replicate 🤷

src/arm/cpu.rs
randomPoison: This is used right below here in `dav1d_get_cpu_flags_arm`. It's also then checked in some other files, though that code hasn't been added in yet.

src/arm/cpu.rs
thedataking: The caller of this function already ensures that the arch is `arm` or `aarch64` which why the condition in C works. The question, as I understand it, is whether we want something close to the logic of the original code (caller ensures we're on ARM) or something that is more robust (e.g. the logic that Andrei suggested).

src/arm/cpu.rs
randomPoison: Nah, you're right. I misread the original C so what you're suggesting is more correct. Fixed.

src/arm/cpu.rs
ahomescu: Sounds good, just wanted to make sure those other places use this definition too.

src/arm/cpu.rs
ahomescu: Oh I didn't realize in my original comment that this also applied to 32-bit ARM. The conditional makes more sense now: 64-bit or Windows or Apple ARM.

ahomescu (approved):

PR #32, opened 2023-03-27
Add asm calls for refmvs

randomPoison:
Add asm logic corresponding to `refmvs.h` in the original C project.


thedataking:
LGTM (did a quick skim). 

Looks like there are some superfluous casts in there (e.g. `as libc::c_int as libc::c_uint`). IDK if you left those so we can do a sweep later of if it was accidental. No big deal.

randomPoison:
@thedataking Extra casts have been cleaned up.

PR #34, opened 2023-03-29
Refactor out the `stdatomic.h` types `memory_order`, `atomic_{u,}int` to `include/stdatomic.rs`

kkysen:


randomPoison (approved):
As with the other PR, I'd prefer if you use glob imports

kkysen:
> As with the other PR, I'd prefer if you use glob imports

I thought about that, but they don't really work that well here since
* a glob import would have to import the whole module, which contains various other things, since C `enum`s turn into just Rust `const`s, not `enum`s with their own scope.  `stdint` and `stddef` were more special in this sense since they correspond with Rust types that are always in scope.
* most of these imports aren't actually used at all (or won't be after I deduplicate more things that use them), and those can be auto-fixed by `rustc` en-masse (though doing so also fixes other things, which we should probably just go ahead and fix?).  Using a glob import would prevent doing this, as a module glob import might use one item, meaning the entire glob import can't be elided automatically.
* I can consider it for future PRs, but rebasing glob imports into all of these is probably more work than the deduplication itself (it was for `std{def,int}`).

PR #48, opened 2023-03-30
Fix some of the trivial warnings

kkysen:
This fixes some of the trivial warnings, namely
* `#[warn(stable_features)]`
* `#[warn(unused_unsafe)]`

Reducing the number of warnings (especially those not in `c2rust_out` like the `unused_unsafe`s) makes other work, especially deduplication with its many, but small and rapid changes, much smoother.  And I don't think these fixes conflict with any of the other work we're doing, as they're quite simple.

### DIFF
--- a/c2rust-lib.rs
+++ b/c2rust-lib.rs
@@ -5,7 +5,6 @@
 #![allow(non_upper_case_globals)]
 #![allow(unused_assignments)]
 #![allow(unused_mut)]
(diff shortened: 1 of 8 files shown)


randomPoison:
Ah, I hadn't seen that you put this up. I ended up cleaning up all of these in #56, so this PR isn't needed anymore I think.

PR #59, opened 2023-04-01
Deduplicate `struct BlockContext` to `src/env.rs`

kkysen:


kkysen:
I noticed that the C source for `struct BlockContext` has aligned fields: https://github.com/memorysafety/rav1d/blob/01ba6ae6556875e44a1be92189306661904d9782/src/env.h#L39-L57

But we haven't added the alignment to the Rust definition.  We need to add them like has been done for other aligned fields, right?

randomPoison:
Correct, we'll have to update the struct to wrap the fields in the appropriate `AlignX` types. That can be done in a separate PR, though, and will be easier to do now that the struct has been deduplicated 😁

randomPoison (approved):

PR #73, opened 2023-04-04
Deduplicate previously unnamed `union Dav1dTaskContext_scratch_interintra_edge` and its 2 unnamed `struct` fields

kkysen:
Now that type is deduplicated, it can be easily `#[repr(align(64))]`ed like in the C source: https://github.com/memorysafety/rav1d/blob/c2fb626189f313dd3ed8f1b74dbcfd80f333a43c/src/internal.h#L433-L442

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -65,3 +65,10 @@ pub union Dav1dTaskContext_cf {
     pub cf_8bpc: [int16_t; 1024],
     pub cf_16bpc: [int32_t; 1024],
 }
+
+#[derive(Copy, Clone)]
+#[repr(C)]
+pub struct Dav1dTaskContext_scratch_interintra_edge_8 {
+    pub interintra_8bpc: [uint8_t; 4096],
+    pub edge_8bpc: [uint8_t; 257],
+}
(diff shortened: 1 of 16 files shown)


src/internal.rs
```diff
@@ -65,3 +66,24 @@ pub union Dav1dTaskContext_cf {
...
+#[derive(Copy, Clone)]
+#[repr(C)]
+pub struct Dav1dTaskContext_scratch_interintra_edge_16 {
+    pub interintra_16bpc: [uint16_t; 4096],
+    pub edge_16bpc: [uint16_t; 257],
+}
+
+#[derive(Copy, Clone)]
+#[repr(C)]
+pub union Dav1dTaskContext_scratch_interintra_edge {
```
randomPoison, line 86: Minor requestion: Could you instead name this struct `Dav1dTaskContext_scratch`? I think the name currently is a bit unnecessarily verbose. Likewise the other two could be shortened to `Dav1dTaskContext_scratch_8bpc` and `Dav1dTaskContext_scratch_16bpc`.

randomPoison (changes requested):

src/internal.rs
randomPoison: Actually nevermind, I see now that this struct more deeply nested than I thought (originally I thought this struct was the `scratch` field in `Dav1dTaskContext`, but looks like it's actually one of the anonymous inner structs).

randomPoison (approved):

src/internal.rs
kkysen: Yeah, I was planning on naming the outer one `Dav1dTaskContext_scratch`, as this one's more deeply nested.

PR #84, opened 2023-04-08
Re-transpilation after naming the unnamed types that don't require other code changes (i.e. naming fields)

kkysen:
Unnamed types, mostly anonymous inner types, that do have named fields/variables can be named without any other change.  That was done in the C source code, then the C re-transpiled to Rust from a branch closer to the initial transpilation, and then the commit `cherry-pick`ed to `main`.  Unnamed types where the field is not named cannot be named without also naming the field, so I'm keeping those separate, though those changes are coming next.

Note: the re-transpilation was `cherry-pick`ed from [`kkysen/retranspile-name-unnamed-simple`](https://github.com/memorysafety/rav1d/tree/kkysen/retranspile-name-unnamed-simple) and merged, as the re-transpilation diff is simpler there.

The type names I chose are generally based on the field name with the parent type as a prefix, `enum` variants' names, etc.

### DIFF
--- a/examples/dav1dplay.c
+++ b/examples/dav1dplay.c
@@ -146,7 +146,7 @@ static void dp_rd_ctx_parse_args(Dav1dPlayRenderContext *rd_ctx,
     // Short options
     static const char short_opts[] = "i:vuzgr:";
 
-    enum {
+    enum arg {
         ARG_THREADS = 256,
         ARG_FRAME_DELAY,
         ARG_HIGH_QUALITY,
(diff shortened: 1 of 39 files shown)


kkysen:
A few things got messed up in the merge that I missed, some changes from #72 and #73.  I'll fix them.

randomPoison:
Can you include the C source changes here? It would be good to commit those as well in case we need to do more transpiles in the future.

kkysen:
> Can you include the C source changes here? It would be good to commit those as well in case we need to do more transpiles in the future.

They're added now.  I wasn't sure if we wanted to change the C sources at all, but I agree it's a good idea, especially since these changes are purely naming types and have no other changes.

PR #117, opened 2023-04-12
Deduplicate inline `fn dav1d_get_cpu_flags` to `src/cpu.rs` and `enum CpuFlags` to `src/x86/cpu.rs`

kkysen:


kkysen:
I'm not sure why it can't find `src::cpu::dav1d_get_cpu_flags` on `--target armv7-unknown-linux-gnueabihf`, as `cargo check --target armv7-unknown-linux-gnueabihf` works fine locally.  Is there an easy way to reproduce what CI runs locally?  Having everything in `rust-toolchain.toml` and `.cargo/config.toml` so it can run anywhere would be nice.

kkysen:
Ahh, asm is disabled.  I think I figured it.

thedataking:
> @thedataking, is there any reason for keeping the arch-specific `enum CpuFlags`,

Yes, the flags are all x86 specific so it makes sense to keep them where they are. A second reason is that we don't want to diverge from the structure of the C code (at least at this stage) since it would complicate debugging and troubleshooting. A third reason is that we've asked you to keep your changes focused on one thing. Finally, on a more general point, you should bias leaving stuff alone unless you have a good working understanding of the code you're changing and there is a need to change it; this reduces the risk that stuff gets messed up inadvertently. For theses reasons, I would strongly recommend you revert bc2c2883835af77033f695ffe91b0d3674d1a82e.

src/cdef_tmpl_16.rs
```diff
@@ -248,13 +248,14 @@ pub struct Dav1dCdefDSPContext {
...
     pub fb: [cdef_fn; 3],
 }
-pub const DAV1D_X86_CPU_FLAG_AVX512ICL: CpuFlags = 16;
-pub const DAV1D_X86_CPU_FLAG_SSE2: CpuFlags = 1;
-pub const DAV1D_X86_CPU_FLAG_AVX2: CpuFlags = 8;
-pub const DAV1D_X86_CPU_FLAG_SSE41: CpuFlags = 4;
-pub const DAV1D_X86_CPU_FLAG_SSSE3: CpuFlags = 2;
-pub type CpuFlags = libc::c_uint;
-pub const DAV1D_X86_CPU_FLAG_SLOW_GATHER: CpuFlags = 32;
+cfg_if! {
```
thedataking, line 251: I would prefer if you could delete all of this and instead add `use crate::src::x86::cpu::*;` at the beginning of `cdef_dsp_init_x86`. That would not only be less code, it would also ensure that the `cfg_if` parts match. Right now, you define these constants if we're on x86 or x86-64 but `cdef_dsp_init_x86` is only defined under a more narrow set of circumstances.  Same fix should be applied in the corresponding 8-bit file.

src/loopfilter_tmpl_16.rs
```diff
@@ -180,14 +180,13 @@ extern "C" {
...
 
-pub const DAV1D_X86_CPU_FLAG_AVX512ICL: CpuFlags = 16;
-pub const DAV1D_X86_CPU_FLAG_SSE2: CpuFlags = 1;
-pub const DAV1D_X86_CPU_FLAG_AVX2: CpuFlags = 8;
-pub const DAV1D_X86_CPU_FLAG_SSSE3: CpuFlags = 2;
-pub type CpuFlags = libc::c_uint;
-pub const DAV1D_X86_CPU_FLAG_SLOW_GATHER: CpuFlags = 32;
-pub const DAV1D_X86_CPU_FLAG_SSE41: CpuFlags = 4;
-
+cfg_if! {
```
thedataking, line 183: Couldn't these definitions go in the block starting on line 15? If so, I think they should. Same comment for the other template file.

thedataking (changes requested):
For the files where these constants are just used in a single function, please star-include the constants direction in those functions (less code, more robust). 

In other cases such as loopfilter_tmpl_*.rs you can put the constants in existing conditionally compiled blocks.

src/loopfilter_tmpl_16.rs
kkysen: I don't think normal `use`s can go in an `extern "C"` block, but I moved them into the `*_init_x86` function like for all the other uses.

src/cdef_tmpl_16.rs
kkysen: I moved all of the `cfg_if!`ed `use`s to inside the `*_init_x86` functions now, for all of them in all of the files.  I didn't realize they were only used in these functions, so I agree it's better to put them inside so the `cfg`s match exactly.  I didn't use glob imports, though, as since this isn't a real Rust `enum`, we'd have to import the whole `mod src::x86::cpu`, which contains various other items, too.

thedataking (approved):
LGTM! This is great, thanks, now I can work on the deduplicated function to fix SIMD tests.

Issue #74, opened 2023-04-04
Test AVX512 instructions in CI


randomPoison:
We've run into a case where tests fail locally but are not failing in CI (see #71). The issue appears to be specific to AVX512 assembly code, such that tests will fail when run on a machine with AVX512 support, but will pass on a machine without it. This likely means that the default GitHub Actions runners that we're using don't have AVX512 support. We should find a way to test against AVX512 instructions in CI to catch similar issues in the future. If possible, we should continue to run tests against a machine without AVX512 to ensure that the code works correctly with different levels of SIMD support.

thedataking:
Note from discussion on #214 : if we self-host on AVX512-capable machine, we should also try to run Argon on each PR.

thedataking:
We now build and run a single input test on a self-hosted AVX512 capable machine. We could do more so leaving this issue open but feel free to close if others feel this is sufficient.

PR #120, opened 2023-04-13
Fix alignment for dav1d_sgr_x_by_x

thedataking:
With this change, all 790 enabled tests pass on a machine with avx512. I will prioritize getting #74 done so CI will catch such issues going forward.


fbossen (approved):
LGTM. I confirm this resolves seg fault issues.

randomPoison (approved):

PR #123, opened 2023-04-13
decode_b: Convert raw pointer to reference

randomPoison:
Convert the `t` argument to `decode_b` from `*mut` to a `&mut`, and simplify all references to that variable. I also removed the `extern "C"` from the function declaration since it is only called from within `decode.rs`.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4485,15 +4485,16 @@ unsafe fn DEBUG_BLOCK_INFO(
         && (*t).bx >= 8 && (*t).bx < 12
     // true
 }
-unsafe extern "C" fn decode_b(
-    t: *mut Dav1dTaskContext,
+
+unsafe fn decode_b(
+    t: &mut Dav1dTaskContext,
     bl: BlockLevel,
     bs: BlockSize,
     bp: BlockPartition,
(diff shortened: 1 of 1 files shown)


kkysen (commented):
This looks good I think, but if we want to retranspile again to get rid of some of the macros, which I was thinking of doing, these little but numerous changes will make the diff much worse.  Maybe we should do this afterward, as it seems like an easy change to tack on afterward.

thedataking (approved):
I skimmed the first ~25% of the change and couldn't spot any problems.

src/decode.rs
```diff
@@ -14735,7 +14736,7 @@ unsafe extern "C" fn decode_sb(
         match bp as libc::c_uint {
             0 => {
                 if decode_b(
-                    t,
+                    &mut *t,
```
kkysen, line 14739: These need a reborrow?

kkysen: Oh, these are calls from another function, nevermind.

kkysen (approved):
If there are no other retranspilaton use cases, which it seems like there aren't, then this LGTM.

PR #124, opened 2023-04-13
decode_b: Cleanup initial variable declarations

randomPoison:
Cleanup the initial variable declarations in `decode_b`, mostly removing redundant type declarations/casts. I also derived `Default` on a few structs to make it easier to initialize.

### DIFF
--- a/src/levels.rs
+++ b/src/levels.rs
@@ -201,7 +201,7 @@ pub const MM_WARP: MotionMode = 2;
 pub const MM_OBMC: MotionMode = 1;
 pub const MM_TRANSLATION: MotionMode = 0;
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Default)]
 #[repr(C)]
 pub struct Av1Block_intra {
     pub y_mode: uint8_t,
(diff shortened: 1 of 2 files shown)


kkysen (commented):
Deriving `Default` is a good idea; we can do that for all `#[repr(C)]` types and eliminate all the zero initialization that way.

src/levels.rs
```diff
@@ -259,7 +259,15 @@ pub union Av1Block_intra_inter {
     pub c2rust_unnamed_0: Av1Block_inter,
 }
 
-#[derive(Copy, Clone)]
+impl Default for Av1Block_intra_inter {
```
kkysen, line 262: Can we derive it in all cases instead of any manual impls?

randomPoison: We can't derive `Default` for unions as far as I can tell, so at least here we need a smol manual impl.

kkysen: Ahh, that's right.  How do we know which variant should be initialized for `Default` then?  How does `c2rust` know?

PR #126, opened 2023-04-13
decode_b: Cleanup instances of case_set

randomPoison:
Translate `case_set` and it's related macros to Rust functions and cleanup instances where `case_set` was expanded within `decode_b`. These aren't all of the instances of `case_set` within `decode_b`, but it's enough of a start to confirm that the implementatino of `case_set` is correct, at least for how it's used within `decode_b`.


src/ctx.rs
```diff
@@ -2,30 +2,100 @@ use crate::include::stdint::uint16_t;
...
+
+#[inline]
+pub unsafe fn set_ctx_rep4(buf: *mut u8, off: isize, val: u64) {
+    set_ctx_rep1::<alias64>(buf, off + 0, val);
+    set_ctx_rep1::<alias64>(buf, off + 8, val);
+    set_ctx_rep1::<alias64>(buf, off + 16, val);
+    set_ctx_rep1::<alias64>(buf, off + 24, val);
+}
+
+pub type SetCtxFn = unsafe fn(*mut u8, isize, u64);
```
kkysen, line 85: Will this function pointer always get inlined and de-virtualized?

src/ctx.rs
```diff
@@ -2,30 +2,100 @@ use crate::include::stdint::uint16_t;
 use crate::include::stdint::uint32_t;
 use crate::include::stdint::uint64_t;
 use crate::include::stdint::uint8_t;
+use std::mem;
+
+pub trait Alias {
+    fn set(&mut self, val: u64);
```
kkysen, line 8: Is it a problem that this uses `u64` for all the `val`s?  In the C macros, the integer size would only be expanded to `u64` for the cases where it's actually needed.

src/ctx.rs
```diff
@@ -2,30 +2,100 @@ use crate::include::stdint::uint16_t;
...
+        self.u64_0 = val;
+    }
+}
+
+#[inline]
+pub unsafe fn set_ctx_rep1<T: Alias>(buf: *mut u8, off: isize, val: u64) {
+    let buf = buf.offset(off);
+    let buf = buf.cast::<T>();
+    // debug_assert!(buf.is_aligned()); // #![feature(pointer_is_aligned)]
+    debug_assert!(buf.cast::<u8>() as usize & mem::align_of::<T>() - 1 == 0);
```
kkysen, line 67: This alignment debug check doesn't work when using `T` that are not extra aligned.  That's why I used `Align16` and `Align32`, as the C code expects it to be aligned to that in those cases.

src/ctx.rs
kkysen: Also, I'm not really sure we need these `aliasN` types at all.  We can just do a direct `*buf = val`, or `buf[i] = val`, right?  I think C and Rust might have different alignment rules and this is why it was done this way in C, to force the compiler to know that these are aligned writes.  In Rust, `*ptr = val` or `ptr.write()` is always aligned, while `ptr.write_unaligned()` is explicit about being unaligned.

kkysen (commented):
I just opened an alternative `case_set` impl in #127 that I was working on earlier and I think addresses the concerns I had with this way, especially with the indirect function pointer used here, which could heavily pessimize performance.

src/ctx.rs
```diff
@@ -2,30 +2,100 @@ use crate::include::stdint::uint16_t;
...
+    F: Fn(&mut D, isize, u64, SetCtxFn),
+{
+    match var {
+        1 => set_ctx(dir, off, 0x01, set_ctx_rep1::<alias8>),
+        2 => set_ctx(dir, off, 0x0101, set_ctx_rep1::<alias16>),
+        4 => set_ctx(dir, off, 0x01010101, set_ctx_rep1::<alias32>),
+        8 => set_ctx(dir, off, 0x0101010101010101, set_ctx_rep1::<alias64>),
+        16 => set_ctx(dir, off, 0x0101010101010101, set_ctx_rep2),
+
+        _ => unreachable!(),
```
kkysen, line 99: I don't think this case is generally reachable, but the C version did nothing in this case, not panic.

src/ctx.rs
randomPoison: Ah, okay. I'll remove it here then since it's not doing anything.

src/ctx.rs
randomPoison: Ah true. Fixed.

src/ctx.rs
randomPoison: Not sure, though I'd hope so given that all of the functions should be getting inlined. I don't think we need to worry about it too much here since you already have improvements lined up.

src/ctx.rs
randomPoison: I'd like to stick with the `aliasN` approach in this PR, just to keep it as close as possible to the original C. It shouldn't matter too much, though, since you already have improvements lined up.

src/ctx.rs
kkysen: Yeah, it should be all good with the future improvements in #127.

src/decode.rs
```diff
@@ -4533,230 +4534,14 @@ unsafe fn decode_b(
...
-                        .as_mut_ptr()
-                        .offset((bx4 + 24 as libc::c_int) as isize) as *mut uint8_t
-                        as *mut alias64))
-                        .u64_0 = const_val_6;
-                }
-                _ => {}
-            }
+
+            let set_ctx = |dir: &mut BlockContext, off, mul, rep_macro: SetCtxFn| {
+                rep_macro(dir.mode.as_mut_ptr() as *mut u8, off, mul * y_mode_nofilt as u64);
```
kkysen, line 4539: I think the `as *mut u8`s are redundant.

kkysen (approved):
Besides the little `as *mut u8` thing, I think this looks good enough now considering we have #127 as an improvement later.  We'll see how this version works once we start adding it to `case_set`s that are actually run.  I am surprised that we seem to have a bunch of these code paths that are fully unused, though; I thought our test suite was pretty comprehensive.

thedataking:
Since this patch has already done through several rounds of review, I did a quick performance comparison between `main` this PR and #127. I reused the test command that Frank previously shared: ` /usr/bin/time target/release/dav1d -i ~/.phoronix-test-suite/installed-tests/pts/dav1d-1.13.0/summer_nature_1080p.ivf -o /dev/null`

Results on `donna`
```
#### benchmarking: main
Decoded 3604/3604 frames (100.0%) - 838.73/25.00 fps (33.55x)
Decoded 3604/3604 frames (100.0%) - 816.25/25.00 fps (32.65x)
Decoded 3604/3604 frames (100.0%) - 819.35/25.00 fps (32.77x)
Decoded 3604/3604 frames (100.0%) - 827.66/25.00 fps (33.11x)
Decoded 3604/3604 frames (100.0%) - 835.20/25.00 fps (33.41x)
#### benchmarking: decode-b/case-set
Decoded 3604/3604 frames (100.0%) - 841.61/25.00 fps (33.66x)
Decoded 3604/3604 frames (100.0%) - 829.99/25.00 fps (33.20x)
Decoded 3604/3604 frames (100.0%) - 844.11/25.00 fps (33.76x)
Decoded 3604/3604 frames (100.0%) - 836.74/25.00 fps (33.47x)
Decoded 3604/3604 frames (100.0%) - 844.58/25.00 fps (33.78x)
#### benchmarking: kkysen/case-set
Decoded 3604/3604 frames (100.0%) - 838.69/25.00 fps (33.55x)
Decoded 3604/3604 frames (100.0%) - 827.71/25.00 fps (33.11x)
Decoded 3604/3604 frames (100.0%) - 820.79/25.00 fps (32.83x)
Decoded 3604/3604 frames (100.0%) - 843.61/25.00 fps (33.74x)
Decoded 3604/3604 frames (100.0%) - 816.23/25.00 fps (32.65x)
```

Results on `amy`
```
#### benchmarking: main
Decoded 3604/3604 frames (100.0%) - 839.52/25.00 fps (33.58x)
Decoded 3604/3604 frames (100.0%) - 841.84/25.00 fps (33.67x)
Decoded 3604/3604 frames (100.0%) - 841.71/25.00 fps (33.67x)
Decoded 3604/3604 frames (100.0%) - 827.67/25.00 fps (33.11x)
Decoded 3604/3604 frames (100.0%) - 834.91/25.00 fps (33.40x)
#### benchmarking: decode-b/case-set
Decoded 3604/3604 frames (100.0%) - 835.58/25.00 fps (33.42x)
Decoded 3604/3604 frames (100.0%) - 826.05/25.00 fps (33.04x)
Decoded 3604/3604 frames (100.0%) - 832.03/25.00 fps (33.28x)
Decoded 3604/3604 frames (100.0%) - 805.15/25.00 fps (32.21x)
Decoded 3604/3604 frames (100.0%) - 819.98/25.00 fps (32.80x)
#### benchmarking: kkysen/case-set
Decoded 3604/3604 frames (100.0%) - 815.72/25.00 fps (32.63x)
Decoded 3604/3604 frames (100.0%) - 815.45/25.00 fps (32.62x)
Decoded 3604/3604 frames (100.0%) - 833.57/25.00 fps (33.34x)
Decoded 3604/3604 frames (100.0%) - 820.78/25.00 fps (32.83x)
Decoded 3604/3604 frames (100.0%) - 829.97/25.00 fps (33.20x)
```

I don't see any performance regression from `main` to #126. I don't see a performance advantage to #127.

thedataking (approved):

PR #127, opened 2023-04-14
Port the `case_set` macro to a Rust `fn` and deduplicate a few of its calls in `decode_b`

kkysen:
This is the implementation of `case_set` I was working on earlier before #126 was opened.  I noticed a couple of concerns in #126, like with the function pointer, which may not necessarily be de-virtualized, and could have a heavy cost.  There were also a few things about alignment and forcing the usage of wider `u64`s when not needed.  This implementation I think should fix those concerns, and is also a lot simpler.  I also did a quick check of the assembly output of `CaseSetter::set_impl` vs. a more straightforward translation with the `aliasN` types and the `rep2`, `rep4` functions, and this way actually allows for further optimization, while also being a lot simpler.  I think this simpler implementation is also easier to copy for the similar `case_set_uptoN(_with_default)` macros.

Also, I deduplicated the same set of `case_set` calls in `decode_b` that #126 did.


randomPoison:
I like the look of this PR on the whole, but since it diverges so much from how the original C is setup I'm not quite sure that what you have here does the same thing as the original C. Normally having passing tests in CI would be enough to convince me that it's correct, but I did a bit of testing and it looks like none of these first few `case_set` invocations in `decode_b` are actually hit in any of the unit tests we currently have setup. So unfortunately this PR and #126 are effectively untested 🙃

Given the lack of test coverage, I'm inclined to stick with the more naïve translation in #126, since it's at least easier to verify that it's doing the same thing as the original C. So my suggestion for how to proceed is this: Let's get the naive version in #126 merged, then you can open a new PR (or rebase this one) rewriting the `case_set` logic. Hopefully then it'll be a bit easier to compare the two Rust versions and confirm that your approach here is doing the same thing. Does that seem reasonable? Hopefully it won't be too much work to rebase these changes on top of the changes in #126.

kkysen:
I think that's a pretty good plan.  We can merge #126 now and then keep this one for later.  It shouldn't be too difficult to rebase it on top of yours.

kkysen:
That said, we should probably replace a `case_set` that actually gets run for both of these PRs, right?

randomPoison:
🤷 Yeah ideally I'd like to make sure that the translated `case_set` stuff is actually getting tested, but I'm not actually sure which instances of it are going to be hit. We may just need to keep cleaning up code until we hit some test failures.

thedataking:
> which may not necessarily be de-virtualized, and could have a heavy cost.

Premature optimization is the root of all evil. I think we should stick to #126 unless there is a clear performance problem.

> but since it diverges so much from how the original C is setup I'm not quite sure that what you have here does the same thing as the original C.

I've also mentioned several times to Khyber that we should not diverge from the structure of the original C since it makes debugging harder.

randomPoison:
This has been superceded by #283.

Issue #115, opened 2023-04-12
SIGSEGV on x86 if --cpumask is set to none, sse2, sse3, sse41


thedataking:
Values that work include -1, avx2 and avx512icl. Once this is resolved, we'll want to run tests with different cpumasks.

fbossen:
Can this issue be closed?

thedataking:
I wasn't able to reproduce this issue after a quick test so it should be safe to close.

Issue #122, opened 2023-04-13
Audit alignment directives


randomPoison:
The original C code contains a number of alignment directives in order to set minimum alignment values for certain struct fields. These are often needed for SIMD instructions inside assembly code, since some SIMD instructions will segfault on unaligned data. The original c2rust transpile did not translate these directives, so we've been manually adding them back in as needed. We should do a full audit of the alignment directives in dav1d and make sure we have translated them all over to rav1d.

PR #129, opened 2023-04-14
Fix alignment of lim_lut and refmvs_block (#115)

fbossen:


kkysen (commented):
If it's the type that's being aligned everywhere we use it, we could just put a `#[repr(align(16))]` on the type and not need to update it everywhere.

thedataking (approved):
LGTM!

fbossen:
> If it's the type that's being aligned everywhere we use it, we could just put a `#[repr(align(16))]` on the type and not need to update it everywhere.

A lot of duplicate changes are the result of `Dav1dFrameContext_lf` being defined in multiple files. If `Dav1dFrameContext_lf` were to be defined only once, then there wouldn't be a need to make multiple changes. For now, it may be better to have alignment directives closely match what is done in the original C code to make auditing (see #122) of such directives easier.

randomPoison:
@fbossen Looks like there's a merge conflict on this now. Can you resolve that? I can merge this once you do.

kkysen:
> A lot of duplicate changes are the result of `Dav1dFrameContext_lf` being defined in multiple files. If `Dav1dFrameContext_lf` were to be defined only once, then there wouldn't be a need to make multiple changes. For now, it may be better to have alignment directives closely match what is done in the original C code to make auditing (see #122) of such directives easier.

Ah, sorry, I didn't realize it's because the containing type wasn't deduplicated yet.  You're right then.

Issue #25, opened 2023-03-20
Fix duplicated _tmpl.rs files


randomPoison:
The original C code contains a number of `_tmpl.c` files that get compiled once for each active bitdepth. This means that if both 8 and 16 bitdepths are enabled those files get compiled twice. To mimic this in the transpiled Rust, we have transpiled these files twice, e.g. `cdef_tmpl.c` gets transpiled to `cdef_tmpl_8.rs` and `cdef_tmpl_16.rs`.

We should see if we can merge the two Rust files back into a single one. Currently there's a lot of duplicated code between the two versions of a given file, with only relatively small differences between the two in some cases.

kkysen:
The templated files are all for bitdepths, right?  And multiple bitdepths can be enabled at the same time, right?  If multiple can be active at once, then it makes deduplicating with features only a lot more complex.  It *may* be easier with macros or generics, though that may get complex as well.  I'd guess that generics would probably be the proper abstraction in the end once things are cleaned up, though I'm much less sure about right now.

The challenges in trying to use features for this, correct me if I'm wrong, we'd have to define e.x. `pixel_8` and `pixel_16` types, and then every item where such a bitdepth-differentiated item is used, duplicate that entire item to have a `*_8` and `*_16` version since we need to have both of them at the same time.  So if we do it this way, it'll still be complex and I'm not sure if we'll have actually deduplicated that much.

The challenges in trying to use macros for this is that all the code will be contained in a macro.  This way will probably be closest to the original C.

The challenges in trying to use generics for this is we can't use generics for `extern "C" fn`s, but I think we could:
* remove the `extern "C"` for functions that are not exported (I believe these are only those defined in `include/dav1d/` marked with `DAVID_API`, right?).
* for functions that are exported, define an 8- and 16-bitdepth `extern "C"` wrapper function that just calls the generic version.
Otherwise, it appears that the generic part of the bitdepths is neatly abstracted in `bitdepths.h` and they all seem amenable to conversion to Rust generics.

If we do go the generics route, I think it would be helpful to first deduplicate all the function declarations into `use`s and then make only the `DAVID_API` ones `pub extern "C"`.

randomPoison:
I think you're right that we can't really do this with `#[cfg(feature)]` alone. I think generics are the right direction to go in, so doing a non-generic `extern "C"` function that just calls the generic function is the right way to go.

[...]

PR #131, opened 2023-04-14
decode_b: Cleanup a chunk of logic

randomPoison:
* Add `frame.rs` corresponding to `frame.h`.
* Add is_inter_or_switch helper method.
* Add helper method to `Av1Block` to more succinctly access unnamed fields.
* Add helper to `Dav1dFrameContext_bd_fn` to call function pointer more succinctly.

### DIFF
--- a/include/common/frame.rs
+++ b/include/common/frame.rs
@@ -1,0 +1,9 @@
+use crate::include::dav1d::headers::Dav1dFrameHeader;
+
+
+/// Checks whether the frame type is `INTER` or `SWITCH`.
+pub fn is_inter_or_switch(frame_header: &Dav1dFrameHeader) -> bool {
+    // Both are defined as odd numbers {1, 3} and therefore have the LSB set.
+    // See also: AV1 spec 6.8.2
+    frame_header.frame_type & 1 != 0
+}
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -1,3 +1,4 @@
+use crate::include::common::frame::is_inter_or_switch;
 use crate::include::stddef::*;
 use crate::include::stdint::*;
 use ::libc;
@@ -542,6 +543,19 @@ pub struct Dav1dFrameContext_bd_fn {
     pub backup_ipred_edge: backup_ipred_edge_fn,
     pub read_coef_blocks: read_coef_blocks_fn,
 }
+
+impl Dav1dFrameContext_bd_fn {
+    pub unsafe fn recon_b_intra(
+        &self,
+        context: *mut Dav1dTaskContext,
+        block_size: BlockSize,
+        flags: EdgeFlags,
+        block: *const Av1Block,
+    ) {
+        self.recon_b_intra.expect("non-null function pointer")(context, block_size, flags, block);
+    }
+}
+
 pub type read_coef_blocks_fn = Option::<
     unsafe extern "C" fn(*mut Dav1dTaskContext, BlockSize, *const Av1Block) -> (),
 >;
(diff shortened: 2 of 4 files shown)


include/common/frame.rs
```diff
@@ -0,0 +1,9 @@
+use crate::include::dav1d::headers::Dav1dFrameHeader;
+
+
+/// Checks whether the frame type is `INTER` or `SWITCH`.
```
kkysen, line 4: As we clean up the code now, should we be adding back the comments now?  It seems like a good opportunity for it now.  Should I do that, too?

src/decode.rs
```diff
@@ -542,6 +543,19 @@ pub struct Dav1dFrameContext_bd_fn {
     pub backup_ipred_edge: backup_ipred_edge_fn,
     pub read_coef_blocks: read_coef_blocks_fn,
 }
+
+impl Dav1dFrameContext_bd_fn {
```
kkysen, line 547: I think this type is still not yet deduplicated as it contains some bitdepth-dependent types.  So we need to be careful in adding code to it.  We'll probably want to define this on all the `bd_fn` types, for example.  But that probably could come later.  Just something to keep in mind.

src/decode.rs
```diff
@@ -542,6 +543,19 @@ pub struct Dav1dFrameContext_bd_fn {
...
+    pub unsafe fn recon_b_intra(
+        &self,
+        context: *mut Dav1dTaskContext,
+        block_size: BlockSize,
+        flags: EdgeFlags,
+        block: *const Av1Block,
+    ) {
+        self.recon_b_intra.expect("non-null function pointer")(context, block_size, flags, block);
```
kkysen, line 555: A lot of these function pointers I think are never null.  Could the `Option` just be dropped instead?

src/levels.rs
```diff
@@ -280,3 +280,9 @@ pub struct Av1Block {
     pub uvtx: uint8_t,
     pub c2rust_unnamed: Av1Block_intra_inter,
 }
+
+impl Av1Block {
+    pub unsafe fn y_mode(&self) -> u8 {
+        self.c2rust_unnamed.c2rust_unnamed.y_mode
```
kkysen, line 286: This is helpful and closely matches how the C code was written.

We can also rename the unnamed fields, too, now.  I didn't before since they weren't deduplicated yet, but now that they are, they can easily be renamed with a refactor.

src/decode.rs
```diff
@@ -4542,40 +4553,28 @@ unsafe fn decode_b(
...
-                    (*r.offset(x as isize))
-                        .ref_0
-                        .ref_0[0 as libc::c_int as usize] = 0 as libc::c_int as int8_t;
-                    (*r.offset(x as isize)).bs = bs as uint8_t;
-                    x += 1;
+            if is_inter_or_switch(&*f.frame_hdr) {
+                let r: *mut refmvs_block = t
+                    .rt
+                    .r[((t.by & 31) + 5 + bh4 - 1) as usize]
+                    .offset(t.bx as isize);
```
kkysen, line 4560: To be clear, were `.as_mut_ptr().offset(...)`s turned into `[...]` indexing (where there is a slice to start with), while bare `.offset(...)`s were kept the same (there's no slice, so indexing can't be done yet)?

src/decode.rs
```diff
@@ -4542,40 +4553,28 @@ unsafe fn decode_b(
...
-                    (*(*rr.offset(y as isize))
-                        .offset((t.bx + bw4 - 1 as libc::c_int) as isize))
-                        .bs = bs as uint8_t;
-                    y += 1;
+
+                let mut rr: *const *mut refmvs_block = &t.rt.r[((t.by & 31) + 5) as usize];
+                for y in 0..bh4 - 1 {
+                    let block = &mut *rr
+                        .offset(y as isize)
+                        .read()
```
kkysen, line 4571: This used to be a `*` deref, not a `ptr::read`.  I believe they're slightly different.

include/common/frame.rs
randomPoison: Yeah, please do! I've been adding them in where I find them in the original C.

src/decode.rs
randomPoison: Good point. I might attempt to deduplicate this next, then, to avoid potential issues.

src/decode.rs
randomPoison: Wrote up #140 to track this.

src/decode.rs
kkysen: To deduplicate this one, I think we'll have to do #25 with generics.  If you find a way to do it more easily, that'd be awesome.  Right now the function pointers all have `pixel`, etc. args, and `pixel` is a bitdepth-dependent type and so there may be multiple of these types existing at once.

src/decode.rs
kkysen: Got it, that make sense.  I started to come across all these instances, too, when I started cleaning up functions as well.

src/decode.rs
randomPoison: Actually looks like there was a better way to express this that allows us to remove those extra raw pointer operations entirely.

kkysen: `ptr::read` is subtly different.  I looked into it and it *may* be the same for the specific type used here, but I think it's safest to stick with the `*` deref in general for these cases so we don't get bitten by one of those subtle differences.

kkysen (approved):
LGTM, and the new way to get around `.read()` looks good, too.

PR #132, opened 2023-04-14

randomPoison:
* Add helper methods to `mv` an `Av1Block` for accessing unnamed fields.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4608,19 +4608,31 @@ unsafe fn decode_b(
 
                     if DEBUG_BLOCK_INFO(f, t)
                     {
+                        use std::fmt;
+
+                        /// Helper struct for printing a number as a signed hexidecimal value.
+                        struct SignAbs(i32);
+
+                        impl fmt::Display for SignAbs {
+                            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+                                let sign = if self.0 < 0 { "-" } else { " " };
+                                write!(f, "{}{:x}", sign, self.0.abs())
+                            }
+                        }
+
                         println!(
-                            "[ {:+x} {:+x} {:+x}\n  {:+x} {:+x} {:+x} ]\n\
-                            alpha={:+x}, beta={:+x}, gamma={:+x}, deta={:+x}, mv=y:{},x:{}",
-                            t.warpmv.matrix[0],
-                            t.warpmv.matrix[1],
-                            t.warpmv.matrix[2],
-                            t.warpmv.matrix[3],
-                            t.warpmv.matrix[4],
-                            t.warpmv.matrix[5],
-                            t.warpmv.u.p.alpha,
-                            t.warpmv.u.p.beta,
-                            t.warpmv.u.p.gamma,
-                            t.warpmv.u.p.delta,
+                            "[ {} {} {}\n  {} {} {} ]\n\
+                            alpha={}, beta={}, gamma={}, deta={}, mv=y:{},x:{}",
+                            SignAbs(t.warpmv.matrix[0]),
+                            SignAbs(t.warpmv.matrix[1]),
+                            SignAbs(t.warpmv.matrix[2]),
+                            SignAbs(t.warpmv.matrix[3]),
+                            SignAbs(t.warpmv.matrix[4]),
+                            SignAbs(t.warpmv.matrix[5]),
+                            SignAbs(t.warpmv.u.p.alpha.into()),
+                            SignAbs(t.warpmv.u.p.beta.into()),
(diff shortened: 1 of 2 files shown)


src/decode.rs
```diff
@@ -4587,172 +4587,42 @@ unsafe fn decode_b(
...
-                        .warpmv
-                        .matrix[5 as libc::c_int
-                        as usize] = b
-                        .c2rust_unnamed
-                        .c2rust_unnamed_0
-                        .c2rust_unnamed
-                        .c2rust_unnamed_0
-                        .matrix[3 as libc::c_int as usize] as libc::c_int
-                        + 0x10000 as libc::c_int;
+                    t.warpmv.matrix[2] = b.matrix()[0] as int32_t + 0x10000;
```
kkysen, line 4595: This one used to be `as c_int`, not `as int32_t`.

src/decode.rs
```diff
@@ -4587,172 +4587,42 @@ unsafe fn decode_b(
...
+                    t.warpmv.matrix[3] = b.matrix()[1] as int32_t;
+                    t.warpmv.matrix[4] = b.matrix()[2] as int32_t;
+                    t.warpmv.matrix[5] = b.matrix()[3] as int32_t + 0x10000;
```
kkysen, line 4598: This one used to be `as c_int`, not `as int32_t`, too.

src/decode.rs
```diff
@@ -4587,172 +4587,42 @@ unsafe fn decode_b(
...
-                            b
-                                .c2rust_unnamed
-                                .c2rust_unnamed_0
-                                .c2rust_unnamed
-                                .c2rust_unnamed_0
-                                .mv2d
-                                .c2rust_unnamed
-                                .x as libc::c_int,
+                        println!(
+                            "[ {:+x} {:+x} {:+x}\n  {:+x} {:+x} {:+x} ]\n\
```
kkysen, line 4612: The `signabs` macro expanded above is slightly different from this.  It prints `-` if negative or else ` `, not `+`, which I believe is what the `+` sign formatting flag does in Rust.  I think we should just deduplicate the `signabs` macro.

kkysen: What does the `x` in `{:+x}` mean?

src/decode.rs
randomPoison: `t.warpmv.matrix` is an array of `int32_t`, so I'd prefer to cast directly to that type here.

src/decode.rs
randomPoison: The `x` in `{:+x}` means that the value is formatted as a hexadecimal number.

You're right, I completely mistranslated what this print output was doing (I entirely forgot the `abs` invocation!). I've changed it to a new version that translates `signabs` into a Rust helper struct, it should behave closer to the original now.

src/levels.rs
```diff
@@ -196,6 +196,16 @@ pub union mv {
...
 
+impl mv {
+    pub fn x(&self) -> i16 {
+        unsafe { self.c2rust_unnamed.x }
+    }
+
+    pub fn y(&self) -> i16 {
+        unsafe { self.c2rust_unnamed.y }
+    }
+}
```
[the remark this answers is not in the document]
randomPoison: Sure, seems like useful cleanup to do. How do you plant to make it fully safe though?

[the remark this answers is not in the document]
kkysen: Okay, I'll wait until after this merges.  I plan to make `x` and `y` getters (and setters will be needed, too, probably) for the `n: u32` field.  The union here seems to be only used for convenience.

src/decode.rs
kkysen: Resolved here: https://github.com/memorysafety/rav1d/pull/132#discussion_r1169106458.

src/decode.rs
kkysen: a0c31084bb60ab3564603136ef06bc1f18eb0a71 looks good.

src/decode.rs
```diff
@@ -4587,172 +4587,54 @@ unsafe fn decode_b(
...
+
+                        /// Helper struct for printing a number as a signed hexidecimal value.
+                        struct SignAbs(i32);
+
+                        impl fmt::Display for SignAbs {
+                            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+                                let sign = if self.0 < 0 { "-" } else { " " };
+                                write!(f, "{}{:x}", sign, self.0.abs())
+                            }
+                        }
```
kkysen, line 4621: `signabs` was used twice in `decode_b`, though re-defined.  I think we can move this to the top of `decode_b`.

kkysen (approved):

PR #133, opened 2023-04-17
Cleanup and make safe `attributes.rs` and `intops.rs`

kkysen:
These functions are already safe, so I just removed the `unsafe` and `extern "C"` and cleaned them up a bit.  They're used all over the place, including in `decode_b` and its children.

### DIFF
--- a/include/common/attributes.rs
+++ b/include/common/attributes.rs
@@ -1,14 +1,14 @@
 #[inline]
-pub unsafe extern "C" fn ctz(mask: libc::c_uint) -> libc::c_int {
+pub fn ctz(mask: libc::c_uint) -> libc::c_int {
     return mask.trailing_zeros() as i32;
 }
 
 #[inline]
-pub unsafe extern "C" fn clz(mask: libc::c_uint) -> libc::c_int {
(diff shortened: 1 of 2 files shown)


include/common/attributes.rs
```diff
@@ -1,14 +1,16 @@
+use std::ffi::{c_int, c_uint, c_ulonglong};
+
 #[inline]
-pub unsafe extern "C" fn ctz(mask: libc::c_uint) -> libc::c_int {
-    return mask.trailing_zeros() as i32;
+pub fn ctz(mask: c_uint) -> c_int {
```
randomPoison, line 4: For these functions that specify their types with the C numeric types, are we able to convert them to the Rust numeric types at this phase? If we're cleaning up these functions, may as well use more idiomatic Rust types if they work for all of our targets.

kkysen: I wasn't sure yet either.  In some PRs I think I may have changed `uint8_t` to `u8` in some places and things like that, but so far I've left the `ffi::c_*` types alone.

randomPoison (approved):

PR #136, opened 2023-04-17
Cleanup and make safe `fn findoddzero`

kkysen:


src/decode.rs
```diff
@@ -2938,20 +2938,16 @@ unsafe extern "C" fn derive_warpmv(
...
-unsafe extern "C" fn findoddzero(
-    mut buf: *const uint8_t,
-    mut len: libc::c_int,
-) -> libc::c_int {
-    let mut n: libc::c_int = 0 as libc::c_int;
-    while n < len {
-        if *buf.offset((n * 2 as libc::c_int) as isize) == 0 {
-            return 1 as libc::c_int;
-        }
-        n += 1;
```
randomPoison, line 2951: If I'm reading this correctly, doesn't the original logic read past end of the buffer? It does `while n < len`, and then indexes buf with `n * 2`, so any values where `n > len / 2` would read past the end of `buf` right? Seems like your implementation is more correct, just seems funny to me that the original logic is incorrect

randomPoison (approved):

src/decode.rs
kkysen: The original is correct, too, but quite misleading.  `len` is not the length, but half the length.  It also checks the even indices even though it says `findoddzero`.  I changed that `+ 1` at the beginning so now it does the same thing in its calls, but it finds the odd zero for real.

PR #145, opened 2023-04-17

randomPoison:
* Add `DAV1D_MAX_SEGMENTS` constant which appears to have not been translated originally.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -2,11 +2,11 @@ use crate::include::common::frame::is_inter_or_switch;
 use crate::include::stddef::*;
 use crate::include::stdint::*;
 use ::libc;
+use crate::src::align::Align16;
 use crate::src::msac::MsacContext;
 use crate::src::cdf::{CdfContext, CdfMvComponent, CdfMvContext};
 use crate::src::ctx::{case_set, SetCtxFn};
-
-use crate::src::align::Align16;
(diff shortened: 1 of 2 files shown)


src/decode.rs
```diff
@@ -4703,45 +4707,41 @@ unsafe fn decode_b(
...
-                b
-                    .seg_id = neg_deinterleave(
+                    .last_active_segid;
+
+                b.seg_id = neg_deinterleave(
                     diff as libc::c_int,
                     pred_seg_id as libc::c_int,
-                    last_active_seg_id.wrapping_add(1 as libc::c_int as libc::c_uint)
-                        as libc::c_int,
+                    last_active_seg_id + 1,
```
kkysen, line 4723: Does `seg_id` increase by 1 generally?  It's not supposed to overflow at all, right?

src/decode.rs
```diff
@@ -4685,16 +4687,18 @@ unsafe fn decode_b(
                         f.prev_segmap,
                         f.b4_stride,
                     );
-                    if seg_id_0 >= 8 as libc::c_int as libc::c_uint {
-                        return -(1 as libc::c_int);
+
+                    if seg_id >= 8 {
+                        return -1;
                     }
```
kkysen, line 4693: Ah, yes, it is bounded, so https://github.com/memorysafety/rav1d/pull/145/files#r1169357964 is fine then.

kkysen (approved):

PR #146, opened 2023-04-17
decode_b: Make seg safe

randomPoison:


src/decode.rs
```diff
@@ -4765,7 +4763,7 @@ unsafe fn decode_b(
     } else {
         b.skip_mode = 0 as libc::c_int as uint8_t;
     }
-    if b.skip_mode as libc::c_int != 0 || !seg.is_null() && (*seg).skip != 0 {
+    if b.skip_mode as libc::c_int != 0 || seg.map(|seg| seg.skip != 0).unwrap_or_default() {
```
kkysen, line 4766: If we have `.unwrap_or(true)`s, then I think we should use `.unwrap_or(false)` here.

kkysen: Same for the others.

src/decode.rs
```diff
@@ -9982,13 +9980,13 @@ unsafe fn decode_b(
...
                 .comp_type = COMP_INTER_NONE as libc::c_int as uint8_t;
-            if !seg.is_null() && (*seg).ref_0 > 0 as libc::c_int {
+            if seg.map(|seg| seg.ref_0 > 0).unwrap_or_default() {
                 b
                     .c2rust_unnamed
                     .c2rust_unnamed_0
                     .r#ref[0 as libc::c_int
-                    as usize] = ((*seg).ref_0 - 1 as libc::c_int) as int8_t;
-            } else if !seg.is_null() && ((*seg).globalmv != 0 || (*seg).skip != 0) {
+                    as usize] = seg.as_ref().unwrap().ref_0 as i8 - 1;
```
kkysen, line 9988: I think this can be an `if let Some(seg) = seg && seg.ref_0 > 0` so get rid of the `seg.unwrap()`.

randomPoison: Oooh good idea, done!

src/decode.rs

PR #148, opened 2023-04-19
`fn obmc_lowest_px`: Cleanup and make mostly safe

kkysen:


randomPoison (approved):

kkysen:
I'll merge this after #149 and its children, as I think there are some merge conflicts that would be easier to fix in this PR rather than the other way around.

PR #153, opened 2023-04-19
Clean up CI a bit

thedataking:


.github/workflows/arm7/envsetup.sh
kkysen, line 1: These could also go in a `.cargo/config.toml`, and then they'd work everywhere, including locally:

```toml
[target.aarch64-unknown-linux-gnu]
linker = "aarch64-linux-gnu-gcc"
ar = "aarch64-linux-gnu-ar"
rustflags = [
    "-C", "link-arg=-fuse-ld=gold", # override any other linker
    "-C", "target-feature=+crt-static",
]

[target.armv7-unknown-linux-gnueabihf]
linker = "arm-linux-gnueabihf-gcc"
ar = "arm-linux-gnueabihf-ar"
rustflags = [
    "-C", "link-arg=-fuse-ld=gold", # override any other linker
    "-C", "target-feature=+crt-static",
]
```

randomPoison: Oh that's wonderful, much better than having to remember to run a shell script. @thedataking if you're onboard with including this, then feel free to add it to this PR 😁

randomPoison (approved):
Looks good, though I wouldn't mind seeing the `.cargo/config.toml` suggestions that @kkysen made added to this PR.

.github/workflows/arm7/envsetup.sh
thedataking: I like the idea overall but it just occurred to me that the scripts we have and this config.toml doesn't do the same thing. We set the C++ and C compilers (which I presume matters if any dependencies call any of those compilers) but the config Khyber proposes sets `ar` and switches the linker from `bfd` to `gold`. I'm not sure whether we should go the minimal route here (only set what we need) or the maximal route (set anything just in case) 🤔

kkysen: Oh, for the `bfd`/`gold` linker, I just had that because I have `mold` set as my default linker in my `~/.cargo/config.toml`, so I wanted to override that.  If the default here is `bfd`, we can use that; I was more just trying to reset it to a linker that worked for cross-compiling.

thedataking: I don't see docs for how to add cc and cxx on `config.toml` so trying the minimal route first.

PR #154, opened 2023-04-19
Add `--release` overflow checks for testing

kkysen:
We eventually want this in a separate profile used for testing (see #155), but this is the simplest fix for now so that we can catch bugs in testing now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,3 +31,6 @@ default = ["asm", "bitdepth_8", "bitdepth_16"]
 asm = []
 bitdepth_8 = []
 bitdepth_16 = []
+
+[profile.release]
+overflow-checks = true


randomPoison:
I think #153 is going to change us to using debug builds for testing, so this PR might be redundant? We probably don't need to merge it unless @thedataking thinks #153 is going to take a while to land.

kkysen:
Yeah, it seems #153 should mostly fix this concern, so I think we can close this PR.  That said, I think #153 enables debug builds by default, so we can easily test it locally, but I think it will still run in release mode in CI.

thedataking:
>  I think it will still run in release mode in CI.
Correct.  What I would suggest for CI:
- Run the QEMU (slow) builds in release mode just like today - they bottleneck the CI runs
- Run the x86 native builds (32 and 64-bit) in debug mode which can catch overflows

kkysen:
That sounds like a great idea.

thedataking:
Implemented in #162 

> * Run the QEMU (slow) builds in release mode just like today - they bottleneck the CI runs
> * Run the x86 native builds (32 and 64-bit) in debug mode which can catch overflows

randomPoison:
Closing this since it's superseded by other PRs. @kkysen reopen if I'm wrong though!

PR #157, opened 2023-04-19
`fn fix_int_mv_precision`: Cleanup and make safe

kkysen:


src/env.rs
```diff
@@ -77,29 +77,23 @@ pub unsafe extern "C" fn get_poc_diff(
...
-        - ((*mv).y as libc::c_int >> 15 as libc::c_int)
-        + 3 as libc::c_int) as libc::c_uint & !(7 as libc::c_uint)) as int16_t;
+fn fix_int_mv_precision(mv: &mut mv) {
+    mv.x = (mv.x - (mv.x >> 15) + 3) & !7;
+    mv.y = (mv.y - (mv.y >> 15) + 3) & !7;
 }
 
 #[inline]
-pub unsafe extern "C" fn fix_mv_precision(hdr: *const Dav1dFrameHeader, mv: *mut mv) {
+pub unsafe extern "C" fn fix_mv_precision(hdr: *const Dav1dFrameHeader, mv: &mut mv) {
```
randomPoison, line 86: Do we still need `unsafe extern "C"` here? If it's fully safe and has been deduplicated, neither of those specifiers should be necessary anymore right?

randomPoison: Ah nevermind, just looked at #158.

randomPoison (approved):

PR #162, opened 2023-04-19
Build rav1d in debug mode for x86 to catch overflows

### LINKED ISSUE #119: Build fails during linking for i686-unknown-linux-gnu unless assembly is disabled

thedataking:
This enables asm for x86 which enables us to test x86 and x86-64 without slowing down CI. Closes #119. 

`build.rs` is probably what warrants most reviewer attention, the rest should be totally unsurprising.


.github/workflows/build-and-test.yml
```diff
@@ -34,25 +40,17 @@ jobs:
...
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      - name: cargo build
-        run: cargo build --release
-      - name: meson test
-        run: |
-          .github/workflows/test.sh
-          cp ${{ github.workspace }}/build/meson-logs/testlog.txt \
-             ${{ github.workspace }}/build/meson-logs/testlog-$(uname -m).txt
-      - name: cargo build for i686 (assembly temporarily disabled)
+      - name: cargo build for ${{ matrix.target }} (assembly temporarily disabled)
```
randomPoison, line 43: Should this still say "assembly temporarily disabled"? Isn't assembly enabled for all targets now?

src/cdef_tmpl_8.rs
```diff
@@ -922,22 +918,26 @@ unsafe extern "C" fn cdef_dsp_init_x86(c: *mut Dav1dCdefDSPContext) {
     (*c).fb[1] = Some(dav1d_cdef_filter_4x8_8bpc_sse4);
     (*c).fb[2] = Some(dav1d_cdef_filter_4x4_8bpc_sse4);
 
-    if flags & DAV1D_X86_CPU_FLAG_AVX2 == 0 {
-        return;
-    }
+    cfg_if!{
+        if #[cfg(target_arch = "x86_64")] {
+            if flags & DAV1D_X86_CPU_FLAG_AVX2 == 0 {
```
randomPoison, line 923: Since there's no `else` in this `cfg_if!` block, it might be cleaner to use a regular `#[cfg]` attribute here and the similar cases you fix in this PR:

```rust
#[cfg(target_arch = "x86_64")]
{
    // ...
}
```

randomPoison (commented):
Thanks for finally cleaning up those CPU flags! I probably shouldn't have been duplicating them in the first place, but hindsight is 20/20 I guess 🤓

randomPoison (approved):

PR #165, opened 2023-04-20
`fn add_compound_extended_candidate`: Cleanup and make safe

kkysen:


src/refmvs.rs
```diff
@@ -486,115 +486,76 @@ unsafe extern "C" fn add_temporal_candidate(
...
-        if cand_ref <= 0 as libc::c_int {
+    let (same, diff) = same.split_at_mut(2);
+
+    // Alternative way with compile-time index checking
+    // for the resulting array refs (rather than slices).
+    //
+    // Safety: `[usize; 4]` and `[[usize; 2]; 2]` should be identical.
+    // let [same_count, diff_count] = unsafe { 
+    //     std::mem::transmute::<&mut [usize; 4], &mut [[usize; 2]; 2]>(same_count)
+    // };
```
randomPoison, line 507: Do we need this comment? Seems like it's documenting a different approach that you're no longer using?

randomPoison (approved):

src/refmvs.rs
kkysen: Yeah, I added it in case we want to implement it that way (trading a transmute for compile-time indexing).  Should I remove it?

randomPoison: I think that's an approach that would be cool to take once safe transmute becomes a thing, but probably not worth the extra unsafe operation when `split_at` already does the trick nicely. So I'd say remove the comment here.

kkysen: Removed in 5ea222199003cae8abd46c2b7236c3e11859c067.

PR #166, opened 2023-04-20
Cleanup trivial redundant casts

kkysen:
I've been cleaning up all the redundant casts and similar things by hand as I clean up individual functions, but it takes a while to do by hand when in most cases the cast is trivial to detect and elide.  So I replaced most of them here with regexes; the remaining ones that are a bit more complex we can do by hand.


randomPoison:
Would it be possible to split the `cargo fmt` pass into a separate PR from the redundant casts cleanup? I'm totally fine with landing the formatting pass now, you're right that it's a good time to do it, but that shouldn't be in the same PR as the redundant casts cleanup.

randomPoison:
Yeah, having a couple of PRs to remove the casts is fine by me 👍

kkysen:
@randomPoison, they're all split up now into this PR, #177, and #178.

PR #172, opened 2023-04-20

randomPoison:


src/decode.rs
```diff
@@ -4851,211 +4851,152 @@ unsafe fn decode_b(
...
                 &mut ts.msac,
                 (*f.frame_hdr).cdef.n_bits as libc::c_uint,
-            ) as libc::c_int;
-            *(t.cur_sb_cdef_idx_ptr).offset(idx as isize) = v as int8_t;
-            if bw4 > 16 as libc::c_int {
-                *(t.cur_sb_cdef_idx_ptr)
-                    .offset((idx + 1 as libc::c_int) as isize) = v as int8_t;
+            ) as i8;
+
+            *(t.cur_sb_cdef_idx_ptr).offset(idx as isize) = v;
```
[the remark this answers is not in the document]
randomPoison: Good catch, fixed!

src/decode.rs
```diff
@@ -4851,211 +4851,152 @@ unsafe fn decode_b(
...
-            prev_delta_lf.as_mut_ptr() as *mut libc::c_void,
-            (ts.last_delta_lf).as_mut_ptr() as *const libc::c_void,
-            4 as libc::c_int as libc::c_ulong,
-        );
-        if have_delta_q != 0 {
-            let mut delta_q: libc::c_int = dav1d_msac_decode_symbol_adapt4(
+
+    // delta-q/lf
+    let shfl = ((*f.seq_hdr).sb128 == 0) as libc::c_int;
+    if t.bx & (31 >> shfl) == 0 && t.by & (31 >> shfl) == 0 {
```
kkysen, line 4890:
```suggestion
    let not_sb128 = ((*f.seq_hdr).sb128 == 0) as libc::c_int;
    if t.bx & (31 >> not_sb128) == 0 && t.by & (31 >> not_sb128) == 0 {
```

I think a more descriptive name is better here.  In other places in the C code, `const int sb128 = f->seq_hdr->sb128` is used, so I think this way is closer.

src/decode.rs
```diff
@@ -4851,211 +4851,152 @@ unsafe fn decode_b(
...
+                let n_bits = 1 + dav1d_msac_decode_bools(&mut ts.msac, 3);
+                delta_q = dav1d_msac_decode_bools(
                     &mut ts.msac,
-                    n_bits as libc::c_uint,
-                ))
-                    .wrapping_add(1 as libc::c_int as libc::c_uint)
-                    .wrapping_add(((1 as libc::c_int) << n_bits) as libc::c_uint)
-                    as libc::c_int;
+                    n_bits
+                ) as libc::c_int + 1 + (1 << n_bits);
```
kkysen, line 4912: I think this used to be unsigned adds (the `.wrapping_add`s).  I think we can move the `as libc::c_int` to around the whole result (`delta_q`), since `n_bits` is also unsigned here.

src/decode.rs
```diff
@@ -4851,211 +4851,152 @@ unsafe fn decode_b(
...
+                        delta_lf = dav1d_msac_decode_bools(
                             &mut ts.msac,
-                            n_bits_0 as libc::c_uint,
-                        ))
-                            .wrapping_add(1 as libc::c_int as libc::c_uint)
-                            .wrapping_add(
-                                ((1 as libc::c_int) << n_bits_0) as libc::c_uint,
-                            ) as libc::c_int;
+                            n_bits as libc::c_uint,
+                        ) as libc::c_int + 1 + (1 << n_bits);
```
kkysen, line 4948: Same here: https://github.com/memorysafety/rav1d/pull/172/files#r1172984709.

src/decode.rs
```diff
@@ -4851,211 +4851,152 @@ unsafe fn decode_b(
...
+    let shfl = ((*f.seq_hdr).sb128 == 0) as libc::c_int;
+    if t.bx & (31 >> shfl) == 0 && t.by & (31 >> shfl) == 0 {
+        let prev_qidx = ts.last_qidx;
+        let have_delta_q =
+            (*f.frame_hdr).delta.q.present != 0
+            && (
+                bs != (if (*f.seq_hdr).sb128 != 0 { BS_128x128 } else { BS_64x64 })
+                || b.skip == 0
+            );
+        let mut prev_delta_lf = ts.last_delta_lf;
```
kkysen, line 4898:
```suggestion
        let prev_delta_lf = ts.last_delta_lf;
```

I think you can drop the `mut` now that we're not using `memcpy`.

PR #176, opened 2023-04-20
decode_b: Make b_dim safe

randomPoison:


src/decode.rs
```diff
@@ -7956,9 +7950,7 @@ unsafe fn decode_b(
             1 => {
                 (*(&mut *(t.l.tx_intra).as_mut_ptr().offset(by4 as isize)
                     as *mut int8_t as *mut alias8))
-                    .u8_0 = (0x1 as libc::c_int
-                    * *b_dim.offset((2 as libc::c_int + 1 as libc::c_int) as isize)
-                        as libc::c_int) as uint8_t;
+                    .u8_0 = (0x1 * b_dim[2 + 1] as libc::c_int) as uint8_t;
```
kkysen, line 7953: What's the point of `0x1 * ` here?  I think the casts are redundant, going from `u8` to `c_int` to `u8` again.

```suggestion
                    .u8_0 = b_dim[2 + 1];
```

kkysen: Oh, I guess the other ones below are multiplied by different things, so this is to keep parity.  We can leave this as is then, and clean it up more later.

src/decode.rs
```diff
@@ -12934,7 +12864,7 @@ unsafe fn decode_b(
                         t,
                         lowest_px,
                         0 as libc::c_int,
-                        b_dim,
+                        b_dim.as_ptr(),
```
kkysen, line 12867: Didn't I change `obmc_lowest_px` to take `b_dim` as `&[u8; 4]` and not a pointer already?  Or is this PR just branched off before I made that change?

kkysen (approved):

src/decode.rs
randomPoison: Yup, that's fixed up after a rebase.

PR #184, opened 2023-04-26
`fn resolve_divisor_{32,64}`: Cleanup and make safe

kkysen:
These could be further cleaned up by parameterizing over `u32` and `u64`, but it's not that important yet.  I would like to replace the `ulog2` and `u64log2` calls with `{u32, u64}::ilog2`, but `ilog2` was stabilized in Rust `1.67`, and our pinned nightly is a bit behind that.  Would it be a good idea to upgrade that to something newer?  Unfortunately, we'll have to stay on nightly for a while longer until we finish getting rid of varargs, extern types, and intrinsic calls.

### DIFF
--- a/src/warpmv.rs
+++ b/src/warpmv.rs
@@ -38,16 +38,15 @@ fn iclip_wmp(v: libc::c_int) -> libc::c_int {
 }
 
 #[inline]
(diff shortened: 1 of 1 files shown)


randomPoison:
I think bumping to the latest nightly at some point would be good, esp if there's new stabilized stuff that we want to use 👍

randomPoison (approved):

PR #189, opened 2023-04-27
Add workflow for checking formatting and run one more formatting pass

randomPoison:
A few things have gotten mis-formatted as a result of bad merges since the initial formatting pass. This PR adds a new workflow to check formatting in CI, and does one final `cargo fmt` run over everything.

### DIFF
--- a/.github/workflows/cargo-fmt.yml
+++ b/.github/workflows/cargo-fmt.yml
@@ -9,5 +9,4 @@ jobs:
         with:
           fetch-depth: 1
       - name: Check formatting
-        run: |
-          cargo fmt --check
+        run: cargo fmt --check
--- a/.github/workflows/cargo-fmt.yml
+++ b/.github/workflows/cargo-fmt.yml
@@ -1,0 +1,38 @@
+name: build and test on x86-64
+on: [push, pull_request]
+jobs:
+  cargo-fmt:
+    runs-on: ubuntu-latest
+    steps:
+      - name: install prerequisites
+        run: |
+          # make sure we don't silently continue
+          set -euo pipefail
+          sudo apt-get update --quiet=2
+          sudo apt-get install --yes meson nasm gcc-multilib
+        env:
+          DEBIAN_FRONTEND: noninteractive
+      - name: git checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+      - name: cache rust toolchain
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.rustup/toolchains
+            ~/.rustup/update-hashes
+            ~/.rustup/settings.toml
+          key: ${{ runner.os }}-rust-toolchain-${{ hashFiles('rust-toolchain', 'rust-toolchain.toml') }}
+      - name: cache rust crates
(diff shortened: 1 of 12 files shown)


randomPoison:
@thedataking I copied one of the existing workflows to make the `cargo fmt` workflow. Please make suggestions if anything could be done better there!

kkysen (commented):
Some of these reformats I already accidentally did in my PRs (I think mainly 63a5b9ec75d3f24702d60180bc2b69523a239ffc in #185, which formats `decode.rs`).  Could we wait to merge those PRs before merging this one, as there's a bunch of PRs stacked on top of that one?

.github/workflows/cargo-fmt.yml
kkysen, line 1: I don't think we need most of this to run just `cargo fmt --check`.

kkysen: We just need the toolchain and `rustfmt` component I believe.

.github/workflows/cargo-fmt.yml
```diff
@@ -0,0 +1,40 @@
...
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      - name: Check formatting
+        run: |
+          rustup component add rustfmt
```
kkysen, line 39: We could add this to `rust-toolchain.toml` instead.  Then it'll always be installed when the toolchain is installed.

.github/workflows/cargo-fmt.yml
```diff
@@ -4,37 +4,10 @@ jobs:
...
         with:
           fetch-depth: 1
-      - name: cache rust toolchain
-        uses: actions/cache@v3
-        with:
-          path: |
-            ~/.rustup/toolchains
-            ~/.rustup/update-hashes
-            ~/.rustup/settings.toml
-          key: ${{ runner.os }}-rust-toolchain-${{ hashFiles('rust-toolchain', 'rust-toolchain.toml') }}
```
kkysen: I think we still need this part, right?

randomPoison: Doesn't seem like it? The workflow run succeeded and ran quickly, so I don't think we need to worry about caching for the fmt workflow.

.github/workflows/cargo-fmt.yml
randomPoison: Done!

.github/workflows/cargo-fmt.yml
kkysen: I guess it doesn't need it, since it'll download them anyways, which doesn't take *that* long, especially compared to the workflows that actually build and test, but if we already have this caching step in other workflows, I think it'd be nice to keep parity between them.  It would speed up formatting (usually only takes a second, not 18-24).

randomPoison: Okay fair enough, I restored the caching step.

PR #197, opened 2023-05-01
`src/env.rs`: Deduplicate (effectively move) inline fns

kkysen:


randomPoison:
Is this change necessary? It seems like those inline functions were only being used in `decode.rs`, so I'm not sure there's any advantage to moving them to `env.rs`.

kkysen:
They are only used in `decode.rs`, but I thought it's useful just to keep the C to Rust source correspondence close where possible.  It also proved a tiny bit useful for some find and replace stuff I could do on the whole file in the next PR without touching the rest of the decode stuff.

randomPoison (approved):

PR #203, opened 2023-05-01

randomPoison:


src/decode.rs
```diff
@@ -3525,280 +3525,231 @@ unsafe fn decode_b(
...
         }
+
         if has_chroma {
             let cfl_allowed = if frame_hdr.segmentation.lossless[b.seg_id as usize] != 0 {
-                (cbw4 == 1 && cbh4 == 1) as libc::c_int
+                cbw4 == 1 && cbh4 == 1
             } else {
-                (cfl_allowed_mask & ((1 as libc::c_int) << bs as libc::c_uint) as libc::c_uint != 0)
-                    as libc::c_int
+                cfl_allowed_mask & (1 << bs) != 0
```
kkysen, line 3542:
```suggestion
                (cfl_allowed_mask & (1 << bs)) != 0
```

For clarity.

src/decode.rs
```diff
@@ -3525,280 +3525,231 @@ unsafe fn decode_b(
...
             };
-            let uvmode_cdf: *mut uint16_t = (ts.cdf.m.uv_mode[cfl_allowed as usize]
-                [b.c2rust_unnamed.c2rust_unnamed.y_mode as usize])
-                .as_mut_ptr();
+
+            let uvmode_cdf = &mut ts.cdf.m.uv_mode[cfl_allowed as usize][b.y_mode() as usize];
             b.c2rust_unnamed.c2rust_unnamed.uv_mode = dav1d_msac_decode_symbol_adapt16(
```
kkysen, line 3546: `uv_mode_mut`?

src/decode.rs
```diff
@@ -3525,280 +3525,231 @@ unsafe fn decode_b(
...
-                    ts.msac.rng,
-                );
+                println!("Post-uvmode[{}]: r={}", b.uv_mode(), ts.msac.rng);
             }
-            b.c2rust_unnamed.c2rust_unnamed.uv_angle = 0 as libc::c_int as int8_t;
-            if b.c2rust_unnamed.c2rust_unnamed.uv_mode as libc::c_int == CFL_PRED as libc::c_int {
-                let sign = (dav1d_msac_decode_symbol_adapt8(
+
+            *b.uv_angle_mut() = 0;
+            if b.uv_mode() as u32 == CFL_PRED {
```
kkysen, line 3557: Should this be `as` whatever enum `CFL_PRED` is part of?

src/decode.rs
```diff
@@ -3525,280 +3525,231 @@ unsafe fn decode_b(
...
-                    b.c2rust_unnamed.c2rust_unnamed.cfl_alpha[0] =
-                        (dav1d_msac_decode_symbol_adapt16(
-                            &mut ts.msac,
-                            (ts.cdf.m.cfl_alpha[ctx as usize]).as_mut_ptr(),
-                            15 as libc::c_int as size_t,
-                        ))
-                        .wrapping_add(1 as libc::c_int as libc::c_uint)
-                            as int8_t;
+                    let ctx = (sign_u == 2) as usize * 3 + sign_v as usize;
+                    b.cfl_alpha_mut()[0] = 1 + dav1d_msac_decode_symbol_adapt16(
```
kkysen, line 3570: Can you keep the C order of the `+ 1` after?

src/decode.rs
```diff
@@ -3525,280 +3525,231 @@ unsafe fn decode_b(
...
-                    b.c2rust_unnamed.c2rust_unnamed.cfl_alpha[1] =
-                        (dav1d_msac_decode_symbol_adapt16(
-                            &mut ts.msac,
-                            (ts.cdf.m.cfl_alpha[ctx_0 as usize]).as_mut_ptr(),
-                            15 as libc::c_int as size_t,
-                        ))
-                        .wrapping_add(1 as libc::c_int as libc::c_uint)
-                            as int8_t;
+                    let ctx = (sign_v == 2) as usize * 3 + sign_u as usize;
+                    b.cfl_alpha_mut()[1] = 1 + dav1d_msac_decode_symbol_adapt16(
```
kkysen, line 3585: Same here.

src/decode.rs
```diff
@@ -3525,280 +3525,231 @@ unsafe fn decode_b(
...
+                        &mut ts.msac,
+                        (ts.cdf.m.cfl_alpha[ctx]).as_mut_ptr(),
+                        15 as libc::c_int as size_t,
```
kkysen, line 3588:
```suggestion
                        15,
```

src/decode.rs
```diff
@@ -3525,280 +3525,231 @@ unsafe fn decode_b(
...
-                let angle_0 = dav1d_msac_decode_symbol_adapt8(
-                    &mut ts.msac,
-                    acdf_0,
-                    6 as libc::c_int as size_t,
-                ) as libc::c_int;
-                b.c2rust_unnamed.c2rust_unnamed.uv_angle = (angle_0 - 3) as int8_t;
+                let acdf = &mut ts.cdf.m.angle_delta[b.uv_mode() as usize - VERT_PRED as usize];
+                let angle = dav1d_msac_decode_symbol_adapt8(&mut ts.msac, acdf.as_mut_ptr(), 6)
+                    as libc::c_int;
+                *b.uv_angle_mut() = (angle - 3) as int8_t;
```
kkysen, line 3613: Above, this was `angle as int8_t - 3`.  And no cast on the line above.

src/decode.rs
```diff
@@ -3525,280 +3525,231 @@ unsafe fn decode_b(
...
+                    ts.cdf.m.pal_y[sz_ctx as usize][pal_ctx].as_mut_ptr(),
+                ) != 0;
+
                 if DEBUG_BLOCK_INFO(f, t) {
-                    printf(
-                        b"Post-y_pal[%d]: r=%d\n\0" as *const u8 as *const libc::c_char,
-                        use_y_pal,
-                        ts.msac.rng,
-                    );
+                    println!("Post-y_pal[{}]: r={}", use_y_pal, ts.msac.rng);
```
kkysen, line 3629: `use_y_pal` is a `bool` now instead of `int`, so it'll print differently here.  That might be okay, but I want sure since it's printed inside of `[]` and I wasn't sure if it's supposed to be an index, in which case casting it to a number is better.

src/decode.rs
```diff
@@ -3525,280 +3525,231 @@ unsafe fn decode_b(
...
+                    ts.cdf.m.pal_uv[pal_ctx as usize].as_mut_ptr(),
+                ) != 0;
+
                 if DEBUG_BLOCK_INFO(f, t) {
-                    printf(
-                        b"Post-uv_pal[%d]: r=%d\n\0" as *const u8 as *const libc::c_char,
-                        use_uv_pal,
-                        ts.msac.rng,
-                    );
+                    println!("Post-uv_pal[{}]: r={}", use_uv_pal, ts.msac.rng);
```
kkysen, line 3645: Same here for `use_uv_pal`.

src/decode.rs
```diff
@@ -3525,280 +3525,231 @@ unsafe fn decode_b(
...
+                    b.y_angle(),
                     ts.msac.rng,
                 );
             }
         }
-        if b.c2rust_unnamed.c2rust_unnamed.pal_sz[0] != 0 {
-            let mut pal_idx: *mut uint8_t = 0 as *mut uint8_t;
+
+        if b.pal_sz()[0] != 0 {
+            let mut pal_idx;
```
kkysen, line 3684: Btw, I'm planning on refactoring `pal_idx` significantly soon.

src/decode.rs
```diff
@@ -3525,280 +3525,231 @@ unsafe fn decode_b(
...
             b.uvtx = dav1d_max_txfm_size_for_bs[bs as usize][f.cur.p.layout as usize];
-            t_dim = &*dav1d_txfm_dimensions
-                .as_ptr()
-                .offset(b.c2rust_unnamed.c2rust_unnamed.tx as isize)
-                as *const TxfmInfo;
-            if frame_hdr.txfm_mode as libc::c_uint
-                == DAV1D_TX_SWITCHABLE as libc::c_int as libc::c_uint
-                && (*t_dim).max as libc::c_int > TX_4X4 as libc::c_int
-            {
+            t_dim = &dav1d_txfm_dimensions[b.tx() as usize];
```
kkysen, line 3734: This is supposed to be a slice I think.  Or just leave it as a raw ptr for now

kkysen: Is it now implicitly converting back to a raw ptr due to the type annotation on `t_dim`?

kkysen: I would make it a slice, though, since it's coming from an array here.

kkysen: Or maybe not?  I need to look closer.  I thought usually this pattern is for a slice.

kkysen: Looking at the C source, I think it's just supposed to be a ref, not a slice.  So nevermind.

kkysen (approved):
LGTM except for the small comments I made.

src/decode.rs
randomPoison: I decided to go the other direction and cast `CFL_PRED` to `u8`, since I think that's the pattern I followed in most other places.

src/decode.rs
randomPoison: I'm not sure what you're referring to? I think the original version of this line was:

```rust
b.c2rust_unnamed.c2rust_unnamed.uv_angle = (angle_0 - 3) as int8_t;
```

src/decode.rs
randomPoison: I changed it deliberately, it doesn't look like `use_y_pal` is being used as an index (it's only used in an `if` block condition), so I figured changing the output here would make things more clear. I can change it back to printing as a number if you think that'd be clearer though.

src/decode.rs
randomPoison: Done

src/decode.rs

src/decode.rs
kkysen: I think it's fine your way, as a `bool`.  I just want to make sure it's deliberate.  It's debug output, and if we need to change it back later for whatever reason, it should be simple.

src/decode.rs
```diff
@@ -3525,280 +3525,233 @@ unsafe fn decode_b(
...
-                (ts.cdf.m.angle_delta[(b.c2rust_unnamed.c2rust_unnamed.y_mode as libc::c_int
-                    - VERT_PRED as libc::c_int) as usize])
-                    .as_mut_ptr();
-            let angle =
-                dav1d_msac_decode_symbol_adapt8(&mut ts.msac, acdf, 6 as libc::c_int as size_t)
-                    as libc::c_int;
-            b.c2rust_unnamed.c2rust_unnamed.y_angle = (angle - 3) as int8_t;
+            let acdf = &mut ts.cdf.m.angle_delta[b.y_mode() as usize - VERT_PRED as usize];
+            let angle = dav1d_msac_decode_symbol_adapt8(&mut ts.msac, acdf.as_mut_ptr(), 6);
+            *b.y_angle_mut() = angle as int8_t - 3;
```
kkysen, line 3533: Ah, I think it's this one that's changed slightly.  I'm not sure of the range of `angle`, but if `angle` ends up being `128`, then `angle as i8 - 3 == -4`, while `(angle - 3) as i8 == 125`.

src/decode.rs
kkysen: See https://github.com/memorysafety/rav1d/pull/203#discussion_r1183195055 actually.  I meant to comment on that one rather than this one (I had them flipped).

src/decode.rs
fbossen: `angle` should be in the range 0 to 5 inclusive

kkysen: In that case, it doesn't really matter for this case, but I've seen similar cases to these break due to reordering arithmetic and casts, so we should generally try to stick closely to the C.

randomPoison: That's a good point, we should probably stick to the original ordering here. Fixed.

randomPoison: Oh nope, that doesn't actually work. In the original logic `angle` is cast to a `c_int`, THEN we do `angle - 3`. When I rewrote it I removed the intermediate cast to `c_int`, and instead cast it directly to `i8`. So what I had was correct, I had just forgot why 😅

kkysen: Oh yeah, I forgot about the C implicit `int` conversions.  It should be all good here now, as `angle` is in `0..=5` anyways.

PR #205, opened 2023-05-01
`fn order_palette`: Cleanup and make mostly safe

kkysen:
The `pal_idx` arg remains a raw ptr and unsafe, and it can be made safe, but since the changes to `pal_idx` propagate significantly throughout the code, I'll save it for a separate PR.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -2044,7 +2044,7 @@ unsafe fn order_palette(
     i: libc::c_int,
     first: libc::c_int,
     last: libc::c_int,
-    order: *mut [uint8_t; 8],
+    order: &mut [[u8; 8]; 64],
     ctx: &mut [u8; 64],
 ) {
     let mut have_top = i > first;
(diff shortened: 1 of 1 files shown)


src/decode.rs
```diff
@@ -2037,151 +2037,83 @@ unsafe extern "C" fn read_pal_uv(
...
-        let have_left = (j > 0) as libc::c_int;
-        if !(have_left != 0 || have_top != 0) {
-            unreachable!();
-        }
-        let mut mask: libc::c_uint = 0 as libc::c_int as libc::c_uint;
+    let mut have_top = i > first;
+
+    assert!(!pal_idx.is_null());
+    pal_idx = pal_idx.offset(first as isize + (i - first) as isize * stride);
+    for ((ctx, order), j) in ctx
```
thedataking, line 2054: I wonder if this could have been written in a way that would make it easier to look at the C code and convince oneself that it does the same as the C code. For instance, do you need to zip everything up? The original C loop iterates over `j` and `n`, whereas you iterate over `ctx` (thus eliminating `n`), `order`, and `j`. 

I'm not overly worried about how you refactor one particular loop but if you generally favor extensive use of functional patterns over an imperative style where the latter is arguably simpler I think it is worth considering.

kkysen: We discussed this in a meeting already, but I also wanted to add that I think this is the most complex iterator example I've written; most of the time I go from

```rust
for i in 0..n {
    let e = &a[i];
}
```

to

```rust
for e in &a[..n] {

}
```

which I've been doing since it's more idiomatic and eliminates bounds checks, which weren't originally present in the C either.  I think in cases like these it's simple enough that it's beneficial, but you're right that as it gets more complex, we have to be increasingly careful.  I'll keep that in mind going forward, rewrite the simple, straightforward ones, and be more wary of more complex changes.

randomPoison (approved):

PR #206, opened 2023-05-03
`fn dav1d_msac_*`: Deduplicate declarations

kkysen:
This also removes `#[no_mangle]`s and `extern "C"`s, which can be done now that their declarations have been deduplicated into `use`s.  It also renames the `*_c` `fn` to `*_rust` now that they're in Rust.

### DIFF
--- a/src/msac.rs
+++ b/src/msac.rs
@@ -145,7 +145,7 @@ unsafe extern "C" fn ctx_norm(s: *mut MsacContext, dif: ec_win, rng: libc::c_uin
     }
 }
 
-pub unsafe extern "C" fn dav1d_msac_decode_bool_equi_c(s: *mut MsacContext) -> libc::c_uint {
+pub unsafe fn dav1d_msac_decode_bool_equi_c(s: *mut MsacContext) -> libc::c_uint {
     let r: libc::c_uint = (*s).rng;
     let mut dif: ec_win = (*s).dif;
     if !(dif >> (::core::mem::size_of::<ec_win>() << 3).wrapping_sub(16) < r as size_t) {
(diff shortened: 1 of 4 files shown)


src/msac.rs
```diff
@@ -282,8 +280,16 @@ pub unsafe extern "C" fn dav1d_msac_decode_symbol_adapt_c(
...
-#[no_mangle]
-pub unsafe extern "C" fn dav1d_msac_decode_bool_adapt_c(
+
+unsafe extern "C" fn dav1d_msac_decode_symbol_adapt_c(
+    s: *mut MsacContext,
+    cdf: *mut uint16_t,
+    n_symbols: size_t,
+) -> libc::c_uint {
+    dav1d_msac_decode_symbol_adapt_rust(s, cdf, n_symbols)
+}
```
randomPoison, line 290: Why do we still need this `_c` version of the function? I would assume this method is only used internally right? So we should just be able to call the `_rust` version directly right?

kkysen: The `_c` version is needed for use in an `unsafe extern "C"` `fn` ptr that roundtrips through asm.

randomPoison: Oh that's fun. I guess we now know for sure that the asm does indeed sometimes call back into C/Rust.

randomPoison (approved):

src/msac.rs
thedataking: Interesting indeed. Might be a good idea to leave a comment lest we forget and muck with the signature.

kkysen: We'll get compile errors from the `fn` ptr `symbol_adapt16` field if we try to change it.

I'm not actually certain the asm calls `.symbol_adapt16`, but it uses other things passed in through `MsacContext`, so I'm guessing it does.

PR #208, opened 2023-05-03
`fn dav1d_msac_*`: Cleanup and make args safe

kkysen:
The major refactor here is making the `cdf: *mut u16` a `&mut [u16]` slice.  All call sites are array refs, so this works nicely (though there are quite a lot of these, ~155).

For `fn dav1d_msac_decode_symbol_adapt_rust`, however, this is trickier, as `cdf` must round-trip through FFI and perhaps an asm call, so slices can't be used.  `cdf.len()` could be added as an extra arg, but that may require changing asm slightly, which I don't want to do at all.  However, `n_symbols` is the last index accessed in `cdf`, so `n_symbols + 1` should be a valid length, possibly smaller.  Thus, an `assert!(n_symbols < cdf.len())` before the `s.symbol_adapt16` call ensures that in `fn dav1d_msac_decode_symbol_adapt_c`, this still holds true, and thus we can safely reconstruct the slice (or a subset) with `unsafe { std::slice::from_raw_parts_mut(cdf, n_symbols + 1) }`.

Since `n_symbols + ` is the length of `cdf` accessed in all of these functions, I could also eliminate the `n_symbols` arg and instead pass it as the `cdf` slice's length, with an array slice at the call sites.  Maybe this is something we should eventually do (or at least an `assert!(n_symbols < cdf.len())`, or `debug_assert!` at least), but for now, I left the `n_symbols` arg to stay closer to the C.  Passing it through the slice would be safer, though, in terms of guarantees when we call asm (the `assert!` would be equally safe, but have a runtime perf cost).

### DIFF
--- a/src/msac.rs
+++ b/src/msac.rs
@@ -231,7 +231,7 @@ pub unsafe fn dav1d_msac_decode_subexp(
 
 unsafe fn dav1d_msac_decode_symbol_adapt_rust(
     s: &mut MsacContext,
-    cdf: *mut uint16_t,
+    cdf: &mut [u16],
     n_symbols: size_t,
 ) -> libc::c_uint {
     let c: libc::c_uint = (s.dif
@@ -245,13 +245,13 @@ unsafe fn dav1d_msac_decode_symbol_adapt_rust(
     if !(n_symbols <= 15) {
         unreachable!();
     }
-    if !(*cdf.offset(n_symbols as isize) as libc::c_int <= 32) {
+    if !(cdf[n_symbols] as libc::c_int <= 32) {
         unreachable!();
     }
     loop {
         val = val.wrapping_add(1);
         u = v;
-        v = r.wrapping_mul((*cdf.offset(val as isize) as libc::c_int >> 6) as libc::c_uint);
+        v = r.wrapping_mul((cdf[val as usize] as libc::c_int >> 6) as libc::c_uint);
         v >>= 7 - 6;
         v = v.wrapping_add(
             (4 as libc::c_int as libc::c_uint)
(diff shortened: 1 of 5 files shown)


src/msac.rs
```diff
@@ -301,38 +296,40 @@ unsafe fn dav1d_msac_decode_symbol_adapt_rust(
...
+    cdf: *mut u16,
     n_symbols: size_t,
 ) -> libc::c_uint {
-    dav1d_msac_decode_symbol_adapt_rust(s, cdf, n_symbols)
+    // # Safety
+    //
+    // This is only called from [`dav1d_msac_decode_symbol_adapt16`],
+    // where there is an `assert!(n_symbols < cdf.len());`.
+    // Thus, `n_symbols + 1` is a valid length for the slice `cdf` came from.
+    #[deny(unsafe_op_in_unsafe_fn)]
```
randomPoison, line 307: Why is this a `#[deny]` attribute? Shouldn't this be an `#[allow]`?

src/msac.rs
```diff
@@ -301,38 +296,40 @@ unsafe fn dav1d_msac_decode_symbol_adapt_rust(
...
-    cdf: *mut uint16_t,
+    cdf: *mut u16,
     n_symbols: size_t,
 ) -> libc::c_uint {
-    dav1d_msac_decode_symbol_adapt_rust(s, cdf, n_symbols)
+    // # Safety
+    //
+    // This is only called from [`dav1d_msac_decode_symbol_adapt16`],
+    // where there is an `assert!(n_symbols < cdf.len());`.
+    // Thus, `n_symbols + 1` is a valid length for the slice `cdf` came from.
```
randomPoison, line 306: Can we also cleanup `dav1d_msac_decode_symbol_adapt16` at the same time so that we don't need this unsafe conversion? Or are you planning to do that in a separate PR?

src/msac.rs
kkysen: Without this, `rustc` says there's an unused `unsafe` in this function since the function itself is `unsafe`.  But the function itself has to stay `unsafe` for the `symbol_adapt16` `fn` ptr.  I want to be clear this is proven safe with an `unsafe` block and `Safety` comment.  `#[deny(unsafe_op_in_unsafe_fn)]` tells `rustc` that `unsafe` operations are not allowed just because a function is `unsafe` and that `unsafe` blocks are still needed.

src/msac.rs
kkysen: The unsafe conversion is necessary here since it roundtrips through asm.  Slices are unstable across FFI boundaries, so an `unsafe` block is required here one way or another.

randomPoison (approved):

PR #210, opened 2023-05-03
Fix logic in assert! statements

fbossen:
Fixes incorrect logic introduced in 0ceca4f89 and d41a06c78 where assert! statements were added.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -2851,9 +2851,7 @@ fn affine_lowest_px(
 ) {
     let h_mul = 4 >> ss_hor;
     let v_mul = 4 >> ss_ver;
-    assert!(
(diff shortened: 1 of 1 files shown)


thedataking:
LGTM! I hit these asserts too earlier today ([commit](https://github.com/memorysafety/rav1d/commit/9eac3abd096f47ea91488e21140868424a3d5f6e)) while working on arm7 assembly. It looks like we've arrived at the same solution.

kkysen:
Sorry for the typos in these; I would've thought they'd be caught in testing.  Any others like these should at least be easily caught by just searching for `assert!(!(` where I accidentally left in the original `!` not.

PR #211, opened 2023-05-03
Fix numerous alignment issues and enable most arm assembly routines

thedataking:
Still need to enable `wiener_filter_neon` and the assembly routines it calls but that can happen in a separate PR.

### DIFF
--- a/.github/workflows/build-and-test-arm7.yml
+++ b/.github/workflows/build-and-test-arm7.yml
@@ -37,8 +37,7 @@ jobs:
       - name: cargo build for armv7-unknown-linux-gnueabihf
         run: |
           rustup target add armv7-unknown-linux-gnueabihf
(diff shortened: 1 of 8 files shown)


randomPoison (approved):

.github/workflows/build-and-test-arm7.yml
```diff
@@ -37,8 +37,7 @@ jobs:
       - name: cargo build for armv7-unknown-linux-gnueabihf
         run: |
           rustup target add armv7-unknown-linux-gnueabihf
-          cargo build --release --target armv7-unknown-linux-gnueabihf \
-            --no-default-features  --features bitdepth_8,bitdepth_16
+          cargo build --release --target armv7-unknown-linux-gnueabihf
```
kkysen, line 40: Do we still have separate tests with asm disabled?  If we don't, we should have one to catch any errors in the no-asm fallback code.

PR #216, opened 2023-05-04

randomPoison:
* Add missing `diridx` argument to `case_set` and friends. The code being updated in this PR is the first time that we've actually needed that argument, so it wasn't added in the initial port of `case_set`.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -3731,13 +3731,11 @@ unsafe fn decode_b(
         }
 
         // update contexts
-
         let y_mode_nofilt = if b.y_mode() == FILTER_PRED as u8 {
             DC_PRED as IntraPredMode
         } else {
             b.y_mode() as IntraPredMode
         };
-
         let mut set_ctx = |dir: &mut BlockContext, diridx, off, mul, rep_macro: SetCtxFn| {
             // NOTE: This corresponds to the following logic in the original C:
             //
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -498,6 +498,15 @@ impl Dav1dFrameContext_bd_fn {
     ) -> libc::c_int {
         self.recon_b_inter.expect("non-null function pointer")(context, block_size, block)
     }
+
+    pub unsafe fn read_coef_blocks(
+        &self,
+        context: *mut Dav1dTaskContext,
+        block_size: BlockSize,
+        block: *const Av1Block,
+    ) {
+        self.read_coef_blocks.expect("non-null function pointer")(context, block_size, block);
+    }
 }
 
 pub type read_coef_blocks_fn =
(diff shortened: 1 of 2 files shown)


src/decode.rs
```diff
@@ -3682,1522 +3693,119 @@ unsafe fn decode_b(
...
+            //
+            // ```
+            // ((uint8_t *) &t_dim->lw)[diridx]
+            // ```
+            //
+            // The original logic casts a pointer to the `lw` field to a `*const u8` and
+            // then does indexing within the struct's fields. This is deeply cursed and
+            // unnecessary since `diridx` will only ever be 0 or 1. So for the Rust version
+            // we manually select which field we want based on `diridx` directly.
+            let lw_lh = match diridx {
```
kkysen, line 3752: This is indeed quite cursed.  I'm surprised they didn't at least use a `union` with an array, which they did for the same thing in other places.

Maybe just `l` is a better name than `lw_lh`?  Though `lw_hw` is also clear enough I think.

src/decode.rs
```diff
@@ -3682,1522 +3693,119 @@ unsafe fn decode_b(
...
+            rep_macro(dir.skip.as_mut_ptr(), off, mul * b.skip as u64);
+
+            // see aomedia bug 2183 for why we use luma coordinates here
+            rep_macro(
+                t.pal_sz_uv[diridx].as_mut_ptr(),
+                off,
+                mul * if has_chroma { b.pal_sz()[1] as u64 } else { 0 },
+            );
+
+            if is_inter_or_switch(frame_hdr) {
```
kkysen, line 3778:
```suggestion
            rep_macro(dir.skip.as_mut_ptr(), off, mul * b.skip as u64);
            // see aomedia bug 2183 for why we use luma coordinates here
            rep_macro(
                t.pal_sz_uv[diridx].as_mut_ptr(),
                off,
                mul * if has_chroma { b.pal_sz()[1] as u64 } else { 0 },
            );
            if is_inter_or_switch(frame_hdr) {
```

I've been trying to keep the extra line breaks the same as the C, as they may convey some semantic meaning.

kkysen (approved):
LGTM.  Just some minor comments about preserving spacing.

Also, let me know when would be a good time to try to make `case_set` safe (I have an idea how to make it entirely safe, aligned writes an all).  Maybe once all the `case_set`s are deduplicated.

Also, could you make the PR titles a bit more descriptive, as they're all the same right now so it's hard to tell them apart at first glance?

randomPoison:
Yeah, good point about all the PR names being the same. I'm not sure how to better name them though since each one is just touching an arbitrary chunk of code within a single function. Maybe I'll just start numbering them 🤷

kkysen:
That should work, numbering, or if there's something standout in the changes, like `diridx` for this one.

PR #217, opened 2023-05-04
decode_b: Cleanup chunk of code

randomPoison:


src/decode.rs
```diff
@@ -3806,1159 +3806,204 @@ unsafe fn decode_b(
...
-                );
-                y_1 += 1;
+
+            for x in 0..bw4 {
+                t.al_pal[0][(bx4 + x) as usize][0] = *pal;
+            }
+
+            for y in 0..bh4 {
+                t.al_pal[1][(by4 + y) as usize][0] = *pal;
             }
```
kkysen, line 3824: We should eventually turn these into a slice + iterator to avoid introducing bounds checks.

src/decode.rs
```diff
@@ -3806,1159 +3806,204 @@ unsafe fn decode_b(
...
-                            ((*t).al_pal[0][(bx4 + x_2) as usize][pl as usize]).as_mut_ptr()
-                                as *mut libc::c_void,
-                            (*pal_0.offset(pl as isize)).as_ptr() as *const libc::c_void,
-                            16 as libc::c_int as libc::c_ulong,
-                        );
-                        x_2 += 1;
+                    &t.scratch.c2rust_unnamed_0.pal
+                };
+
+                for pl in 0..2 {
```
kkysen, line 3843:
```suggestion
                for pl in 1..=2 {
```

This comes from this:

```rust
                let mut pl = 1;
                while pl <= 2 {
```

src/decode.rs
```diff
@@ -3806,1159 +3806,204 @@ unsafe fn decode_b(
...
-                        memcpy(
-                            ((*t).al_pal[1][(by4 + y_2) as usize][pl as usize]).as_mut_ptr()
-                                as *mut libc::c_void,
-                            (*pal_0.offset(pl as isize)).as_ptr() as *const libc::c_void,
-                            16 as libc::c_int as libc::c_ulong,
-                        );
-                        y_2 += 1;
+
+                    for y in 0..bh4 {
+                        (*t).al_pal[1][(by4 + y) as usize][pl] = pal[pl];
```
kkysen, line 3849:
```suggestion
                        t.al_pal[1][(by4 + y) as usize][pl] = pal[pl];
```

`t` is used in the above `x` loop, not `(*t)`.

src/decode.rs
```diff
@@ -3806,1159 +3806,204 @@ unsafe fn decode_b(
...
-                .offset(0),
-            &mut ts.cdf.dmv,
-            0 as libc::c_int,
-        );
+            b.mv_mut()[0].y = -(512 << (*f.seq_hdr).sb128) as int16_t;
+            b.mv_mut()[0].x = 0;
+        }
+
+        let r#ref: mv = b.mv()[0];
+        read_mv_residual(t, &mut b.mv_mut()[0], &mut ts.cdf.dmv, 0 as libc::c_int);
```
kkysen, line 3888:
```suggestion
        let r#ref = b.mv()[0];
        read_mv_residual(t, &mut b.mv_mut()[0], &mut ts.cdf.dmv, 0);
```

src/decode.rs
```diff
@@ -3806,1159 +3806,204 @@ unsafe fn decode_b(
...
+                off,
+                mul * b_dim[2 + diridx] as u64,
+            );
+            rep_macro(dir.mode.as_mut_ptr(), off, mul * DC_PRED as u64);
+            rep_macro(dir.pal_sz.as_mut_ptr(), off, 0);
+
+            // see aomedia bug 2183 for why this is outside `if has_chroma {}`
+            let pal_sz_uv: &mut [u8; 32] = &mut t.pal_sz_uv[diridx];
+            rep_macro(pal_sz_uv.as_mut_ptr(), off, 0);
+
```
kkysen, line 3992:
```suggestion
        let mut set_ctx = |dir: &mut BlockContext, diridx: usize, off, mul, rep_macro: SetCtxFn| {
            rep_macro(
                dir.tx_intra.as_mut_ptr() as *mut u8,
                off,
                mul * b_dim[2 + diridx] as u64,
            );
            rep_macro(dir.mode.as_mut_ptr(), off, mul * DC_PRED as u64);
            rep_macro(dir.pal_sz.as_mut_ptr(), off, 0);
            // see aomedia bug 2183 for why this is outside `if has_chroma {}`
            rep_macro(t.pal_sz_uv[diridx].as_mut_ptr(), off, 0);
```

The type wasn't inferred correctly for `pal_sz_uv` because `diridx` was already inferred.  Specifying that, `pal_sz_uv` is inferred correctly.

kkysen (changes requested):
The `1..=2` suggested change above is a bug, so we need to fix that.  Other than that, I had some other small comments.

One thing is that I've generally been converting all simple indexing loops into a slice + iterator to preserve having no bounds checking (I commented on the first one I saw here).  This isn't that important now as it's still totally correct with indexing, but since it's a minor change, I thought I might as well just do them as I come across them.  When they'd need a more complex rewrite than just a slice and iterator, then I think we can leave for them for now.  What do you think?

src/decode.rs
randomPoison: Oh yup, you're absolutely right.

src/decode.rs
randomPoison: I gave it a try just now, but it felt less readable to me so for now let's keep it closer to the C version.

randomPoison:
I think the loop conversions you've been doing have been good so far 👍 I'm trying to keep an eye out for opportunities to do so, but I think I've been leaning towards keeping things closer to the C in general. For the couple in this PR I think they're a bit more readable if kept closer to the C style loop, but in general I'm not opposed to converting things to be more Rust-style.

kkysen (approved):
Okay, that's fine.  The other things LGTM now.

PR #218, opened 2023-05-04
decode_b: Cleanup a chunk of logic

randomPoison:


src/refmvs.rs
```diff
@@ -79,6 +79,20 @@ pub struct refmvs_refpair {
...
+            r#ref: [ref_0, ref_1],
+        }
+    }
+}
+
+impl From<[i8; 2]> for refmvs_refpair {
+    fn from(from: [i8; 2]) -> Self {
+        refmvs_refpair { r#ref: from }
+    }
+}
```
kkysen, line 94: Do we need both of these?  The second one seems good and sufficient.

src/levels.rs
```diff
@@ -190,7 +190,7 @@ pub const INTER_INTRA_NONE: InterIntraType = 0;
 /// which includes `&self` methods, including autogenerated ones like [`PartialEq::eq`].
 ///
 /// [`refmvs_block`]: crate::src::refmvs::refmvs_block
-#[derive(Clone, Copy, PartialEq, Eq, Default)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
```
kkysen, line 193: The previous format string was `y=%d,x=%d`, so the `Debug` (or `Display`) `impl` should at least preserve that.  That said, I don't think we should use a `Debug` `impl` for this at all, as `mv` is commonly formatted in other ways elsewhere, such as

```rust
            println!(
                "Post-dmv[{}/{},ref={}/{}|{}/{}]: r={}",
                b.mv()[0].y,
                b.mv()[0].x,
                r#ref.y,
                r#ref.x,
                mvstack[0].mv.mv[0].y,
                mvstack[0].mv.mv[0].x,
                ts.msac.rng,
            );
```

in #217.

kkysen (changes requested):
See the above comments.

src/refmvs.rs
kkysen: If we want the `new` method, I think it should at least take a single `[i8; 2]` argument.

src/levels.rs
randomPoison: Yeah that's a fair point. I'll remove the `Debug` impl and revert the `println!` where I was using it to be the same as in the C version.

PR #219, opened 2023-05-04
Fix unused compiler warnings

### LINKED ISSUE #13: Cleanup warnings
The transpiled Rust generates a lot of warnings currently. We should do a cleanup pass and determine if any of the warnings indicate actual problems, fix all the actual issues, and suppress any remaining warnings. Once all warnings have been removed, we should update CI to deny all warnings (tracked separately in #220).

fbossen:
Resolves #13

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,12 +1,15 @@
+#[cfg(feature = "asm")]
 use core::panic;
+#[cfg(feature = "asm")]
 use std::env;
+#[cfg(feature = "asm")]
 use std::path::Path;
 
 fn main() {
-    let arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
-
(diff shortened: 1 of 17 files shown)


kkysen (approved):
The changes LGTM, but I'm not sure it fully resolves #13 since that also includes updating CI to deny warnings.

I also changed the PR title since I think "silence" implies just turning the warnings off (`allow`ing them), but this properly fixes the remaining ones, which seem to be all from unused items.

thedataking:
@kkysen I suggest we mark #13 as closed and track denying warnings in another, low-priority issue?  I don't think denying warnings in CI is likely to help much although its a fine thing to do generally. At this point, we'll catch many more issues by running the Argon tests in CI, getting test coverage for AVX512, etc.

PR #222, opened 2023-05-05
`struct MsacContext`, `fn ctx_{refill,norm}`: Make safe

kkysen:
This adds a safe API around `MsacContext::buf_{pos,end}` and uses it to make `fn ctx_{refill,norm}` safe.

Note that the lifetime requirements for `buf_{pos,end}` are moved to `fn dav1d_msac_init`'s `data` arg's `Safety` comment.  Later, we'll add a proper lifetime here, but that's much more involved as it propagates through a lot of function calls and types.

### DIFF
--- a/src/msac.rs
+++ b/src/msac.rs
@@ -5,6 +5,7 @@ use crate::include::stddef::*;
 use crate::include::stdint::*;
 use cfg_if::cfg_if;
 use std::mem;
+use std::ops::Range;
 
 #[cfg(all(feature = "asm", target_arch = "x86_64"))]
 extern "C" {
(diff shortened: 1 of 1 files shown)


src/msac.rs
```diff
@@ -66,8 +67,8 @@ pub type ec_win = size_t;
 #[derive(Copy, Clone)]
 #[repr(C)]
 pub struct MsacContext {
-    pub buf_pos: *const uint8_t,
-    pub buf_end: *const uint8_t,
+    buf_pos: *const uint8_t,
+    buf_end: *const uint8_t,
```
randomPoison, line 71: Why store this as a start/end pointer combo? Why not store the length of the slice instead, since that's ultimately what you need to reconstruct it.

kkysen: Asm uses `MsacContext` so I can't change it at all.  Otherwise I'd definitely do something like this.

randomPoison (approved):

PR #226, opened 2023-05-05
Run Argon tests nightly (Pacific Time)

### LINKED ISSUE #214: Test rav1d against Argon test vectors
- Argon test vectors [download link](https://storage.googleapis.com/downloads.aomedia.org/assets/zip/argon_coveragetool_av1_base_and_extended_profiles_v2.1.1.zip). More info [here](https://aomedia.org/av1-video-decoder-verification-tool/).
- This takes a long time so we might want to schedule this as an overnight job.
- Run this for x86-64 only or for other targets as well?
- Blocked on #213
- dav1d contains a helpful script `tests/dav1d_argon.bash` that we might want to cherry pick into this repo.

thedataking:
Also triggers when pushing any branch with `argon` in the name. Note:
- limited to x86 release builds with/without assembly plus overflow checks due to the high number of tests. 
- adds `dav1d_argon.bash` script using commits from upstream.
- we might want to add notification on failures at some point since regular PRs will not be subject to these tests.

Closes #214.


kkysen:
Does the time that it runs matter?  Do we want it to run at night when we're not working?  If that's the case, maybe early morning would be better since we tend to work pretty late.

thedataking:
@kkysen time doesn't matter; I've updated it. If it looks good to you otherwise, feel free to merge it in – we can always further adjust when they run.

kkysen (approved):

.github/workflows/build-and-test-x86-extra.yml
```diff
@@ -0,0 +1,77 @@
...
+        id: cache-argon
+        uses: actions/cache@v3
+        with:
+          path: |
+            tests/argon
+          key: argon_coveragetool_av1_base_and_extended_profiles_v2.1.1.zip
+      - name: cargo build for ${{ matrix.target }} ${{ matrix.build.name }}
+        run: |
+          cargo clean
+          rustup target add ${{ matrix.target }} --toolchain nightly-2022-08-08
```
kkysen, line 61: I don't think it's worth fixing right now in this PR since it's already done elsewhere, but I don't think we need to hardcode the toolchain version here.  It's specified in our `rust-toolchain.toml` and we'd like to keep that as the single source of truth so nothing gets out of sync.

thedataking: Yeah, I agree. I removed them and it looks like we still get the desired toolchain (m worry was that we'd pull down the stable toolchain for no reason) https://github.com/memorysafety/rav1d/pull/236

PR #227, opened 2023-05-05
`static dav1d_cpu_flags{,_mask}`: Cleanup and make safe

kkysen:
I also cleaned up and made safe most of the rest of `src/cpu.rs`.


kkysen:
`arm7` CI keeps failing, but I'm not sure why, as I can't reproduce it locally.  All tests pass locally with

```sh
cargo b -r --target armv7-unknown-linux-gnueabihf && (meson setup build --reconfigure -Dtest_rust_path=../target/armv7-unknown-linux-gnueabihf/release/dav1d && cd build && meson test --no-rebuild)
```

thedataking:
After looking at the C code and the way `dav1d_cpu_flags` and `dav1d_cpu_flags_mask` is accessed, there does not appear to be a justification to switch to using atomic types here.

randomPoison:
@thedataking I think we need atomics because the flags are stored in a `static` variable. If we don't use atomics it would need to be `static mut` and unsafe to access. So I think the switch to using atomics here is the smart move.

thedataking:
> If we don't use atomics it would need to be static mut and unsafe to access.

I assume you mean unsafe to access from multiple threads. Single threaded access should be perfectly fine. Pressumably that why the domain experts (the dav1d developers) chose not to synchronize access to these variables. 

> I think the switch to using atomics here is the smart move.

I understand that Rust generally discourages static mutable variables but I don't think we should spend time and energy to convert the `dav1d` API which isn't thread safe to one that is. We want to remain drop-in compatible so we can assume the calling application uses the API as intended. Concerns if we go further down this path:

1) it introduces a additional complexity. It could make it harder to backport additional threading fixes. The first backport was made easier since C and Rust code did not diverge w.r.t. threading.
2) it adds performance overheads which we can't even detect since they may only show up on highly-threaded use cases. 
3) the goal of the `rav1d` port is to prevent untrusted inputs from corrupting the decoder and the hosting application; what is being suggested here is trying to prevent bugs from trusted developers using the `dav1d` API incorrectly. That's a productivity win at best; not a security win. Why take on additional work given that we're pressed for time as is?
4) the assembly code is inherently unsafe so there is only so much migrating to Rust buys us; achieving absolute safety is not an option so I don't think that – on balance – the pros outweighs the cons.

thedataking:
`dav1d_obmc_masks` needed to be 16 bit aligned. I pushed a fix. Since it seems your PR was intended to be a non-functional change (modulo the use of atomics), it remains an open question what surfaced the alignment issue.

> arm7 CI keeps failing, but I'm not sure why, as I can't reproduce it locally.

QEMU 7.0.0 is needed to reproduce; you are most likely running 6.2.0 locally. I just built it from sources; it was straightforward. Then you can run something like:

```
$ ~/bin/qemu-arm -g 9000  ./target/armv7-unknown-linux-gnueabihf/release/dav1d -i tests/dav1d-test-data/8-bit/data/00000001.ivf --verify 98b8c18a74a27a4fc1436f08b73b9270
```

The `-g 9000` argument tells QEMU to wait for a connection from gdb. In another terminal you can can run the following commands

```
$ gdb-multicarch
(gdb) target remote localhost:9000
(gdb) file target/armv7-unknown-linux-gnueabihf/release/dav1d
```

after which point you can continue execution with `c` and see the crash location with `disas`, `bt`, etc.

thedataking (approved):
LGTM under two assumptions:
- The `dav1d` developers possibly did not use atomic types since they don't want to require C11. If so, that shouldn't stop us.
- Additional divergence from `dav1d` w.r.t. threading/synchronization will need to be reviewed *very* carefully. This is an area where testing is hard and messing up is too easy.

kkysen:
> LGTM under two assumptions:
> 
> * The `dav1d` developers possibly did not use atomic types since they don't want to require C11. If so, that shouldn't stop us.

That's a good point, and definitely seems possible.

> * Additional divergence from `dav1d` w.r.t. threading/synchronization will need to be reviewed _very_ carefully. This is an area where testing is hard and messing up is too easy.

I agree.  We'll need to be very careful around these areas.  Hopefully #232 and #233 at least get rid of the last alignment issues and so we won't keep running into them.  That said, we *also* need to be very careful about `dav1d`'s potential data races.  We're rewriting it in Rust for a reason, and so we can't just assume the C code didn't have data races.  For the specific case in this PR, it's pretty simple and also doesn't seem to be on any hot paths, but I assume there are more complex cases we'll come across as well.

kkysen:
> `dav1d_obmc_masks` needed to be 16 bit aligned. I pushed a fix.

Thanks for tracking it down!

I decided to fix #122 fully with #232 and #233 now, so hopefully this won't be an issue in the future anymore either.  I also cherry-picked your fix for `dav1d_obmc_masks` in #232 as I came across it.

> Since it seems your PR was intended to be a non-functional change (modulo the use of atomics), it remains an open question what surfaced the alignment issue.

I'm just guessing here, but somehow the atomics usage surfaced this bug, which to me seems like a good thing, and if it continues to do that with other subtle bugs, that would be great!

> > arm7 CI keeps failing, but I'm not sure why, as I can't reproduce it locally.
> 
> QEMU 7.0.0 is needed to reproduce; you are most likely running 6.2.0 locally. I just built it from sources; it was straightforward. Then you can run something like:
> 
> ```
> $ ~/bin/qemu-arm -g 9000  ./target/armv7-unknown-linux-gnueabihf/release/dav1d -i tests/dav1d-test-data/8-bit/data/00000001.ivf --verify 98b8c18a74a27a4fc1436f08b73b9270
> ```
> 
> The `-g 9000` argument tells QEMU to wait for a connection from gdb. In another terminal you can can run the following commands
> 
> ```
> $ gdb-multiarch
> (gdb) target remote localhost:9000
> (gdb) file target/armv7-unknown-linux-gnueabihf/release/dav1d
> ```
> 
> after which point you can continue execution with `c` and see the crash location with `disas`, `bt`, etc.

Thanks for explaining how to debug these things on `qemu`!  I'll keep it in mind for next time.

thedataking:
> We're rewriting it in Rust for a reason, and so we can't just assume the C code didn't have data races. 

The main reason we are doing a Rust port is because `dav1d` has to parse untrusted untrusted data which can trigger memory corruption (see point 3 in my earlier [comment](https://github.com/memorysafety/rav1d/pull/227#issuecomment-1537208019)). We did not offer to refactor/change the way threading is implemented under the current contract; please treat it as out of scope for now and focus on the memory safety wins.

> For the specific case in this PR, it's pretty simple 

It might have been simple to propose the change but reviewing isn't simple. I first checked the ARM memory model which is weaker than the x86 memory model which guarantees that naturally aligned loads and stores up to 64 bits are atomic. ARM does the same in a singleprocessing system but not in a multiprocessing system [as far as I can tell](https://developer.arm.com/documentation/ddi0406/cb/Application-Level-Architecture/Application-Level-Memory-Model/Memory-types-and-attributes-and-the-memory-order-model/Atomicity-in-the-ARM-architecture?lang=en#CHDICIIH).

Second, we now have to reason about how they are used throughout the program. This is why I suggested we stick to the way `dav1d` handles these variables and focus on memory safety wins. Consider these function calls (I've just looked at the x86/8-bit case and _hope_ all other cases are analogous):

```Rust
// `dav1d_parse_obus` calls `dav1d_submit_frame` calls
dav1d_cdef_dsp_init_8bpc(&mut (*dsp).cdef); // calls `cdef_dsp_init_x86` calls `dav1d_get_cpu_flags`
dav1d_intra_pred_dsp_init_8bpc(&mut (*dsp).ipred); // calls `intra_pred_dsp_init_x86` calls `dav1d_get_cpu_flags`
dav1d_itx_dsp_init_8bpc(&mut (*dsp).itx, bpc); // calls `dav1d_itx_dsp_init_8bpc`  calls `itx_dsp_init_x86`
dav1d_loop_filter_dsp_init_8bpc(&mut (*dsp).lf); // ... 
dav1d_loop_restoration_dsp_init_8bpc(&mut (*dsp).lr, bpc);
dav1d_mc_dsp_init_8bpc(&mut (*dsp).mc);
dav1d_film_grain_dsp_init_8bpc(&mut (*dsp).fg);
```

What happens if a second thread thread calls `dav1d_set_cpu_flags_mask` while a first thread is anywhere in the above code sequence? Assuming that `dsp` is not shared among threads, the worst that can happen is that `dsp` may be inconsistently initialized because different calls to `*_dsp_init_8bpc` saw different values of `dav1d_cpu_flags_mask`. Assuming `dav1d_set_cpu_flags_mask` is much less frequently than `dav1d_submit_frame`, `dsp` should eventually reach a consistent state. I'm relying on a lot of assumptions here, perhaps some of which are wrong... 🤷 

I've only discussed `dav1d_cpu_flags_mask` so far. What about `dav1d_cpu_flags`? It is only ever set in `dav1d_init_cpu` which is called from `init_internal` which in turn is called from `pthread_once` (at the beginning of `dav1d_open`) which ensures `dav1d_cpu_flags` is set in a single-threaded context. As far as I can tell, it happens before other threads are created with calls to `pthread_create` (later in `dav1d_open`); so on second thought I still don't see a reason to use atomics for `dav1d_cpu_flags`.

thedataking (changes requested):
Upon further review, it appears that `dav1d_cpu_flags` is only every written to once per-process (thanks to `pthread_once`) before any worker threads are created. I'd prefer it we could put a note on this variable to explain why all concurrent accesses are reads (which is always safe) instead of changing its type to something that indicates it may be read and written to concurrently.

randomPoison:
@thedataking I'm still inclined to say that using atomics here is the correct thing to do. The use of atomics here doesn't really have anything to do with thread safety imo, rather it's to avoid making `dav1d_cpu_flags` and `dav1d_cpu_flags_mask` mutable static variables, which require an `unsafe` block to access. We could use [lazy_static](https://crates.io/crates/lazy_static) to achieve the same result without atomics, but this case is simple enough that I think the use of atomics is appropriate. Your concerns about deviating from the original C are noted, but in this case I think the changes are minor and warranted in order to keep our Rust code safe where possible.

randomPoison (approved):

randomPoison:
@kkysen I discussed this with @thedataking. He's onboard with using atomics here to avoid using a `static mut` variable, but could you at a comment to `dav1d_cpu_flags`/`dav1d_cpu_flags_mask` to note why they're atomic in Rust but not in the original C? Once that's added this PR is good to merge.

kkysen:
> @kkysen I discussed this with @thedataking. He's onboard with using atomics here to avoid using a `static mut` variable, but could you at a comment to `dav1d_cpu_flags`/`dav1d_cpu_flags_mask` to note why they're atomic in Rust but not in the original C? Once that's added this PR is good to merge.

I've added docs now explaining why atomics were used for the masks in 17ebca706cac09ba45d7e32436e94bec48d674a6.

kkysen:
> The main reason we are doing a Rust port is because `dav1d` has to parse untrusted untrusted data which can trigger memory corruption (see point 3 in my earlier [comment](https://github.com/memorysafety/rav1d/pull/227#issuecomment-1537208019)). We did not offer to refactor/change the way threading is implemented under the current contract; please treat it as out of scope for now and focus on the memory safety wins.

Thanks for explaining that more clearly.  I was previously treating memory and data race safety as more intertwined (they can affect each other, though not often I believe?), so I think that's where some of the confusion came from.  I can focus more specifically on memory safety going forward.

Issue #225, opened 2023-05-05
Backport performance optimizations from david 1.2.0


thedataking:
**NOTE**: this issue does not track work to backport correctness fixes from upstream dav1d. That's tracked separately in #224.

It seems several improvements have been made to upstream david in releases 1.1.0 and 1.2.0 which add or improve the assembly routines such as https://code.videolan.org/videolan/dav1d/-/commit/9d64591854764dfb76459f3f0455305155e80e7f. We should port these over to ensure performance parity with upstream.

PR #228, opened 2023-05-05
Backport x86 asm changes from dav1d 1.2.0

fbossen:
Partially addresses #225


kkysen (changes requested):
We also need to re-transpile the C changes to Rust.  @thedataking, I assume you just did this recently for #215, so maybe you have some advice for doing it easily?  They're also simple enough to do manually (the larger `tests/checkasm/*` change doesn't need to be since we didn't transpile that in the first place), but we need to be careful doing it that way.

fbossen:
I will do the port to Rust manually. As you say, the changes look simple enough

fbossen:
Closing this PR. Will resubmit separate PRs as more C code actually needs to be pulled in.

thedataking:
> I assume you just did this recently for https://github.com/memorysafety/rav1d/pull/215, so maybe you have some advice for doing it easily?

I simply re-transpiled the affected files using the `--filter` flag and made sure I only patched in the Rust lines corresponding to the changes in C files. Careful review will hopefully catch any inadvertent changes to the Rust.

PR #232, opened 2023-05-08
Fix missing aligns

### LINKED ISSUE #122: Audit alignment directives

kkysen:
Fixes #122.

This aligns all remaining statics, fields, and variables that use `dav1d`'s C `ALIGN` macro to specify their alignment.

### DIFF
--- a/src/align.rs
+++ b/src/align.rs
@@ -8,122 +8,38 @@
 
 use std::ops::{Index, IndexMut};
 
-#[derive(Copy, Clone)]
-#[repr(C, align(64))]
-pub struct Align64<T>(pub T);
+macro_rules! def_align {
+    ($align:literal, $name:ident) => {
+        #[derive(Copy, Clone)]
+        #[repr(C, align($align))]
+        pub struct $name<T>(pub T);
 
-impl<T> From<T> for Align64<T> {
-    fn from(from: T) -> Self {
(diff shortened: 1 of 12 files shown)


thedataking (approved):
Thanks for going over the alignment directives, this will surely save us time! LGTM.

(I didn't review the changes to `align.rs` since I think that got looked over in #127.)

randomPoison (approved):
A minor thought with the `AlignX` stuff: We could avoid adding a lot of `.0`s to the code by adding `as_ptr()` and `as_mut_ptr()` methods (I added the `Index` impls for the same reason). I'm not gonna require you do that though since you've already added all the necessary `.0`s 😅

kkysen:
I was thinking about that as well as I was in the middle of updating all the `.0`s, but I thought there still is use in making it explicit.  One of my ideas for making `case_set` would rely on knowing the alignment of the buffer passed to it, so in that case, explicitness may help.

PR #233, opened 2023-05-08
`struct refmvs_block`: Align a wrapper struct for an inner packed struct

kkysen:
Align by aligning a wrapper struct around an inner packed struct, as Rust doesn't allow packed and aligned types (see https://github.com/rust-lang/rust/issues/59154).

### DIFF
--- a/src/refmvs.rs
+++ b/src/refmvs.rs
@@ -87,12 +87,16 @@ pub struct refmvs_mvpair {
 
 #[derive(Copy, Clone)]
 #[repr(C, packed)]
-pub struct refmvs_block {
+pub struct refmvs_block_unaligned {
     pub mv: refmvs_mvpair,
     pub r#ref: refmvs_refpair,
     pub bs: uint8_t,
     pub mf: uint8_t,
 }
+
+#[derive(Copy, Clone)]
+#[repr(C, align(4))]
+pub struct refmvs_block(pub refmvs_block_unaligned);
 
 #[derive(Copy, Clone)]
 #[repr(C)]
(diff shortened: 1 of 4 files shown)


kkysen:
@ahomescu, @rinon, I saw that you had opened the Rust issue for this, https://github.com/rust-lang/rust/issues/59154, and commented on it, so I wanted to ask, does this (the aligned wrapper struct) seem like a valid way to get around it for this case?

src/refmvs.rs
```diff
@@ -87,13 +87,17 @@ pub struct refmvs_mvpair {
 
 #[derive(Copy, Clone)]
 #[repr(C, packed)]
-pub struct refmvs_block {
+pub struct refmvs_block_unaligned {
```
randomPoison, line 90: Could we get a comment on here noting why we need this double-struct setup in Rust? Would be good for future reference to note why we're deviating from the C here.

randomPoison (approved):
Just one comment requesting a comment, feel free to merge after you add that.

src/refmvs.rs
kkysen: Fixed in da6781d0a95e1cc2b35a5c26aea3da80a9566800.

src/refmvs.rs
```diff
@@ -85,15 +85,28 @@ pub struct refmvs_mvpair {
     pub mv: [mv; 2],
 }
 
+/// For why this unaligned, see the aligned [`refmvs_block`] below.
 #[derive(Copy, Clone)]
 #[repr(C, packed)]
-pub struct refmvs_block {
+pub struct refmvs_block_unaligned {
```
ahomescu, line 91: Does the definition need to be `pub`? It could be `pub(crate)`.

src/refmvs.rs
```diff
@@ -85,15 +85,28 @@ pub struct refmvs_mvpair {
...
+/// but this (aligned types within a packed type) is not yet allowed in Rust
+/// (see [rust-lang/rust#59154](https://github.com/rust-lang/rust/issues/59154)),
+/// as different C compilers treat this differently.
+/// To get around this limitation, we split `struct refmvs_block`
+/// into an inner packed [`refmvs_block_unaligned`]
+/// and an outer aligned [`refmvs_block`]
+/// that is just a wrapper over the real [`refmvs_block_unaligned`].
+#[derive(Copy, Clone)]
+#[repr(C, align(4))]
+pub struct refmvs_block(pub refmvs_block_unaligned);
```
ahomescu, line 108: Does the inner value need `pub`?

ahomescu (approved):

src/refmvs.rs
kkysen: It could be `pub(crate)`, but basically everything is still `pub` at this point whether it needs to be or not, so I haven't been fixing those at all.

src/refmvs.rs
kkysen: It needs to at least be `pub(crate)`, but like https://github.com/memorysafety/rav1d/pull/233#discussion_r1187959647, we haven't been cleaning up `pub`s yet, and probably won't be for a little while.

PR #240, opened 2023-05-08
decode_b: Deduplicate assign_comp_mv

randomPoison:
Recreate the inline macro `assign_comp_mv` and deduplicate the two expansions of it in the Rust code.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4185,22 +4185,18 @@ unsafe fn decode_b(
             assert!(b.drl_idx() >= NEAREST_DRL as u8 && b.drl_idx() <= NEARISH_DRL as u8);
 
             has_subpel_filter = imin(bw4, bh4) == 1 || b.inter_mode() != GLOBALMV_GLOBALMV as u8;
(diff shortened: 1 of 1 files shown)


src/decode.rs
```diff
@@ -4185,104 +4185,40 @@ unsafe fn decode_b(
             assert!(b.drl_idx() >= NEAREST_DRL as u8 && b.drl_idx() <= NEARISH_DRL as u8);
 
             has_subpel_filter = imin(bw4, bh4) == 1 || b.inter_mode() != GLOBALMV_GLOBALMV as u8;
```
kkysen, line 4187: In C, this line is after the macro and right before `assign_comp_mv(0);`.  Could you move it there?

kkysen (approved):
LGTM once that one line is moved after the macro/closure definition like it is in C.

src/decode.rs
kkysen: Oh, I missed that it uses it in the closure.  In that case, this is the right way.

PR #247, opened 2023-05-10
`fn DEBUG_BLOCK_INFO`: Deduplicate and cleanup

kkysen:
Since `fn DEBUG_BLOCK_INFO` depends on types (`Dav1dFrameContext`, `Dav1dTaskContext`) that haven't been deduplicated/genericized over bitdepth yet, we can't have a single definition for it, even though it only uses fields that are not bitdepth-dependent.  Thus, I defined a macro that defines the `fn` in each module that uses it.  Once we do bitdepth genericization or split out the non-bitdepth-dependent parts of these types, we can consolidate the `fn`-defining macro into just a normal `fn`.

Also, making some other functions safe requires `DEBUG_BLOCK_INFO` to take refs instead of raw ptrs, which is why I decided to fully deduplicate and clean this up now.

### DIFF
--- a/src/recon.rs
+++ b/src/recon.rs
@@ -1,6 +1,29 @@
 use crate::src::msac::dav1d_msac_decode_bool_equi;
 use crate::src::msac::MsacContext;
 
+/// This is a macro that defines a function
+/// because it takes `Dav1dFrameContext` and `Dav1dTaskContext` as arguments,
+/// which have not yet been deduplicated/genericized over bitdepth.
+macro_rules! define_DEBUG_BLOCK_INFO {
+    () => {
+        unsafe fn DEBUG_BLOCK_INFO(
+            f: *const Dav1dFrameContext,
+            t: *const Dav1dTaskContext,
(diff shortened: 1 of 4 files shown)


src/recon.rs
```diff
@@ -1,6 +1,25 @@
 use crate::src::msac::dav1d_msac_decode_bool_equi;
 use crate::src::msac::MsacContext;
 
+/// This is a macro that defines a function
```
thedataking, line 4: If you are describing remaining work, I'd recommend starting the comment (and similar comments) with `TODO: ` such that we can easily grep through the code to find outstanding work. This is also what we're doing on line 9.

kkysen: Good idea, I'll add that.

kkysen: Added the `TODO` in cb05c1d68cf17c2a26597425757cb80d66be5177 along with a longer comment explaining how we might remove the macro.

randomPoison (approved):

PR #254, opened 2023-05-12
`fn read_pal_uv`: Cleanup and make mostly safe

kkysen:


kkysen:
CI is failing on arm7, but not locally on `donna`, so I think it's an alignment bug that I'm not seeing due to the older local version of qemu.  This looks similar to one I fixed with #253, so I'll wait until that merges and see if that was the same bug.

randomPoison (approved):

PR #261, opened 2023-05-19
`fn decomp_tx`: Cleanup and make mostly safe

kkysen:
I still need to make `txa` a safe slice/array ref (see the discussion below), but doing so in the way it's done in C would be UB in Rust, so I'm going to change the implementation a bit to make it safe, as @fbossen suggested below in a follow-up PR.

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -108,3 +108,30 @@ pub unsafe fn case_set<D, F>(
         _ => {}
     }
 }
+
+#[inline]
+pub unsafe fn case_set_upto16<D, F>(
+    var: libc::c_int,
+    dir: &mut D,
+    diridx: usize,
+    off: isize,
+    set_ctx: &mut F,
+) where
+    F: FnMut(&mut D, usize, isize, u64, SetCtxFn),
+{
+    match var {
+        1 => set_ctx(dir, diridx, off, 0x01, set_ctx_rep1::<alias8>),
+        2 => set_ctx(dir, diridx, off, 0x0101, set_ctx_rep1::<alias16>),
+        4 => set_ctx(dir, diridx, off, 0x01010101, set_ctx_rep1::<alias32>),
+        8 => set_ctx(
+            dir,
+            diridx,
+            off,
+            0x0101010101010101,
+            set_ctx_rep1::<alias64>,
+        ),
+        16 => set_ctx(dir, diridx, off, 0x0101010101010101, set_ctx_rep2),
+
+        _ => {}
+    }
+}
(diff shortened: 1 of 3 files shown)


kkysen:
@fbossen, could you take a look at `txa` here?  I'm not exactly sure how C is accessing and casting it here: https://github.com/memorysafety/rav1d/blob/05bceedfc8aed8aa0fae805adc2aaedb4e3aed86/src/lf_mask.c#L54-L62

From what I can tell, the indexes here aren't indexes really but slices.  But the part that doesn't make sense to me is that the inner two dimensions (the 32 sized ones) are being indexed by the non-zero `htw4` and `hth4`, but then it's being cast to a pointer to a 32x32 array.

Also, it seems I can't request you for review normally.  Is there some permissions thing you need for that?

fbossen:
The `decomp_tx` function could be written differently, where instead of doing this "funny" pointer arithmetic, two additional parameters `offset0` and `offset1` are passed to it.

`decomp_tx((uint8_t(*)[2][32][32]) &txa[0][0][hth4][htw4], ...` would become `decomp_tx(txa, hth4+offset0, htw4+offset1, ...` (`offset0` and `offset1` would be zero in `mask_edges_inter`) and accesses such as `txa[a][b][y][x]` would become `txa[a][b][y+offset0][x+offset1]`

Whereas an array of size `[2][2][32][32]` is passed to `decomp_tx`, the function will access only a part/slice of the array.

Let me know whether that makes sense.

As for permissions to request formal reviews, you may have to designate me as a collaborator on the project to give me some write access. The Github help pages say: "You can assign multiple people to each issue or pull request, including yourself, anyone who has commented on the issue or pull request, anyone with write permissions to the repository, and organization members with read permissions to the repository."

kkysen:
> The `decomp_tx` function could be written differently, where instead of doing this "funny" pointer arithmetic, two additional parameters `offset0` and `offset1` are passed to it.
> 
> `decomp_tx((uint8_t(*)[2][32][32]) &txa[0][0][hth4][htw4], ...` would become `decomp_tx(txa, hth4+offset0, htw4+offset1, ...` (`offset0` and `offset1` would be zero in `mask_edges_inter`) and accesses such as `txa[a][b][y][x]` would become `txa[a][b][y+offset0][x+offset1]`
> 
> Whereas an array of size `[2][2][32][32]` is passed to `decomp_tx`, the function will access only a part/slice of the array.
> 
> Let me know whether that makes sense.

Yeah, that makes sense.  I think that's kind of what I was realizing.  The `(uint8_t(*)[2][32][32])` cast is not accurate, but it's not actually accessed out of bounds, so it's probably (? not sure about the specifics of C's UB rules) okay.  But making that safe in Rust with references would definitely be UB, so I'm going to convert it to your suggested offset approach.

That said, I think I'll open a separate PR to make that change, so I'll mark this one as not a draft anymore.

randomPoison (approved):

PR #264, opened 2023-05-20
`fn decomp_tx`: Reformulate `txa` accesses so it can be made safe

kkysen:
This reformulates the usage of `txa` in `decomp_tx` to a way that can be easily made safe (and is actually simpler, too).  Previously, the multidimensional array `txa` was offset in each recursive call but still passed as a pointer to the original sized array.  Making this array pointer a reference would be UB, since it didn't actually point to the whole array anymore, but to somewhere in the middle of it.  Instead, the correct offsets are calculated from `{x,y}_off`, which are existing args, and applied at all `txa` use sites.

Note that this was also previously UB in Rust, as the fake array pointer was dereferenced and that extended out of bounds.

### DIFF
--- a/src/lf_mask.rs
+++ b/src/lf_mask.rs
@@ -78,8 +78,8 @@ unsafe fn decomp_tx(
     debug_assert!(depth <= 2);
     let t_dim = &dav1d_txfm_dimensions[from as usize];
 
-    let txa_y_off = y_off * t_dim.h;
-    let txa_x_off = x_off * t_dim.w;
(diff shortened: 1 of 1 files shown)


kkysen:
@fbossen, does this look good for what you suggested in https://github.com/memorysafety/rav1d/pull/261#issuecomment-1554548477?  It turned out even simpler I think, since I could calculate `{x,y}0` from `{x,y}_off`, which were already args.

randomPoison (approved):

PR #272, opened 2023-05-25
`fn decode_b`: Replace `union` accesses through `c2rust_unnamed*` fields with method accessors

kkysen:


src/decode.rs
```diff
@@ -3735,12 +3735,11 @@ unsafe fn decode_b(
...
         } else {
-            b.c2rust_unnamed.c2rust_unnamed_0.comp_type = COMP_INTER_NONE as libc::c_int as uint8_t;
+            *b.comp_type_mut() = COMP_INTER_NONE as libc::c_int as uint8_t;
             if let Some(seg) = seg.filter(|seg| seg.r#ref > 0) {
-                b.c2rust_unnamed.c2rust_unnamed_0.r#ref[0 as libc::c_int as usize] =
-                    seg.r#ref as i8 - 1;
+                b.r#ref_mut()[0 as libc::c_int as usize] = seg.r#ref as i8 - 1;
             } else if let Some(_) = seg.filter(|seg| seg.globalmv != 0 || seg.skip != 0) {
-                b.c2rust_unnamed.c2rust_unnamed_0.r#ref[0] = 0 as libc::c_int as int8_t;
+                b.r#ref_mut()[0] = 0 as libc::c_int as int8_t;
```
randomPoison, line 3742: We don't need the `r#` specificier here, `ref_mut` isn't a keyword. Same goes for the other references to this method.

randomPoison (changes requested):

src/decode.rs
kkysen: Oh yeah, good catch.

kkysen: Fixed now in 932559ad9fcf01bb05389347f46b834523564c40.

PR #274, opened 2023-05-25
`fn decode_b`: Translate `printf`s to `println!`s

kkysen:


src/decode.rs
```diff
@@ -3784,11 +3784,7 @@ unsafe fn decode_b(
                     }
                 }
                 if DEBUG_BLOCK_INFO(f, t) {
-                    printf(
-                        b"Post-ref[%d]: r=%d\n\0" as *const u8 as *const libc::c_char,
-                        b.r#ref()[0] as libc::c_int,
-                        ts.msac.rng,
-                    );
+                    println!("Post-ref[{}]: r={}", b.r#ref()[0], ts.msac.rng,);
```
randomPoison, line 3787: remove trailing comma plox

randomPoison (approved):

src/decode.rs
kkysen: 👍 , but what is plox?

kkysen: Fixed in cb5a5ad4c4842ef67f0af2afc8b900e9df04155c.

PR #277, opened 2023-05-30
`splat_mv_fn`: Make safer

kkysen:
This adds wrapper `fn`s around the FFI/asm `fn`s so things can be made safer without changing any asm signatures.

Once fewer things have to be `#[repr(C)]`, then the `fn` ptr can be made non-`extern "C"` and the args even safer (e.x. slices will work, but aren't safe currently over C's ABI).

Once the inner `*mut refmvs_block` raw ptr is made safe (a slice or an obviously correct separate length), then these functions can be made fully safe (with `unsafe` blocks, but checked ones).

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -1979,8 +1979,8 @@ unsafe fn splat_oneref_mv(
     t: &mut Dav1dTaskContext,
     bs: BlockSize,
     b: &Av1Block,
-    bw4: libc::c_int,
-    bh4: libc::c_int,
+    bw4: usize,
+    bh4: usize,
 ) {
     let mode = b.inter_mode() as InterPredMode;
     let tmpl = Align16(refmvs_block(refmvs_block_unaligned {
@@ -1999,7 +1999,7 @@ unsafe fn splat_oneref_mv(
     c.refmvs_dsp.splat_mv(
         &mut t.rt.r[((t.by & 31) + 5) as usize..],
         &tmpl.0,
-        t.bx,
+        t.bx as usize,
         bw4,
         bh4,
     );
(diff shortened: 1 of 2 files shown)


kkysen:
@randomPoison, if this approach looks good, I'm going to do something similar for the other `fn` ptrs called from `decode_b`.

thedataking:
I spent a second looking at the `wrap_splat_mv` macros. Do I have this correctly?: Right now they don't do anything but the goal is to add checks on the input arguments before passing them to the assembly routines. (Initially, I was wondering why you added the macros until I re-read your commit message.) 

If I'm understanding correctly, just move on and ignore me. It might have been easier to review if you held off introducing the macro's until you actually need them to hold checks rather than adding them ahead of time.  Or you could add a TODO note stating your intentions for the macro to hold argument checks. Anyway, no big deal, just sharing my thoughts.

kkysen:
> I spent a second looking at the `wrap_splat_mv` macros. Do I have this correctly?: Right now they don't do anything but the goal is to add checks on the input arguments before passing them to the assembly routines.

In 60299638813b6df42178ed92b331f368d9ffa68b, they don't do anything yet, but in the subsequent commits in this PR, they do.  The next 3 commits wouldn't be possible without the wrappers while still staying asm compatible.  It's not for checks on the arguments per se, but small modifications to the arguments (e.x. casting refs to ptrs).

randomPoison (approved):

kkysen:
Hi @thedataking, can you benchmark before and after this PR?  I just remembered I had tried this approach as well in the past, and I'm wondering if this is also bad for performance.  If it can be inlined, I wouldn't see why it would be, but it might not be inlined.

PR #283, opened 2023-06-02
`fn case_set*`: Simplify and make safe `fn case_set*`s as `struct CaseSet::{one,many}`

kkysen:
I'll add more explanation of the changes here in a bit.


kkysen:
@fbossen, I would request you as a reviewer, but the permissions still aren't set up right for that, so I'm tagging you here instead.

fbossen:
For benchmarking it is generally preferable to use a larger bitstream than a bunch of tiny ones (such as the argon test bitstreams). Some benchmark results were provided in the dav1d commit introducing the caseset macro https://code.videolan.org/videolan/dav1d/-/commit/e3b5d4d044506f9e0e95e79b3de42fd94386cc61.

kkysen:
> For benchmarking it is generally preferable to use a larger bitstream than a bunch of tiny ones (such as the argon test bitstreams). Some benchmark results were provided in the dav1d commit introducing the `case_set` macro https://code.videolan.org/videolan/dav1d/-/commit/e3b5d4d044506f9e0e95e79b3de42fd94386cc61.

That makes sense, and thanks for finding the commit that introduced `case_set`; that helps a lot.

I'm not sure how to test "decoding first 1000 frames of Chimera (1080p, 8bit)", though.  If you let me know what the command to run that is, I can benchmark things to check my safe Rust implementation.

kkysen:
I see now that in the e3b5d4d044506f9e0e95e79b3de42fd94386cc61 commit introducing `case_set`, `BlockContext`'s fields were all 8-byte aligned.  So it's possible that the alignment there was helpful for performance and we need to keep taking advantage of it.  In Rust, we can only safely do aligned writes if we're sure they are aligned, which can't statically verify since there are runtime offsets used.  To avoid the runtime cost, one idea I have is to mask off the lower 3 bits but only check the lower 3 bits are 0 in debug mode.  Like this:

```rust
let align = 8;
let mask = align - 1;
debug_assert!(offset & mask == 0, "{offset} is not {align}-byte aligned");
let aligned_offset = offset & ~mask;
```

Hopefully the `&` would be cheap enough compared to 2 `&`s and a predicted branch.

Alternatively, the aligned fields of `BlockContext` may be helpful enough on their own, as then they'd fit into cache boundaries much better, and so maybe we don't need aligned writes.

I'd like to see what benchmarking has to say more definitely about this.

fbossen:
> I'm not sure how to test "decoding first 1000 frames of Chimera (1080p, 8bit)", though. If you let me know what the command to run that is, I can benchmark things to check my safe Rust implementation. Mainly I'm not sure where "Chimera" is, but a full command to run would be perfect, too. Thanks!

You would run a command such as `time target/release/dav1d -i chimera_8b_1080p.ivf -l 1000 -o /dev/null` multiple times (say 3 or 5) and pick the median or best time. I shared the "Chimera" file in the Slack channel on March 9.

kkysen:
Here are the benchmarks with `hyperfine`:

Original `case_set`:

```shell
Benchmark 1: ./target/release/dav1d -i ./tests/large/chimera_8b_1080p.ivf -l 1000 -o /dev/null
  Time (mean ± σ):      1.650 s ±  0.008 s    [User: 14.936 s, System: 0.693 s]
  Range (min … max):    1.632 s …  1.659 s    10 runs
```

Using `small_memset` as in this PR:

```shell
Benchmark 1: ./target/release/dav1d -i ./tests/large/chimera_8b_1080p.ivf -l 1000 -o /dev/null
  Time (mean ± σ):      1.642 s ±  0.011 s    [User: 14.958 s, System: 0.702 s]
  Range (min … max):    1.624 s …  1.658 s    10 runs
```

Using only `memset` (i.e. `.fill`):

```shell
Benchmark 1: ./target/release/dav1d -i ./tests/large/chimera_8b_1080p.ivf -l 1000 -o /dev/null
  Time (mean ± σ):      1.687 s ±  0.013 s    [User: 15.274 s, System: 0.655 s]
  Range (min … max):    1.664 s …  1.710 s    10 runs
```

Using only `memset` is clearly slower (1.687 s), but the safe Rust version using `small_memset` but no aligned writes (1.642 s) is slightly faster than the original `case_set` (1.650 s).  Given that, I think the `small_memset` version in this PR is the best, as it's the fastest and relatively simple.

kkysen:
Updated benchmarks after freeing up disk space as they were previously very full (see https://github.com/memorysafety/rav1d/pull/285#issuecomment-1579310135) and I don't want that to interfere with anything.  The results are mostly very similar, though:

Original `case_set`:

```rust
kkysen in 🌐 donna in rav1d on  kkysen/preparation-for-safe-fns-case-set is 📦 v0.0.0 via 🦀 v1.71.0-nightly via 🐏 4GiB/63GiB | 108MiB/2GiB took 17s457ms
at 11:56:19 AM ❯ c b -r && hyperfine './target/release/dav1d -i ./tests/large/chimera_8b_1080p.ivf -l 1000 -o /dev/null'
    Finished release [optimized] target(s) in 0.01s
Benchmark 1: ./target/release/dav1d -i ./tests/large/chimera_8b_1080p.ivf -l 1000 -o /dev/null
  Time (mean ± σ):      1.659 s ±  0.010 s    [User: 15.025 s, System: 0.648 s]
  Range (min … max):    1.647 s …  1.683 s    10 runs
```

Using `small_memset` as in this PR:

```rust
kkysen in 🌐 donna in rav1d on  kkysen/safe-fns-case_set is 📦 v0.0.0 via 🦀 v1.71.0-nightly via 🐏 4GiB/63GiB | 108MiB/2GiB took 21ms
at 11:54:56 AM ❯ c b -r && hyperfine './target/release/dav1d -i ./tests/large/chimera_8b_1080p.ivf -l 1000 -o /dev/null'
   Compiling c2rust_out v0.0.0 (/home/kkysen/work/rust/rav1d)
    Finished release [optimized] target(s) in 0.78s
Benchmark 1: ./target/release/dav1d -i ./tests/large/chimera_8b_1080p.ivf -l 1000 -o /dev/null
  Time (mean ± σ):      1.638 s ±  0.008 s    [User: 14.925 s, System: 0.674 s]
  Range (min … max):    1.624 s …  1.652 s    10 runs
```

Using only `memset` (i.e. `.fill`):

```rust
kkysen in 🌐 donna in rav1d on  kkysen/case-set-as-memset is 📦 v0.0.0 via 🦀 v1.71.0-nightly via 🐏 3GiB/63GiB | 108MiB/2GiB took 13ms
at 12:08:18 PM ❯ c b -r && hyperfine './target/release/dav1d -i ./tests/large/chimera_8b_1080p.ivf -l 1000 -o /dev/null'
   Compiling c2rust_out v0.0.0 (/home/kkysen/work/rust/rav1d)
    Finished release [optimized] target(s) in 0.81s
Benchmark 1: ./target/release/dav1d -i ./tests/large/chimera_8b_1080p.ivf -l 1000 -o /dev/null
  Time (mean ± σ):      1.692 s ±  0.010 s    [User: 15.333 s, System: 0.688 s]
  Range (min … max):    1.679 s …  1.709 s    10 runs
```

randomPoison (approved):

fbossen:
The PR and the benchmarks LGTM. At this point in time I don't think it's worth trying to further optimize to take advantage of alignment.

thedataking:
My two cents (after taking a quick look): 

1. The PR has lots of commentary but the code to be merged does not. I don't think it is good to introduce `small_memset` without explaning why a regular memset isn't optimal. @kkysen please don't rely on folks reading the PR message, the code should be able to stand on its own.

2. I'm not sure results from one machine justifies rolling our own memset given that these low-level routines can vary greatly from one C library implementation to another, from one compiler to another, from one architecture to another. I would have used whatever the standard library provides absent any strong evidence to the contrary. That would save you from having to document your hand-rolled `small_memset` and keep things as simple as possible.

3. The replacement code for `case_set` is a lot more compact but I'm not sure it is as easy to grok. I would optimize for readability over compactness. This is a general comment. I've seen you fold/compact code with macros elsewhere and I personally don't think it is worth it. I realize this is a matter of taste but I don't like having to mentally expand macros in my head to understand what is going on.

If it was up to me, you'd take action on 2. Failing that, I think `small_memset` deserves a comment before merging.

kkysen:
I'll add the explanations from the PR description as comments in the code.  I had written that after opening the PR and forgot to go back and add it.

kkysen:
> This definitely needs to be documented in the code. This is complex enough that no one is going to understand what this is doing or how to use it without comments. Otherwise it does look correct, afaict.

> 1. The PR has lots of commentary but the code to be merged does not. I don't think it is good to introduce `small_memset` without explaning why a regular memset isn't optimal. @kkysen please don't rely on folks reading the PR message, the code should be able to stand on its own.

I've now added the documentation from the PR description to the code (and a little bit more) in 15b4d6eafb61efc7077311b5fa5646f2dc1156c2.

> 2. I'm not sure results from one machine justifies rolling our own memset given that these low-level routines can vary greatly from one C library implementation to another, from one compiler to another, from one architecture to another. I would have used whatever the standard library provides absent any strong evidence to the contrary. That would save you from having to document your hand-rolled `small_memset` and keep things as simple as possible.

There does seem to be strong evidence to the contrary, though, from both my benchmarks on `donna` in Rust, and the benchmarks done in e3b5d4d044506f9e0e95e79b3de42fd94386cc61 on some other machine in C.  In both, the small powers of 2 optimization done in `small_memset`, the unsafe Rust `case_set*` `fn`s, and the C `case_set*` macros are faster than just calling `memset`.  Had my benchmarks shown little difference between the implementations, I would've liked to choose `memset`, but that's not what I nor the `dav1d` authors observed.  Given they did they did this optimization for a reason and I'm able to reproduce the benchmarks, I think we should preserve them in Rust.

> 3. The replacement code for `case_set` is a lot more compact but I'm not sure it is as easy to grok. I would optimize for readability over compactness. This is a general comment. I've seen you fold/compact code with macros elsewhere and I personally don't think it is worth it. I realize this is a matter of taste but I don't like having to mentally expand macros in my head to understand what is going on.

I've tried to make it a bit simpler to read by using an `as_array` inner `fn` rather than the more confusing `set!` macro in 7cc882869e4bc73c840a1ced59d50ec4446e7b5f.  Is it also easier to read for you this way?  Even with the `set!` macro, I did already think it was far easier to understand than the previous `case_set*` code, which took a very long time to properly understand, and which I still do not fully understand (like the multiplying bytes by 1).

> If it was up to me, you'd take action on 2. Failing that, I think `small_memset` deserves a comment before merging.

I've also documented `small_memset` in 15b4d6eafb61efc7077311b5fa5646f2dc1156c2 now.

fbossen:
It's common to have a specialized `memset` function in code like this (i.e., a video decoder) to maximize processing speed. I've certainly done this before. @kkysen Can you push your `kkysen/case-set-as-memset` branch so that I can run some tests on my end?

kkysen:
> @kkysen Can you push your `kkysen/case-set-as-memset` branch so that I can run some tests on my end?

Sorry, just saw this.  It's pushed now: [`kkysen/case-set-as-memset`](https://github.com/memorysafety/rav1d/tree/kkysen/case-set-as-memset).

kkysen:
@thedataking, I've made the optimization docs a lot more concise in b846093ac808b6c56de371d4fc3094143f053a7d.  Does that look better?

fbossen:
> It's pushed now: [`kkysen/case-set-as-memset`](https://github.com/memorysafety/rav1d/tree/kkysen/case-set-as-memset).

Thanks. I ran tests with some more sequences and CPU configurations. The specialized `small_memset` is consistently faster than the generic `memset` although the amount by which it is faster varies a bit.

PR #285, opened 2023-06-05
Add `--profile incr-release`

kkysen:
Add `--profile incr-release`, which is `--release` with incremental compilation, which vastly reduces compile time for minor perf hits, and is good for iterative testing locally.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,3 +31,13 @@ default = ["asm", "bitdepth_8", "bitdepth_16"]
 asm = []
 bitdepth_8 = []
 bitdepth_16 = []
+
+# `release` takes ~2min to build.
+# For many use cases during iterative development,
+# such as running the argon tests,
+# way faster compile times is worth it for a small perf hit,
+# which can be achieved with incremental compilation,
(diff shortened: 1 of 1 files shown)


thedataking (approved):
LGTM (I assume this only takes effect if you pass -incr-release to cargo). Two general nitpicks:

- Why not break multi-line comments like this https://doc.rust-lang.org/rust-by-example/hello/comment.html (e.g. lines 10-17?)
> `# `release` takes ~2min to build.` 

- this observation is true for your machine today but could change at any time. A comment like `Enabling incremental brings compile times down to seconds at the expense of slightly reduced performance` conveys the same message without being specific to your environment.

kkysen:
It turns out the painfully slow release compile times were due to the disk being nearly full.  Now that there's a lot more space, clean release builds take ~52 s and rebuilds take ~1 s wheres with a full disk they were both taking ~2 min (all on `donna`).  So I'm closing this PR, as it's not really needed at all.  If builds ever start taking that long again, we should check disk space.  I'm not sure, though, why it just took forever rather than caused any errors, which is what usually happens.

PR #295, opened 2023-06-13
WIP: Deduplicate bitdepth-dependent types

randomPoison:
All the deduplication work you know in love, but now LOTS MORE! 🎉

### DIFF
--- a/src/cdef_apply_tmpl_16.rs
+++ b/src/cdef_apply_tmpl_16.rs
@@ -12,65 +12,6 @@ pub type coef = int32_t;
 use crate::include::stdatomic::atomic_int;
 use crate::include::stdatomic::atomic_uint;
 use crate::src::r#ref::Dav1dRef;
-#[derive(Copy, Clone)]
-#[repr(C)]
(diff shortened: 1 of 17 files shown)


kkysen:
When I last looked at the remaining type deduplication, the remaining blocker was `Dav1dDSPContext` and all of its bitdepth-dependent `fn` ptr fields.  `Dav1dContext` owns the `Dav1dDSPContext`s and `Dav1dFrameContext` contains a pointer to it.  Most of these `fn` ptrs have `*mut pixel`s and many `*mut coef`s, along with some other bitdepth-dependent types like `const left_pixel_row`.  I think some of them also have an extra `c_int` arg.  These are much more difficult to deduplicate as you can't cast between `fn` ptr types (except to cast `fn` safety).  Thus, we'd need either wrapper `fn`s that do the arg casting, or a `union` of the two possible `fn` ptr types.  If you can think of another solution for that, that'd be great, but I couldn't think of any.

randomPoison:
Replaced by #308

PR #303, opened 2023-06-14
`fn decode_b`: Revert code deleted in #132

kkysen:
See also https://github.com/memorysafety/rav1d/commit/19cb8394582b5af123ac8b2cecdcfd097904a66b#r118089193.


randomPoison (approved):
Good catch on this!

PR #304, opened 2023-06-20
`trait BitDepth`: Add bitdepth generics

kkysen:


include/common/intops.rs
```diff
@@ -41,9 +41,19 @@ pub fn iclip(v: c_int, min: c_int, max: c_int) -> c_int {
     }
 }
 
+#[inline]
+pub fn clip_u8(v: u8) -> u8 {
+    clip(v, u8::MIN, u8::MAX)
+}
```
randomPoison, line 47: This... doesn't seem to do anything? This seems like something that would be used on a wider integer type to clip it into the u8 range (i.e. what `iclip_u8` is doing). When run for a `u8` value it should always be a no-op, right?

randomPoison (approved):
Seems good to me, just a couple of minor comments.

include/common/intops.rs
kkysen: Hmm, good point.  It was probably clipping a larger int into a `u8`.  I'll get rid of this and just keep the `iclip_u8`, which I think does it correctly.

include/common/bitdepth.rs
```diff
@@ -0,0 +1,155 @@
...
+    fn bitdepth_max(&self) -> Self::Pixel;
+
+    fn get_intermediate_bits(&self) -> u8;
+
+    const PREP_BIAS: i16;
+}
+
+pub struct BitDepth8 {
+    bitdepth_max: (),
+}
```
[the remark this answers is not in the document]
kkysen: Perhaps not, but I thought it would be useful to have clear parity between the two.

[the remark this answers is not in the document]
kkysen: I think I might rearrange some of the definitions so this becomes `bitdepth_max: BitDepth8::BitDepthMax` so there's only one definition of `BitDepthMax` for each bitdepth.  Same thing for some of the other associated types.

include/common/intops.rs
kkysen: Actually, I think you're right that `iclip_pixel` for `BitDepth8` is just a nop.

kkysen: I've left `clip_u8` now, but as a narrowing conversion from `T` to `u8`.  That turns into a nop for `BitDepth8` where `T = Self::Pixel = u8`, and so I've also merged the `fn iclip_pixel` definitions as they're clearly equivalent now.  I also fixed `fn bitdepth_max` for `BitDepth8`, which was not working for some reason.

kkysen:
@randomPoison, could re-review this?  A bunch of little things changed.

randomPoison (approved):
Re-approved after changes! :shipit:

PR #306, opened 2023-06-20
`trait BitDepth`: Enable generic `as` casts in bitdepth-dependent `fn`s

kkysen:
An example of using this can be seen in `fn decode_coefs` in [`kkysen/deduplicate-fn-decode_coefs`](https://github.com/memorysafety/rav1d/tree/kkysen/deduplicate-fn-decode_coefs), but that code can't be finished until we deduplicate types, so it's only here as a WIP branch for now.

A complete example is in #311.

### DIFF
--- a/include/common/bitdepth.rs
+++ b/include/common/bitdepth.rs
@@ -2,6 +2,74 @@ use std::fmt::{self, Display, Formatter};
 
 use crate::include::common::intops::clip;
 
+pub trait FromPrimitive<T> {
+    fn from_prim(t: T) -> Self;
+}
+
+pub trait ToPrimitive<T> {
+    fn to_prim(self) -> T;
+}
+
+pub trait AsPrimitive {
+    fn as_<T>(self) -> T
+    where
+        Self: ToPrimitive<T>;
+}
+
+impl<T, U> ToPrimitive<U> for T
+where
+    U: FromPrimitive<T>,
+{
+    fn to_prim(self) -> U {
+        FromPrimitive::from_prim(self)
+    }
+}
+
+impl<U> AsPrimitive for U {
+    fn as_<T>(self) -> T
+    where
+        Self: ToPrimitive<T>,
+    {
+        self.to_prim()
+    }
+}
+
+macro_rules! impl_FromPrimitive {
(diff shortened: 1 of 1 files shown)


randomPoison:
I'm not sure I understand what this is enabling? It'd be helpful to also see a bit of code that makes use of these new traits in order to better understand what they're supposed to do.

kkysen:
> I'm not sure I understand what this is enabling? It'd be helpful to also see a bit of code that makes use of these new traits in order to better understand what they're supposed to do.

The code I was testing it on, `fn decode_ceofs`, needs deduplicated types to work, so I have a WIP branch I can link to here, but it doesn't fully compile because of the duplicate types.  I can also see if there's another function I can genericize as an example, but most of them take one of the duplicated types as an argument.

randomPoison:
I gave #311 a look so I see why you need these generic conversions. Kinda wish there were built-in traits for this, but ah well. Have you looked at the num_traits crate? Looks like they have [similar traits](https://docs.rs/num-traits/latest/num_traits/cast/trait.AsPrimitive.html) that you define here, so if their traits work for our needs then it's probably better to use that crate rather than defining our own traits.

kkysen:
> I gave #311 a look so I see why you need these generic conversions. Kinda wish there were built-in traits for this, but ah well. Have you looked at the num_traits crate? Looks like they have [similar traits](https://docs.rs/num-traits/latest/num_traits/cast/trait.AsPrimitive.html) that you define here, so if their traits work for our needs then it's probably better to use that crate rather than defining our own traits.

I looked at `num_traits` and tried to get it to work, but it's slightly different in a way that matters.  Their `.as_()` method has the `T` as a generic parameter for the type being implemented, but for my `as_()`, the `as_()` function itself is generic, meaning you can do `.as_::<T>()`, which is very useful in approximately a normal `as T` cast.  Furthermore, they don't have the equivalent `From` version, as their `FromPrimitive` and `ToPrimitive` are very different from mine, and support `Option`al conversion from/to any primitive based on if that, similar to `.try_into()` from `core::convert::TryInto`.

randomPoison (approved):

PR #307, opened 2023-06-20
Move Dav1dTaskContext and friends into internal.rs

randomPoison:
This PR adds the definitions for `Dav1dTaskContext` and its contained types into `internal.rs`. This actually *increases* duplication because these types are still re-defined in every file where they're used. In order to deduplicate any of these types we need all of their related type declarations. Duplicating the type definitions into `internal.rs` allows us to then incrementally deduplicate the type definitions across mulitple PRs, instead of having to do all of them in a single PR.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -1,13 +1,30 @@
+use crate::include::dav1d::common::Dav1dDataProps;
 use crate::include::dav1d::data::Dav1dData;
+use crate::include::dav1d::dav1d::Dav1dDecodeFrameType;
+use crate::include::dav1d::dav1d::Dav1dEventFlags;
+use crate::include::dav1d::dav1d::Dav1dInloopFilterType;
+use crate::include::dav1d::dav1d::Dav1dLogger;
+use crate::include::dav1d::headers::Dav1dContentLightLevel;
+use crate::include::dav1d::headers::Dav1dFilmGrainData;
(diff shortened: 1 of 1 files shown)


src/internal.rs
```diff
@@ -374,3 +714,267 @@ pub struct Dav1dTaskContext_task_thread {
...
+    pub filter_sbrow_deblock_cols: filter_sbrow_fn,
+    pub filter_sbrow_deblock_rows: filter_sbrow_fn,
+    pub filter_sbrow_cdef: Option<unsafe extern "C" fn(*mut Dav1dTaskContext, libc::c_int) -> ()>,
+    pub filter_sbrow_resize: filter_sbrow_fn,
+    pub filter_sbrow_lr: filter_sbrow_fn,
+    pub backup_ipred_edge: backup_ipred_edge_fn,
+    pub read_coef_blocks: read_coef_blocks_fn,
+}
+
+impl Dav1dFrameContext_bd_fn {
```
kkysen, line 876: Do we want these here, too, given we'd re-implement these with the new approach anyways?

kkysen (commented):
From which file did you move these?  If they're 8-bpc, could you note that in a comment at the top?

Also, could you squash that small formatting commit into the main one?

Also, is it worth re-ordering these items to their order as declared in C?  I've done that for the other items in `internal.rs`.

PR #308, opened 2023-06-20
Devirtualize looprestorationfilter_fn

randomPoison:
Define a new `LoopRestorationFilterFn` enum to wrap both bitdepth variants of the `looprestorationfilter_fn` function type, and update usages accordingly. This approach allows us to deduplicate the definitions of `Dav1dLoopRestorationDSPContext` while avoiding issues with needing to cast between function pointer types (since the original C was doing some fuckery with type-erasing function pointers that doesn't translate nicely to Rust).


src/looprestoration_tmpl_16.rs
```diff
@@ -350,7 +334,7 @@ unsafe extern "C" fn padding(
     mut dst: *mut pixel,
     mut p: *const pixel,
     stride: ptrdiff_t,
-    mut left: *const [pixel; 4],
+    mut left: const_left_pixel_row,
```
kkysen, line 337: What happened here?  Did the transpiler mess up?

src/looprestoration_tmpl_16.rs
```diff
@@ -1034,13 +1018,13 @@ unsafe extern "C" fn loop_restoration_dsp_init_x86(
         return;
     }
 
-    (*c).wiener[0] = Some(dav1d_wiener_filter7_16bpc_ssse3);
-    (*c).wiener[1] = Some(dav1d_wiener_filter5_16bpc_ssse3);
+    (*c).wiener[0] = LoopRestorationFilterFn::Bpc16(dav1d_wiener_filter7_16bpc_ssse3);
```
kkysen, line 1021: Oh, you did it this way.  I was thinking of having something like `LoopRestorationFilterFn::BPC16_SSSE3` so there are no more `fn` ptrs at all.  Doing it this way, we'd have both the `enum` `match` and the indirect `fn` ptr call and we don't get the other benefits from entirely removing the `fn` ptrs.

src/internal.rs
```diff
@@ -373,19 +373,61 @@ pub type cdef_fn = Option<
...
-        LrEdgeFlags,
-    ) -> (),
->;
+pub type pixel_8bpc = uint8_t;
+pub type coef_8bpc = int16_t;
+
+pub type pixel_16bpc = uint16_t;
+pub type coef_16bpc = int32_t;
+
+pub type const_left_pixel_row<Pixel> = *const [Pixel; 4];
```
kkysen, line 382: I think we should avoid generics here until we do them more properly in #304.  The advantage of using it here is more minimal.

src/looprestoration_tmpl_16.rs
randomPoison: Oh, that's a good point. Let me rework this to do away with fn pointers.

src/internal.rs
```diff
@@ -373,19 +373,61 @@ pub type cdef_fn = Option<
...
+    *const pixel_16bpc,
+    libc::c_int,
+    libc::c_int,
+    *const LooprestorationParams,
+    LrEdgeFlags,
+    libc::c_int,
+) -> ();
+
+#[derive(Clone, Copy)]
+#[repr(C)]
```
kkysen, line 410: Once these are data-less `enum`s, I think we can make it `#[repr(u8)]` instead of `#[repr(C)]`.  `#[repr(u8)]` is also stable, just smaller.

src/internal.rs
```diff
@@ -975,6 +995,6 @@ pub struct Dav1dCdefDSPContext {
 #[derive(Copy, Clone)]
 #[repr(C)]
 pub struct Dav1dLoopRestorationDSPContext {
-    pub wiener: [looprestorationfilter_fn; 2],
-    pub sgr: [looprestorationfilter_fn; 3],
+    pub wiener: [LoopRestorationFilterFn; 2],
+    pub sgr: [LoopRestorationFilterFn; 3],
 }
```
kkysen, line 1000: This is originally defined in `src/looprestoration.h`, so can you put it in `src/looprestoration.rs`?

src/looprestoration_tmpl_16.rs
```diff
@@ -1092,27 +1076,27 @@ unsafe extern "C" fn loop_restoration_dsp_init_arm(
...
-            (*c).wiener[0] = Some(dav1d_wiener_filter7_16bpc_neon);
-            (*c).wiener[1] = Some(dav1d_wiener_filter5_16bpc_neon);
+            (*c).wiener[0] = LoopRestorationFilterFn::Wiener7_Neon;
+            (*c).wiener[1] = LoopRestorationFilterFn::Wiener5_Neon;
         } else {
             // TODO(perl): enable assembly routines here
-            // (*c).wiener[0] = Some(dav1d_wiener_filter_neon);
-            // (*c).wiener[1] = Some(dav1d_wiener_filter_neon);
+            // (*c).wiener[0] = LoopRestorationFilterFn::Bpc16(dav1d_wiener_filter_neon);
+            // (*c).wiener[1] = LoopRestorationFilterFn::Bpc16(dav1d_wiener_filter_neon);
```
kkysen, line 1084: It would be nice to get these arm asm functions ported first to properly test all the changes here.  cc @thedataking.

src/internal.rs
```diff
@@ -373,19 +374,38 @@ pub type cdef_fn = Option<
...
-        libc::c_int,
-        *const LooprestorationParams,
-        LrEdgeFlags,
-    ) -> (),
->;
+pub type pixel_8bpc = uint8_t;
+pub type coef_8bpc = int16_t;
+
+pub type pixel_16bpc = uint16_t;
+pub type coef_16bpc = int32_t;
```
kkysen, line 381: Now that we have `trait BitDepth`, we can just use `BitDepth{8,16}::{Pixel,Coef}` for these.

src/internal.rs
kkysen: Similarly to https://github.com/memorysafety/rav1d/pull/308/files#r1245938890, we can now move these definitions to inside of `trait BitDepth`.

src/looprestoration_tmpl_16.rs
kkysen: Did you forget to update the `::Bpc16`s?

src/looprestoration_tmpl_8.rs
```diff
@@ -1057,26 +1041,26 @@ unsafe extern "C" fn loop_restoration_dsp_init_arm(
...
-            (*c).wiener[0] = Some(dav1d_wiener_filter7_8bpc_neon);
-            (*c).wiener[1] = Some(dav1d_wiener_filter5_8bpc_neon);
+            (*c).wiener[0] = LoopRestorationFilterFn::Wiener7_Neon;
+            (*c).wiener[1] = LoopRestorationFilterFn::Wiener5_Neon;
         } else {
             // TODO(perl): enable assembly routines here
-            // (*c).wiener[0] = Some(dav1d_wiener_filter_neon);
-            // (*c).wiener[1] = Some(dav1d_wiener_filter_neon);
+            // (*c).wiener[0] = LoopRestorationFilterFn::Bpc8(dav1d_wiener_filter_neon);
+            // (*c).wiener[1] = LoopRestorationFilterFn::Bpc8(dav1d_wiener_filter_neon);
```
kkysen, line 1049: Same here.

src/looprestoration.rs
```diff
@@ -23,3 +27,188 @@ pub union LooprestorationParams {
...
+            $($(
+                #[cfg(all(feature = "asm", $arch))]
+                $var_name,
+            )*)*
+        }
+
+        impl $enum_name {
+            #[cfg(feature = "bitdepth_16")]
+            #[inline]
+            pub fn call_16bpc(
```
kkysen, line 66: These need to unify bitdepths into one generic signature, because the callers will eventually be deduplicate into generic functions.  I used an extension `trait` of `BitDepth` to accomplish this.

src/looprestoration.rs
```diff
@@ -23,3 +27,188 @@ pub union LooprestorationParams {
...
+                lpf: *const pixel_16bpc,
+                w: libc::c_int,
+                h: libc::c_int,
+                params: *const LooprestorationParams,
+                edges: LrEdgeFlags,
+                bitdepth_max: libc::c_int,
+            ) {
+                match self {
+                    $(
+                        Self::$rust_var => unsafe {
```
kkysen, line 80: I don't think we want to mark these `unsafe` yet.  So far I've been leaving them `unsafe` until the `*_c`/`*_rust` versions are safe, so at least if we can show they're equivalent, it'd be safe (I'm not sure how `checkasm` works).

src/looprestoration.rs
```diff
@@ -23,3 +27,188 @@ pub union LooprestorationParams {
     pub filter: Align16<[[int16_t; 8]; 2]>,
     pub sgr: LooprestorationParams_sgr,
 }
+
+macro_rules! looprestoration_filter_fn_enum {
```
kkysen, line 31: This is a cool macro.  I didn't think of making them this long, especially in the pattern/match part of it.

PR #309, opened 2023-06-20
CI: Set timeout multiplier to 2 for debug builds

thedataking:


kkysen (approved):
LGTM.  2x for debug builds should be enough I think given it only sometimes timed out before.

PR #310, opened 2023-06-21
`MsacContext::symbol_adapt16`: Devirtualize

kkysen:
This is a small, simple example of devirtualizing a `fn` ptr into an `enum` in code I've already cleaned up.  Since it still works and there are no segfaults, I assume assembly is not calling the `fn` ptr directly (though there is untested assembly).

### DIFF
--- a/src/msac.rs
+++ b/src/msac.rs
@@ -134,7 +134,7 @@ pub fn dav1d_msac_decode_uniform(s: &mut MsacContext, n: libc::c_uint) -> libc::
 
 #[cfg(all(feature = "asm", target_arch = "x86_64"))]
 #[inline(always)]
-unsafe fn msac_init_x86(s: &mut MsacContext) {
+fn msac_init_x86(s: &mut MsacContext) {
     use crate::src::cpu::dav1d_get_cpu_flags;
     use crate::src::x86::cpu::DAV1D_X86_CPU_FLAG_AVX2;
     use crate::src::x86::cpu::DAV1D_X86_CPU_FLAG_SSE2;
--- a/src/msac.rs
+++ b/src/msac.rs
@@ -61,6 +61,14 @@ extern "C" {
 pub type ec_win = size_t;
 
 #[derive(Copy, Clone)]
+#[repr(u8)]
+pub enum FnSymbolAdapt16 {
+    Rust,
+    Sse2,
+    Avx2,
+}
+
+#[derive(Copy, Clone)]
 #[repr(C)]
 pub struct MsacContext {
     buf_pos: *const uint8_t,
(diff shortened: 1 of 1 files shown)


randomPoison (approved):

kkysen:
@thedataking, I should revert 4bea6a04b09ffae62530d04c46791f0f1f4384f3 and 908703ea1bc70a3de24e2c130fe72be6a64cb006 now, right?  I was also wondering if we can similarly pass other extra args to the `*_c` `fn`, such as slice lengths, so that it can be made pretty trivially safe, if that won't have any perf overhead.

cc @randomPoison

thedataking:
> @thedataking, I should revert https://github.com/memorysafety/rav1d/commit/4bea6a04b09ffae62530d04c46791f0f1f4384f3 and https://github.com/memorysafety/rav1d/commit/908703ea1bc70a3de24e2c130fe72be6a64cb006 now, right?

yes, please.  I wasn't aware some devirtualized code already had landed.

> I was also wondering if we can similarly pass other extra args for the *_c fn, such as slice lengths, so that it can be made pretty trivially safe,

the answer is probably yes, but now you have me wondering why `_c` functions  need extra args. How do they avoid going out of bounds currently? 🤔 Take `dav1d_msac_decode_symbol_adapt_c` for example, couldn't you get a slice from `cdf` and `n_symbols`? For `dav1d_msac_decode_bool_adapt_c` you know the size implicitly. Do you have any interesting examples?

kkysen:
> the answer is probably yes, but now you have me wondering why `_c` functions need extra args. How do they avoid going out of bounds currently? 🤔 Take `dav1d_msac_decode_symbol_adapt_c` for example, couldn't you get a slice from `cdf` and `n_symbols`? For `dav1d_msac_decode_bool_adapt_c` you know the size implicitly. Do you have any interesting examples?

In this example, I can use `n_symbols` as the length.  That's what the C code uses.  But to make this safe in Rust we need to check that `n_symbols < cdf.len()`, and that's why there's an `assert!`.  Presumably, there could be some complex whole-program proof that `n_symbols < cdf.len()`, but those are easy to get wrong, and that's why we're rewriting it in Rust.  Thus, by passing `cdf.len()` as an extra arg, we can recreate the `cdf` slice without having to add the extra `assert!(n_symbols < cdf.len())` check.  For other functions, the relationship with something like `n_symbols` might be even more complex, and so passing the actual slice length would be much simpler and much more obviously safe.  For example, in many functions I've cleaned up, I've had to do things like `std::slice::from_raw_parts(ptr, if h == 0 { 0 } else { (h - 1) * stride + w })` to get the length that is actually accessed by the function.  It would be far simpler in those cases to just have the actual slice length in the first place.

thedataking:
> Thus, by passing cdf.len() as an extra arg, we can recreate the cdf slice without having to add the extra assert!(n_symbols < cdf.len())

I looked at the interface for `dav1d_msac_decode_symbol_adapt_c` and `dav1d_msac_decode_symbol_adapt16`. I agree that it probably makes sense to pass the cdf slice length to guard against error.

>  For other functions, the relationship with something like n_symbols might be even more complex, and so passing the actual slice length would be much simpler and much more obviously safe.

I understand what you're saying but I'd prefer we avoid passing arguments whenever we can compute the slice length. Conceptually, the simplest is that all these routines have exactly the same interface as the C and assembly code but you have convinced me (in case of `dav1d_msac_decode_symbol_adapt_c`) that straying from that is sometimes a net win.

PR #313, opened 2023-06-25
`fn {clip,iclip_pixel}`: Allow them to clip into a smaller type

kkysen:
This reflects actual usage of `iclip_pixel`.

I used an `unsafe` in `fn clip`, as `.try_into().unwrap()` doesn't always get optimized out, and it's way slower when not.  This should be safe as long as the `U: Into<T>` and `T: TryInto<U>` `impl`s are well-formed (they definitely are for all primitives).

### DIFF
--- a/include/common/intops.rs
+++ b/include/common/intops.rs
@@ -1,5 +1,4 @@
 use std::ffi::{c_int, c_uint, c_ulonglong};
-use std::fmt::Debug;
 
 use crate::include::common::attributes::clz;
 use crate::include::common::attributes::clzll;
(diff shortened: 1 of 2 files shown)


include/common/intops.rs
```diff
@@ -31,24 +30,39 @@ pub fn umin(a: c_uint, b: c_uint) -> c_uint {
...
     } else {
-        v
+        let v = v.try_into();
+        // # Safety
+        // `min <= v <= max`, `min: U`, `max: U`,
+        // and for all `u: U`, `u.into().try_into() == Ok(u)`,
+        // so `v` must be in `U`, too.
+        //
+        // Note that `v.try_into().unwrap()` is not always optimized out.
+        unsafe { v.unwrap_unchecked() }
```
randomPoison, line 56: Rather than using `try_into` and unwrapping, would it work to use the traits you're adding in #306 to do this as an `as` cast instead? Your logic for why this is safe seems sound to me, but I think we could avoid the unsafety entirely here since this is effectively just doing integer conversions.

kkysen: Hmm, maybe.  I was thinking about that, too, but I was thinking `.as_()` should be restricted to primitives and perhaps we may want a wrapper type for `Pixel` or `Coef` or something.  I'll think about it.

randomPoison (approved):

include/common/intops.rs
kkysen: I'll come back to this once I've ported more bitdepth-dependent code and have a better idea of what the correct abstraction is, as neither are quite right for this case.

PR #314, opened 2023-06-25
`fn put_8tap_c`: Deduplicate w/ generics and cleanup/re-translate, including macros used

kkysen:
`src` is left as a raw ptr due to the negative offset.  It's easier to wait to fix it once the pointers become slices top-down than to figure out exactly how to adjust the indices and calculate the length here.


randomPoison (approved):

randomPoison:
Did you make changes to the C code in order to convert the macros to functions? If so could you also include those changes in this PR?

kkysen:
> Did you make changes to the C code in order to convert the macros to functions? If so could you also include those changes in this PR?

No, by re-translate I meant human translation.  I just ported these by hand and figured out the types from the call sites.

PR #319, opened 2023-06-27
Fix issues preventing macOS from building and passing tests

thedataking:
Frank's fixes for macOS to enable benchmarking + a few portability fixes of my own.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,6 +3,7 @@ use crate::include::stdint::*;
 
 use crate::stderr;
 use ::libc;
+use cfg_if::cfg_if;
 extern "C" {
     fn memcpy(_: *mut libc::c_void, _: *const libc::c_void, _: size_t) -> *mut libc::c_void;
     fn memset(_: *mut libc::c_void, _: libc::c_int, _: size_t) -> *mut libc::c_void;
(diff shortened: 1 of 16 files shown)


src/lib.rs
```diff
@@ -755,7 +756,32 @@ use crate::src::levels::BL_64X64;
...
+
+        const fn pthread_once_init() -> pthread_once_t {
+            let init = _opaque_pthread_once_t {
+                __sig: 0x30b1bcba as libc::c_int as libc::c_long,
+                __opaque: [0 as libc::c_int as libc::c_char, 0, 0, 0, 0, 0, 0, 0],
+            };
+            init
+        }
+    }
+}
```
kkysen, line 784: Can you put this in `crate::include::pthread`?

kkysen (approved):
LGTM once that `pthread_once_init` code is moved to `crate::include::pthread`.  There's also other code that'd ideally be made fully portable (i.e. anywhere we use our own version instead of `libc`'s), but since a lot of that code will eventually be ported to idiomatic Rust, it's probably not worth it now.  So if these changes pass all tests on macOS, then it LGTM.

fbossen (approved):
I confirm this resolves issues with building/running on macOS

Issue #75, opened 2023-04-04
Setup checkasm tests


randomPoison:
We should port the checkasm tests from dav1d over to rav1d. There's already a Rust version of the checkasm test harness available in the [checkasm](https://crates.io/crates/checkasm) crate, so we should be able to just hook up tests for all of the assembly that has been ported to Rust so far.

PR #320, opened 2023-06-28
`src/mc.rs`: Rename `*_c` `fn`s to `*_rust` as they're now in Rust

kkysen:


thedataking:
The code under `tests/checkasm` relies on the `_c` suffixes. Might be worth hanging onto the original names until #75 is closed. If you want to rename the functions now, I guess a note on #75 could also mitigate any potential confusion.

kkysen:
> The code under `tests/checkasm` relies on the `_c` suffixes. Might be worth hanging onto the original names until #75 is closed. If you want to rename the functions now, I guess a note on #75 could also mitigate any potential confusion.

Oh, I've already renamed other `*_c` `fn`s to `*_rust` in prior PRs.  I need to do something like this for the upcoming `*bilin*` `fn`s since I need to keep the `*_c` ones while I cleanup and genericize the `*_rust` ones, as the `*_c` ones are still being used as `fn` ptrs (for just a little while longer).  If we need the `*_c` ones for `checkasm`, I think it's better to create `*_c` wrappers calling the `*_rust` ones so that they can be non-generic, `extern "C"`, `#[no_mangle]`, etc.  I'm not really sure how `checkasm` works, though, or how the Rust port of it works, but I think we should be fine either way, as renaming these back or adding wrappers (which I already have to do for some) would be pretty simple.

thedataking (approved):

PR #322, opened 2023-06-28
`struct Dav1dMCDSPContext`: Devirtualize `mc{,t}{,_scaled}` `fn` ptrs

kkysen:
This devirtualizes all of the `fn` ptrs in the `Dav1dMCDSPContext::{mc{,t}{,_scaled}` arrays, all at once, which includes matching on both the `Dav1dFilter2dRust` and `FnVersion` (which multiply to 250 `fn` ptrs).  So sorry for the huge PR, but I didn't see a way of breaking up the changes into smaller pieces without also creating a ton of extra intermediary work.

I used the `paste!` proc-macro to reduce a lot of the boilerplate, though there still is significant boilerplate from bitdepths and from the 4 functions.  If you see a way to reduce it further, that'd be great, but I didn't see another way (one idea I had got thwarted by `dav1d` naming its identifier parts starting with numbers, like `8tap` and `8bpc`, which means they're not Rust `ident`s).

The way I designed `FnVersion` and `FnAsmVersion`, they probably can be used by other devirtualized `fn` ptrs, too, though there may also be other x86 variants used in other functions that would need to be added.


src/mc.rs
```diff
@@ -1,12 +1,1025 @@
...
+                        src: *const <BitDepth8 as BitDepth>::Pixel,
+                        src_stride: ptrdiff_t,
+                        w: c_int,
+                        h: c_int,
+                        mx: c_int,
+                        my: c_int,
+                    );
+                }
+
+                let (dst, dst_stride, src, src_stride, w, h, mx, my) = args;
```
kkysen, line 147: (from #323) can enclose over `args`, so don't need this.

src/log.rs
```diff
@@ -178,13 +178,11 @@ pub type itxfm_fn = Option<
...
 pub type coef = ();
+use crate::src::mc::Dav1dMCDSPContextRust;
 #[derive(Copy, Clone)]
 #[repr(C)]
 pub struct Dav1dMCDSPContext {
-    pub mc: [mc_fn; 10],
-    pub mc_scaled: [mc_scaled_fn; 10],
-    pub mct: [mct_fn; 10],
-    pub mct_scaled: [mct_scaled_fn; 10],
+    pub rust: Dav1dMCDSPContextRust,
```
thedataking, line 185: Why do we need to have types and fields named `rust` and `Dav1dMCDSPContextRust`? We should name stuff based on its purpose (what it does) not the implementation language (how it does it).

src/mc.rs
```diff
@@ -1,12 +1,1025 @@
...
+                (Sharp, Sharp) => asm_fn!(asm, args, _8tap_scaled_sharp),
+            },
+            Filter2dRust::BiLinear => asm_fn!(asm, args, _bilin_scaled),
+        }
+    }
+}
+
+#[derive(Copy, Clone, Default)]
+#[repr(C)]
+pub struct Dav1dMCDSPContextRust {
```
thedataking, line 824: Can't this be `Dav1dMCDSPContext`? if that name is already taken, can't you get rid of the old one?

src/log.rs
kkysen: They're just temporary.  I wanted to only devirtualize a few fields of that struct to start with, but once the rest are, too, these Rust-named structs and fields will go away.  Is there another name you think would be good?

thedataking: I think it would be good to deduplicate `Dav1dMCDSPContext` first (can land as a separate PR) and then just do to the outer/original structure what you want. This PR isn't easy to review to begin with and the introduction of temporary structures doesn't make it easier. 

If you must  – I'd advise advise against temporary changes if only done for convenience since it makes for a harder to understand commit history – please *always* label temporary stuff like so `// TODO(kkysen): Remove once ...`. This has two benefits:

1. It leaves our future selves with breadcrumbs so we don't miss stuff we need to cleanup
2. It communicates to others what your plans are. I won't nitpick or worry so much about stuff that goes away down the line.

src/mc.rs
kkysen: The old one is still duplicated since it's function pointers haven't been devirtualized yet.  I didn't want to inline each field of the Rust one into all the duplicated old ones, so I put them into one struct, the Rust one.  If there's a better name I could switch to that, but I think it's fine since it's temporary and will go away pretty soon once I finish devirtualizing the rest of the struct.  I can definitely add a comment though.

src/log.rs
kkysen: I'll definitely add a comment to make it clear it's temporary.  We can't deduplicate it first, though.  The devirtualization is needed to deduplicate it since it gets rid of the bitdepth-dependent function pointers, which make deduplication extremely difficult.

The other alternative is to avoid the temporary is to devirtualize the whole struct at once, but that would be a very large PR so I wanted to split it up.

thedataking:
What is the benefit of getting rid of the function pointers again? It seems we're going from a mode of operation where you set the function pointers once and then the right function gets called until the configuration changes (which may be never). 

If I'm reading this code right, each function call inspects the current configuration and then dispatches to the correct function. If that true, the C version is much more elegant.

I vaguely remember that one of the reasons we're doing all this work is to paper over the differences between the assembly code and the safe Rust versions (since the two have different function signatures). Seeing how large this PR is, it makes me wonder if there is a simpler better way to achieve the same goals. For instance, couldn't we keep the interface to all these functions the same - and for the Rust versions, reconstruct safe Rust types from raw pointers – that would let us avoid the per-function call dispatch that (I think) is being used here.

kkysen:
> Remaining question: by getting rid of the function pointers, are we replacing the indirect function calls in the C code with switches/matches

Yeah.

> (such that we do more work per function call)?

Not necessarily.  Often a match/switch + direct calls are faster than an indirect call, though I plan to benchmark this for any noticeable regressions.  This way also would have the small extra benefit of no match if assembly is disabled.

> If so, is that strictly necessary? The dispatch through function pointers in the C version seems like a good and elegant solution.

It's an elegant solution with the templates in C, but it ends up being pretty ugly in Rust.  The function pointers are different per bitdepth, so we can't deduplicate the remaining types.  Since we can't cast between different function pointers, which we do for other bitdepth-different pointers, this makes it very difficult.  Nicole and I tried a while to get that approach to work, but it was very difficult, and the potential solution preserving the function pointers would be more unsafe (e.x. using unions) and have just as much or possibly more boilerplate.  I'm also gonna add more of an explanation for why we're using this solution in the PR description (I could put this in the code, too, but I'm not sure where it'd go) (@rinon also suggested maybe a blog post eventually).

> This technique is also commonly used in high-performance interpreters.

I might be thinking of something else, but I thought it was computed gotos or tail calls that were used a lot, with the emphasis on having direct calls, since there's a cost to indirect calls, while also avoiding the switch case.

thedataking:
>  Often a match/switch + direct calls are faster than an indirect call, though I plan to benchmark this for any noticeable regressions.

Indirect branches are fast on all modern CPUs thanks to branch prediction and speculative execution. The branches we are talking about here is one instruction which gets essential perfect prediction since the target rarely changes. The C implementation uses indirect calls which should tell you something about performance. As I mentioned, some interpreters (Perl I think) is basically a loop that dispatches on function pointers each of which represent an instruction.

> It's an elegant solution with the templates in C, but it ends up being pretty ugly in Rust.

The code we have here isn't elegant either. The calls to the assembly functions now go through two layers of macros; not very readable. Sometimes the more verbose but easier to read version is preferable. If having a few unsafe blocks in the Rust code, allows substantial simplification, I would strongly consider exploring that.

> This way also would have the small extra benefit of no match if assembly is disabled.

You can assume assembly will be enabled most of the time so this isn't much of a benefit. The C/Rust versions are basically there to test the correctness of the optimized routines but using them in production burns way too much energy vs the SIMD versions.

kkysen:
> The code we have here isn't elegant either. The calls to the assembly functions now go through two layers of macros; not very readable. Sometimes the more verbose but easier to read version is preferable. If having a few unsafe blocks in the Rust code, allows substantial simplification, I would strongly consider exploring that.

We both already tried though, and when we talked a few weeks ago before trying this way, we decided this seemed simpler to implement.  If you really want, we can go back and try the union of function pointers way, but it's going to be ugly as well, and it'll also have quite a lot of unsafe blocks, not just a few.

The function pointers to assembly functions also have to be `extern "C"`, which also means we can't put any references, slices, or safe types in any of the structs that are passed in any of these function pointers.  If we make them not `extern "C"`, then we fix this issue but have to generate shims for all of the assembly functions, which is a ton of code and will take a while to implement as well.  We could use macros like we do in this PR, but then we have the same macro complexity.

We have thought about a lot of approaches to this problem and tried them out on some function pointers, and we concluded that the devirtualization approach would be the overall best.  If you can think of another potential solution, we can consider that if it solves things better, but out of all the solutions we've thought of so far, I still think the devirtualization approach is the best.

Also, the static dispatch we have here will also be branch predicted just like the dynamic dispatch would be.  I doubt that there'll be a meaningful difference between the two in performance, but I'll check.

thedataking:
Your [previous comment](https://github.com/memorysafety/rav1d/pull/322#issuecomment-1636643776), is very helpful, thanks @kkysen! Having the additional context helped me wrap my head around this. Understood that you've arrived at the present approach after spending much effort trying to use function pointers.

One last question: is the difference between the function signatures for the assembly routines and the fallback functions a source of complexity? If so, would it be possible to always dispatch using the unsafe function signatures and only have wrappers (recovering safe types from raw pointers) for the fallback routines? I imagine the following benefits:

1. It takes less work to shim the fallback routines (vs. shimming everything).
2. There would only be an overhead when using the fallback routines (rare, don't care).
3. The safety arguments would presumably be trivial: recovering slice from raw args is trivial for instance.
4. (need to double check this): `checkasm` assumes that the assembly and fallback routines have equivalent function signatures. 

Maybe you already considered all of the above or know of a reason why it won't work (or won't allow simplification)?

> Also, the static dispatch we have here will also be branch predicted just like the dynamic dispatch would be. I doubt that there'll be a meaningful difference between the two in performance, but I'll check.

Yes, I doubt what you have here will be meaningfully slower. My point was that we shouldn't get rid of indirect calls out of concerns for performance*. You've made it clear that there are other reasons you chose the current approach so all good.

* Modern CPUs mainly lose performance to mis-prediction, mis-speculation, and when they have to wait for memory (typically when the access pattern is unpredictable and/or has poor locality). None of these concerns apply here AFAICT.

kkysen:
I'm glad that made sense.  I'm going to add some of that to the PR description, as I think it wasn't that clear before unless someone remembered all our previous discussions.

> One last question: is the difference between the function signatures for the assembly routines and the fallback functions a source of complexity? If so, would it be possible to always dispatch using the unsafe function signatures and only have wrappers (recovering safe types from raw pointers) for the fallback routines?

I'm not sure what the exact reason was, but it might have been about not being able to change the types (e.x. replace a pointer with a `Vec`) that already in the function pointer signatures, not changing the signatures themselves.  I can check more closely when I'm back to my laptop.  I did do it this way for `src/masc.rs` (though with more unsafe), so it can work for some function pointers, but I remember a bunch of the others didn't work when I tried this way.

PR #323, opened 2023-06-30
Devirtualize looprestoration function pointers

randomPoison:


src/looprestoration.rs
```diff
@@ -23,3 +29,288 @@ pub union LooprestorationParams {
...
+                    left,
+                    lpf,
+                    w,
+                    h,
+                    params,
+                    edges,
+                    self.bitdepth_max().into(),
+                )
+            };
+        }
```
kkysen, line 242: Oh, you're allowed to do this?  I had thought you can't due to macro hygiene, but if you can, that'd be great.  Is it because the macro is defined inside those variables' scope already, so it's like a closure?

src/looprestoration.rs
```diff
@@ -23,3 +29,288 @@ pub union LooprestorationParams {
...
+
+    /// Converts a raw SGR filter index into the corresponding filter type.
+    pub fn sgr(raw: usize) -> Self {
+        match raw {
+            0 => Self::Sgr5x5,
+            1 => Self::Sgr3x3,
+            2 => Self::SgrMix,
+            _ => unreachable!(),
+        }
+    }
```
kkysen, line 94: These may (?) be able to take `bool`s instead, having cursorily looked at the call sites.

src/looprestoration.rs
```diff
@@ -23,3 +29,288 @@ pub union LooprestorationParams {
     pub filter: Align16<[[int16_t; 8]; 2]>,
     pub sgr: LooprestorationParams_sgr,
 }
+
+#[derive(Copy, Clone)]
+#[repr(C)]
+pub struct Dav1dLoopRestorationDSPContext {
+    pub wiener: FnVersion,
+    pub sgr: FnVersion,
```
kkysen, line 37: I think we can mark these temporarily `pub` until `looprestoration` template deduplication.  I'll do the same in my PR.

src/looprestoration_tmpl_16.rs
```diff
@@ -2,6 +2,7 @@ use crate::include::stddef::*;
 use crate::include::stdint::*;
 #[cfg(all(feature = "asm", any(target_arch = "arm", target_arch = "aarch64"),))]
 use crate::src::align::Align16;
+use crate::src::mc::FnVersion;
```
kkysen, line 5: Maybe we should move `FnVersion` and `FnAsmVersion` to `src/cpu.rs`?

kkysen (commented):
Overall, this looks really great I think.  I'll take a closer look at everything once it's not a draft anymore.  The macros and asm matching look very clean (as clean as can be).

src/looprestoration.rs
randomPoison: Yeah, I'm pretty sure it only works because the macro is defined within the fn, so it effectively "captures" its environment kinda like a closure does. Funnily enough, I only even noticed that this works because you accidentally forgot a `$` in one of your macros in `mc.rs`, but the code still compiles because there's a variable named `args` in scope.

kkysen: > Funnily enough, I only even noticed that this works because you accidentally forgot a `$` in one of your macros in `mc.rs`, but the code still compiles because there's a variable named `args` in scope.

Oh, oops!

I might use this approach then.  I was purposefully passing `args` through everything since I thought I couldn't capture it.

src/looprestoration_tmpl_16.rs
randomPoison: Done!

src/looprestoration.rs
```diff
@@ -23,3 +30,327 @@ pub union LooprestorationParams {
...
+        left: *const [Self::Pixel; 4],
+        lpf: *const Self::Pixel,
+        w: libc::c_int,
+        h: libc::c_int,
+        params: *const LooprestorationParams,
+        edges: LrEdgeFlags,
+    );
+}
+
+#[cfg(feature = "bitdepth_8")]
```
kkysen, line 114: I didn't do this myself.  Should `BitDepth8` itself have a `#[cfg(feature = "bitdepth_8")]`, too?  Are 8bpc asm functions built when the `bitdepth_8` feature is not built?  If they are, it would simplify things not to need these includes, but if they're only built when needed, then we should probably have the `#[cfg]` on `BitDepth8` and all of its `impl`s.

src/looprestoration_tmpl_16.rs
```diff
@@ -1092,24 +883,21 @@ unsafe extern "C" fn loop_restoration_dsp_init_arm(
...
     cfg_if! {
         if #[cfg(target_arch = "aarch64")] {
-            (*c).wiener[0] = Some(dav1d_wiener_filter7_16bpc_neon);
-            (*c).wiener[1] = Some(dav1d_wiener_filter5_16bpc_neon);
+            (*c).wiener = FnVersion::NEON;
         } else {
             // TODO(perl): enable assembly routines here
-            // (*c).wiener[0] = Some(dav1d_wiener_filter_neon);
-            // (*c).wiener[1] = Some(dav1d_wiener_filter_neon);
+            // (*c).wiener = FnVersion::NEON;
```
kkysen, line 889: See #326.

src/looprestoration_tmpl_8.rs
```diff
@@ -1057,18 +852,14 @@ unsafe extern "C" fn loop_restoration_dsp_init_arm(
...
     cfg_if! {
         if #[cfg(target_arch = "aarch64")] {
-            (*c).wiener[0] = Some(dav1d_wiener_filter7_8bpc_neon);
-            (*c).wiener[1] = Some(dav1d_wiener_filter5_8bpc_neon);
+            (*c).wiener = FnVersion::NEON;
         } else {
             // TODO(perl): enable assembly routines here
-            // (*c).wiener[0] = Some(dav1d_wiener_filter_neon);
-            // (*c).wiener[1] = Some(dav1d_wiener_filter_neon);
+            // (*c).wiener = FnVersion::NEON;
```
kkysen, line 858: See #326.

src/looprestoration.rs
```diff
@@ -23,3 +30,327 @@ pub union LooprestorationParams {
...
+        };
+
+        bd.call(
+            fn_version, filter, p, stride, left, lpf, w, h, params, edges,
+        )
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum FilterFn {
```
kkysen, line 69: Should this be `LoopRestorationFilterFn`?  Also, I was writing them with the `Fn` first to be more similar to how `fn` precedes the `fn` name in `fn foo`; what do you think about that?

src/lr_apply_tmpl_16.rs
```diff
@@ -650,9 +636,11 @@ unsafe extern "C" fn lr_stripe(
...
                 + (*filter.offset(1))[2] as libc::c_int)
                 * 2) as int16_t;
-        lr_fn = (*dsp).lr.wiener[((*filter.offset(0))[0] as libc::c_int
-            | (*filter.offset(1))[0] as libc::c_int
-            == 0) as libc::c_int as usize];
+        lr_fn = FilterFn::wiener(
+            ((*filter.offset(0))[0] as libc::c_int | (*filter.offset(1))[0] as libc::c_int == 0)
+                as libc::c_int as usize,
+        )
+        .into();
```
kkysen, line 643: What's the `.into()` for?

src/looprestoration.rs
```diff
@@ -23,3 +30,327 @@ pub union LooprestorationParams {
...
+            1 => Self::Sgr3x3,
+            2 => Self::SgrMix,
+            _ => unreachable!(),
+        }
+    }
+}
+
+pub trait BitDepthFnLoopRestoration: BitDepth {
+    unsafe fn call(
+        &self,
```
kkysen, line 100: We don't want this publically callable, but `BitDepthFnLoopRestoration` has to be `pub` since it appears in necessarily `pub methods.  In #322, I used a `pub struct Token(())` type with a private constructor, though there might be a cleaner way to accomplish it.

randomPoison:
Closing this because it seems like we're going to go with the approach used in #332.

PR #324, opened 2023-07-04
Fix macOS build on x86 hosts

thedataking:
We need to set PREFIX to one in config.asm such that the assembly routine names get mangled correctly, otherwise linking will fail.


kkysen (approved):
LGTM assuming this fixes the build errors on macOS as you say (none of this is in CI, right?).

PR #325, opened 2023-07-08
CI: Test on x86_64-apple-darwin

thedataking:
Since performance testing happens on macOS (for aarch64), it would be nice to automatically test breakage on macOS although GitHub actions do not offer any arch64-apple-darwin test runners at the moment.


randomPoison (approved):

kkysen (approved):
LGTM and makes sense for getting as close to `aarch64-apple-darwin` testing in CI as we can.

PR #326, opened 2023-07-12
Add armv7 assembly for wiener_filter

### LINKED ISSUE #180: Fix arm7 builds with assembly enabled
Steps:
- Fix build errors (need to update `build.rs` and fix any linker errors)
- Test arm7 with assembly enabled in CI

thedataking:
Closes #180.

### DIFF
--- a/src/looprestoration_tmpl_8.rs
+++ b/src/looprestoration_tmpl_8.rs
@@ -288,6 +288,30 @@ extern "C" {
     );
 }
 
+#[cfg(all(feature = "asm", target_arch = "arm"))]
+extern "C" {
+    fn dav1d_wiener_filter_h_8bpc_neon(
+        dst: *mut int16_t,
+        left: *const [pixel; 4],
+        src: *const pixel,
+        stride: ptrdiff_t,
+        fh: *const int16_t,
+        w: intptr_t,
+        h: libc::c_int,
+        edges: LrEdgeFlags,
+    );
+    fn dav1d_wiener_filter_v_8bpc_neon(
+        dst: *mut pixel,
+        stride: ptrdiff_t,
+        mid: *const int16_t,
+        w: libc::c_int,
+        h: libc::c_int,
+        fv: *const int16_t,
+        edges: LrEdgeFlags,
+        mid_stride: ptrdiff_t,
+    );
+}
+
 use crate::src::tables::dav1d_sgr_x_by_x;
 
 pub type pixel = uint8_t;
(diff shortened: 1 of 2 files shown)


src/looprestoration_tmpl_8.rs
```diff
@@ -1060,9 +1154,8 @@ unsafe extern "C" fn loop_restoration_dsp_init_arm(
             (*c).wiener[0] = Some(dav1d_wiener_filter7_8bpc_neon);
             (*c).wiener[1] = Some(dav1d_wiener_filter5_8bpc_neon);
         } else {
-            // TODO(perl): enable assembly routines here
```
fbossen, line 1063: Should this change also be made for the 16-bit path?

thedataking: thanks, good catch! 😅

kkysen:
@randomPoison, adding you since this affects #323.

src/looprestoration_tmpl_8.rs
```diff
@@ -1037,7 +1061,77 @@ unsafe extern "C" fn loop_restoration_dsp_init_x86(
...
     }
 }
-
+// This calculates things slightly differently than the reference version.
+// This version calculates roughly this:
+// int32_t sum = 0;
+// for (int i = 0; i < 7; i++)
+//     sum += mid[idx] * fv[i];
+// sum = (sum + rounding_off_v) >> round_bits_v;
+// This function assumes that the width is a multiple of 8.
```
kkysen, line 1070: This comment is for `dav1d_wiener_filter_v`, not `wiener_filter_neon`.  I think it's fine to leave out for now; we'll add the comments back as we clean up, devirtualize, and genericize `looprestoration`.

src/looprestoration_tmpl_8.rs
```diff
@@ -1037,7 +1061,77 @@ unsafe extern "C" fn loop_restoration_dsp_init_x86(
...
+    dst: *mut pixel,
+    stride: ptrdiff_t,
+    left: *const [pixel; 4],
+    mut lpf: *const pixel,
+    w: libc::c_int,
+    h: libc::c_int,
+    params: *const LooprestorationParams,
+    edges: LrEdgeFlags,
+) {
+    let filter: Align16<[[int16_t; 8]; 2]> = (*params).filter;
```
kkysen, line 1082: `filter` isn't aligned in the C version, and is a ptr to `[i16; 8]`, not an array of it.  So this should be a reference/pointer to `(*params).filter`, not a copy of its array.  Same for the 16 bpc one.

kkysen (changes requested):
I gave it a close review.  The `filter` variable needs to be fixed and the long comment moved/left for later, but everything else LGTM.

src/looprestoration_tmpl_8.rs
thedataking: Good catch!

src/looprestoration_tmpl_8.rs
```diff
@@ -288,6 +288,30 @@ extern "C" {
...
 }
 
+#[cfg(all(feature = "asm", target_arch = "arm"))]
+extern "C" {
+    fn dav1d_wiener_filter_h_8bpc_neon(
+        dst: *mut int16_t,
+        left: *const [pixel; 4],
+        src: *const pixel,
+        stride: ptrdiff_t,
+        fh: *const [int16_t; 8],
```
kkysen, line 298: What you had previously, `*const int16_t`, was correct, as `const int16_t fh[8]` in an argument decays to `const int16_t *fh`, and that's how `c2rust transpile` translates it.

src/looprestoration_tmpl_8.rs
```diff
@@ -1037,7 +1061,70 @@ unsafe extern "C" fn loop_restoration_dsp_init_x86(
...
+    let filter: *const [int16_t; 8] = (*params).filter.0.as_ptr();
+    let mut mid: Align16<[int16_t; 68 * 384]> = Align16([0; 68 * 384]);
+    let mut mid_stride: libc::c_int = w + 7 & !7;
+    // Horizontal filter
+    dav1d_wiener_filter_h_8bpc_neon(
+        &mut *mid.0.as_mut_ptr().offset((2 * mid_stride) as isize),
+        left,
+        dst,
+        stride,
+        filter.offset(0),
```
kkysen, line 1084: This would be translated as `(*filter.offset(0)).as_ptr()` when the `fh` arg is `*const int16_t`.

kkysen (changes requested):
The change to `filter` looks good, but you had the `f{h,v}` arg types right before.

thedataking:
> The change to `filter` looks good, but you had the `f{h,v}` arg types right before.

I'm not convinced the current types are wrong. Andrei and I just checked that `*const [i16; 8]` has the same size as `*const i16` so what gets passed to the assembly is still just a pointer to the beginning of the array but the call sites are simpler.

If you want it to match the transpiler output, I can change it back, the way we call the assembly is changing anyway right?

kkysen:
> I'm not convinced the current types are wrong. Andrei and I just checked that `*const [i16; 8]` has the same size as `*const i16` so what gets passed to the assembly is still just a pointer to the beginning of the array but the call sites are simpler.
> 
> If you want it to match the transpiler output, I can change it back, the way we call the assembly is changing anyway right?

Yeah, the way we ultimately call the assembly is gonna change, but I'd much rather match the transpiler output that I've been cleaning up the whole time, as I can recognize its output patterns very quickly now.  And since the C function is declared to take a `const int16_t fh[8]`, which decays to `const int16_t *fh`, I think it'd be best to match that even if the assembly is ultimately passed the same pointer.  Note that the C functions didn't declare it as `const int16_t (*fh)[8]`, which would be `*const [int16_t; 8]`.

kkysen (approved):

Issue #77, opened 2023-04-04
Port test data tests to Rust


randomPoison:
Currently we're running the test data tests by hacking the Meson test runner to invoke the rav1d Rust executable instead of the dav1d C executable. This is good enough for testing locally and in CI, but it's not ideal for a few reasons:

* Not idiomatic Rust, would be better to use `cargo test`.
* Doesn't automatically rebuild, you have to remember to run `cargo build --release` before running tests or they'll run against an out-of-date binary.
* Runs against release builds instead of debug builds, which makes debugging harder.
* We don't want to have to maintain the Meson scripts long term.
* It would also be good to have some examples of invoking rav1d from Rust, and writing Rust tests would be a good way of doing that.

We should port these tests over to the standard Rust test harness so that we can run our test suite with `cargo test`. Since there are 790 test files, this is probably best done by writing a one-off script to generate a test for each file, along with the correct MD5 checksums.

eli-schwartz:
Some of this could be solved, even in the short term, by having the meson.build files execute `cargo build`.

...

Meson also supports rust-based tests in the rust module: `rs_mod = import('rust'); rs_mod.test('name', exe_or_lib)`. This assumes you use meson's native support for rust, which should be quite suitable as long as you don't have a complex dependency tree of crates (which currently need to be manually wrapped) -- fortunately rav1d doesn't seem to have many so this may be practical.

One advantage of using meson here is that it actually supports installation steps.

thedataking:
Thanks @eli-schwartz for the input. I've implemented your suggestion to execute `cargo build` from meson in #153 as a stopgap solution. For a build system, meson is surprisingly reasonable to work with as someone with no prior exposure.

thedataking:
See https://github.com/memorysafety/rav1d/pull/605#issuecomment-1841981944 re. profile for cargo tests.

PR #327, opened 2023-07-12
Add transpiled seek_stress integration test

### LINKED ISSUE #76: Port seek stress tests
dav1d included a suite of seek-stress tests that use a custom runner to seek randomly within the test files. These tests are run against all of the `.ivf` test data files. Because these tests require a custom runner that links statically against `libdav1d`, we can't hack Meson to have it run against rav1d. We'll have to port the seek-stress runner to Rust and then set these tests up as Rust tests. The source code can be found at [`tests/seek_stress.c`](https://github.com/memorysafety/rav1d/blob/c2fb626189f313dd3ed8f1b74dbcfd80f333a43c/tests/seek_stress.c).

thedataking:
Closes #76. This is the easiest way to run all the stress tests so hopefully okay until we address #77. `seek_stress.rs` could be cleaned up (at least deleting all the duplicated types and externs) but that is probably easier to review in a separate PR.

### DIFF
--- a/.github/workflows/build-and-test-aarch64.yml
+++ b/.github/workflows/build-and-test-aarch64.yml
@@ -50,7 +50,8 @@ jobs:
             --platform linux/arm64 \
             ghcr.io/immunant/rav1d/debian-bullseye-aarch64:latest \
             .github/workflows/test.sh \
-                -r ../target/aarch64-unknown-linux-gnu/release/dav1d
+                -r ../target/aarch64-unknown-linux-gnu/release/dav1d \
+                -s ../target/aarch64-unknown-linux-gnu/release/seek_stress
       - name: upload build artifacts
         if: ${{ !cancelled() }}
         uses: actions/upload-artifact@v3
--- a/.github/workflows/build-and-test-arm7.yml
+++ b/.github/workflows/build-and-test-arm7.yml
@@ -50,7 +50,8 @@ jobs:
             --platform linux/arm/v7 \
             ghcr.io/immunant/rav1d/debian-bullseye-arm7:latest \
             .github/workflows/test.sh -t 2 \
-                -r ../target/armv7-unknown-linux-gnueabihf/release/dav1d
+                -r ../target/armv7-unknown-linux-gnueabihf/release/dav1d \
+                -s ../target/armv7-unknown-linux-gnueabihf/release/seek_stress
       - name: upload build artifacts
         if: ${{ !cancelled() }}
         uses: actions/upload-artifact@v3
--- a/.github/workflows/build-and-test-x86.yml
+++ b/.github/workflows/build-and-test-x86.yml
@@ -50,7 +50,10 @@ jobs:
           cargo build --target ${{ matrix.target }} ${{ matrix.build.flags }}
       - name: meson test for ${{ matrix.target }} ${{ matrix.build.name }}
         run: |
-          .github/workflows/test.sh -r ../target/${{ matrix.target }}/${{ matrix.build.name }}/dav1d -t ${{ matrix.build.timeout_multiplier }}
+          .github/workflows/test.sh \
+              -r ../target/${{ matrix.target }}/${{ matrix.build.name }}/dav1d \
+              -s ../target/${{ matrix.target }}/${{ matrix.build.name }}/seek_stress \
+              -t ${{ matrix.build.timeout_multiplier }}
           cp ${{ github.workspace }}/build/meson-logs/testlog.txt \
              ${{ github.workspace }}/build/meson-logs/testlog-${{ matrix.target }}-${{ matrix.build.name }}.txt
       - name: upload build artifacts
@@ -94,7 +97,8 @@ jobs:
           cargo build --release
       - name: meson test for x86_64-apple-darwin
         run: |
-          .github/workflows/test.sh -r ../target/release/dav1d
+          .github/workflows/test.sh -r ../target/release/dav1d \
+              -s ../target/release/seek_stress
           cp ${{ github.workspace }}/build/meson-logs/testlog.txt \
               ${{ github.workspace }}/build/meson-logs/testlog-x86_64-apple-darwin.txt
       - name: upload build artifacts
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -5,12 +5,14 @@ set -o pipefail
 
 timeout_multiplier=1
 rust_test_path=
+seek_stress_test_rust_path=
 debug_opt=
-while getopts t:r:d flag
+while getopts t:r:d:s: flag
 do
     case "${flag}" in
         t) timeout_multiplier=${OPTARG};;
-        r) rust_test_path=${OPTARG};;
+        r) rust_test_path="-Dtest_rust_path=${OPTARG}";;
+        s) seek_stress_test_rust_path="-Dseek_stress_test_rust_path=${OPTARG}";;
         d) debug_opt="-Ddebug=true";;
     esac
 done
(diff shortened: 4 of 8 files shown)


kkysen:
> This is the easiest way to run all the stress tests so hopefully okay until we address #77.

I'm not that familiar with the seek stress tests and how they're integrated with `meson` and everything, but would it be less work overall to just do #77 now?

thedataking:
> would it be less work overall to just do https://github.com/memorysafety/rav1d/issues/77 now?

I think this PR should land first since it increases test coverage and might catch errors we make elsewhere. #77 is a cleanup task which is sort of orthogonal to this PR. I suggest we first get test coverage up and then port all tests to Cargo once there are no more important tasks we need to do.

thedataking:
I'm investigating the test failures for the arm7 CI runs. It appears that it surfaces a bug in our port.

Repro (with `qemu-system-arm` etc. on x86_64 Linux):
```
$ cargo build --target armv7-unknown-linux-gnueabihf
$ RUST_BACKTRACE=1 target/armv7-unknown-linux-gnueabihf/release/seek_stress -q --threads 1 -i tests/dav1d-test-data/8-bit/data/00000169.ivf --muxer null
thread 'main' panicked at 'attempt to calculate the remainder with a divisor of zero', /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/core/src/num/mod.rs:1097:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/std/src/panicking.rs:578:5
   1: core::panicking::panic_fmt
             at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/core/src/panicking.rs:67:14
   2: core::panicking::panic
             at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/core/src/panicking.rs:117:5
   3: seek_stress::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
$ RUST_BACKTRACE=1 target/armv7-unknown-linux-gnueabihf/debug/seek_stress -q --threads 1 -i tests/dav1d-test-data/8-bit/data/00000169.ivf --muxer null
Failed to seek: Success
Failed to seek: Success
Failed to seek: Success
Failed to seek: Success
Failed to seek: Success
Failed to seek: Success
```

Error manifests differently in CI and between release and debug builds. Since aarch64 passes, this could be 32-bit related.

kkysen:
> ```
> $ cargo build --target armv7-unknown-linux-gnueabihf
> $ RUST_BACKTRACE=1 target/armv7-unknown-linux-gnueabihf/release/seek_stress -q --threads 1 -i tests/dav1d-test-data/8-bit/data/00000169.ivf --muxer null
> ```

Did you mean to `cargo build` without `--release` and then run `.../release/seek_stress`?  Does the error still happen if the binary is up-to-date?

thedataking:
> Does the error still happen if the binary is up-to-date?

Yup, but my repro instructions were wrong, should be right now. Best guess, something is wrong with`ivf.rs`.

kkysen:
Seems like it's this `.wrapping_rem` in `tests/seek_stress.rs`:

```rust
1060   │             pts = llround(
1061   │                 (xor128_rand() as libc::c_uint).wrapping_rem(total) as libc::c_double
1062   │                     * spf
1063   │                     * 1000000000.0f64,
1064   │             ) as uint64_t;
```

from

```shell
❯ RUST_BACKTRACE=1 target/armv7-unknown-linux-gnueabihf/release-with-debug-info/seek_stress -q --threads 1 -i tests/dav1d-test-data/8-bit/data/00000169.ivf --muxer null
thread 'main' panicked at 'attempt to calculate the remainder with a divisor of zero', /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/core/src/num/mod.rs:1097:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/std/src/panicking.rs:578:5
   1: core::panicking::panic_fmt
             at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/core/src/panicking.rs:67:14
   2: core::panicking::panic
             at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/core/src/panicking.rs:117:5
   3: core::num::<impl u32>::wrapping_rem
             at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/core/src/num/uint_macros.rs:1323:13
   4: seek_stress::main_0
             at ./tests/seek_stress.rs:1061:49
   5: seek_stress::main
             at ./tests/seek_stress.rs:1218:30
   6: core::ops::function::FnOnce::call_once
             at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

and from this C `tests/seek_stress.rs`:

```c
 196   │         pts = FRAME_OFFSET_TO_PTS(xor128_rand() % total);
```

Is it possible that `total` is uninitialized in the C version, but in the translated version, we explicitly zero-initialize `total`, and then when we do `% total`, we get this division by 0 error?  That seems like it could be a bug in the C code.

thedataking:
> Is it possible that `total` is uninitialized in the C version,

One way to find out is to add something like `assert(total);` in the C version after the call to `input_open(...)`.

kkysen:
There's UB somewhere at least.  `mut total` starts as `0`, then becomes `6` after `input_open(..., &mut total, ...)`.  Then after one `% total` with `total = 6`, `total` becomes `0` again and `% total` is run without calling `input_open` again.  `total` is only written there in `input_open`, so I added a `let total = total` after that to make it immutable, and then `seek_stress` segfaults instead.

thedataking:
It looks like `dav1d` compiles with `_FILE_OFFSET_BITS` set to 64 which means we need to use the 64-bit variants of `ftello` and `fseeko` (and operate on `libc::off64_t`) even on 32-bit builds. I'll put up a separate PR it that turns out to be the fix.

kkysen:
> It looks like `dav1d` compiles with `_FILE_OFFSET_BITS` set to 64 which means we need to use the 64-bit variants of `ftello` and `fseeko` (and operate on `libc::off64_t`) even on 32-bit builds. I'll put up a separate PR if that turns out to be the fix.

I'd guess that's not the fix, though that seems like it'll need fixing anyways.  The error here is super fragile.  It goes away if `total` is put into a `Box`, it goes away if I put `dbg!(total)`s in `seek` (the `seek` call is where `total` becomes a large number and then 0), or sometimes it prints infinite spaces if I put the `dbg!(total)`s in the wrong place.

I'm guessing `input_open` may keep a reference to the `total` pointer for longer, but I can't find where it does if it does.  I think there's also some weird interaction with the stack, as when `total` is on the heap it runs fine.  I also tried setting a breakpoint on writes but qemu said hardware breakpoints aren't supported.

thedataking:
> The error here is super fragile.

I still think these are 32-bit related problems. I'm fixing some of the same problems we had getting librav1d working for 32-bit. I'll push these as a separate commit. Not sure it makes sense for both of us to stare at this in the meanwhile.

kkysen:
> > The error here is super fragile.
> 
> I still think these are 32-bit related problems. I'm fixing some of the same problems we had getting librav1d working for 32-bit. I'll push these as a separate commit. Not sure it makes sense for both of us to stare at this in the meanwhile.

Okay, hopefully that works.  I already hit a dead-end mostly.

kkysen (approved):
LGTM.  I'm glad the 32-bit fixes were able to fix that weird crash.

PR #330, opened 2023-07-18
`fn padding`: Deduplicate w/ generics and cleanup

randomPoison:
Deduplicate `padding` into `looprestoration.rs` and make it generic over bitdepth.

### DIFF
--- a/src/looprestoration.rs
+++ b/src/looprestoration.rs
@@ -1,3 +1,4 @@
+use crate::include::common::bitdepth::BitDepth;
 use crate::include::stddef::ptrdiff_t;
 use crate::include::stdint::int16_t;
 use crate::include::stdint::uint32_t;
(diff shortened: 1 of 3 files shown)


src/looprestoration_tmpl_16.rs
```diff
@@ -1109,6 +938,9 @@ unsafe extern "C" fn wiener_filter_neon(
     edges: LrEdgeFlags,
     bitdepth_max: libc::c_int,
 ) {
+    use crate::src::looprestoration::LR_HAVE_BOTTOM;
```
kkysen, line 941: Can't these be top-level imports?

randomPoison: They can be, but they'd have to be conditionally imported to avoid warnings since they're only used within `#[cfg(arm)]` code. I figured importing them within the functions they're used was the cleaner approach, but I can make them top-level instead if you prefer.

kkysen: Ah, it's fine as is then.

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+// 256 * 1.5 + 3 + 3 = 390
+const REST_UNIT_STRIDE: usize = 390;
+
+// TODO Reuse p when no padding is needed (add and remove lpf pixels in p)
+// TODO Chroma only requires 2 rows of padding.
+// TODO(randomPoison): Temporarily pub until remaining looprestoration fns have
+// been deduplicated.
+#[inline(never)]
+pub(crate) unsafe fn padding<BD: BitDepth>(
+    mut dst: &mut [BD::Pixel; 70 /*(64 + 3 + 3)*/ * REST_UNIT_STRIDE],
```
kkysen, line 150: Is it this way in C?  Why not just use `(64 + 3 + 3)` instead `70`?

kkysen (commented):
Is this cleaned up only to the extent needed to work with `trait BitDepth`'s APIs?

src/looprestoration.rs
randomPoison: Yeah, this comment is there in a bunch of the C code so I reproduced in Rust.

randomPoison:
Yup, I didn't do any cleanup yet. I was planning on doing that after deduplication.

src/looprestoration.rs
kkysen: Is it something we want to clean up (when we clean up this function)?  I'm surprised `rustfmt` didn't eat this comment, as it has for other similar ones on array lengths.

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
     fn dav1d_wiener_filter7_16bpc_neon,
     fn dav1d_wiener_filter5_16bpc_neon,
 }
+
+// 256 * 1.5 + 3 + 3 = 390
+const REST_UNIT_STRIDE: usize = 390;
+
+// TODO Reuse p when no padding is needed (add and remove lpf pixels in p)
+// TODO Chroma only requires 2 rows of padding.
```
kkysen, line 145: Can you make these 2 from C doc comments?  I guess you can do that in the cleanup PR, though.  I usually add back comments in the cleanup PRs.

randomPoison: Oh, for this one I *did* cleanup `padding` as much as I could. That's why these C comments are here.

In general I'd prefer not to make TODOs doc comments. They're not meant to be part of the generated docs, and so should be regular comments.

src/looprestoration.rs
randomPoison: I was planning to leave the comment in here since it was in the original C. I think we're trying to preserve the C comments as much as possible, right?

src/looprestoration.rs
kkysen: Oh true.  For non TODO-like comments, those should be doc comments, right?

randomPoison: Well, they should be doc comments if they're meant to be part of the function's rendered docs. But sometimes you want to leave comments that are more of an implementation detail than part of the function's public API. Mostly this comes up with `TODO` comments, but sometimes it's appropriate to leave e.g. a `NOTE` comment that you don't want to be part of the public docs.

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+            BD::pixel_copy(&mut dst_l[REST_UNIT_STRIDE..], left, 3);
+            BD::pixel_copy(&mut dst_l[2 * REST_UNIT_STRIDE..], left, 3);
+        }
+    }
+
+    let mut dst_tl = &mut dst_l[3 * REST_UNIT_STRIDE..];
+    if edges & LR_HAVE_BOTTOM != 0 {
+        // Copy next loop filtered rows
+        let below_1 = std::slice::from_raw_parts(lpf.offset(6 * stride as isize), unit_w as usize);
+        let below_2 = std::slice::from_raw_parts(lpf.offset(7 * stride as isize), unit_w as usize);
```
kkysen, line 198:
```suggestion
        let lpf = std::slice::from_raw_parts(lpf, 7 * stride + unit_w as usize);
        let below_1 = &lpf[6 * stride...];
        let below_2 = &below_1[stride..];
```

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+    unit_w += 3 * have_left + 3 * have_right;
+    let mut dst_l = &mut dst[(3 * (have_left == 0) as libc::c_int) as usize..];
+    p = p.offset(-((3 * have_left) as isize));
+    lpf = lpf.offset(-((3 * have_left) as isize));
+
+    if edges & LR_HAVE_TOP != 0 {
+        // Copy previous loop filtered rows
+        let above_1 = std::slice::from_raw_parts(lpf, stride + unit_w as usize);
+        let above_2 = &above_1[stride..];
+
```
kkysen, line 174:
```suggestion
```

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+        BD::pixel_copy(dst_l, above_1, unit_w as usize);
+        BD::pixel_copy(&mut dst_l[REST_UNIT_STRIDE..], above_1, unit_w as usize);
+        BD::pixel_copy(&mut dst_l[2 * REST_UNIT_STRIDE..], above_2, unit_w as usize);
+    } else {
+        // Pad with first row
+        let p = std::slice::from_raw_parts(p, unit_w as usize);
+
```
kkysen, line 181:
```suggestion
```

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+        BD::pixel_copy(dst_l, p, unit_w as usize);
+        BD::pixel_copy(&mut dst_l[REST_UNIT_STRIDE..], p, unit_w as usize);
+        BD::pixel_copy(&mut dst_l[2 * REST_UNIT_STRIDE..], p, unit_w as usize);
+
```
kkysen, line 185:
```suggestion
```

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+            BD::pixel_copy(&mut dst_l[2 * REST_UNIT_STRIDE..], left, 3);
+        }
+    }
+
+    let mut dst_tl = &mut dst_l[3 * REST_UNIT_STRIDE..];
+    if edges & LR_HAVE_BOTTOM != 0 {
+        // Copy next loop filtered rows
+        let below_1 = std::slice::from_raw_parts(lpf.offset(6 * stride as isize), unit_w as usize);
+        let below_2 = std::slice::from_raw_parts(lpf.offset(7 * stride as isize), unit_w as usize);
+
```
kkysen, line 199:
```suggestion
```

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+            below_2,
+            unit_w as usize,
+        );
+    } else {
+        // Pad with last row
+        let src = std::slice::from_raw_parts(
+            p.offset(((stripe_h - 1) as isize * stride as isize) as isize),
+            unit_w as usize,
+        );
+
```
kkysen, line 221:
```suggestion
```

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+            &mut dst_tl[(stripe_h + 1) as usize * REST_UNIT_STRIDE..],
+            src,
+            unit_w as usize,
+        );
+        BD::pixel_copy(
+            &mut dst_tl[(stripe_h + 2) as usize * REST_UNIT_STRIDE..],
+            src,
+            unit_w as usize,
+        );
+
```
kkysen, line 237:
```suggestion
```

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+        if have_left != 0 {
+            let left = &(*left.offset((stripe_h - 1) as isize))[1..];
+
```
kkysen, line 240:
```suggestion
```

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+    if have_left == 0 {
+        // Pad 3x(STRIPE_H+6) with first column
+        for j in 0..stripe_h as usize + 6 {
+            let offset = j * REST_UNIT_STRIDE;
+            let val = dst[3 + offset];
+            BD::pixel_set(&mut dst[offset..], val, 3);
+        }
+    } else {
+        let dst = &mut dst[3 * REST_UNIT_STRIDE..];
+
```
kkysen, line 287:
```suggestion
```

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+            &mut dst_tl[(stripe_h + 2) as usize * REST_UNIT_STRIDE..],
+            below_2,
+            unit_w as usize,
+        );
+    } else {
+        // Pad with last row
+        let src = std::slice::from_raw_parts(
+            p.offset(((stripe_h - 1) as isize * stride as isize) as isize),
+            unit_w as usize,
+        );
```
kkysen, line 220:
```suggestion
        let p = std::slice::from_raw_parts(p, (stripe_h as usize - 1) * stride + unit_w as usize);
        let src = &p[(stripe_h as usize - 1) * stride..];
```

I think it's a bit better to convert raw pointers to slices as soon as possible, and then do the rest of the operations all safely.

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+// been deduplicated.
+#[inline(never)]
+pub(crate) unsafe fn padding<BD: BitDepth>(
+    mut dst: &mut [BD::Pixel; 70 /*(64 + 3 + 3)*/ * REST_UNIT_STRIDE],
+    mut p: *const BD::Pixel,
+    stride: ptrdiff_t,
+    mut left: *const [BD::Pixel; 4],
+    mut lpf: *const BD::Pixel,
+    mut unit_w: libc::c_int,
+    stripe_h: libc::c_int,
```
kkysen, line 156: Can you make `stride`, `unit_w`, `stripe_h` `usize`s?  I've been doing that, too.  It makes the code a lot cleaner vs. casts at use site, and these values aren't negative.

src/looprestoration.rs
```diff
@@ -136,3 +137,160 @@ decl_looprestorationfilter_fns! {
...
+            let mut pad = &mut dst_l[unit_w as usize + j * REST_UNIT_STRIDE..];
+            BD::pixel_set(pad, row_last, 3);
+        }
+    }
+
+    if have_left == 0 {
+        // Pad 3x(STRIPE_H+6) with first column
+        for j in 0..stripe_h as usize + 6 {
+            let offset = j * REST_UNIT_STRIDE;
+            let val = dst[3 + offset];
```
kkysen, line 282: I think this needs a note explaining why `dst[3 + offset]` is used instead of `dst_l[offset]`.  From what I see, `dst_l` can't be used since there'd be multiple mutable borrows of `dst`, and the 3 comes from `let mut dst_l = &mut dst[(3 * (have_left == 0) as libc::c_int) as usize..];`.  Since `have_left == 0` in this branch, `3 * (have_left ==0) as libc::c_int == 3 * 1 == 3`.  Also, can you make `dst` and `dst_l` non-`mut` (the variable, not the slice)?  Then it's clear those slices don't move and the recalculation of `dst_l` is simple.

kkysen (approved):
@randomPoison, ignore the previous comments for now.  Everything's correct, but since you were cleaning up `fn padding`, too, I added extra comments.  We can merge this PR first and do those cleanups in a separate PR.

Also, can you squash merge this so each commit stays CI-passing?

PR #333, opened 2023-07-19
Remove Option from looprestorationfilter_fn

randomPoison:


kkysen (approved):
I think you forgot to remove the `Some`s for arm code.  Otherwise it LGTM.

PR #332, opened 2023-07-18
`struct Dav1dLoopRestorationDSPContext`: Deduplicate via type-erased `fn` pointers

randomPoison:
Alternative to #323 that approaches the problem of deduplicating bitdepth-dependent fn pointers by type-erasing the bitdepth-dependent parts. To accomplish this, the relevant `extern "C"` declarations were moved into `looprestoration.rs`, where `pixel` has been defined as `c_void`. For the 8bpc versions, an extra `bitdepth_max: c_int` argument has been added so that the 8bpc and 16bpc fns have the same signature, with the understanding that this extra argument will simply be ignored by the called asm routine. For fn pointers that are defined in Rust code, type-erased wrappers have been added..

PR #334, opened 2023-07-20
`fn wiener_{c => rust}`: Deduplicate w/ generics

randomPoison:


src/looprestoration.rs
```diff
@@ -294,3 +298,106 @@ pub(crate) unsafe fn padding<BD: BitDepth>(
         }
     };
 }
+
+// TODO(randompoison): Temporarily public until init logic is deduplicated.
+pub(crate) unsafe extern "C" fn wiener_erased<BD: BitDepth>(
```
kkysen, line 303: Should this be named `wiener_c`?  Since part of its purpose is to be `extern "C"`, which `wiener_rust` doesn't need to.  But it's also doing the type erasure, so maybe it should be `wiener_c_erased`?

src/looprestoration.rs
```diff
@@ -294,3 +298,106 @@ pub(crate) unsafe fn padding<BD: BitDepth>(
...
+        lpf.cast(),
+        w,
+        h,
+        params,
+        edges,
+        bd,
+    )
+}
+
+unsafe extern "C" fn wiener_rust<BD: BitDepth>(
```
kkysen, line 328: We can drop the `extern "C"` now.  The `*_rust` `fn`s shouldn't need to be `extern "C"` at all.  You could save this for a PR cleaning up `wiener_rust`, though, or just do it now.

src/mc_tmpl_8.rs
```diff
@@ -2037,7 +2037,7 @@ unsafe extern "C" fn put_8tap_regular_c(
         mx as usize,
         my as usize,
         DAV1D_FILTER_8TAP_REGULAR | (DAV1D_FILTER_8TAP_REGULAR << 2),
-        BitDepth8::new(()),
+        BitDepth8::new(8),
```
kkysen, line 2040: We don't need to add this I don't think.  `trait BitDepth` already has `fn bitdepth` and `fn bitdepth_max` methods.

include/common/bitdepth.rs
```diff
@@ -90,7 +90,7 @@ pub trait BitDepth: Clone + Copy {
         + ToPrimitive<c_int>
         + ToPrimitive<c_uint>;
 
-    type BitDepthMax;
+    type BitDepthMax: Copy + FromPrimitive<c_int>;
```
kkysen, line 93: Where do we need these traits?

src/looprestoration.rs
```diff
@@ -294,3 +298,106 @@ pub(crate) unsafe fn padding<BD: BitDepth>(
...
+    stride: ptrdiff_t,
+    left: *const libc::c_void,
+    mut lpf: *const libc::c_void,
+    w: libc::c_int,
+    h: libc::c_int,
+    params: *const LooprestorationParams,
+    edges: LrEdgeFlags,
+    bitdepth_max: libc::c_int,
+) {
+    let bd = BD::new(bitdepth_max.as_());
```
kkysen, line 314: Can we instead add `fn from_c(bitdepth_max: c_int) -> Self` to `trait BitDepth`?  For `BitDepth8` this would just call `Self::new(())` and for `BitDepth16` it would call `Self::new(bitdepth_max.as_())`?  That way we keep a constant bitdepth for `BitDepth8`, as I just don't want that to accidentally interfere with any optimizations later on.

src/looprestoration.rs
randomPoison: I wanted to use the `_erased` naming convention that I started using in #332 for the type-erased wrapper functions. I thought more generally we were replacing the `_c` suffix with `_rust`?

kkysen: I was keeping the `*_c` for `fn`s that have to be `extern "C"`.  The `*_erased` convention is useful, too, though, so I think `wiener_c_erased` might be good here.

src/looprestoration.rs
randomPoison: Yup, that works too. Done!

src/looprestoration.rs

src/looprestoration.rs
randomPoison: I think that's a bit redundant, since `_erased` is only going to be used for functions that are `extern "C"` callbacks. Do we need that `_c` suffix?

kkysen: I think it's just useful for searching.

kkysen (approved):
Can you rename `wiener_erased` to `wiener_c_erased`?  And then also rebase onto #338 (changes `padding` call args (adds some casts)) and #339 (adds `BitDepth::from_c` with docs)?  With that, the rest of the changes here LGTM.

randomPoison:
I don't think it makes sense to rebase this on top of #338 and #339 (if I'm correctly understanding what you're asking). I'd rather that you land those as cleanup PRs after this one lands.

kkysen:
> I don't think it makes sense to rebase this on top of #338 and #339 (if I'm correctly understanding what you're asking). I'd rather that you land those as cleanup PRs after this one lands.

For #338 that's fine; order doesn't really matter.  I was hoping to do #339 first, though, since you've been squash merging and I wanted it as its own commit/PR.  It doesn't really matter so much so if you'd rather merge this first, that's okay, but I thought it'd make the history more understandable.

PR #335, opened 2023-07-20
`fn boxsum{3, 5}`: Deduplicate w/ generics

randomPoison:


src/looprestoration_tmpl_8.rs
```diff
@@ -134,144 +134,12 @@ use crate::src::looprestoration::padding;
 use crate::src::looprestoration::wiener_erased;
 use crate::src::looprestoration::Dav1dLoopRestorationDSPContext;
 use crate::src::looprestoration::LooprestorationParams;
+use crate::src::looprestoration::{boxsum3, boxsum5};
```
kkysen, line 137: Can you split these into one line each?  I've been doing that since there's a ton of imports usually and I find it easier to read that way when there are so many, plus the diffs are way cleaner.

kkysen: Same for 16bpc, too.

src/looprestoration.rs
```diff
@@ -401,3 +402,138 @@ unsafe fn wiener_rust<BD: BitDepth>(
         j_0 += 1;
     }
 }
+
+pub(crate) unsafe fn boxsum3<BD: BitDepth>(
```
kkysen, line 406: Temp comments?

kkysen (approved):
A few minor comments.  Otherwise, LGTM.

src/looprestoration_tmpl_8.rs
randomPoison: Done!

src/looprestoration.rs
randomPoison: Added!

PR #336, opened 2023-07-21
`fn selfguided_filter`: Deduplicate w/ generics

randomPoison:


src/looprestoration.rs
```diff
@@ -1,13 +1,17 @@
...
+use crate::include::common::intops::imax;
+use crate::include::common::intops::umin;
 use crate::include::stddef::ptrdiff_t;
 use crate::include::stdint::int16_t;
 use crate::include::stdint::int32_t;
 use crate::include::stdint::uint16_t;
 use crate::include::stdint::uint32_t;
 use crate::src::align::Align16;
 
+use super::tables::dav1d_sgr_x_by_x;
```
kkysen, line 13: Can you make it a `crate::` import and put it in the block with the rest?  I think it helps with consistency when there are so many imports generally.

src/looprestoration_tmpl_16.rs
```diff
@@ -353,15 +186,15 @@ unsafe extern "C" fn sgr_5x5_rust(
     let mut tmp: [pixel; 27300] = [0; 27300];
     let mut dst: [coef; 24576] = [0; 24576];
     padding::<BitDepth16>(&mut tmp, p, stride, left, lpf, w, h, edges);
-    selfguided_filter(
+    selfguided_filter::<BitDepth16>(
```
kkysen, line 189:
```suggestion
    selfguided_filter(
```

Shouldn't be needed since you specify `BitDepth16::from_c(bitdepth_max)` as an arg.

kkysen (approved):
A few minor comments.  Otherwise, LGTM.

src/looprestoration.rs

src/looprestoration_tmpl_16.rs
randomPoison: Done!

PR #338, opened 2023-07-23
`fn padding`: Further cleanup

kkysen:
These are mostly from my comments on #330 further cleaning up `fn padding`, plus a few other things.

I also think there are other cleanups we can do that would get rid of most of the bounds checking that was added to `fn padding`.  I'll wait until #334 merges first, though.


randomPoison (approved):
This all looks good, though this PR needs a rebase after my other ones merged.

PR #339, opened 2023-07-24
`trait BitDepth`: Add `fn from_c` to uniformly convert from C's repr (w/ docs)

kkysen:
This is the same as I suggested for ab5262ca84cdbfb5457e2a2b4328c6b119d9a7b4 in #334, but I wanted to add it as a separate PR and also add docs so it's clear why we're doing it this way.

### DIFF
--- a/include/common/bitdepth.rs
+++ b/include/common/bitdepth.rs
@@ -96,14 +96,14 @@ pub trait BitDepth: Clone + Copy {
 
     fn new(bitdepth_max: Self::BitDepthMax) -> Self;
 
-    /// While [`Self::new`] is the implementation specific way to construct a [`BitDepth`],
-    /// [`Self::from_c`] is a uniform way to construct a [`BitDepth`]
-    /// from its C representation, which is a [`c_int`].
+    /// While [`BitDepth::new`] is the implementation specific way to
+    /// construct a [`BitDepth`], [`BitDepth::from_c`] is a uniform way to
+    /// construct a [`BitDepth`] from its C representation, a [`c_int`].
     ///
-    /// Since we're deduplicating [`BitDepth`]-dependent `fn` ptr types through type erasure,
-    /// and always passing the `bitdepth_max` last argument
-    /// even for `8bpc` `fn`s where it's superfluous (it's constant),
-    /// we need to convert from that `bitdepth_max: c_int` arg back to a [`BitDepth`].
+    /// Since [`BitDepth`]-dependent `fn` ptr types use type erasure, they
+    /// always pass the `bitdepth_max` last argument (even for `8bpc``fn`s
+    /// where it's superfluous (it's constant)), so we need to convert from
+    /// that `bitdepth_max: c_int` arg back to a [`BitDepth`].
     fn from_c(bitdepth_max: c_int) -> Self;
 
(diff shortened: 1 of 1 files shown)


include/common/bitdepth.rs
```diff
@@ -96,6 +96,17 @@ pub trait BitDepth: Clone + Copy {
...
 
+    /// While [`Self::new`] is the implementation specific way to construct a [`BitDepth`],
+    /// [`Self::from_c`] is a uniform way to construct a [`BitDepth`]
+    /// from its C representation, which is a [`c_int`].
+    ///
+    /// Since we're deduplicating [`BitDepth`]-dependent `fn` ptr types through type erasure,
+    /// and always passing the `bitdepth_max` last argument
+    /// even for `8bpc` `fn`s where it's superfluous (it's constant),
+    /// we need to convert from that `bitdepth_max: c_int` arg back to a [`BitDepth`],
+    /// which is what this `fn` is for.
```
thedataking, line 107: > which is what this `fn` is for.

redundant. 

General comment: you should wrap long comments consistently to the text looks justified around col 80.

```
    /// While [`Self::new`] is the implementation specific way to construct a [`BitDepth`],
    /// [`Self::from_c`] is a uniform way to construct a [`BitDepth`] from its C representation
    /// representation, which is a [`c_int`].
    ///
    /// Since we're deduplicating [`BitDepth`]-dependent `fn` ptr types through type erasure,
    /// and always passing the `bitdepth_max` last argument even for `8bpc` `fn`s where
    /// it's superfluous (it's constant), we need to convert from that `bitdepth_max: c_int` 
    /// arg back to a [`BitDepth`],
```

> Since we're deduplicating

Comments should just document what the function does, not talk about the development process.

kkysen: > > which is what this `fn` is for.
> 
> redundant.

👍

> General comment: you should wrap long comments consistently to the text looks justified around col 80.

Oh, I always try to wrap them at semantic breaks.  That way you get much cleaner diffs when you edit the comments.  I can wrap them at 80 columns, but I'd prefer to wrap at semantic breaks rather than hard wrapping at word boundaries.

> Comments should just document what the function does, not talk about the development process.

I thought it's helpful to explain why we convert certain C constructs to different Rust ones, like this interface between the Rust API and the C way of doing it.

kkysen: > ```
>     /// While [`Self::new`] is the implementation specific way to construct a [`BitDepth`],
>     /// [`Self::from_c`] is a uniform way to construct a [`BitDepth`] from its C representation
>     /// representation, which is a [`c_int`].
>     ///
>     /// Since we're deduplicating [`BitDepth`]-dependent `fn` ptr types through type erasure,
>     /// and always passing the `bitdepth_max` last argument even for `8bpc` `fn`s where
>     /// it's superfluous (it's constant), we need to convert from that `bitdepth_max: c_int` 
>     /// arg back to a [`BitDepth`],
> ```

I wasn't sure if this was a suggestion since some of the lines are longer than 80, but I changed the comment to wrap at column 80 now.  And I tried to change the language so it's not talking about the development process, but more so how things are implemented.  Does that look better now?

randomPoison (approved):

PR #341, opened 2023-07-25
`fn dav1d_sgr_filter1_neon`: Deduplicate w/ generics

randomPoison:
This is the first bit of bitdepth deduplication that I've touched that needs to call into bitdepth-specific asm. ~~I used an extension trait to handle this. It's more boilerplate than I'd like, but that can probably be addressed with a macro later if needed. @kkysen let me know if you have a better suggestion.~~ An extension `trait` with associated `const`s in now used.  See https://github.com/memorysafety/rav1d/pull/341#pullrequestreview-1559981624 for an incoming follow-up.

### DIFF
--- a/src/looprestoration.rs
+++ b/src/looprestoration.rs
@@ -956,44 +956,37 @@ unsafe fn sgr_mix_rust<BD: BitDepth>(
     }
 }
 
+type fn_dav1d_sgr_box3_h_neon<BD> = unsafe extern "C" fn(
+    sumsq: *mut int32_t,
+    sum: *mut int16_t,
+    left: *const [<BD as BitDepth>::Pixel; 4],
+    src: *const <BD as BitDepth>::Pixel,
+    stride: ptrdiff_t,
+    w: libc::c_int,
+    h: libc::c_int,
+    edges: LrEdgeFlags,
+);
+
+type fn_dav1d_sgr_finish_filter1_neon<BD> = unsafe extern "C" fn(
+    tmp: *mut int16_t,
+    src: *const <BD as BitDepth>::Pixel,
+    stride: ptrdiff_t,
+    a: *const int32_t,
+    b: *const int16_t,
+    w: libc::c_int,
+    h: libc::c_int,
+);
+
 // TODO(randomPoison): Temporarily pub until all usages can be made private.
 #[cfg(all(feature = "asm", any(target_arch = "arm", target_arch = "aarch64")))]
 pub(crate) trait BitDepthLooprestorationArm: BitDepth {
-    fn dav1d_sgr_box3_h_neon(
-        sumsq: *mut int32_t,
-        sum: *mut int16_t,
-        left: *const [Self::Pixel; 4],
-        src: *const Self::Pixel,
-        stride: ptrdiff_t,
-        w: libc::c_int,
-        h: libc::c_int,
-        edges: LrEdgeFlags,
-    );
-
-    fn dav1d_sgr_finish_filter1_neon(
-        tmp: *mut int16_t,
-        src: *const Self::Pixel,
-        stride: ptrdiff_t,
-        a: *const int32_t,
-        b: *const int16_t,
-        w: libc::c_int,
-        h: libc::c_int,
-    );
+    const dav1d_sgr_box3_h_neon: fn_dav1d_sgr_box3_h_neon<Self>;
+    const dav1d_sgr_finish_filter1_neon: fn_dav1d_sgr_finish_filter1_neon<Self>;
 }
 
 #[cfg(all(feature = "asm", any(target_arch = "arm", target_arch = "aarch64")))]
 impl BitDepthLooprestorationArm for BitDepth8 {
-    #[inline(always)]
-    fn dav1d_sgr_box3_h_neon(
-        sumsq: *mut int32_t,
-        sum: *mut int16_t,
-        left: *const [Self::Pixel; 4],
-        src: *const Self::Pixel,
-        stride: ptrdiff_t,
-        w: libc::c_int,
-        h: libc::c_int,
-        edges: LrEdgeFlags,
-    ) {
+    const dav1d_sgr_box3_h_neon: fn_dav1d_sgr_box3_h_neon<Self> = {
         extern "C" {
             fn dav1d_sgr_box3_h_8bpc_neon(
                 sumsq: *mut int32_t,
@@ -1007,19 +1000,10 @@ impl BitDepthLooprestorationArm for BitDepth8 {
             );
         }
 
-        unsafe { dav1d_sgr_box3_h_8bpc_neon(sumsq, sum, left, src, stride, w, h, edges) }
-    }
+        dav1d_sgr_box3_h_8bpc_neon
+    };
 
-    #[inline(always)]
-    fn dav1d_sgr_finish_filter1_neon(
-        tmp: *mut int16_t,
-        src: *const Self::Pixel,
-        stride: ptrdiff_t,
-        a: *const int32_t,
-        b: *const int16_t,
-        w: libc::c_int,
-        h: libc::c_int,
-    ) {
+    const dav1d_sgr_finish_filter1_neon: fn_dav1d_sgr_finish_filter1_neon<Self> = {
         extern "C" {
             fn dav1d_sgr_finish_filter1_8bpc_neon(
                 tmp: *mut int16_t,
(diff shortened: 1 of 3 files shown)


randomPoison:
The other approach I can think of taking is to still use an extension trait, but make the items associated constants (instead of methods). That'd maybe make things less verbose, e.g. we wouldn't need to put `#[inline(always)]` everywhere at least.

src/looprestoration.rs
```diff
@@ -934,3 +955,194 @@ unsafe fn sgr_mix_rust<BD: BitDepth>(
...
+        w: libc::c_int,
+        h: libc::c_int,
+        edges: LrEdgeFlags,
+    ) {
+        extern "C" {
+            fn dav1d_sgr_box3_h_8bpc_neon(
+                sumsq: *mut int32_t,
+                sum: *mut int16_t,
+                left: *const [<BitDepth8 as BitDepth>::Pixel; 4],
+                src: *const <BitDepth8 as BitDepth>::Pixel,
```
kkysen, line 1002: I thought we were gonna make these type erased so that we didn't need to do the bitdepth trait thing.

randomPoison: In this case since the `extern "C"` declaration is inside of a bitdepth-specific trait impl, we don't need to do type erasure. I wish I could do that as `Self::Pixel` instead of `<BitDepth8 as BitDepth>::Pixel`, but Rust doesn't allow the former (because reasons). Not doing type erasure here doesn't matter much, though, it just saves a couple of extra casts 🤷

kkysen: Wouldn't type erasure fix the need to have a whole trait and impls?

Let me look more closely at the PR, though.  I've only looked at it cursorily so far.

randomPoison: The main reason for the trait is that we need to call functions with bitdepth-dependent names from a generic context. Type erasure is more related to being able to store function pointers in a bitdepth-generic way.

kkysen:
> The other approach I can think of taking is to still use an extension trait, but make the items associated constants (instead of methods). That'd maybe make things less verbose, e.g. we wouldn't need to put `#[inline(always)]` everywhere at least.

I think making them associated constants would be a lot cleaner.  There'd be much less repetition in that case, right?

Another potential alternative could be to unconditionally compile both the 8- and 16-bitdepth versions of all assembly functions.  Then a simple `match` on `BD::BITDEPTH` would suffice, and the unused functions should be optimized out by the linker.

src/looprestoration.rs
```diff
@@ -934,3 +955,194 @@ unsafe fn sgr_mix_rust<BD: BitDepth>(
...
+                stride: ptrdiff_t,
+                w: libc::c_int,
+                h: libc::c_int,
+                edges: LrEdgeFlags,
+            );
+        }
+
+        unsafe { dav1d_sgr_box3_h_8bpc_neon(sumsq, sum, left, src, stride, w, h, edges) }
```
kkysen, line 1010: We should propagate the `unsafe` here.  I've been putting safety comments on all `unsafe {}`s.

kkysen:
@randomPoison, I've prototyped a couple of alternatives:

* With both bitdepths always compiled and `BitDepth::BPC: BPC` added:
```rust
pub enum BPC {
    BPC8,
    BPC16,
}

unsafe fn dav1d_sgr_box3_h_neon<BD: BitDepth>(
    sumsq: *mut int32_t,
    sum: *mut int16_t,
    left: *const [BD::Pixel; 4],
    src: *const BD::Pixel,
    stride: ptrdiff_t,
    w: libc::c_int,
    h: libc::c_int,
    edges: LrEdgeFlags,
) {
    macro_rules! asm_fn {
        ($name:ident) => {{
            extern "C" {
                fn $name(
                    sumsq: *mut int32_t,
                    sum: *mut int16_t,
                    left: *const libc::c_void,
                    src: *const libc::c_void,
                    stride: ptrdiff_t,
                    w: libc::c_int,
                    h: libc::c_int,
                    edges: LrEdgeFlags,
                );
            }

            $name
        }};
    }
    (match BD::BPC {
        BPC::BPC8 => asm_fn!(dav1d_sgr_box3_h_8bpc_neon),
        BPC::BPC16 => asm_fn!(dav1d_sgr_box3_h_16bpc_neon),
    })(sumsq, sum, left.cast(), src.cast(), stride, w, h, edges)
}
```

* As associated constants:
```rust
type fn_dav1d_sgr_box3_h_neon<BD> = unsafe extern "C" fn(
    sumsq: *mut int32_t,
    sum: *mut int16_t,
    left: *const [<BD as BitDepth>::Pixel; 4],
    src: *const <BD as BitDepth>::Pixel,
    stride: ptrdiff_t,
    w: libc::c_int,
    h: libc::c_int,
    edges: LrEdgeFlags,
);

trait BitDepthLoopRestoration: BitDepth {
    const dav1d_sgr_box3_h_neon: fn_dav1d_sgr_box3_h_neon<Self>;
}

impl BitDepthLoopRestoration for BitDepth8 {
    const dav1d_sgr_box3_h_neon: fn_dav1d_sgr_box3_h_neon<Self> = {
        extern "C" {
            fn dav1d_sgr_box3_h_8bpc_neon(
                sumsq: *mut int32_t,
                sum: *mut int16_t,
                left: *const [<BitDepth8 as BitDepth>::Pixel; 4],
                src: *const <BitDepth8 as BitDepth>::Pixel,
                stride: ptrdiff_t,
                w: libc::c_int,
                h: libc::c_int,
                edges: LrEdgeFlags,
            );
        }

        dav1d_sgr_box3_h_8bpc_neon
    };
}

impl BitDepthLoopRestoration for BitDepth16 {
    const dav1d_sgr_box3_h_neon: fn_dav1d_sgr_box3_h_neon<Self> = {
        extern "C" {
            fn dav1d_sgr_box3_h_16bpc_neon(
                sumsq: *mut int32_t,
                sum: *mut int16_t,
                left: *const [<BitDepth16 as BitDepth>::Pixel; 4],
                src: *const <BitDepth16 as BitDepth>::Pixel,
                stride: ptrdiff_t,
                w: libc::c_int,
                h: libc::c_int,
                edges: LrEdgeFlags,
            );
        }

        dav1d_sgr_box3_h_16bpc_neon
    };
}
```

What do you think?

randomPoison:
I don't think always compiling both 8bpc and 16bpc functions is a good idea unless we're REALLY confident that the compiler will be able to eliminate the unused functions. I think the safest approach is to continue to conditionally compile them.

I do think the associated const approach seems cleaner, though, so I'll update my PR to use that approach. I especially like the way that you declare the `extern "C"` blocks inline 👍

src/looprestoration.rs
randomPoison: Switching to using associated consts also addresses this 👍

kkysen:
> I don't think always compiling both 8bpc and 16bpc functions is a good idea unless we're REALLY confident that the compiler will be able to eliminate the unused functions.

The `cc` crate we're using automatically compiles with `-ffunction-sections`, and the linker removes unused sections (I checked when compiling; the unused symbols are gone).  So I do think we can be quite confident this will happen.

> I do think the associated const approach seems cleaner, though, so I'll update my PR to use that approach. I especially like the way that you declare the `extern "C"` blocks inline 👍

I also checked if the associated const approach has no overhead, since it's calling through `fn` ptrs, which are normally indirect, but they're in a `const`:

[Godbolt](https://godbolt.org/z/Edn3xsd4Y)

```rust
extern "C" {
    fn s_asm(x: i32) -> i32;
}

// #[inline(never)]
// unsafe extern "C" fn s_asm(x: i32) -> i32 {
//     x + 10
// }

trait Trait {
    const ASM: unsafe extern "C" fn(x: i32) -> i32;

    unsafe fn asm(x: i32) -> i32;
}

struct S;

impl Trait for S {
    const ASM: unsafe extern "C" fn(x: i32) -> i32 = s_asm;

    unsafe fn asm(x: i32) -> i32 {
        s_asm(x)
    }
}

fn generic_const<T: Trait>(x: i32) -> i32 {
    unsafe {
        T::ASM(x) + T::ASM(x * 2)
    }
}

fn generic_fn<T: Trait>(x: i32) -> i32 {
    unsafe {
        T::asm(x) + T::asm(x * 2)
    }
}

pub fn concrete_const(x: i32) -> i32 {
    generic_const::<S>(x)
}

pub fn concrete_fn(x: i32) -> i32 {
    generic_fn::<S>(x)
}
```

Godbolt says they produce identical asm, so that approach should be all good.

kkysen:
Actually, the `match` and `enum BPC` way still works even without compiling both bitdepths' assembly.  The compiler can see the other `match` arm is impossible and so it doesn't try to link to the asm function in that case.

src/looprestoration.rs
```diff
@@ -934,3 +955,159 @@ unsafe fn sgr_mix_rust<BD: BitDepth>(
...
+            0 as *const [BD::Pixel; 4],
+            lpf.offset((6 * BD::pxstride(stride as usize)) as isize),
+            stride,
+            w,
+            2 as libc::c_int,
+            edges,
+        );
+    }
+    dav1d_sgr_box3_v_neon(sumsq, sum, w, h, edges);
+    dav1d_sgr_calc_ab1_neon(a, b, w, h, strength, bd.bitdepth_max().as_());
```
kkysen, line 1111:
```suggestion
    dav1d_sgr_calc_ab1_neon(a, b, w, h, strength, bd.bitdepth_max().into());
```

`.into()` is better since we know it's lossless, and we have `BitDepth::Pixel: Into<i32>`.

kkysen (approved):
As we discussed, I think the `match` way is a bit more clean and flexible vs. the associated `const` way.  It co-locates the bitdepth cases better, is more concise, and doesn't leave a `pub trait`, all while being just as optimized since the `match` is at compile time.  The associated `const` way is still good, though, and since all the PRs for cleaning up `looprestoration.rs` after this use that approach, they're good for now, and I'll switch them to the `match` way later (I already did it for this PR).  So this PR LGTM now, except for the small `.as_()` vs. `.into()` comment.  Can you fix that and then merge (normal merge is good, no squash necessary)?

PR #342, opened 2023-07-25
`MsacContext::symbol_adapt16`: Re-virtualize and simplify by passing extra args

kkysen:
eca5a4addd48b54b7005e040b82137ab2e1a66d6 reverts 4bea6a04b09ffae62530d04c46791f0f1f4384f3 and 908703ea1bc70a3de24e2c130fe72be6a64cb006 from #310.  The rest of the commits simplify the unsafe by passing the extra `cdf.len()` arg so an `assert!` isn't needed and the unsafety is extremely simpler, as well as passing `&mut` instead of `*mut`, which seems to work (if not, I can remove this commit).

### DIFF
--- a/src/msac.rs
+++ b/src/msac.rs
@@ -17,11 +17,13 @@ extern "C" {
         s: *mut MsacContext,
         cdf: *mut uint16_t,
         n_symbols: size_t,
+        _cdf_len: usize,
     ) -> libc::c_uint;
     fn dav1d_msac_decode_symbol_adapt16_sse2(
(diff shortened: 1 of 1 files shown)


randomPoison (approved):

thedataking (approved):
LGTM 🚀 

(I was initially puzzled by the different uses of safety syntaxes but it looks consistent with https://std-dev-guide.rust-lang.org/policy/safety-comments.html I'll assume we're all good.)

PR #344, opened 2023-07-30
threading: Fix a race on task_thread.init_done

fbossen:
Backports a fix from dav1d 1.2.1

### DIFF
--- a/src/thread_task.c
+++ b/src/thread_task.c
@@ -327,6 +327,7 @@ int dav1d_task_create_tile_sbrow(Dav1dFrameContext *const f, const int pass,
         f->task_thread.pending_tasks.tail->next = &tasks[0];
     f->task_thread.pending_tasks.tail = prev_t;
     atomic_store(&f->task_thread.pending_tasks.merge, 1);
+    atomic_store(&f->task_thread.init_done, 1);
     pthread_mutex_unlock(&f->task_thread.pending_tasks.lock);
 
     return 0;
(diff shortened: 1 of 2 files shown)


src/thread_task.c
```diff
@@ -730,14 +731,11 @@ void *dav1d_worker_task(void *data) {
                             dav1d_decode_frame_exit(f, DAV1D_ERR(ENOMEM));
                             f->n_tile_data = 0;
                             pthread_cond_signal(&f->task_thread.cond);
-                            atomic_store(&f->task_thread.init_done, 1);
-                            continue;
```
kkysen, line 734: Did you also delete the `continue` from the Rust version?  I don't see a continue there to begin with though.  Did it get turned into a state machine?

fbossen: This is a good question. In the C code, the `continue` statement doesn't seem to make a difference. It is inside a loop `for (int p = 1; p <= 2; p++)`. The code block with `continue` is executed only when `p==2` and there is no statement in the loop after the `continue` statement. Maybe the initial C to Rust translation detected this behaviour and did way with `continue`.

randomPoison (approved):

src/thread_task.c
kkysen: Yeah, probably.  The `continue` was deleted anyways, so it should all be good now.

kkysen (approved):

Issue #286, opened 2023-06-07
Backport performance optimizations from david 1.2.1


thedataking:
Release notes: https://code.videolan.org/videolan/dav1d/-/tags/1.2.1 

We probably want to ensure that the changes in 1.2.0 are back-ported first (tracked by #224 and #225) .

fbossen:
At this time, I believe only the following are missing:
- https://code.videolan.org/videolan/dav1d/-/commit/18dee9c72dfe6e4f6146667c83524b50af736148
- https://code.videolan.org/videolan/dav1d/-/commit/253e46604e7f655b31329247e8af58b6d46f527e

PR #345, opened 2023-07-31
Shrink intra_edge structs

fbossen:
Backport performance improvement from dav1d 1.2.1. Partially solves #286.

Use 8 bits to represent EdgeFlags instead of 32 bits


randomPoison (approved):

kkysen (approved):
I'm not sure why the original transpiled code was doing transmutes, but it all LGMT now.

PR #349, opened 2023-07-31
arm64: ipred: backport asm code from dav1d 1.2.1

fbossen:
Partially addresses #286


randomPoison (approved):
How did you go about translating the new code to Rust? Is this all hand-translated, or did you use C2Rust and diff?

fbossen:
> How did you go about translating the new code to Rust? Is this all hand-translated, or did you use C2Rust and diff?

By hand. Although I did copy a few things from similar functions that were generated using C2Rust.

PR #350, opened 2023-08-01
`src/cdf.rs`: Revert CDF macro expansion with const functions

barrbrain:
This is a pretty large change, so the first few commits separate the transformation steps.


src/cdf.rs
```diff
@@ -710,6 +710,49 @@ use crate::src::levels::N_MV_JOINTS;
 use crate::include::common::intops::imin;
 use crate::src::levels::N_COMP_INTER_PRED_MODES;
 use crate::src::r#ref::dav1d_ref_inc;
+const fn cdf0d<const P: usize, const N: usize>(probs: [u16; P]) -> [u16; N] {
+    assert!(P < N);
+    let mut cdf0d = [0; N];
+    let mut i = 0;
+    while i < P {
+        cdf0d[i] = (32768 - probs[i] as u32) as u16;
```
barrbrain:
```suggestion
        cdf0d[i] = probs[i].wrapping_neg();
```

thedataking (approved):
This improves readability a lot, thanks @barrbrain!

I'm fine merging this but it would be good if @randomPoison had a chance to take a look too.

src/cdf.rs
thedataking: It isn't clear to me that using `wrapping_neg` is equivalent to the previous computation. Perhaps that's why the comment shows as outdated.

barrbrain: Yeah, I mixed up the range of u16 and i16. It turns out the conversions are unnecessary.

barrbrain: I'm considering this small change in behavior so that the probability range is enforced and zero values map back to zero. I think that where my code currently differs from the original C code.

src/cdf.rs
```diff
@@ -753,699 +753,678 @@ const fn cdf3d<const P: usize, const N: usize, const M: usize, const L: usize, c
     }
     cdf3d
 }
-pub const fn av1_default_cdf() -> CdfModeContext {
-    let mut init = CdfModeContext {
-        y_mode: Align32(cdf1d([
+static mut av1_default_cdf: CdfModeContext = CdfModeContext {
```
kkysen: Why do we need/want `static mut`s here instead of a `const fn`?  `static mut`s are `unsafe` to access.

barrbrain: I was simply matching existing code patterns. I'm happy to drop the `mut` from all of the constants.

randomPoison (approved):
This looks good to me! A big improvement in readability, thanks @barrbrain ❤️

kkysen (commented):
If the `static`s can be immutable then that's great; it's just an issue if they're `static mut`s.  I think I've changed most of the other `static mut`s to just `static` (they're originally `static const` in C, but `c2rust transpile` misses keeping the `const`), though I may have missed some.

I also checked the compile times, as when I tried changing some things in `cdf.rs` before, there were large differences in compile time for some reason since the file is so large, but there is little to no regression in compile times with this PR, so it looks good on that front.

Thanks for the cleanup!

The explicit `Align*`s look better now, too.

One remaining question: the original macros were named `CDF[0-9]`, but you named their `const fn` equivalents `cdf[0-9]d`.  Is the `d` for dimension?  If so, it seems weird that `0d` creates a 1-D array, `1d` creates a 2-D array, and so on.  Does the `d` mean something else?

barrbrain:
> One remaining question: the original macros were named `CDF[0-9]`, but you named their `const fn` equivalents `cdf[0-9]d`. Is the `d` for dimension? If so, it seems weird that `0d` creates a 1-D array, `1d` creates a 2-D array, and so on. Does the `d` mean something else?

The `d` is for dimension. In my early drafts, I named the functions `cdf`, `cdfs`, `cdfs2d`, `cdfs3d`. My logic being that a CDF is a 1-D array and the inputs of the later functions are N-D arrays of CDFs. I see that without naming the CDF types, this is unclear.

kkysen:
> The `d` is for dimension. In my early drafts, I named the functions `cdf`, `cdfs`, `cdfs2d`, `cdfs3d`. My logic being that a CDF is a 1-D array and the inputs of the later functions are N-D arrays of CDFs. I see that without naming the CDF types, this is unclear.

Ah, that makes sense.  Do you think we should add something like `type CDF<const N: usize> = [u16; N];` and use that?  Otherwise, a comment before the `cdf*` `const fn`s explaining that should suffice.

PR #353, opened 2023-08-03
`src/qm.rs`: Tidy static tables

barrbrain:
Remove no-op cast `as libc::c_int as uint8_t` from cell values and run `cargo fmt`.
Remove `mut` from tables not modified in `dav1d_init_qm_tables`.


kkysen (approved):
Thanks again!  That's a lot of lines removed; it's awesome!

PR #354, opened 2023-08-03
`src/scan.rs`: Tidy static tables

barrbrain:
Remove no-op cast `as libc::c_int as uint16_t` from cell values and run `cargo fmt`.
Remove `mut` from static tables which are never modified. Use Rust `u16` type, equivalent to `uint16_t`.
Remove unnecessary `unsafe` block from `dav1d_scans`.


kkysen (approved):
LGMT.  Thanks!

In the future, we can also make `dav1d_scans` contain `&'static [u16]`s instead of `*const u16`s.  The presence of the raw pointer in the `static`, even though it's `const`, is what's still making it `unsafe`.

PR #362, opened 2023-08-07
Revert broken commit

randomPoison:
Reverts commit 49035f4d4142b3a7f0a53f7dd966437d1f6b2c43, which broke the argon tests. I'm not sure what exactly about these changes caused the build to break, but I've confirmed that the argon tests pass with these changes reverted.


kkysen (approved):
I'll take a look at this when I'm back.  It's probably a casting issue.  I think I made some assumptions based on the values used by the other tests that aren't true for the argon tests.

fbossen:
I submitted #363 to fix an overflow issue where a sum of u8 values could exceed the u8 range.

PR #363, opened 2023-08-08
Fix overflow issue introduced in 49035f4

fbossen:
Intermediate values may exceed u8 range. Use u16 instead. Also modify add to saturating add to avoid possibiliy of overflow (unconfirmed).

### DIFF
--- a/src/mc.rs
+++ b/src/mc.rs
@@ -866,8 +866,10 @@ pub unsafe fn w_mask_rust<BD: BitDepth>(
     {
         let mut x = 0;
         while x < w {
-            let m =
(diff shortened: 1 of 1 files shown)


kkysen (commented):
The `u16` intermediary looks like it could definitely be the issue.  Thanks for figuring that out.  The saturating add I'm less sure about though; the original code didn't do anything like that.  If you think it's more correct, though, that's fine.  But there are lots of other similar adds in other functions that aren't saturating adds either.

randomPoison (approved):
This looks good, let's try landing this before #362 to see if we can avoid fully reverting the changes.

PR #366, opened 2023-08-08
Deduplicate `itx_tmpl` with generics

endbr64:


randomPoison (approved):

src/itx.rs
```diff
@@ -0,0 +1,141 @@
...
+    eob: libc::c_int,
+    w: libc::c_int,
+    h: libc::c_int,
+    shift: libc::c_int,
+    first_1d_fn: itx_1d_fn,
+    second_1d_fn: itx_1d_fn,
+    has_dconly: libc::c_int,
+    bd: BD,
+) {
+    let bitdepth_max: libc::c_int = bd.bitdepth_max().to_prim();
```
kkysen, line 29: We shouldn't need to use the `.{to,from}_prim()` methods.  I think the `.as_()` method suffices (let me know if it doesn't work) and is the one meant to be used, whereas the prim ones are more for the implementation.

src/itx_tmpl_16.rs
```diff
@@ -4975,113 +4977,19 @@ unsafe extern "C" fn inv_txfm_add_c(
...
-        y_0 += 1;
-        c = c.offset(w as isize);
-    }
-    memset(
-        coeff as *mut libc::c_void,
-        0 as libc::c_int,
-        (::core::mem::size_of::<coef>() as libc::c_ulong)
-            .wrapping_mul(sw as libc::c_ulong)
-            .wrapping_mul(sh as libc::c_ulong),
+    return inv_txfm_add_rust(
```
kkysen, line 4980: nit: Don't need a `return` for the last statement in a function.

src/itx.rs
```diff
@@ -0,0 +1,141 @@
...
+use crate::include::stddef::*;
+use crate::include::stdint::*;
+
+extern "C" {
+    fn memset(_: *mut libc::c_void, _: libc::c_int, _: libc::c_ulong) -> *mut libc::c_void;
+}
+pub type itx_1d_fn =
+    Option<unsafe extern "C" fn(*mut int32_t, ptrdiff_t, libc::c_int, libc::c_int) -> ()>;
+
+pub unsafe extern "C" fn inv_txfm_add_rust<BD: BitDepth>(
```
kkysen, line 16: This only deduplicates this function in the module, right?  Could you name the commit and PR something like "`fn inv_txfm_add_rust`: Deduplicate with generics" for the next ones?

endbr64: Will do, thanks.

PR #367, opened 2023-08-09
`looprestoration`: Dedup init fns w/ generics

randomPoison:
Deduplicate the init functions using the `BPC` enum matching approach introduced in #355. I used a macro to generate the wrapper functions to cut down on boilerplate. There a couple of asm functions that don't have both 8bpc and 16bpc versions, so those are handled differently.

### DIFF
--- a/src/looprestoration.rs
+++ b/src/looprestoration.rs
@@ -1,6 +1,6 @@
 use crate::include::common::bitdepth::AsPrimitive;
 use crate::include::common::bitdepth::BitDepth;
-#[cfg(all(feature = "asm", any(target_arch = "arm", target_arch = "aarch64")))]
+#[cfg(feature = "asm")]
 use crate::include::common::bitdepth::BPC;
 use crate::include::common::intops::iclip;
 use crate::include::common::intops::imax;
@@ -8,6 +8,7 @@ use crate::include::common::intops::umin;
 use crate::include::stddef::ptrdiff_t;
 use crate::include::stdint::int16_t;
 use crate::include::stdint::int32_t;
+#[cfg(all(feature = "asm", target_arch = "arm"))]
 use crate::include::stdint::intptr_t;
 use crate::include::stdint::uint16_t;
 use crate::include::stdint::uint32_t;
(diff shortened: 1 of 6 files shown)


fbossen (approved):
LGTM

src/looprestoration.rs
```diff
@@ -1719,3 +1741,130 @@ unsafe extern "C" fn sgr_filter_mix_neon<BD: BitDepth>(
...
+        c.wiener[0] = dav1d_wiener_filter7_avx512icl::<BD>;
+        c.wiener[1] = match BD::BPC {
+            // With VNNI we don't need a 5-tap version.
+            BPC::BPC8 => c.wiener[0],
+
+            #[cfg(feature = "bitdepth_16")]
+            BPC::BPC16 => dav1d_wiener_filter5_16bpc_avx512icl,
+
+            #[cfg(not(feature = "bitdepth_16"))]
+            BPC::BPC16 => unreachable!(),
```
kkysen, line 1806: I think we can do this without an `unreachable!` arm.  For my previous matches on `BPC`, I didn't have a cfg and it still handled everything correctly when only compiling for one bitdepth.

src/looprestoration.rs
```diff
@@ -58,93 +59,127 @@ pub struct Dav1dLoopRestorationDSPContext {
...
-                    bitdepth_max: libc::c_int,
-                );
-            )*
+/// Generates a generic wrapper function that delegates to the appropriate
+/// bitdepth-specific extern function.
+macro_rules! decl_looprestorationfilter_bd_fn {
+    ($name:ident, $suffix:ident) => {
+        paste::paste! {
+            #[cfg(feature = "asm")]
+            unsafe extern "C" fn [<$name $suffix>]<BD: BitDepth>(
```
kkysen, line 68: We need to check the perf of this extra call.  The asm function won't be inlined unless we have cross language LTO for asm (not sure if we can; it'd be great if we can, though) or something like BOLT probably.  We should be able to do this without this extra call, though.  When I'm back I can open a PR for a way to do this.

src/looprestoration.rs
```diff
@@ -58,93 +59,127 @@ pub struct Dav1dLoopRestorationDSPContext {
...
+                        }
+                        [<$name _16bpc $suffix>]
+                    }
+                })(dst, dst_stride, left, lpf, w, h, params, edges, bitdepth_max)
+            }
         }
     };
 }
 
+decl_looprestorationfilter_bd_fn!(dav1d_wiener_filter7, _ssse3);
```
kkysen, line 119: If we don't globally declare these but leave them as local expressions in the init function, then we won't need the wrapper functions.

src/looprestoration.rs
```diff
@@ -1719,3 +1741,130 @@ unsafe extern "C" fn sgr_filter_mix_neon<BD: BitDepth>(
...
+        c.sgr[0] = sgr_filter_5x5_neon_erased::<BD>;
+        c.sgr[1] = sgr_filter_3x3_neon_erased::<BD>;
+        c.sgr[2] = sgr_filter_mix_neon_erased::<BD>;
+    }
+}
+
+#[cold]
+pub fn dav1d_loop_restoration_dsp_init<BD: BitDepth>(
+    c: &mut Dav1dLoopRestorationDSPContext,
+    _bpc: libc::c_int,
```
kkysen, line 1854:
```suggestion
    bpc: libc::c_int,
```

randomPoison:
Woops, I somehow missed that @kkysen had added feedback here and merged the PR before addressing those comments. I'll open up another PR to do that.

kkysen:
I think I made comments after you merged it already.  But you had an older PR for the same thing but before the match way, so I think I had some comments on that one that you fixed some of but there were still a couple comments I had, plus some new ones I think.

PR #368, opened 2023-08-09
`fn wiener_rust`: Cleanup generated code

randomPoison:


src/looprestoration.rs
```diff
@@ -372,64 +376,49 @@ unsafe fn wiener_rust<BD: BitDepth>(
...
-        let mut i = 0;
-        while i < w {
-            let mut sum = (1 as libc::c_int) << bitdepth + 6;
+    let round_bits_h = 3 + (bitdepth == 12) as libc::c_int * 2;
+    let rounding_off_h = 1 << round_bits_h - 1;
+    let clip_limit = 1 << bitdepth + 1 + 7 - round_bits_h;
+    let chunks = tmp
+        .chunks_exact(REST_UNIT_STRIDE)
+        .zip(hor.chunks_exact_mut(REST_UNIT_STRIDE));
+    for (_, (tmp, hor)) in (0..h + 6).zip(chunks) {
```
kkysen, line 391: A slice plus enumerate would be better.  Simpler, more efficient implementation I'm pretty sure.

src/looprestoration.rs
```diff
@@ -372,64 +376,49 @@ unsafe fn wiener_rust<BD: BitDepth>(
...
-    let mut j = 0;
-    while j < h + 6 {
-        let mut i = 0;
-        while i < w {
-            let mut sum = (1 as libc::c_int) << bitdepth + 6;
+    let round_bits_h = 3 + (bitdepth == 12) as libc::c_int * 2;
+    let rounding_off_h = 1 << round_bits_h - 1;
+    let clip_limit = 1 << bitdepth + 1 + 7 - round_bits_h;
+    let chunks = tmp
+        .chunks_exact(REST_UNIT_STRIDE)
```
kkysen, line 389: I added a `.ext_as_chunks()` method, which is the unstable std method added in our code to make it stable.  It's an array version of chunks basically, so there won't be bounds checking on the array chunks.

kkysen: Generally I'd recommend splitting these kinds of changes.  There's the basic cleanup ones that are unlikely to change any semantics or cause errors, then there's cast elision which in certain cases could lead to overflow errors, there's conversion to safe references which is straightforward usually, there's conversion to safe slices, which is more complex when we have to guess the size, and then there's these bounds checking optimizations.  I think it's better to split them into separate commits since there's very different chances of errors for each.  IMO it's also easier to follow the changes that way.

randomPoison: Is `ext_as_chunks` in the project? I'm not seeing it. Or do you mean that you plan to add it?

src/looprestoration.rs
kkysen: Sorry it's hard to type on my phone lol.  `chunks[..h + 6].iter().enumerate()`.

src/looprestoration.rs
kkysen: Oh, I was in the middle of a PR adding it.  I think I didn't get to opening it yet.

randomPoison: Okay then let's hold off on updating this PR to use `as_chunks`. We can come back and clean this up after you've had a chance to add a stable version of `as_chunks`, but I don't don't want to use unstable features in the meantime.

src/looprestoration.rs
randomPoison: This requires `as_chunks` right? I don't want to use unstable features so I can't use that approach for now. Is this something we can cleanup later?

kkysen: Oh I misread the iterator code.  I thought `chunks` was a slice.  In that case you can do `chunks.iter().take(h + 6).enumerate()`.

src/looprestoration.rs
```diff
@@ -373,25 +372,30 @@ unsafe extern "C" fn wiener_c_erased<BD: BitDepth>(
...
-    params: *const LooprestorationParams,
+    params: &LooprestorationParams,
     edges: LrEdgeFlags,
     bd: BD,
 ) {
-    let mut tmp: [BD::Pixel; 27300] = [0.as_(); 27300];
-    let mut tmp_ptr: *mut BD::Pixel = tmp.as_mut_ptr();
+    // Wiener filtering is applied to a maximum stripe height of 64 + 3 pixels
+    // of padding above and below
+    let mut tmp = [0.as_(); 70 /*(64 + 3 + 3)*/ * REST_UNIT_STRIDE];
```
kkysen, line 398:
```suggestion
    let mut tmp = [0.into(); 70 /*(64 + 3 + 3)*/ * REST_UNIT_STRIDE];
```

src/looprestoration.rs
```diff
@@ -404,64 +408,49 @@ unsafe fn wiener_rust<BD: BitDepth>(
...
-    let round_bits_h = 3 as libc::c_int + (bitdepth == 12) as libc::c_int * 2;
-    let rounding_off_h = (1 as libc::c_int) << round_bits_h - 1;
-    let clip_limit = (1 as libc::c_int) << bitdepth + 1 + 7 - round_bits_h;
-    let mut j = 0;
-    while j < h + 6 {
-        let mut i = 0;
-        while i < w {
-            let mut sum = (1 as libc::c_int) << bitdepth + 6;
+    // Values stored between horizontal and vertical filtering don't
+    // fit in a uint8_t.
```
kkysen, line 412:
```suggestion
    // fit in a u8.
```

src/looprestoration.rs
```diff
@@ -404,64 +408,49 @@ unsafe fn wiener_rust<BD: BitDepth>(
...
+    let mut hor = [0u16; 70 /*(64 + 3 + 3)*/ * REST_UNIT_STRIDE];
```
kkysen, line 413:
```suggestion
    let mut hor = [0; 70 /*(64 + 3 + 3)*/ * REST_UNIT_STRIDE];
```

Doesn't seem necessary.

src/looprestoration.rs
```diff
@@ -404,64 +408,49 @@ unsafe fn wiener_rust<BD: BitDepth>(
...
+    let round_bits_h = 3 + (bitdepth == 12) as libc::c_int * 2;
+    let rounding_off_h = 1 << round_bits_h - 1;
+    let clip_limit = 1 << bitdepth + 1 + 7 - round_bits_h;
+    let chunks = tmp
+        .chunks_exact(REST_UNIT_STRIDE)
+        .zip(hor.chunks_exact_mut(REST_UNIT_STRIDE));
+    for (tmp, hor) in chunks.take((h + 6) as usize) {
+        for i in 0..w as usize {
+            let mut sum = 1 << bitdepth + 6;
+
```
kkysen, line 426:
```suggestion
```

To match the C spacing.

src/looprestoration.rs
```diff
@@ -404,64 +408,49 @@ unsafe fn wiener_rust<BD: BitDepth>(
...
+            if BD::BPC == BPC::BPC8 {
+                sum += tmp[i + 3].as_::<libc::c_int>() * 128;
```
kkysen, line 428:
```suggestion
                sum += tmp[i + 3].into() * 128;
```

The `Pixel` to `i32` conversion is infallible so we have `Pixel: Into<i32>`.

src/looprestoration.rs
```diff
@@ -404,64 +408,49 @@ unsafe fn wiener_rust<BD: BitDepth>(
...
             }
 
-            let mut k = 0;
-            while k < 7 {
-                sum += (*tmp_ptr.offset((i + k) as isize)).as_::<libc::c_int>()
-                    * (*filter.offset(0))[k as usize] as libc::c_int;
-                k += 1;
+            for k in 0..7 {
+                sum += tmp[i + k].as_::<libc::c_int>() * filter[0][k] as libc::c_int;
```
kkysen, line 432:
```suggestion
                sum += tmp[i + k].into() * filter[0][k] as libc::c_int;
```

src/looprestoration.rs
```diff
@@ -404,64 +408,49 @@ unsafe fn wiener_rust<BD: BitDepth>(
...
             }
```
kkysen, line 433:
```suggestion
            for (&tmp, &filter) in std::iter::zip(&tmp[i..][..7], &filter[0][..7]) {
                sum += tmp.into() * filter as libc::c_int;
            }
```

Bounds checking elision.

kkysen: We can save this for later, though.

src/looprestoration.rs
```diff
@@ -404,64 +408,49 @@ unsafe fn wiener_rust<BD: BitDepth>(
...
-                sum += (*tmp_ptr.offset((i + 3) as isize)).as_::<libc::c_int>() * 128;
+    let filter = &params.filter.0;
+    let bitdepth = bd.bitdepth().as_::<libc::c_int>();
+    let round_bits_h = 3 + (bitdepth == 12) as libc::c_int * 2;
+    let rounding_off_h = 1 << round_bits_h - 1;
+    let clip_limit = 1 << bitdepth + 1 + 7 - round_bits_h;
+    let chunks = tmp
+        .chunks_exact(REST_UNIT_STRIDE)
+        .zip(hor.chunks_exact_mut(REST_UNIT_STRIDE));
+    for (tmp, hor) in chunks.take((h + 6) as usize) {
```
kkysen, line 423:
```suggestion
    for (tmp, hor) in tmp
        .chunks_exact(REST_UNIT_STRIDE)
        .zip(hor.chunks_exact_mut(REST_UNIT_STRIDE))
        .take((h + 6) as usize)
    {
```

I think having it all as one expression is easier to read.  I don't usually see intermediate iterator variables very much.

kkysen (approved):
I left a few small comments.  After those it LGMT.  There's still a bit more cleanup we can do here, but those are around cast elision and bounds-checking elision, not making things safe, so we can leave it for later.

fbossen (approved):

src/looprestoration.rs
randomPoison: I don't think we need to worry about matching the C spacing exactly. In some places I'm going to insert some extra lines for readability.

PR #371, opened 2023-08-10
Minor cleanup in `itx_tmpl`

endbr64:
Addressing comments from @kkysen after #366 was already merged.

### DIFF
--- a/src/itx.rs
+++ b/src/itx.rs
@@ -26,7 +26,7 @@ pub unsafe extern "C" fn inv_txfm_add_rust<BD: BitDepth>(
     has_dconly: libc::c_int,
     bd: BD,
 ) {
-    let bitdepth_max: libc::c_int = bd.bitdepth_max().to_prim();
+    let bitdepth_max: libc::c_int = bd.bitdepth_max().as_();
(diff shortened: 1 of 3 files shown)


src/itx.rs
```diff
@@ -40,8 +40,8 @@ pub unsafe extern "C" fn inv_txfm_add_rust<BD: BitDepth>(
     let is_rect2: libc::c_int = (w * 2 == h || h * 2 == w) as libc::c_int;
     let rnd = (1 as libc::c_int) << shift >> 1;
     if eob < has_dconly {
-        let mut dc: libc::c_int = BD::Coef::to_prim(*coeff.offset(0));
-        *coeff.offset(0) = BD::Coef::from_prim(0);
+        let mut dc: libc::c_int = BD::Coef::as_(*coeff.offset(0));
```
kkysen, line 43: It should also always work in suffix position like `as` does, though you may need an extra parentheses sometimes.

kkysen (approved):

src/itx.rs
endbr64: Thanks, done.

PR #372, opened 2023-08-11
`fn selfguided_filter`: Cleanup w/ generics and introduce CursorMut

randomPoison:
In order to cleanup `selfguided_filter`, I'd like to introduce a new `CursorMut` type that allows us to safely support negative indexing into arrays. `CursorMut` wraps a slice and an index, allowing you to sub-slice an array without forgetting what the original allocation was. This allows us to do negative indexing relative to the current index as long as the final index is still within bounds of the original allocation. I suspect this pattern will be useful as we cleanup more code and find where blocks of memory are allocated.


src/cursor.rs
```diff
@@ -0,0 +1,116 @@
...
+        &mut self.data[self.index + index]
+    }
+}
+
+impl<'a, T> Index<isize> for CursorMut<'a, T> {
+    type Output = T;
+
+    fn index(&self, index: isize) -> &Self::Output {
+        let index = self.index as isize + index;
+        assert!(index > 0 && (index as usize) < self.data.len());
```
kkysen, line 91: `>= 0`?

kkysen: The `< self.data.len()` check should be handled by the indexing, so I don't think we need to repeat it.

src/cursor.rs
```diff
@@ -0,0 +1,116 @@
...
+        &mut self.data[index as usize]
+    }
+}
+
+impl<'a, T> Index<i32> for CursorMut<'a, T> {
+    type Output = T;
+
+    fn index(&self, index: i32) -> &Self::Output {
+        &self[index as isize]
```
kkysen, line 108: Why is this indexing directly and not with the `index` offset?

src/looprestoration.rs
```diff
@@ -601,185 +605,145 @@ unsafe fn boxsum5<BD: BitDepth>(
...
+    let mut sumsq = [0; 68 /*(64 + 2 + 2)*/ * REST_UNIT_STRIDE];
+    // By inverting A and B after the boxsums, B can be of size coef instead
+    // of int32_t
+    let mut sum = [0.as_::<BD::Coef>(); 68 /*(64 + 2 + 2)*/ * REST_UNIT_STRIDE];
+
     let step = (n == 25) as libc::c_int + 1;
     if n == 25 {
-        boxsum5::<BD>(sumsq.as_mut_ptr(), sum.as_mut_ptr(), src, w + 6, h + 6);
+        // TODO: Update `boxsum5` to take safe arguments.
+        unsafe {
```
kkysen, line 630: Can you add safety comments for all `unsafe` blocks?

src/cursor.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    pub fn as_ptr(&self) -> *const T {
+        self.as_slice().as_ptr()
+    }
+
+    pub fn as_mut_ptr(&mut self) -> *mut T {
+        self.as_mut_slice().as_mut_ptr()
+    }
+
+    pub fn clone<'b>(&'b mut self) -> CursorMut<'b, T> {
```
kkysen, line 30: Can you just `impl Clone`?

src/cursor.rs
```diff
@@ -0,0 +1,116 @@
...
+        CursorMut {
+            data: self.data,
+            index: self.index,
+        }
+    }
+}
+
+impl<'a, T> Add<usize> for CursorMut<'a, T> {
```
kkysen, line 38: `impl Add` and `impl Sub` this way doesn't seem that intuitive/idiomatic in Rust.  Do you think we should do it this way?  It is more similar to the C way, but I worry that it's easy to overlook hidden operations in doing this.  When optimizing code to avoid bounds checks later, for example, it'd be more difficult to see where they occur.  What might be more idiomatic is to allow slicing with a range like normal slices do, but I haven't fully thought that out yet.

src/cursor.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn index(&self, index: isize) -> &Self::Output {
+        let index = self.index as isize + index;
+        assert!(index > 0 && (index as usize) < self.data.len());
+        &self.data[index as usize]
+    }
+}
+
+impl<'a, T> IndexMut<isize> for CursorMut<'a, T> {
+    fn index_mut(&mut self, index: isize) -> &mut Self::Output {
+        let index = self.index as isize + index;
```
kkysen, line 98: Can you extract the index calculation and checking into it's own method so we just do it once don't get off by 1 errors in multiples places that we need to fix?

src/cursor.rs
randomPoison: It's converting `index` to be an `isize` and then deferring to the `Index<isize>` impl. The `Index<isize>` version checks to make sure that the final index is valid, and I didn't want to repeat that logic here.

src/cursor.rs
randomPoison: We can't because `CursorMut` contains a `&mut [T]`, which can't be cloned. This method also works slightly different than the one provided by `Clone` in that it reborrows `self.data` rather than cloning it, which is why it also takes a `&mut self` instead of `&self`.

kkysen: I'm confused as to what the point of this method is then.  Is it just shortening the lifetime?

src/cursor.rs
kkysen: Do we want to keep the Rust code as similar looking to the C code?  Operator overloading isn't normally done like this in Rust.  It's also harder to change when it's done this for, vs. using a method, like `.offset`, that's much easier to refactor.

src/cursor.rs
```diff
@@ -0,0 +1,116 @@
...
+            index: self.index,
+        }
+    }
+}
+
+impl<'a, T> Add<usize> for CursorMut<'a, T> {
+    type Output = Self;
+
+    fn add(mut self, rhs: usize) -> Self::Output {
+        self.index += rhs;
```
kkysen, line 42: If we keep the operator overloads, could you have the non assign impls delegate to the assign ones?  So this would be `self += rhs; self`, for example.

src/cursor.rs
```diff
@@ -0,0 +1,116 @@
+use std::ops::{Add, AddAssign, Index, IndexMut, Sub, SubAssign};
+
+// TODO: AddAssign
+pub struct CursorMut<'a, T> {
+    data: &'a mut [T],
```
kkysen, line 5: What about making this unsized, so it's like an even fatter slice?  Then either `&` or `&mut` can be added to it so it's more flexible, and I think `Clone` should work normally (not sure).  I'm not sure if this would work, but if it does I think it'd be cleaner.

src/cursor.rs
```diff
@@ -0,0 +1,116 @@
...
+    index: usize,
+}
+
+impl<'a, T> CursorMut<'a, T> {
+    pub fn new(data: &'a mut [T]) -> Self {
```
kkysen, line 10: Should we have a `From` `impl` for this conversion?

src/cursor.rs
randomPoison: Since `CursorMut` isn't `Copy`, the various arithmetic operations move the value. If you want to still use an older cursor after doing some math operations to create a new one, you need to explicitly clone the cursor. This happens in `selfguided_filter`:

```rust
// Create `AA` from `A` and then use `AA`.
let mut AA = A.clone() - REST_UNIT_STRIDE
for ... {
    AA += step * REST_UNIT_STRIDE;
}

// Go back to using `A`...
```

src/cursor.rs
randomPoison: In general we want to keep the Rust code similar to the C as much as possible. In some cases we can convert the pointer arithmetic operations into safe slice operations, but I don't think that's always going to be possible, or at least can't always be done in a clean way.

src/cursor.rs
randomPoison: I don't think that'd be all that helpful in this case. We'd be pulling 2 lines of code into a helper method that is only used in 2 places, so I think that'd harm readability more than it'd help.

src/cursor.rs
kkysen: I agree we generally want similar semantics where safe, but is it desirable to keep a similar syntax even when not idiomatic?  It might be, but I'm not sure if it is.

src/cursor.rs
randomPoison: I don't think that would work? I thought there wasn't a way to construct those DSTs currently? At least that's what the nomicon says, idk if it's out of date. But I think it's probably best to avoid using DSTs because language support for them is kind of lacking, and that's going to add maintenance complexity that we don't need.

src/cursor.rs
randomPoison: Sure, seems reasonable. Done.

src/cursor.rs
randomPoison: Done.

src/looprestoration.rs
randomPoison: Done.

src/cursor.rs
randomPoison: Ah, good catch. Fixed.

src/cursor.rs
kkysen: Yeah, you're right.  I didn't realize constructing DSTs was still so unstable.  That's unfortunate.

src/cursor.rs
kkysen: I'm still not sure about the specifics why this is needed/only works this way, but it should be fine for now.

src/cursor.rs
kkysen: We discussed this over and decided it's good when the checks are only debug assets so that arithmetic doesn't have hidden bounds checking costs.  The real bounds check still happens when actually indexing.

kkysen (approved):
See https://github.com/memorysafety/rav1d/pull/372/files#r1309209051.

PR #374, opened 2023-08-15
`fn sgr_{3x3, 5x5, mix}_rust`: Cleanup generated code

randomPoison:


fbossen (approved):

src/looprestoration.rs
```diff
@@ -871,8 +871,14 @@ unsafe fn sgr_5x5_rust<BD: BitDepth>(
     edges: LrEdgeFlags,
     bd: BD,
 ) {
-    let mut tmp: [BD::Pixel; 27300] = [0.as_(); 27300];
-    let mut dst: [BD::Coef; 24576] = [0.as_(); 24576];
+    // Selfguided filter is applied to a maximum stripe height of 64 + 3 pixels
+    // of padding above and below
+    let mut tmp = [0.as_(); 70 /*(64 + 3 + 3)*/ * REST_UNIT_STRIDE];
```
kkysen, line 876: Do you need the `.as_()` here?  Shouldn't the `0` literal be able to be any integer?

randomPoison: Yeah we need the `.as_()` cast. These are arrays of `BD::Pixel` and `BD::Coef`, and since we're in a generic context the compiler won't let us initialize them from raw integers.

kkysen: Ah.  Does `.from()` work?  That's guaranteed non truncating so I usually use that if possible.  I think we have a `From<u8>` for that.

PR #375, opened 2023-08-15
Delete `looprestoration_tmpl_{8, 16}.rs`

randomPoison:
These files have been fully deduplicated so are no longer needed.


kkysen (commented):
Can you delete the `mod` declarations, too?

Also, it'd be nice if these would be marked as unused automatically so we don't forget.  Can we make the modules that are not the public APIs not `pub`?  I think that should fix that?

randomPoison:
The `mod` declarations were already deleted, I had just forgotten to delete the actual files in the previous PR. And yeah, modules that aren't part of the public API can be made private.

PR #376, opened 2023-08-15
`struct Dav1dCdefDSPContext`: Deduplicate and type-erase fn pointers

randomPoison:


kkysen (approved):
LGTM.  Did we want to have a macro for the extern decls all with the same signature?  We can do that when we generecize the init fns though in a compatible way.

randomPoison:
Yeah, I chose to hold off on using a macro for now to keep the PR simple. Doing that when we do the init fns makes sense to me.

PR #381, opened 2023-08-24
CI: Copy logs even if tests fail

thedataking:


kkysen (approved):

kkysen:
@thedataking, I noticed in https://github.com/memorysafety/rav1d/actions/runs/5959507449/job/16165347727?pr=382 that if building fails, it tries to copy the logs, which then fails, too.  I think we only need the logs if we get to running the tests.  It's not really a big deal, though, since the build is already failing in this case.

PR #382, opened 2023-08-24
`fn decode_sb`: Cleanup majority of `fn`

kkysen:
This cleans up the majority of `fn decode_sb`, all the simple cleanups that comprise most of the changes.  There are still some things I want to change, but those are more involved, so I want to keep them as separate PRs.

For example, I want to change some of the `enum` sizes separately (`enum`s that were translated as type aliases).  And I want to try to turn `EdgeNode` into an `enum` or at least `union`, as it can be an `EdgeTip` or `EdgeBranch`.  And some other small but trickier things.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4190,9 +4190,7 @@ unsafe fn decode_sb(
     let have_h_split = f.bw > t.bx + hsz;
     let have_v_split = f.bh > t.by + hsz;
     if !have_h_split && !have_v_split {
-        if !((bl as libc::c_uint) < BL_8X8 as libc::c_int as libc::c_uint) {
-            unreachable!();
-        }
+        assert!((bl as libc::c_uint) < BL_8X8 as libc::c_int as libc::c_uint);
         return decode_sb(
             t,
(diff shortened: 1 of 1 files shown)


randomPoison (approved):

src/decode.rs
```diff
@@ -4179,626 +4179,465 @@ unsafe fn decode_b(
...
-        if 0 as libc::c_int != 0 && bl as libc::c_uint == BL_64X64 as libc::c_int as libc::c_uint {
-            printf(
-                b"poc=%d,y=%d,x=%d,bl=%d,r=%d\n\0" as *const u8 as *const libc::c_char,
-                (*(*f).frame_hdr).frame_offset,
-                (*t).by,
-                (*t).bx,
-                bl as libc::c_uint,
-                (*ts).msac.rng,
+    if t.frame_thread.pass != 2 {
+        if bl == BL_64X64 {
```
fbossen, line 4204: `0 as libc::c_int != 0` is always false, so the condition should be `false && bl == BL_64X64`? This appears to have a significant impact on performance

kkysen: Ah, I think I must've misread that as `false ||` instead of `&&`.  I'll fix that.  Thanks for tracking it down!

PR #395, opened 2023-08-25
`fn dav1d_msac_init`: Make safe

kkysen:
I'm not sure if this works correctly, but I tried to make this safe by adding a phantom lifetime to `MsacContext` for the `buf`'s lifetime, which originally was a slice, but is made into pointers to be `#[repr(C)]` and stay layout-compatible with asm.  I'm particularly not sure about the `'static` I stuck in the `MsacContext` in its containing `struct`.  Is that any less safe/checked than before?

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4647,8 +4647,7 @@ unsafe fn setup_tile(
     ts.last_delta_lf.fill(0);
     dav1d_msac_init(
         &mut ts.msac,
-        data,
-        sz,
+        std::slice::from_raw_parts(data, sz),
         (*f.frame_hdr).disable_cdf_update != 0,
     );
     ts.tiling.row = tile_row as libc::c_int;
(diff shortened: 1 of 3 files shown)


randomPoison:
I'm not sure it makes sense to add the lifetime to `MsacContext`. If we were threading that lifetime all the way from the source slice to `MsacContext` then it would work, but since we're using `slice::from_raw_parts` to construct the slice from a pointer, it's going to pull a lifetime from thin air that will always match. So adding the lifetime isn't going to correctly track the lifetime of the slice.

kkysen:
I plan on, after this, to thread the lifetime up until at least `Dav1dData`, which seems like it might be a ref counted slice/buf.  I want to do it incrementally, though.

randomPoison:
Can you do threading the lifetime in a single PR? I think any incremental work there is going to leave us with lifetimes that don't make sense, so it'd probably be best to do a single PR where we can see how the lifetimes connect.

kkysen:
> Can you do threading the lifetime in a single PR? I think any incremental work there is going to leave us with lifetimes that don't make sense, so it'd probably be best to do a single PR where we can see how the lifetimes connect.

Okay, I'll try and see how it goes.

PR #396, opened 2023-08-25
`struct Dav1dInvTxfmDSPContext`: Type-erase `fn` ptrs and deduplicate

randomPoison:
This PR deduplicates the `Dav1dInvTxfmDSPContext` and type-erases the function pointers that are stored in it. To do this I rewrote the macros that the C code uses to declare both the extern asm fns and the C fns that are wrappers around `inv_txfm_add_rust`.


kkysen:
Before I even review, damn that's a lot of lines removed.  Nice!

src/itx.rs
```diff
@@ -309,6 +309,94 @@ extern "C" {
...
+        paste::paste! {
+            // TODO(legare): Temporarily pub until init fns are deduplicated.
+            pub(crate) unsafe extern "C" fn [<inv_txfm_add_ $type1 _ $type2 _ $w x $h _c_erased>] <BD: BitDepth> (
+                mut dst: *mut DynPixel,
+                stride: ptrdiff_t,
+                coeff: *mut DynCoef,
+                eob: libc::c_int,
+                bitdepth_max: libc::c_int,
+            ) {
+                use crate::src::itx_1d::*;
```
kkysen, line 339: What's the glob needed for?

src/itx.rs
```diff
@@ -137,3 +139,277 @@ pub unsafe extern "C" fn inv_txfm_add_rust<BD: BitDepth>(
         dst = dst.offset(BD::pxstride(stride) as isize);
     }
 }
+
+pub type itxfm_fn = Option<
```
kkysen, line 143: We can remove the `Option`, right?  We've been doing that for other ones already.

src/itx.rs
```diff
@@ -137,3 +139,277 @@ pub unsafe extern "C" fn inv_txfm_add_rust<BD: BitDepth>(
...
+            dst: *mut DynPixel,
+            dst_stride: ptrdiff_t,
+            coeff: *mut DynCoef,
+            eob: libc::c_int,
+            bitdepth_max: libc::c_int,
+        );
+    };
+
+    ($prefix:ident, $opt:ident) => {
+        paste::paste! {
```
kkysen, line 166: Is there a reason why we don't do `use paste::paste;` and just `paste!` here?

src/itx.rs
```diff
@@ -137,3 +139,277 @@ pub unsafe extern "C" fn inv_txfm_add_rust<BD: BitDepth>(
...
+    decl_itx_fn!(dav1d_inv_txfm_add_wht_wht_4x4, _sse2);
+}
+
+#[cfg(all(feature = "asm", any(target_arch = "arm", target_arch = "aarch64")))]
+extern "C" {
+    decl_itx_fns!(_neon);
+}
+
+macro_rules! inv_txfm_fn {
+    ($type1:ident, $type2:ident, $w:literal, $h:literal, $shift:literal, $has_dconly:literal) => {
```
kkysen, line 329: I think `$has_dconly` can just be `expr`, not `literal`, since it's not used to make an `ident`.  And also, can we use `true` and `false` instead of `1` and `0`?  We can cast to `c_int` for now if we need to and fix that later, but for the literals, `true`/`false` is definitely much better.

kkysen: It could also be a `const` generic parameter instead I think.  That would be even better.  We can still generate separate `fn` ptrs out of it the same way we do with `::<BitDepth8>` now.

kkysen: Also, I think the same suggestions above apply also to `$shift`.

kkysen: Actually, since this is for the `_c_erased` ones, a bunch more can be made into `const` generic params instead.

kkysen (approved):
This is a ton of code removed; that's awesome!  Most of my comments can be applied on top, later potentially, like using more `const` generic parameters instead of macro parameters.  I think where we can, it'd be better to keep as much out of macros as possible.

My only potential concern is how easily can the declaring macros be made into expression-like macros?  What I mean is, in #379, I do the bitdepth selection inline in a `match`, and these have to be done inline so that both asm functions are in scope (so it type checks) and then it successfully links through dead code elimination.  I haven't dug into the macros used here too much yet, so do you think it'll be hard to convert it to that style (or any other way that does the bitdepth selection with zero-cost)?

src/itx.rs
randomPoison: Oh, I wanted to avoid doing that in this PR and instead do it when the init fns get cleaned up. I wanted to avoid having to update all of the init logic manually since that's hundreds and hundreds of lines that are all going to get rewritten with a macro down the line most likely. So I figured I'd avoid doing that here to keep this PR simpler.

src/itx.rs
randomPoison: Mostly for simplicity, since we'd also have to have a `#[cfg(feature = "asm")]` on the import to avoid warnings, so it seemed cleaner to just use the absolute path directly.

src/itx.rs
randomPoison: What's the advantage of making `shift` and `has_dconly` const generics? And are you talking about making it a generic parameter to `inv_txfm_add_rust` or on the generated `_c_erased` fn? I don't really see the advantage either way.

> I think `$has_dconly` can just be `expr`, not `literal`, since it's not used to make an `ident`.

Does that matter? Having it be a literal works just the same in this case.

randomPoison:
> My only potential concern is how easily can the declaring macros be made into expression-like macros? What I mean is, in #379, I do the bitdepth selection inline in a `match`, and these have to be done inline so that both asm functions are in scope (so it type checks) and then it successfully links through dead code elimination.

For the inline `match`, does the extern function also have to be declared inline for that approach to work? I don't think these macros would translate cleanly to expression-like macros (e.g. the way the looprestoration one did) because the whole point of these macros is that they're using nested invocations to expand out a bunch of functions at once. But if the extern fn doesn't need to be declared inline then that shouldn't be a problem?

src/itx.rs
randomPoison: I've updated the macros to use `bool` literals instead of `1`/`0`.

src/itx.rs
kkysen: Ah, that's fine then.  I also noticed in #411 that using non-absolute paths in macros isn't good since it'll resolve where the macro is expanded, so any `use`s have to be put inside the macro.

kkysen:
> For the inline `match`, does the extern function also have to be declared inline for that approach to work?

I think so.  It's relying on dead-code elimination to remove the `extern` block (or at least, enough to tell `rustc` that it doesn't need to try to link to that `fn`).  I'll have to see if it works with separately declared `fn`s.

kkysen:
Actually, I was wrong.  It still works with the `fn`s declared separately.  I think it only tries to link to `extern`s that are actually used, and so if it's a dead-code eliminated `match` arm, then it won't try to link to it.  In that case, all should be good here.

src/itx.rs
randomPoison: The generated function references `david_inv_*` functions defined in `itx_1d`. I can change it to use an absolute path if you think that matters.

kkysen: If it's possible to make this `use` just import a single function, that'd be great.  It'd be much clearer what it's depending on.

src/itx.rs
kkysen: > What's the advantage of making `shift` and `has_dconly` const generics? And are you talking about making it a generic parameter to `inv_txfm_add_rust` or on the generated `_c_erased` fn? I don't really see the advantage either way.

I just thought it's slightly cleaner.  I looked at it more closely when doing #420 and I think it's overkill.

> > I think `$has_dconly` can just be `expr`, not `literal`, since it's not used to make an `ident`.
> 
> Does that matter? Having it be a literal works just the same in this case.

No, not really.

PR #399, opened 2023-08-26
`fn read_restoration_info`: Refactor out inner `fn msac_decode_lr_subexp`

kkysen:


kkysen:
@fbossen, does this look like something reasonable to do?

randomPoison (approved):

fbossen:
> @fbossen, does this look like something reasonable to do?

Looks reasonable to me

fbossen (approved):
LGTM

PR #406, opened 2023-08-27
`fn dav1d_init_qm_tables`: Further cleanup

kkysen:
In preparation for making it safe.


randomPoison (commented):
I'm a bit confused by this PR. It looks like you reordered a number of the steps in this function, but it's not clear to me why. What was the rationale there?

kkysen:
In 45a0321, I re-ordered those statements so the the initializations come before the assignments.  It doesn't matter when everything's mutable, but in the next PRs I change it to by value so it has to come before.  Also, I thought it was in-general clearer to do the initialization before assignment.

randomPoison (approved):

PR #412, opened 2023-08-28
`src/{ref,mem,cdf,data,picture,refmvs,log,thread_task,cpu}.rs`: Deduplicate declarations into `use` imports

kkysen:


randomPoison (approved):
This all seems reasonable to me. My only thought is that we want to be careful about removing the `#[no_mangle]` and `extern "C"` attributes because we don't want to break anything that's supposed to be part of the C API. I don't think any of the functions you touched here are part of the public API, but we should be conservative about which functions we cleanup since we don't have any tests for the C API yet.

PR #416, opened 2023-08-29
`struct Dav1dMCDSPContext`: Type-erase `fn` ptrs and deduplicate

randomPoison:
The various callback functions (for both asm routines and the C/Rust fallback fns) are now declared using macros.

Since all that's left in the `mc_tmpl_*.rs` files are the init fns, I'm gonna follow this PR up with one that deduplicates those and removes the `Option` wrappers around the fn pointer types.


randomPoison:
Oh wait, just realized I still haven't deleted the duplicate instances of `Dav1dMCDSPContext` 🤦 Fixing that now.

src/mc.rs
```diff
@@ -1230,3 +1217,1063 @@ pub unsafe fn resize_rust<BD: BitDepth>(
...
+        ptrdiff_t,
+        libc::c_int,
+        libc::c_int,
+        libc::c_int,
+        libc::c_int,
+        libc::c_int,
+        libc::c_int,
+    ) -> (),
+>;
+#[derive(Copy, Clone)]
```
kkysen, line 1388: Depending on if this PR or #417 lands first, we should get rid of this.

src/mc.rs
```diff
@@ -1230,3 +1217,1063 @@ pub unsafe fn resize_rust<BD: BitDepth>(
...
+) {
+    w_mask_rust(
+        dst.cast(),
+        dst_stride as usize,
+        tmp1,
+        tmp2,
+        w as usize,
+        h as usize,
+        mask,
+        sign != 0,
```
kkysen, line 1751: I had `debug_assert!(sign == 0 || sign == 1);`s for these casts previously.  I wasn't totally sure they're always `bool`s.

src/mc.rs
```diff
@@ -1230,3 +1217,1063 @@ pub unsafe fn resize_rust<BD: BitDepth>(
...
+    };
+}
+
+#[cfg(feature = "asm")]
+macro_rules! decl_fns {
+    ($ty:ident, $name:ident) => {
+        decl_fns!($ty, $name, sse2);
+        decl_fns!($ty, $name, ssse3);
+        decl_fns!($ty, $name, avx2);
+        decl_fns!($ty, $name, avx512icl);
```
kkysen, line 2162: Make sure avx is only enabled for `x86_64`.

Also, why not include `neon` here with `#[cfg]`s?

src/mc.rs
```diff
@@ -1230,3 +1217,1063 @@ pub unsafe fn resize_rust<BD: BitDepth>(
...
+    };
+
+    ($ty:ident, $name:ident, $suffix:ident) => {
```
kkysen, line 2165: I think `$ty` is kind of confusing since `ty` usually means an actual type.  Maybe `$mc_kind`?

src/mc.rs
kkysen: #417 and #418 merged so could you remove this and any other warnings that show now?

src/mc.rs
randomPoison: Changed.

src/mc.rs
randomPoison: Added those asserts back in.

src/mc.rs
randomPoison: Done.

src/mc.rs
randomPoison: I've added the `#[cfg]`s to make avx only for `x86_64`. I also tried merging the `neon` declarations in here, but that resulted in a bunch of extra unused fn declarations, so I opted to keep it separate. Doesn't really matter all that much though, since the macro also generates a bunch of extra declarations for x86 also 🤷

kkysen: > I've added the `#[cfg]`s to make avx only for `x86_64`.

👍 

I also tried merging the `neon` declarations in here, but that resulted in a bunch of extra unused fn declarations, so I opted to keep it separate.

👍 

> Doesn't really matter all that much though, since the macro also generates a bunch of extra declarations for x86 also 🤷

What do you mean?  I didn't see any unused warnings when compiling for x86 or x86_64.

src/mc.rs
```diff
@@ -1229,3 +1216,1069 @@ pub unsafe fn resize_rust<BD: BitDepth>(
...
+    pub blend_v: blend_dir_fn,
+    pub blend_h: blend_dir_fn,
+    pub warp8x8: warp8x8_fn,
+    pub warp8x8t: warp8x8t_fn,
+    pub emu_edge: emu_edge_fn,
+    pub resize: resize_fn,
+}
+
+macro_rules! filter_fns {
+    ($mc_kind:ident, $type_h:expr, $type_v:expr) => {
```
kkysen, line 1407: I didn't realize there was another `$mc_kind`.  This one is different so it's kinda confusing.  What about `$type_names` for this one, since it's a combination of the names of `$type_h` and `$type_v`?

src/mc.rs
```diff
@@ -1229,3 +1216,1069 @@ pub unsafe fn resize_rust<BD: BitDepth>(
...
+        decl_fns!($mc_kind, $name, ssse3);
+
+        #[cfg(target_arch = "x86_64")]
+        decl_fns!($mc_kind, $name, avx2);
+
+        #[cfg(target_arch = "x86_64")]
+        decl_fns!($mc_kind, $name, avx512icl);
+    };
+
+    ($mc_kind:ident, $name:ident, $suffix:ident) => {
```
kkysen, line 2170: Could you use `$asm` instead of `$suffix`?  `$suffix` seems too general and makes it harder to read since macros are already quite hard to read.  Also, could you put this arm (is that what they're called) before the other one so that they're in the order in which they're called?

src/mc.rs
```diff
@@ -1229,3 +1216,1069 @@ pub unsafe fn resize_rust<BD: BitDepth>(
...
+            #[cfg(feature = "bitdepth_8")]
+            decl_fn!($mc_kind, [<$name _8bpc_ $suffix>]);
+            #[cfg(feature = "bitdepth_16")]
+            decl_fn!($mc_kind, [<$name _16bpc_ $suffix>]);
+        }
+    };
+}
+
+#[cfg(all(feature = "asm", any(target_arch = "x86", target_arch = "x86_64")))]
+#[allow(dead_code)] // Macro invocations generate more fn declarations than are actually used.
```
kkysen, line 2181: Is it possible to avoid this?  It seems bad to massively over-declare functions that don't exist.

kkysen: If you base it on what happens in the `dsp_init` `fn`s, there shouldn't be any extra ones.

src/mc.rs
kkysen: > What do you mean? I didn't see any unused warnings when compiling for x86 or x86_64.

Nevermind.  I didn't see the `#[allow(dead_code)]`.

src/mc.rs
kkysen: Actually, ignore this for now.  Once the other little things are fixed, you can merge this, and we'll fix this later (since it may be simpler to do it along with #426 (see https://github.com/memorysafety/rav1d/pull/426/files#r1316922243).

src/mc.rs
randomPoison: I'm gonna leave this one `mc_kind` now that the other one has been changed to `$fn_kind`.

src/mc.rs
```diff
@@ -1229,3 +1216,1069 @@ pub unsafe fn resize_rust<BD: BitDepth>(
...
+            dx: libc::c_int,
+            mx: libc::c_int,
+            bitdepth_max: libc::c_int,
+        );
+    };
+}
+
+#[cfg(feature = "asm")]
+macro_rules! decl_fns {
+    ($mc_kind:ident, $name:ident) => {
```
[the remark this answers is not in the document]
randomPoison: Done.

src/mc.rs
randomPoison: Done.

PR #418, opened 2023-08-30
`#![allow(*)]`: Remove unneeded

kkysen:
This removes all the `#![allow(*)]`s that aren't needed or unused.

Warning on `unused_mut` and `unused_assignments` helps catch a lot of little mistakes so it's nice to have, and fixing this cleans up a lot of code we'd otherwise have to manually clean up when going function-by-function.

Warning on `dead_code` helps catch a bunch of little mistakes as well as a lot of over-declaring asm `fn`s that don't exist.  I've fixed most of them (they're mostly `avx{2,512icl}` on `x86`), except for in `mc_*.rs` since #416 is in the middle of that so I don't want to overly touch it, and `itx.rs`, where way more asm `fn`s than actually exist are declared.

`mutable_transmutes` and `non_snake_case` (for the `dav1d` and `seek_stress` binaries) are just wholly unused, so we can remove those.

### DIFF
--- a/src/msac.rs
+++ b/src/msac.rs
@@ -175,6 +175,10 @@ fn ctx_norm(s: &mut MsacContext, dif: ec_win, rng: libc::c_uint) {
     }
 }
 
+#[cfg_attr(
+    all(feature = "asm", any(target_arch = "x86_64", target_arch = "aarch64")),
+    allow(dead_code)
+)]
 fn dav1d_msac_decode_bool_equi_rust(s: &mut MsacContext) -> bool {
     let r = s.rng;
     let mut dif = s.dif;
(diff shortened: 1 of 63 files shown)


src/msac.rs
```diff
@@ -175,6 +175,10 @@ fn ctx_norm(s: &mut MsacContext, dif: ec_win, rng: libc::c_uint) {
     }
 }
 
+#[cfg_attr(
+    all(feature = "asm", any(target_arch = "x86_64", target_arch = "aarch64")),
+    allow(dead_code)
+)]
```
randomPoison, line 181: Why use `cfg_attr` here to `allow(dead_code)` instead of just `cfg`ing the entire function out when it's unused? Same for the other functions in this file where you did `cfg_attr`.

randomPoison (approved):

src/msac.rs
kkysen: If I `#[cfg]` the whole thing out it doesn't type check unless it's specifically enabled and rust-analyzer stops working, so it's much nicer to have it enabled and just dead code eliminated.

PR #419, opened 2023-08-30
`decl_itx_fn*!`: Simplify/refactor

kkysen:


randomPoison (approved):
This rewrite looks fine but was it necessary to do? It doesn't seem like there was any real issue with the original version, and it's not clear to me that the rewritten version is substantially better. Was this change necessary to unblock some other work?

kkysen:
> This rewrite looks fine but was it necessary to do? It doesn't seem like there was any real issue with the original version, and it's not clear to me that the rewritten version is substantially better. Was this change necessary to unblock some other work?

No, not really.  I did it at the same time as #420, though, and after I tried to split it up so that #420 was just the main semantic changes.

For 51512fb

kkysen:
Ugh I pressed the wrong button.

PR #421, opened 2023-08-30
Remove error/cleanup-like `goto`s in C

kkysen:
The relooper algorithm `c2rust transpile` uses to translate `goto`s makes a mess of the control flow to the point where it takes a long time to untangle the state machines into normal control flow.  Most of these `goto`s use fairly simple control flow, being mostly used for error cleanup.  So they can be removed by either inlining (if the `goto` is very short), outlining and then calling the outline function at each `goto`, or in a few cases, just reorganize the structured control flow a little bit.

I intend to re-transpile and merge back these `goto`-less functions, as I'm pretty sure that'll be easier than untangling the relooper state machines (I usually end up just re-translating by hand, but some of these functions are quite long).  I'll probably add this to this same PR, but I want to put up these changes to the C first for some initial review.  In particular the ones where I re-organized control flow, as the inlining and outlining are much more straightforward.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -3489,6 +3489,38 @@ static int get_upscale_x0(const int in_w, const int out_w, const int step) {
     return x0 & 0x3fff;
 }
 
+static int dav1d_submit_frame_error(const int res, Dav1dFrameContext *const f,
+        Dav1dContext *const c, Dav1dThreadPicture *const out_delayed) {
+    atomic_init(&f->task_thread.error, 1);
+    dav1d_cdf_thread_unref(&f->in_cdf);
+    if (f->frame_hdr->refresh_context)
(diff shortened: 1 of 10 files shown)


randomPoison (approved):
I'm gonna approve this PR as is, probably makes sense to do the re-translation in separate PRs.

kkysen:
> I'm gonna approve this PR as is, probably makes sense to do the re-translation in separate PRs.

I can do that in a separate PR if you think that's better.  I want to test this changed C code first, though.  I'll merge once I do that.

kkysen:
Fixed some typos and a few warnings and the tests all pass now with `-Dtest_rust=false`, so this should be good.  I'll re-transpile and merge the functions with `goto`s back in in a follow-up PR.

PR #423, opened 2023-08-30
`fn wiener_rust`: Fix perf regression

fbossen:
Use 64-bit arithmetic to compute indices to avoid repeated bound checks. Performance regression was observed after #368

### DIFF
--- a/src/looprestoration.rs
+++ b/src/looprestoration.rs
@@ -349,13 +349,13 @@ unsafe fn wiener_rust<BD: BitDepth>(
     let round_bits_v = 11 - (bitdepth == 12) as libc::c_int * 2;
     let rounding_off_v = 1 << round_bits_v - 1;
     let round_offset = 1 << bitdepth + (round_bits_v - 1);
-    for j in 0..h {
-        for i in 0..w {
+    for j in 0..h as usize {
+        for i in 0..w as usize {
             let mut sum = -round_offset;
+            let z = &hor[j * REST_UNIT_STRIDE + i..(j + 7) * REST_UNIT_STRIDE];
 
             for k in 0..7 {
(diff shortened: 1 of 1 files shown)


kkysen (commented):
Hmm, I'm surprised this makes a difference since the bounds checks are still happening.

I can also do a PR for this that would eliminate a lot of the bounds checks here entirely.

It's good to know that making the indices `usize`s instead of `i32`s helps, as that's something I've been doing as I clean up functions.

src/looprestoration.rs
```diff
@@ -349,13 +349,13 @@ unsafe fn wiener_rust<BD: BitDepth>(
...
-    for j in 0..h {
-        for i in 0..w {
+    for j in 0..h as usize {
+        for i in 0..w as usize {
             let mut sum = -round_offset;
 
-            for k in 0..7 {
-                sum += hor[((j + k) * REST_UNIT_STRIDE as libc::c_int + i) as usize] as libc::c_int
-                    * filter[1][k as usize] as libc::c_int;
+            for k in 0..7 as usize {
```
kkysen, line 356: I think the `as usize` on the constant shouldn't be needed but I may be wrong.

fbossen:
> Hmm, I'm surprised this makes a difference since the bounds checks are still happening.

It looks like the bound checks were happening for every iteration of `k`. With the change, bound checks happen much less frequently.

src/looprestoration.rs
fbossen: Not needed indeed

fbossen:
I've defined a slice to avoid the repeated bound checks. Code now runs much faster as it is getting properly vectorized.

randomPoison (approved):

kkysen (approved):
LGTM.  I also wanna check if there's more bounds checks we can remove from here and other similar code working with `REST_UNIT_STRIDE`.

PR #425, opened 2023-08-31
Re-transpile `fn`s with `goto`s removed now

kkysen:
#421 removed a lot of the simple error/cleanup `goto`s in C.  This re-transpiles those functions and merges the new versions back in with the relooper state machines now removed.

I only re-transpiled and merged back functions used by the `dav1d` binary (the others aren't as important).

I also included the `retranspile.sh` script I used to help with this, as we may want to do this again for something (e.x. more `goto`s, upstream changes to `dav1d`, etc.).

### DIFF
--- a/retranspile.md
+++ b/retranspile.md
@@ -1,0 +1,16 @@
+# How to Re-Transpile `fn`s After Changes to the C code
+
+1. Make your changes to the C code.
+2. Run `./retranspile.sh transpile` to actually re-transpile.
+3. For each `fn` `${fn_name}`,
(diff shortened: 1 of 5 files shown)


randomPoison (approved):
Looks great! Excellent to see all of that awful relooper control flow gone 🎉 The only weird thing I noticed is there is still some weird control flow in `dav1d_parse_obus`. I tried looking at the original C to see if there's an obvious reason for this (e.g. are there still gotos), but I didn't spot anything that would explain it 🤷

PR #429, opened 2023-08-31
`struct Dav1dIntraPredDSPContext`: Type-erase `fn` ptrs and deduplicate

randomPoison:


src/ipred.rs
```diff
@@ -6,3 +12,167 @@ pub unsafe extern "C" fn get_upsample(
...
+    };
+
+    ($fn_kind:ident, $name:ident, $suffix:ident) => {
+        paste::paste! {
+            #[cfg(feature = "bitdepth_8")]
+            decl_fn!($fn_kind, [<dav1d_ $name _8bpc_ $suffix>]);
+            #[cfg(feature = "bitdepth_16")]
+            decl_fn!($fn_kind, [<dav1d_ $name _16bpc_ $suffix>]);
+        }
+    };
```
kkysen, line 147: Could you do the same as in https://github.com/memorysafety/rav1d/pull/416#discussion_r1316887931?

Also, the `$fn_kind` works well here.  And I like that pattern in general a lot (vs. separate macros, which I probably would've done).

src/ipred.rs
```diff
@@ -6,3 +12,167 @@ pub unsafe extern "C" fn get_upsample(
...
+}
+
+#[cfg(feature = "asm")]
+#[allow(dead_code)] // Macro declares more fns than actually exist.
```
kkysen, line 151: Could you do the same as https://github.com/memorysafety/rav1d/pull/416#discussion_r1316889886?

kkysen (approved):
LGTM after https://github.com/memorysafety/rav1d/pull/429/files#r1316905709.  For https://github.com/memorysafety/rav1d/pull/429/files#r1316906548, we can fix it later.

PR #430, opened 2023-08-31
`struct Dav1dIntraPredDSPContext`: Remove (most) `Option`s

randomPoison:
I wasn't able to remove one of the `Option`s because there's actually a bit of code that checks if that field is `None` in `dav1d_submit_frame`, and it wasn't clear to me if there was another flag that we can use as an appropriate sentinel.


kkysen (approved):
I saw that `.is_none` in `fn dav1d_submit_frame`.  We might need to keep the `Option` for that one, but I'll see.  I'm cleaning up `fn dav1d_submit_frame` right now.

PR #442, opened 2023-09-09
`src/lib.rs`: Only make `mod`s `pub` if they need to be

kkysen:
To remove these `pub`s, I just looked for `rg 'DAV1D_API '` for public `fn`s.  There aren't very many of them.  I made sure to keep those `mod`s `pub`, as well as almost everything in `include/` since that's where the public types are.  And any other `mod` that had to be `pub` to compile I kept as well.

This leaves a lot of dead code warnings.  Most of the legitimate ones are cleaned up here, and in following PRs I'll fix the remaining ones, which were left out during transpilation and are actually used.


randomPoison (approved):
How did you determine which modules need to be public? We don't have a clear picture of what the public API looks like yet, do we?

kkysen:
> How did you determine which modules need to be public? We don't have a clear picture of what the public API looks like yet, do we?

I just looked for `rg 'DAV1D_API '` for public `fn`s.  There aren't very many of them.  I made sure to keep those `mod`s `pub`, as well as almost everything in `include/` since that's where the public types are.  And any other `mod` that had to be `pub` to compile I kept as well.

PR #448, opened 2023-09-11
`fn transpose`: Replace with calls to `const fn transposed`

kkysen:
@randomPoison, do you think I should move `const fn transposed` somewhere else more general?  I may also add other similar `const fn` utilities, like an integral `for` loop macro (since `for` loops don't work in `const fn`s).


randomPoison (approved):
Seems like it's mostly used in `qm.rs`, so probably fine to leave it there.

PR #450, opened 2023-09-11
`fn insert_border`: `const`ify

kkysen:
This one is trickier, uglier, and has higher overhead, but it should all run at compile-time.


randomPoison (approved):
What's the reason for constifying this one? It doesn't seem like it's being used in a const context. Is the goal to make `dav1d_init_wedge_masks` const?

kkysen:
> What's the reason for constifying this one? It doesn't seem like it's being used in a const context. Is the goal to make `dav1d_init_wedge_masks` const?

Yeah, the goal is to make this whole file `const`.

PR #456, opened 2023-09-11
`struct Dav1dFilmGrainDSPContext`: Type-erase `fn` ptrs and deduplicate

randomPoison:


src/filmgrain.rs
```diff
@@ -16,3 +23,55 @@ pub unsafe extern "C" fn get_random_number(
 pub unsafe extern "C" fn round2(x: libc::c_int, shift: uint64_t) -> libc::c_int {
     return x + ((1 as libc::c_int) << shift >> 1) >> shift;
 }
+
+pub type EntryRow = [DynEntry; 82]; // [entry; 82]
```
kkysen, line 27: Is this in the C?  It seems like originally it was `[DynEntry; GRAIN_WIDTH]`.

src/filmgrain_tmpl_8.rs
```diff
@@ -1,218 +1,245 @@
...
+        luma_row: *const DynPixel,
         luma_stride: ptrdiff_t,
         uv_pl: libc::c_int,
         is_id: libc::c_int,
+        bitdepth_max: libc::c_int,
+    );
+    fn dav1d_generate_grain_y_8bpc_avx2(
+        buf: *mut EntryRow,
+        data: *const Dav1dFilmGrainData,
+        bitdepth_max: libc::c_int,
```
kkysen, line 133: Where did this new `fn` come from?

src/filmgrain_tmpl_8.rs
```diff
@@ -1,218 +1,245 @@
...
     );
```
kkysen, line 134: Where did this new `fn` come from?

src/filmgrain_tmpl_8.rs
```diff
@@ -234,23 +261,30 @@ extern "C" {
...
+        buf_y: *const EntryRow,
         data: *const Dav1dFilmGrainData,
         uv: intptr_t,
+        bitdepth_max: libc::c_int,
+    );
+    fn dav1d_generate_grain_y_8bpc_neon(
+        buf: *mut EntryRow,
+        data: *const Dav1dFilmGrainData,
+        bitdepth_max: libc::c_int,
     );
```
kkysen, line 281: And also this one.

src/filmgrain_tmpl_8.rs
kkysen: Oh, my bad, it just got re-formatted.

kkysen (commented):
It LGTM except for the `type EntryRow = [DynEntry; 82]`.  Is there a reason you added that instead of just using `[DynEntry; 82]` everywhere, which is what the C does (with `GRAIN_WIDTH`)?  This might be something we want to add, but I'd prefer to do it separately if it's not in the C and isn't needed.

randomPoison:
I had added `EntryRow` to be consistent with `const_pixel_row` and friends, but it's not important to have. I've removed it.

kkysen:
> I had added `EntryRow` to be consistent with `const_pixel_row` and friends, but it's not important to have. I've removed it.

Maybe it's useful to have later, but I think we'd want it as a generic type alias like the others.

kkysen (approved):

PR #457, opened 2023-09-11
`fn dav1d_decode_frame_init`: Fix perf by using `.iter()` instead of `.into_iter()`

### LINKED ISSUE #455: Performance regression after #438
A significant performance regression was observed after merging #438 when assembly is not used (i.e., disabled at either compile or run time).

A substantial amount of time is newly spent in `core::array::drain::drain_array_with` which is called from `rav1d::src::mc::prep_8tap_rust`

fbossen:
* Fixes #455.


kkysen (approved):
Damn, I'm surprised this makes that much of a difference for performance but I'm glad it was a simple fix.

PR #473, opened 2023-09-19
`fn fill2d_16x2`: Prepare for `const`ification

kkysen:
`fn fill2d_16x2` is difficult to make `const` directly since it takes `&'static mut` args that are referenced in the return value.  I'm going to switch to a different API that can make this work, but these are some initial steps to prepare for `const`ification before that.

### DIFF
--- a/src/wedge.rs
+++ b/src/wedge.rs
@@ -260,7 +260,6 @@ const fn init_chroma<const LEN_LUMA: usize, const LEN_CHROMA: usize>(
 
 #[cold]
 fn fill2d_16x2<const LEN_444: usize, const LEN_422: usize, const LEN_420: usize>(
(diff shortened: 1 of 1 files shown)


randomPoison (approved):
Huh, so `dst` and `masks_444` were pointing to the same thing? Does that mean that the code was UB before since we were aliasing mutable references? Either way glad you fixed that ^__^

kkysen:
> Huh, so `dst` and `masks_444` were pointing to the same thing? Does that mean that the code was UB before since we were aliasing mutable references? Either way glad you fixed that ^__^

Yeah, I think so.  And originally it was okay since they were raw pointers.  I'm not sure why the borrow checker didn't stop that (me making them references), though, maybe because they were statics.

PR #477, opened 2023-09-19
`enum Wedge{Direction,MasterLine}Type`: Make `enum`s

kkysen:
This makes them real `enum`s instead of just integer constants.  To replicate the `const N_*: usize`s, I used `strum`'s `#[derive(EnumCount)]`.

This isn't super necessary for this file, but I wanted to get the hang of converting the integer constant "enum"s to real `enum`s, as it is necessary for performance in other places (due to the stronger `enum` guarantees), and this is an easy place to do as I just finished completely cleaning up the file.


randomPoison (approved):
Good change, I'm in favor of using strum to generate the enum count 👍

PR #482, opened 2023-09-20
`struct Dav1dContext`: Deduplicate (last one and all its circular dependents)

kkysen:
`struct Dav1dContext` circularly contains references to:
* `struct Dav1dFrameContext`
  * `struct Dav1dFrameContext_bd_fn`
    * `type recon_b_intra_fn`
    * `type recon_b_inter_fn`
    * `type filter_sbrow_fn`
    * `type backup_ipred_edge_fn`
    * `type read_coef_blocks_fn`
* `struct Dav1dTaskContext`

which are also all deduplicated in the same commit since they are self-referential to each other.

Since `struct Dav1dContext` is a publically incomplete type in C, we (for now) make it `pub` with a `pub use` unlike the rest of `src::internal`.  The correct way to expose such an opaque type is through type alias impl trait (TAIT), but that's still unstable, so this should suffice now until we've progressed further.  The next commit also does this for `struct Dav1dRef`, too, as that's the other type that `dav1d` exposes as an incomplete type.


include/dav1d/dav1d.rs
```diff
@@ -1,3 +1,10 @@
+// TODO(kkysen)
+// In C, these are incomplete types.
+// This can be mimicked in Rust with type alias impl trait,
+// but that is still unstable, so the types are just `pub` for now.
+pub use crate::src::internal::Dav1dContext;
+pub use crate::src::r#ref::Dav1dRef;
```
randomPoison, line 6: I'm not sure I understand why we need TAIT for this? If these types are part of the public API but the fields shouldn't be exposed, shouldn't we just make the fields private (or `pub(crate)` if needed internally)? C doesn't have privacy modifiers so exposing the type as an incomplete type is the only way to hide fields, but in Rust we have proper privacy control.

randomPoison (approved):

include/dav1d/dav1d.rs
kkysen: Hmm, I was previously thinking that we still expose other things like the size of the type if we just make all the fields private, but that might be the case for TAIT, too, now that I think about it.  Making all the fields `pub(crate)` is probably enough.

kkysen: I've now changed it in bac5f540c4cde6e1c5ecdf84c940dac5e30c81d3 to make all the fields `pub(crate)` instead.

PR #497, opened 2023-09-23
`static av1_intra_prediction_edges`: Make non-`mut` by replacing `#[derive(BitfieldStruct)]` with `bools_bitfield_struct!` for `const fn`s

kkysen:
`#[derive(c2rust_bitfields::BitfieldStruct)]` doesn't generate `const fn`s (https://github.com/immunant/c2rust/issues/1027), so it can't be used in a `const` block to initialize a `static`.  Thus, the `static` has to made `mut` and `unsafe` and initialized at runtime.

I opened https://github.com/immunant/c2rust/issues/1027 and looked into making the generated `fn`s `const`, but they used a lot of `trait`s, which can't contain `const fn`s, so it seemed like non-trivial work.  I also tried looking for other off-the-shelf bitfield crates, but they all either didn't support `const fn`s or still relied on `trait`s within them, and thus only could create `const fn`s with `nightly` features.

Thus, I rolled my own `macro_rules!` macro, `bools_bitfield_struct!` that generates `const fn`s.  The macro is specific for bitfields that are always `bool`s, which makes things a bit simpler and have a cleaner API.  And it also doesn't have to worry about being `#[repr(C)]` at all, which might makes things a bit simpler, too.  Overall, though, it's quite a simple macro, so it seemed much easier to do it this.  Since `rav1d` only used bitfields here, I kept the macro private in this file, but if we ever need bitfields that are all `bool`s elsewhere, we can extract the macro.  The `c2rust-bitfields` dependency is also now removed as it's unused.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,30 +3,9 @@
 version = 3
 
 [[package]]
-name = "c2rust-bitfields"
-version = "0.18.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b43c3f07ab0ef604fa6f595aa46ec2f8a22172c975e186f6f5bf9829a3b72c41"
-dependencies = [
- "c2rust-bitfields-derive",
-]
-
-[[package]]
-name = "c2rust-bitfields-derive"
-version = "0.18.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d3cbc102e2597c9744c8bd8c15915d554300601c91a079430d309816b0912545"
-dependencies = [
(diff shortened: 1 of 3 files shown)


kkysen:
Mostly superceded by #501, but I'm leaving this open (and intending to merge in order) so the history doesn't get overly complicated.

randomPoison:
I wouldn't normally want to add a custom macro for this, but since you're ultimately replacing it with the bitflags crate I'll approve this one as-is so that you can preserve merge order 👍

randomPoison (approved):

PR #504, opened 2023-09-23
`fn dav1d_get_cpu_flags`: Add compile-time CPU feature detection 

kkysen:
This adds `fn CpuFlags::compile_time_detect` for compile-time CPU feature detection for dead code elimination.  `dav1d_get_cpu_flags` does this in `src/cpu.h`, too, but we only transpiled the `SSE2` case, which is the only one enabled by default for the default target.  This was done in C for dead code elimination, so it should be useful for us, too, even though it seems to override the purpose of `fn dav1d_set_cpu_flags_mask` (a `DAV1D_API`).

### DIFF
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -1,12 +1,10 @@
+use crate::src::const_fn::const_for;
 use crate::src::internal::Dav1dContext;
 use bitflags::bitflags;
 use std::ffi::c_int;
 use std::ffi::c_uint;
 use std::sync::atomic::AtomicU32;
 use std::sync::atomic::Ordering;
-
-#[cfg(feature = "asm")]
-use cfg_if::cfg_if;
 
 #[cfg(not(any(
     target_arch = "x86",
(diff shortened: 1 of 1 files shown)


randomPoison (approved):

thedataking:
I'm not convinced this PR preserves dav1d functionality. 

In dav1d, the compile-time feature detection (see `dav1d_get_cpu_flags`) is guarded by `TRIM_DSP_FUNCTIONS` which is set for release builds (or explicitly) but not for debug builds. It looks like `cpu.rs` isn't feature gated in the same way.

thedataking:
The implementation of compile-time feature translation doesn't match one to one either, IDK, if it is of consequence:

```Rust
#[cfg(target_feature = "sse2")]
CpuFlags::SSE2,
```
vs

```C
#elif ARCH_X86_64 || defined(__SSE2__) || \
      (defined(_M_IX86_FP) && _M_IX86_FP >= 2)
    flags |= DAV1D_X86_CPU_FLAG_SSE2;
```

It appears that the C version will enable the SSE2 flag in a wider range of circumstances. I think a comment would have been appropriate if it was somehow decided that the Rust implementation equals the C one.

kkysen:
> I'm not convinced this PR preserves dav1d functionality.
> 
> In dav1d, the compile-time feature detection (see `dav1d_get_cpu_flags`) is guarded by `TRIM_DSP_FUNCTIONS` which is set for release builds (or explicitly) but not for debug builds. It looks like `cpu.rs` isn't feature gated in the same way.

Ohh, I must've missed that.

> With this change, it looks like the `--cpumask` command line option may not work anymore? Also, if compiling on an AVX512 machine and running on an older CPU would result in a crash?

I missed this comment, too, whoops!

So what needs to be fixed is just `cfg`ing it based on `debug_assertions`?  Or should a feature be added for it.  Though I'm not sure we can default a feature based on `debug_assertions`.

thedataking:
> So what needs to be fixed is just cfging it based on debug_assertions?

Not sure what `debug_assertions` refer to but the logic in `dav1d` is this:

```meson
cdata.set10('TRIM_DSP_FUNCTIONS', get_option('trim_dsp') == 'true' or
    (get_option('trim_dsp') == 'if-release' and get_option('buildtype') == 'release'))
```

I'm worried this could turn out to be a time sink since we do not know if dead code elimination will happen the same way for rav1d as it does in dav1d. I also don't know if there is a _straightforward_ way to replicate this behavior in rav1d.

There are two potential problems that Frank flagged:

1. Will compiling on a machine with AVX512 cause crash on a host that doesn't have AVX512.
2. Will compiling on a machine without AVX512 regress performance on a host that has AVX512.

(Similar for other features but AVX512 is the obvious difference between modern x86 hosts.)

This can be determined by simply experimenting with the current code.

kkysen:
> Not sure what `debug_assertions` refer to

Whether you're compiling in debug mode or not.  This is what `std` uses in things like `debug_assert!`.

> but the logic in `dav1d` is this:
> 
> ```meson
> cdata.set10('TRIM_DSP_FUNCTIONS', get_option('trim_dsp') == 'true' or
>     (get_option('trim_dsp') == 'if-release' and get_option('buildtype') == 'release'))
> ```

We could have a `--feature trim_dsp` and check the same with `#[cfg(any(feature = "trim_dsp"), not(debug_assertions)))]`.

> I'm worried this could turn out to be a time sink since we do not know if dead code elimination will happen the same way for rav1d as it does in dav1d. I also don't know if there is a _straightforward_ way to replicate this behavior in rav1d.

I think I checked that the functions got DCEd previously when I wrote this PR.  It does seem important for code size if you compile for a higher microarchitecture level.

> 1. Will compiling on a machine with AVX512 cause crash on a host that doesn't have AVX512.

So when you compile, Rust doesn't do feature detection for your current CPU.  It assumes a baseline CPU, [`x86_64-v1`](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels), which only statically enables `CpuFlags::SSE2`.  Unless you pass `-C target-cpu=native` and you have a CPU with AVX512, or you pass `-C target-features=+avx512f` (and all of the other AVX512 features) on any CPU, the resulting binary should run on any CPU supporting `x86_64-v1`.  If you did turn on `avx512*` features, then the binary would `SIGILL` on older CPUs without AVX512, but that would happen even without assembly if there was autovectorization.

> 2. Will compiling on a machine without AVX512 regress performance on a host that has AVX512.

No, it should detect the AVX512 at runtime.  Unless you did something like `-C target-cpu=native`.

PR #513, opened 2023-10-06
`enum Rav1dError`: Convert internal error-handling to use `Rav1d{Error,Result}`

kkysen:
This adds an `enum Rav1dError`, which is a narrowed version of the overloaded `c_int` that `dav1d` uses for errors.  `dav1d` uses `errno` values for errors (except for `-1` sometimes).  This PR does the same (the same constants), but with `enum Rav1dError` instead of `c_int`.  Since `Rav1dError` has a niche at `0`, this makes `type Rav1dResult<T = ()> = Result<T, Rav1dError>` equivalent to how `dav1d` uses `c_int`, with `0` meaning no error, negative values meaning an error, and positive values meaning a stateful result.  `struct Dav1dResult` encapsulates this in Rust using a `#[repr(transparent)]` new-type, and there are efficient conversions between `{D <=> R}av1dResult`.

`Rav1dResult` is used as the return type for all error-returning `fn`s that are not part of the `#[no_mangle] extern "C"` `DAV1D_API`, which now use `Dav1dResult` instead of `c_int`.  This latter change is done to be more clear about errors while remaining ABI compatible with `dav1d`.  It is not, however, API compatible with Rust consumers of the `DAV1D_API` as now they get `Dav1dResult`s instead of `c_int`s.  The difference is trivial, a `.0`, but it is a difference.  I'm assuming this is okay since users of the `DAV1D_API` would be written in C (or other non-Rust language using C as the interop layer) and thus consume the `dav1d` headers, which remain ABI compatible with `rav1d`.  Rust users, on the other hand, would use the fully Rust-ified safe API, including `Rav1dResult`.

To make sure we are staying ABI compatible, it would be nice to also build the `tools/` and `tests/` binaries from C and link against `rav1d`, though trying the same with Chromium also would more than suffice.

This PR also uses the `errno` `const`s from `libc` rather than the hardcoded ones from transpilation, which used the Linux versions of them.  `ENOPROTOOPT` is not cross-platform, so this PR now fixes that bug.

### DIFF
--- a/include/dav1d/picture.rs
+++ b/include/dav1d/picture.rs
@@ -11,6 +11,8 @@ use crate::include::dav1d::headers::Rav1dFrameHeader;
 use crate::include::dav1d::headers::Rav1dITUTT35;
 use crate::include::dav1d::headers::Rav1dMasteringDisplay;
 use crate::include::dav1d::headers::Rav1dSequenceHeader;
+use crate::src::error::Dav1dResult;
+use crate::src::error::Rav1dResult;
 use crate::src::r#ref::Rav1dRef;
 use libc::ptrdiff_t;
 use libc::uintptr_t;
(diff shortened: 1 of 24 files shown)


kkysen:
I'm not sure what's failing on `macOS`.  I thought it was that I forgot to symbolicate `92` into `libc::ENOPROTOOPT` in `tools/*`, but that didn't fix the issue.

randomPoison (approved):

PR #520, opened 2023-10-12
`fn get_num_threads`: Cleanup and make safe

kkysen:


src/lib.rs
```diff
@@ -204,40 +203,31 @@ unsafe fn get_stack_size_internal(_thread_attr: *const pthread_attr_t) -> usize
...
+        rav1d_num_logical_processors().clamp(1, 256)
+    };
+    let n_fc = if s.max_frame_delay != 0 {
+        cmp::min(s.max_frame_delay as usize, n_tc)
     } else {
-        (if *n_tc < 50 as c_uint {
-            fc_lut[(*n_tc).wrapping_sub(1 as c_int as c_uint) as usize] as c_int
-        } else {
-            8 as c_int
-        }) as c_uint
```
randomPoison, line 230: What happened to this bit of logic? It looks like you completely changed the logic here in the cleaned up version.

randomPoison (approved):

src/lib.rs
kkysen: Yeah, I simplified it in 93f4416.  That's what the comments said it was doing, so I did that using a sqrt instead of a lookup table which is IMO a lot more confusing.

PR #522, opened 2023-10-12
`Cargo.toml`: Cleanup

kkysen:
This just cleans up `Cargo.toml`, which was still mostly generated by `c2rust transpile`.  The package is now named `rav1d` (like the `lib` was already changed to) and I set the version to `0.2.0` for future publishing (and so things like #519 don't need to patch it).  The existing `rav1d` crate is at `0.1.0`, and so since we plan on publishing there, we can start at `0.2.0`.  I also fixed some doc comments so `cargo test` can run.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,21 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b4682ae6287fcf752ecaabbfcc7b6f9b72aa33933dc23a554d853aea8eea8635"
 
 [[package]]
-name = "c2rust_out"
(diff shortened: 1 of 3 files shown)


kkysen:
@endbr64, would `0.2.0` work for building `chrome` or is `>= 1.0.0` needed for some reason?

endbr64:
> @endbr64, would `0.2.0` work for building `chrome` or is `>= 1.0.0` needed for some reason?

Should work fine, but I'll need to update the patches.

randomPoison (approved):

Issue #224, opened 2023-05-05
Backport correctness fixes from upstream dav1d 1.2.0


thedataking:
**Note**: this issue is explicitly about correctness fixes, not performance optimizations, or other changes such as support for other architectures or testing/infra improvements.

#213 showed that it can be necessary to backport critical fixes made upstream into our port. It seems there is at least a few more important fixes that we'll want to pull into rav1d:

- https://code.videolan.org/videolan/dav1d/-/commit/cf617fdae0b9bfabd27282854c8e81450d955efa
- https://code.videolan.org/videolan/dav1d/-/commit/f89dbc071704e26478b6d431f18244223b3808cb

These two fixes could also be relevant
- https://code.videolan.org/videolan/dav1d/-/commit/92d8b8154270bfac2115711571482a22aadbc1be
- https://code.videolan.org/videolan/dav1d/-/commit/98b0c96d2143e9264e3399a08dfdb0faa44796bf

PR #528, opened 2023-10-15
backport from `dav1d` 1.2.0: picture: allow storing an array of Dav1dITUTT35 entries

fbossen:
Partially addresses #224

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -30,7 +30,7 @@ project('dav1d', ['c'],
                       'b_ndebug=if-release'],
     meson_version: '>= 0.49.0')
 
-dav1d_soname_version       = '6.8.0'
+dav1d_soname_version       = '6.9.0'
 dav1d_api_version_array    = dav1d_soname_version.split('.')
 dav1d_api_version_major    = dav1d_api_version_array[0]
 dav1d_api_version_minor    = dav1d_api_version_array[1]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,6 +129,7 @@ dependencies = [
  "raw-cpuid",
  "strum",
  "to_method",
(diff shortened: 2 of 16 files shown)


randomPoison (approved):

meson.build
```diff
@@ -30,7 +30,7 @@ project('dav1d', ['c'],
                       'b_ndebug=if-release'],
     meson_version: '>= 0.49.0')
 
-dav1d_soname_version       = '6.8.0'
+dav1d_soname_version       = '6.9.0'
```
kkysen, line 33: Do we need to update this somewhere?

kkysen:
I think the code this changes has subtly changed a lot after my other PRs merged, so I think it should be re-reviewed after being rebased.

fbossen:
> I think the code this changes has subtly changed a lot after my other PRs merged, so I think it should be re-reviewed after being rebased.

There have been quite a few changes since this was initially written. Definitely needs another round of reviews. I am using `DRav1d<Vec<Rav1dITUTT35>, Vec<Dav1dITUTT35>>` to store the ITU-T T.35 data but am not convinced it's the best approach.

fbossen:
@kkysen Can you have another look at this when you have time? I am sure there are better ways of doing this. Also, some more recent commits in dav1d build on top of this, so it may be good to have this integrated in a not too distant future.

kkysen (commented):
@fbossen, can I take over this and fix the conversion issues?  I know the conversion code very well.

fbossen:
> @fbossen, can I take over this and fix the conversion issues? I know the conversion code very well.

Please go ahead

kkysen:
Ugh, why did this get automatically closed?

PR #534, opened 2023-10-23
`struct {D,R}av1dFilmGrainData`: Make separate `struct`s and rustify `Rav1dFilmGrainData`

kkysen:
`Dav1dFilmGrainData` is passed to asm, so I previously made `{D,R}av1dFilmGrainData` the same type (i.e. just aliases), but I think we can still properly separate them.  To avoid conversion overhead that's not in the `dav1d`-`rav1d` boundary (which already has overhead and isn't a bottleneck), I updated the assembly field offsets so that we can pass `Rav1dFilmGrainData` directly to asm, which proved to be fairly simple actually.  Then I made a bunch of `Rav1dFilmGrainData` fields properly `bool`s rather than `c_int`s, simplifying things and shrinking the `struct` size.

Note that we decided to revert the asm changes in this PR, but due to merge conflicts with the ton of other PRs building on this one, they are reverted later in #549, which is approved now.


randomPoison (commented):
A couple of things:

* I'm not sure I understand why we're now passing `Rav1dFilmGrainData` to asm. If I'm understanding your PR description, it sounds like it has to do with when we would do the conversion between the Rav1d and Dav1d versions? Could you elaborate on why we can't do the conversion at the same point where we do it for the other types?
* What purpose is `Dav1dFilmGrainData` serving at this point? My understanding is that it would be the internal representation of the data that we pass to asm. Now that the Rav1d version is being passed to asm, do we need the Dav1d version at all?

I'd also prefer if @fbossen (or someone else with asm experience) reviews this before merging. The changes on the Rust side look fine, but I don't know ARM/x86 assembly well enough to meaningfully review the other changes.

kkysen:
> * I'm not sure I understand why we're now passing `Rav1dFilmGrainData` to asm. If I'm understanding your PR description, it sounds like it has to do with when we would do the conversion between the Rav1d and Dav1d versions? Could you elaborate on why we can't do the conversion at the same point where we do it for the other types?

So all the other types aren't passed to asm; only primitive types and their arrays are passed to asm, except for `Dav1dFilmGrainData` that is.  So if we don't change the asm so that we can pass `Rav1dFilmGrainData`, we'll need to convert between the two when to call into asm, which has a (small but not non-existent) overhead.

> * What purpose is `Dav1dFilmGrainData` serving at this point? My understanding is that it would be the internal representation of the data that we pass to asm. Now that the Rav1d version is being passed to asm, do we need the Dav1d version at all?

`Dav1dFilmGrainData` is still part of the `DAV1D_API`, so we still need to convert to it, though that conversion happens at the same time as all the other ones so it's no worse for perf.  It's probably not really meant to be used in the public API, but we still need to keep it to be strictly ABI compatible.

randomPoison:
> So if we don't change the asm so that we can pass `Rav1dFilmGrainData`, we'll need to convert between the two when to call into asm, which has a (small but not non-existent) overhead.

Ah, so if I'm understanding it correctly, we're using `Rav1dFilmGrainData` internally, which would have to be converted back to a `Dav1dFilmGrainData` before each call to asm. My initial thought was "why don't we just use `Dav1dFilmGrainData` internally instead, then we wouldn't have to do any extra conversions", but then I realized that this is exactly what the code was doing before 🤦‍♀️

Though this does make me wonder, what's the advantage of `Rav1dFilmGrainData`? If we need to keep `Dav1dFilmGrainData` around as part of the C API anyway, why not keep using it internally so that we don't have to update the asm? I'm not inherently opposed to to switching to `Rav1dFilmGrainData` or making updates to the asm, it's just that anything that requires changes to the asm are inherently more risky and should be considered carefully. What advantages do we get from `Rav1dFilmGrainData` other than being able to rearrange a few fields to make it more compact?

kkysen:
> Though this does make me wonder, what's the advantage of `Rav1dFilmGrainData`? If we need to keep `Dav1dFilmGrainData` around as part of the C API anyway, why not keep using it internally so that we don't have to update the asm? I'm not inherently opposed to to switching to `Rav1dFilmGrainData` or making updates to the asm, it's just that anything that requires changes to the asm are inherently more risky and should be considered carefully. What advantages do we get from `Rav1dFilmGrainData` other than being able to rearrange a few fields to make it more compact?

The advantages are using idiomatic Rust types, like `bool` for these fields, and making the type smaller and more compact (less memory, better cache utilization).  There are also some other fields that can be modified, not just these `bool` ones.  That said, it's not a huge advantage for sure, but I realized that the asm changes needed were much simpler than I had thought.  There may be other similar places, though, where we can make small asm tweaks like this that would have bigger impacts, like allowing safer wrappers (I imagine a slice-like `#[repr(C)]` type and we just need to skip the length arg in asm, which might be fairly trivial).  Another choice could be leaving in the conversions when we call asm, as I didn't see a noticeable perf regression on `donna` when I did some cursory benchmarks.

randomPoison:
Okay, then let's let @fbossen give this a look and confirm if the asm changes are correct. If he feels comfortable reviewing asm changes, and the changes are relatively small like this, then I'm okay with making similar changes going forward. Though we should in general be cautious about changing things that require changes to the asm, and should generally prioritize other cleanup first if possible.

fbossen:
Looks like the Argon tests are failing on my ARM box. I can try to track the error down. However, I am not keen on modifying the asm code and the data structures passed to it. Let me look at this some more in the morning.

fbossen:
> It may be that old assembly objects got mistakenly cached; I always run `touch build.rs` after any assembly changes. That's something we should probably fix. Let me know if it's that, as I didn't see any crashes.

Much better after touching `build.rs`. Thanks for pointing that out. The changes look correct to me. I am still not convinced that changing the asm code is the way to go here. Would it make sense to keep the memory layout of the C code but change the data structures to use `bool`s (e.g., replace a `c_int` with a `bool` and 3 zero-valued `bool`s)? Trying to save a couple bytes in memory isn't terribly important here. It's not like we're instantiating thousands of these structures. Whatever you do here, I don't expect any noticeably change speed-wise.

fbossen:
An alternative could be to first push structure and asm changes into dav1d such as to avoid diverging asm code

thedataking:
> I am still not convinced that changing the asm code is the way to go here. 

> Trying to save a couple bytes in memory isn't terribly important here. It's not like we're instantiating thousands of these structures. Whatever you do here, I don't expect any noticeably change speed-wise.

Please listen to @fbossen on this one. I think changing the asm files is unjustified and opens the door to further changes. We  no longer benefit from upstream testing and improvements.

kkysen:
> We no longer benefit from upstream testing and improvements.

That's a good point; I hadn't considered that enough.

> An alternative could be to first push structure and asm changes into dav1d such as to avoid diverging asm code

That does seem like a much better way to go about things like this.  Then we'll keep things in sync, benefit from their upstream testing and further improvements on top of it, and their review of if whatever change is correct and beneficial.

> I am still not convinced that changing the asm code is the way to go here.

👍.  I trust you more to make decisions about the asm.  And I think the risk is greater and the benefits smaller than I initially thought.

> Whatever you do here, I don't expect any noticeably change speed-wise.

Given that and me not noticing any perf difference in 6e9a8f8, does going back to the version that just does the conversions with no asm changes make sense?

fbossen:
> Given that and me not noticing any perf difference in [6e9a8f8](https://github.com/memorysafety/rav1d/commit/6e9a8f8de800b3da91675ced3f898a6c38a7629d), does going back to the version that just does the conversions with no asm changes make sense?

Doing the conversions and no asm changes sounds good to me

kkysen:
I realized that reverting the asm changes here is a pain since I have to rebase all of the other PRs, a lot of which have merge conflicts, so I'm gonna just merge this PR after I open a PR reverting the asm changes here (`git checkout main src/x86/* src/arm/*`).

PR #542, opened 2023-10-23
`src/filmgrain.rs`: Fix `GrainLut` and `Scaling` args and make them refs

kkysen:


src/internal.rs
```diff
@@ -115,10 +115,12 @@ pub(crate) struct Rav1dContext_frame_thread {
     pub next: c_uint,
 }
 
+pub type GrainLut<Entry> = [[Entry; GRAIN_WIDTH]; GRAIN_HEIGHT + 1];
```
randomPoison, line 118: Is this typedef meant to be used somewhere?

randomPoison (approved):

src/internal.rs
kkysen: Yeah it's used in a bunch of the filmgrain signatures.

PR #549, opened 2023-10-30
`struct {D,R}av1dFilmGrainData`: Revert asm changes from #534 and fix a `dav1d` alignment bug

kkysen:
# Asm Change Reversion

As discussed in #534, changing the asm at all means we lose upstream testing, review, and other improvements.  Given that e27877e5f, which made `{D <=> R}av1dFilmGrainData` conversions, had no perf regressions, this commit reverts back to that behavior.

The asm changes were reverted with `git checkout main -- src/x86/* src/arm/*`, where `main` is bb68fb8.

I'm doing this as a follow-up PR after all of the intermediary ones to avoid a ton of merge conflicts.  So once this is approved, I can merge all of the PRs up to and including this one.

# `dav1d` Alignment Bug Fix

This PR also fixes a alignment bug in `dav1d` surfaced/triggered by this asm reversion, which places `Dav1dFilmGrainData`s on the stack in a different place (i.e. alignment) than e27877e5f did.

x86 asm uses `psrad` on a pointer to `Dav1dFilmGrainData`::ar_coeff_shift`.  When `psrad`'s shift operand is a memory address, i.e. a `XMMWORD PTR`, it loads 128 bits from it, shifts by the lower 64 bits, and requires the 128 bits to be 128-bit/16-byte aligned.

Previously, and still in `dav1d`, `ar_coeff_shift` is only 8-byte aligned, as is `Dav1dFilmGrainData`.  However, in `dav1d`, `Dav1dFilmGrainData` is part of `Dav1dFrameHeader`, which is allocated with `malloc`.  `malloc` happens to return 16-byte aligned pointers usually, but is not required to, so this is UB and will segfault if not aligned.

Due to the `Rav1dFilmGrainData` to `Dav1dFilmGrainData` conversion done now, the `Dav1dFilmGrainData` is stored on the stack, and often will not be 16-byte aligned, and thus will often segfault.

To fix this, `ar_coeff_shift` must be 16-byte aligned.  This cannot be done only for the field without changing the offsets of
the subsequent fields, however, so we instead align `Dav1dFilmGrainData` itself with `#[repr(align(16))]`.  `ar_coeff_shift` is at offset `0xB0`/`176`, which is divisible by 16.

`psrad` also loads a full 128 bits, not just the 64 bits of `ar_coeff_shift`, even if it doesn't read them, so we must ensure that the following 64 bits are also deferenceable.  They indeed are in `dav1d`, but we must be careful, as `ar_coeff_shift` being the last field would read 8 bytes out of bounds and be UB.

Note that this bug fix comes before the next commit, which actually triggered the bug due to putting `Dav1dFilmGrainData` on the stack and at a slightly different spot than previously.  However, this was always a bug and UB all along, just not triggered by a segfault yet.  Putting this bug fix commit first means all the commits will continue to pass all tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ target/
 /retranspile/
 
 .gdb_history
+.gdbinit
--- a/include/dav1d/headers.rs
+++ b/include/dav1d/headers.rs
@@ -1031,15 +1031,11 @@ impl From<Rav1dLoopfilterModeRefDeltas> for Dav1dLoopfilterModeRefDeltas {
 }
 
 #[derive(Clone)]
-#[repr(C)]
 pub struct Rav1dFilmGrainData {
     pub seed: c_uint,
     pub num_y_points: c_int,
     pub y_points: [[u8; 2]; 14],
     pub chroma_scaling_from_luma: bool,
-    pub overlap_flag: bool,
-    pub clip_to_restricted_range: bool,
-    // 1 byte of padding
     pub num_uv_points: [c_int; 2],
     pub uv_points: [[[u8; 2]; 10]; 2],
     pub scaling_shift: c_int,
@@ -1051,6 +1047,8 @@ pub struct Rav1dFilmGrainData {
     pub uv_mult: [c_int; 2],
     pub uv_luma_mult: [c_int; 2],
     pub uv_offset: [c_int; 2],
+    pub overlap_flag: bool,
+    pub clip_to_restricted_range: bool,
 }
 
 /// Must be 16-byte aligned for `psrad` on [`Self::ar_coeff_shift`].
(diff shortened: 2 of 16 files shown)


randomPoison (approved):

fbossen (approved):
LGTM. BTW, I believe the alignment issue may arise only on older 32-bit systems. Alignment requirements have been relaxed over time but I don't quite remember when that happened.

kkysen:
> BTW, I believe the alignment issue may arise only on older 32-bit systems. Alignment requirements have been relaxed over time but I don't quite remember when that happened.

Ahh.  I've been testing on `aarch64-unknown-linux-gnu` through `qemu`; would I run into 32-bit things?

PR #556, opened 2023-11-07
`static dav1d_dq_tbl`: Format like in C

thedataking:


kkysen (approved):
I updated the PR name to explain we're formatting it how it is in C (reformat sounds like we just ran `rustfmt`), but the code change itself LGTM.

thedataking:
> I updated the PR name

Great, thank you, that is indeed more descriptive!

PR #557, opened 2023-11-08
`fn rav1d_parse_sequence_header`: get rid of `current_block`

thedataking:
We forgot to re-transpile the `goto`-less `dav1d_parse_sequence_header` to remove the `current_block_*` state machine.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -499,7 +499,16 @@ pub(crate) unsafe fn rav1d_parse_sequence_header(
     ptr: *const u8,
     sz: usize,
 ) -> Rav1dResult {
-    let mut current_block: u64;
+    unsafe fn rav1d_parse_sequence_header_error(
+        res: Rav1dResult,
+        mut c: *mut Rav1dContext,
+        buf: *mut Rav1dData,
+    ) -> Rav1dResult {
+        rav1d_data_unref_internal(buf);
+        rav1d_close(&mut c);
+        res
+    }
+
     let mut buf: Rav1dData = {
         let init = Rav1dData {
             data: 0 as *const u8,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -494,12 +494,21 @@ unsafe extern "C" fn dummy_free(data: *const u8, user_data: *mut c_void) {
     }
 }
 
+unsafe fn dav1d_parse_sequence_header_error(
```
kkysen, line 497:
```suggestion
unsafe fn rav1d_parse_sequence_header_error(
```

I've been naming `fn`s that use `Rav1d*` types `rav1d_*` and `fn`s that use `Dav1d*` types `dav1d_*`.

kkysen (commented):
Oh did I forget to re-transpile this change?  Did you make these changes from re-transpiling?  Cause I see in `src/lib.c` `dav1d_parse_sequence_header_error` is there already.

src/lib.rs
```diff
@@ -530,45 +539,30 @@ pub(crate) unsafe fn rav1d_parse_sequence_header(
...
-                    current_block = 10647346020414903899;
-                }
-            }
-        }
+
+    while buf.sz > 0 {
+        let res = rav1d_parse_obus(c, &mut buf, 1 as c_int);
+        let Ok(res) = res else {
+            return dav1d_parse_sequence_header_error(res.map(|_| ()), c, &mut buf);
+        };
```
kkysen, line 550:
```suggestion
        let res = match res {
            Ok(res) => res,
            Err(res) => return dav1d_parse_sequence_header_error(Err(res), c, &mut buf),
        }
```

Is this clearer?

kkysen (approved):
LGTM once `dav1d_parse_sequence_header_error` is renamed to `rav1d_parse_sequence_header_error`.  You can also make it an inner `fn error` inside of `fn rav1d_parse_sequence_header` if you want; I did that for a few others of these, though eventually we'll want to handle this with `Drop`.

src/lib.rs
thedataking: yes, this is much better, I didn't like that I had to call res.map(...) to convert between the error types. Thanks!

thedataking:
> Did you make these changes from re-transpiling?

I did this by hand, I added `dav1d_parse_sequence_header_error` by hand, I didn't see an existing copy in the Rust sources.

kkysen (approved):

PR #559, opened 2023-11-09
`fn fguv_32x32xn_*`: Make `{uv,is_uv,uv_pl}` args `bool`s, fixing a perf regression due to bounds checks

kkysen:
This makes `{uv,is_uv,uv_pl}` args `bool`s, casting to `c_int`/`usize` for FFI/indexing.

By making the arg a `bool` in `fn fguv_32x32xn_rust` and doing the `as usize` cast inside the `fn`[1], (4[2]) bounds checks of `uv` into length-2 arrays are eliminated, fixing a perf regression.  This was checked on `aarch64-unknown-linux-gnu` asm.

[1] Doing the `bool` to `usize` cast inside of the `add_noise_uv` closure didn't make a difference in the number of bounds checks.
[2] 27 => 23 `bl core::panicking::panic_bounds_check`s

### DIFF
--- a/src/fg_apply.rs
+++ b/src/fg_apply.rs
@@ -238,7 +238,7 @@ pub(crate) unsafe fn rav1d_apply_grain_row<BD: BitDepth>(
                 row,
                 luma_src,
                 r#in.stride[0],
-                pl,
+                pl != 0,
                 is_id,
                 bd,
             );
@@ -258,7 +258,7 @@ pub(crate) unsafe fn rav1d_apply_grain_row<BD: BitDepth>(
                     row,
                     luma_src,
                     r#in.stride[0],
-                    pl,
+                    pl != 0,
                     is_id,
                     bd,
                 );
--- a/src/filmgrain.rs
+++ b/src/filmgrain.rs
@@ -153,7 +153,7 @@ impl FnFGUV32x32xN {
         row_num: usize,
         luma_row: *const BD::Pixel,
         luma_stride: ptrdiff_t,
-        uv_pl: usize,
+        is_uv: bool,
         is_id: bool,
         bd: BD,
     ) {
@@ -165,7 +165,7 @@ impl FnFGUV32x32xN {
         let bh = bh as c_int;
         let row_num = row_num as c_int;
         let luma_row = luma_row.cast();
-        let uv_pl = uv_pl as c_int;
+        let uv_pl = is_uv as c_int;
         let is_id = is_id as c_int;
         let bd = bd.into_c();
         (self.get())(
@@ -706,13 +706,13 @@ unsafe fn fguv_32x32xn_rust<BD: BitDepth>(
     row_num: usize,
     luma_row: *const BD::Pixel,
     luma_stride: ptrdiff_t,
-    uv: usize,
+    is_uv: bool,
     is_id: bool,
     is_sx: bool,
     is_sy: bool,
     bd: BD,
 ) {
-    let [sx, sy] = [is_sx, is_sy].map(|it| it as usize);
+    let [uv, sx, sy] = [is_uv, is_sx, is_sy].map(|it| it as usize);
 
     let rows = 1 + (data.overlap_flag && row_num > 0) as usize;
     let bitdepth_min_8 = bd.bitdepth() - 8;
@@ -894,7 +894,7 @@ unsafe extern "C" fn fguv_32x32xn_c_erased<
         row_num,
         luma_row,
         luma_stride,
-        uv_pl,
+        uv_pl != 0,
         is_id,
         IS_SX,
         IS_SY,


fbossen:
For some reason, on my aarch64 Darwin system, the bounds check goes away only when `let uv = is_uv as usize` is done inside the `add_noise_uv` closure (which isn't inlined)

fbossen:
BTW, I am building with `cargo build --release --no-default-features --features=bitdepth_8,bitdepth_16` to default to the Rust code (no asm)

kkysen:
> BTW, I am building with `cargo build --release --no-default-features --features=bitdepth_8,bitdepth_16` to default to the Rust code (no asm)

I've been using `cargo asm --lib --target aarch64-unknown-linux-gnu --rust --no-default-features --features bitdepth_8,bitdepth_16`, and `cargo asm` defaults to `--release`, so this should be the same except for the OS.  It's weird that they're different.

I can move the `let uv = is_uv as usize;` inside the closure, but I'd like to figure out what's causing the difference so I can avoid it in the future, as there's a lot of similar `bool` indexing in the rest of `rav1d`.

kkysen:
> For some reason, on my aarch64 Darwin system, the bounds check goes away only when `let uv = is_uv as usize` is done inside the `add_noise_uv` closure (which isn't inlined)

So to me it seems like the issue is that on `aarch64-apple-darwin`, the closure `add_noise_uv` isn't inlined, which means `let uv = is_uv as usize` has to be inside the closure for the bounds checks to be elided.  If the closure is inlined, like it is on `aarch64-unknown-linux-gnu`, then LLVM can see the `let uv = is_uv as usize` in the `fn` and elide the bounds checks.  If `move` is applied to the closure, does that make it inlined on macOS?  And does that then fix the bounds checks?

`add_noise_uv` should be inlined for perf reasons regardless, right?  It is in C at least since it's done with a macro.  Also, making the closures `move` might not work once they are made safe due to borrow checker reasons, so that might not be a long-term solution (I'd have to think harder to see if there would actually be a conflict).

fbossen:
I tried on an x86-64 system running Ubuntu. The behaviour is the same as on aarch64 Darwin: no inlining and need to do the type conversion inside the closure

fbossen:
Output from `perf`:
```
       │     let combined = avg.as_::<c_int>() * data.uv_luma_mult[uv]                                                                                                 
  0.00 │       mov    0x50(%rdi),%rcx                                                                                                                               
  0.00 │       mov    (%rcx),%r8                                                                                                                            
  0.00 │       cmp    $0x2,%r8                                                                                                                                               
       │     ↓ jae    170                                                                                                                                                      
       │     + (*src).as_::<c_int>() * data.uv_mult[uv];                                                                                                                    
  3.85 │       movzwl (%r10,%rsi,2),%ebp                                                                                                                                   

```

fbossen:
Adding `move` doesn't impact inlining and bounds check

thedataking:
> add_noise_uv should be inlined for perf reasons regardless, right? It is in C at least since it's done with a macro.

I'm not sure that replacing macros with closures and hoping for them to be inlined is the way to go. If the changes made here make a difference to performance, that's a very brittle strategy that could very well be undone by rolling to a newer `rustc` or looking at another target; it also goes against the principle of least surprise.

Did you consider other viable solutions (rust macros or just duplicating the code expanded from the C macro)?

kkysen:
> Output from `perf`:
> 
> ```assembly
>        │     let combined = avg.as_::<c_int>() * data.uv_luma_mult[uv]                                                                                                 
>   0.00 │       mov    0x50(%rdi),%rcx                                                                                                                               
>   0.00 │       mov    (%rcx),%r8                                                                                                                            
>   0.00 │       cmp    $0x2,%r8                                                                                                                                               
>        │     ↓ jae    170                                                                                                                                                      
>        │     + (*src).as_::<c_int>() * data.uv_mult[uv];                                                                                                                    
>   3.85 │       movzwl (%r10,%rsi,2),%ebp                                                                                                                                   
> ```

If you're still getting the non-inlined closure and bounds checks on `x86_64-unknown-linux-gnu`, that's weird.  I don't see that at all.  I see this (for roughly that same section):

```asm
                // /home/kkysen/work/rust/rav1d/src/filmgrain.rs : 786
                let combined = avg.as_::<c_int>() * data.uv_luma_mult[uv]
        imul ecx, dword ptr [r9 + 4*r10 + 136]

        mov r8d, dword ptr [r9 + 4*r10 + 128]
                // /home/kkysen/work/rust/rav1d/src/filmgrain.rs : 787
                + (*src).as_::<c_int>() * data.uv_mult[uv];
        imul r8d, edx
                // /home/kkysen/work/rust/rav1d/src/filmgrain.rs : 786
                let combined = avg.as_::<c_int>() * data.uv_luma_mult[uv]
        add r8d, ecx

                // /home/kkysen/work/rust/rav1d/src/filmgrain.rs : 789
                (combined >> 6) + data.uv_offset[uv] * ((1 as c_int) << bitdepth_min_8),
        sar r8d, 6

        add r8d, dword ptr [r9 + 4*r10 + 144]
```

And the only `call`s I see are to panicking functions, no closure or any other `rav1d` functions.

But I checked the asm by disassembling `target/release/dav1d` and I see totally different asm, probably similar to yours, so I'm not sure what's going on with `cargo asm` now (though if we can get that asm, it seems a lot better).

kkysen:
Okay, I think I figured it out.  `cargo asm` was compiling things with `-Ccodegen-units=1`, which we weren't (by default I think it's 16 for `release`/non-incremental builds).  If we add `[profile.release] codegen-units = 1` to our `Cargo.toml`, the asm from a normal `cargo build` now matches what `cargo asm` is showing.  I'll add another PR for that, and then can you check if the inlining and bounds checks are fixed?

randomPoison:
@kkysen Do we still need this PR, or does the `codegen-units = 1` change address the perf regression? Or is that what you're asking Frank to confirm once #563 gets merged?

kkysen:
> @kkysen Do we still need this PR, or does the `codegen-units = 1` change address the perf regression? Or is that what you're asking Frank to confirm once #563 gets merged?

We still definitely need this PR.  It's important that the `bool` to `usize` cast is done within the same function as where it's used to index to avoid bounds checks.  #563 is also needed to fully enable this fix, though, as it fixes the inlining issue.  And later I'll explore converting the closures to macros to guarantee inlining (`#[inline(always)]` on closures is still unstable), but I'll do that after I've refactored and cleaned up the code around and in the closure.

kkysen:
@fbossen, this one's good to merge now that #563 fixes the inlining, right?

fbossen:
> @fbossen, this one's good to merge now that #563 fixes the inlining, right?

Yes

fbossen (approved):

PR #560, opened 2023-11-09
`fn fguv 32x32xn_rust`: Elide `val` bounds checks, fixing a perf regression

kkysen:
This elides bounds checks from indexing with `val` by clipping the index to a valid one.  Initially, I copied #558's approach, which uses `cmp::min(val, scaling.as_ref().len() - 1)`, and this removes 3 bounds checks.  But since the scaling size is a `const` and is `1 << BD::SCALING_BITS`, we can just truncate the other bits with `&`, which avoids a `csel` and 2 more bounds checks.

### DIFF
--- a/include/common/bitdepth.rs
+++ b/include/common/bitdepth.rs
@@ -119,7 +119,8 @@ pub trait BitDepth: Clone + Copy {
         + ToPrimitive<c_int>;
 
     type Scaling: AsRef<[u8]> + AsMut<[u8]> + ArrayDefault + Copy;
-    const SCALING_SIZE: usize;
+    const SCALING_BITS: usize;
+    const SCALING_SIZE: usize = 1 << Self::SCALING_BITS;
 
     type BitDepthMax;
 
@@ -206,7 +207,7 @@ impl BitDepth for BitDepth8 {
     type Entry = i8;
 
     type Scaling = [u8; Self::SCALING_SIZE];
-    const SCALING_SIZE: usize = 1 << 8;
+    const SCALING_BITS: usize = 8;
 
     type BitDepthMax = ();
 
@@ -287,7 +288,7 @@ impl BitDepth for BitDepth16 {
     type Entry = i16;
 
     type Scaling = [u8; Self::SCALING_SIZE];
-    const SCALING_SIZE: usize = 1 << 12;
+    const SCALING_BITS: usize = 12;
 
     type BitDepthMax = Self::Pixel;
 
--- a/src/filmgrain.rs
+++ b/src/filmgrain.rs
@@ -795,7 +795,7 @@ unsafe fn fguv_32x32xn_rust<BD: BitDepth>(
             // `val` isn't out of bounds, so we can
             // eliminate extra panicking code by bit-truncating `val`.
             let noise = round2(
-                scaling.as_ref()[val as usize & ((1 << BD::SCALING_BITS) - 1)] as c_int * grain,
+                scaling.as_ref()[val as usize % scaling.as_ref().len()] as c_int * grain,
                 data.scaling_shift,
             );
             *dst = iclip((*src).as_::<c_int>() + noise, min_value, max_value).as_::<BD::Pixel>();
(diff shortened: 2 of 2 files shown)


fbossen (approved):
LGTM

randomPoison:
I'm a bit confused by your description on this PR 🤔 You mention that the approach taken in #558 is to use `cmp::min` to truncate `val` to the array length removes 3 bounds, and that this change using `&` removes 2 more. But I'm unclear on a couple of things:

* In the first case, how does using `cmp::min` allow for bounds checks to be elided? `cmp::min` isn't a const fn, so isn't the compiler still going to need to include the bounds check, even if we know that the clamped index will already be in bounds?
* What do you mean by saying that 3 bounds checks are removed in that case? There's only one index operation happening, so as I understand it there's only one bounds check to be elided? Or is this a case where making the change in one place allows the bounds check to be removed in multiple other places? I'm not seeing how that would be the case, though.
* What do you mean when you say 2 additional checks are elided by the change in this PR? This is basically the same as the last point, but I'm not seeing where there are 2 additional checks to remove.

None of this needs to block the PR since @fbossen already approved, mostly I just want to make sure I understand what the changes are doing wrt this performance regression yall are fixing :)

kkysen:
> I'm a bit confused by your description on this PR 🤔 You mention that the approach taken in #558 is to use `cmp::min` to truncate `val` to the array length removes 3 bounds, and that this change using `&` removes 2 more. But I'm unclear on a couple of things:
> 
> * In the first case, how does using `cmp::min` allow for bounds checks to be elided? `cmp::min` isn't a const fn, so isn't the compiler still going to need to include the bounds check, even if we know that the clamped index will already be in bounds?

So there is still a `cmp` checking against the array length, but since the out-of-bounds case is now another number instead of a call to `core::panicking::panic_bounds_check`, `csel` (a conditional select) is used instead of a branch to a block that calls `core::panicking::panic_bounds_check`.  So there kind of still is a bounds check, it's just cheaper, and the load always happens unconditionally now (since the index after the `cmp::min` is guaranteed to be in bounds).

> * What do you mean by saying that 3 bounds checks are removed in that case? There's only one index operation happening, so as I understand it there's only one bounds check to be elided? Or is this a case where making the change in one place allows the bounds check to be removed in multiple other places? I'm not seeing how that would be the case, though.

I'm not sure exactly.  I determined this by counting the number of `bl core::panicking::panic_bounds_checks` were in the arm64 asm.  Sometimes loops get unrolled and stuff so one bounds check in the source code can turn into multiple in the assembly.

> * What do you mean when you say 2 additional checks are elided by the change in this PR? This is basically the same as the last point, but I'm not seeing where there are 2 additional checks to remove.

I'm not sure why the actual calls to `core::panicking::panic_bounds_check` were removed (what I was counting), but doing an `and` instead of a `csel` is definitely significantly cheaper.

Also, I think I'll replace the `& ((1 << BD::SCALING_BITS) - 1)` with `% scaling.as_ref().len()`, since I think it's clearer what it's doing and why it will result in an in-bounds index, but it will still be optimized the same into an `and`

fbossen:
> Also, I think I'll replace the `& ((1 << BD::SCALING_BITS) - 1)` with `% scaling.as_ref().len()`, since I think it's clearer what it's doing and why it will result in an in-bounds index, but it will still be optimized the same into an `and`

Have you looked at the generated asm when using the `%` operator?

kkysen:
> Have you looked at the generated asm when using the `%` operator?

Yeah, it's still using `and` for it, the same as before.

kkysen:
> > Have you looked at the generated asm when using the `%` operator?
> 
> Yeah, it's still using `and` for it, the same as before.

On `x86_64`, it's not using `and`, but it's the same either way (no `and` even with `&`), some `mov`s and a `movzx` that I'm not sure about:

```asm
                // /home/kkysen/work/rust/rav1d/src/filmgrain.rs : 798
                scaling.as_ref()[val as usize & ((1 << BD::SCALING_BITS) - 1)] as c_int * grain,
        mov ecx, ecx

        mov r15, qword ptr [rsp + 144]
        movzx ecx, byte ptr [r15 + rcx]
		
                // /home/kkysen/work/rust/rav1d/src/filmgrain.rs : 798
                scaling.as_ref()[val as usize % scaling.as_ref().len()] as c_int * grain,
        mov ecx, ecx

        mov r15, qword ptr [rsp + 144]
        movzx ecx, byte ptr [r15 + rcx]
```

Either way it's not doing a bounds check, though.

PR #562, opened 2023-11-09
`mod {cdef,{cdef,lf,lr}_apply}_tmpl_{8,16}`: Always compile both and dead-code eliminate

kkysen:
Since these bitdepth-specific `fn`s are Rust `fn`s, not asm `fn`s, we can't use `bd_fn!` in the same way, so when `mod recon` got deduplicated, it broke compiling for a single bitdepth. If we just compile both bitdepths for these `mod`s and rely on dead code elimination, it works again. And once we deduplicate these `mod`s, it will all be for naught since they'll be generic instead.

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -30,13 +30,13 @@ pub mod include {
 pub mod src {
     pub mod align;
     mod cdef;
-    #[cfg(feature = "bitdepth_16")]
+    #[cfg_attr(not(feature = "bitdepth_16"), allow(dead_code))]
     mod cdef_apply_tmpl_16;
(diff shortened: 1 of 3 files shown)


randomPoison (approved):
This seems like an awkward solution, but sounds like it's only meant to be temporary? If that's the case I'm fine with this as a stopgap solution until we deduplicate these modules.

kkysen:
> This seems like an awkward solution, but sounds like it's only meant to be temporary? If that's the case I'm fine with this as a stopgap solution until we deduplicate these modules.

Yeah, it should just be temporary.

PR #563, opened 2023-11-10
`Cargo.toml`: Set `codegen-units = 1` for `release`

kkysen:
This significantly improves code generation and optimization. In C, translation units correspond to each `*.c` file, and thus things are deliberately grouped together. In Rust, codegen units are chosen with a heuristic and functions are assigned to a codegen unit (passed to LLVM). Thus, to ensure related functions are in the same codegen unit (and are thus inlineable), we should set this to 1. This will also make optimization much more reliable, since a far away change can't suddenly affect an optimization elsewhere.

We probably also want to look at enabling LTO (though we'll need to enable LTO for `dav1d` for equivalent comparisons).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,3 +49,6 @@ bitdepth_16 = []
 # unless we enable the basic optimizations (`opt-level = 1`),
 # which don't increase compile-time that much.
 opt-level = 1
+
+[profile.release]
+codegen-units = 1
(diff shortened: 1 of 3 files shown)


kkysen:
This should hopefully fix the inlining issue in #559.  @fbossen, can you confirm on your aarch64 machine?  I'll also try out a macro (vs. a closure) version to see how that compares, but I want to check if `codegen-units = 1` fixes the inlining issue regardless.

fbossen:
I confirm this results in inlining of the closure on my aarch64 system.

randomPoison (approved):

PR #564, opened 2023-11-11
`fn fguv_32x32xn_rust`: Fix a `bd.iclip_pixel` call in the `add_noise_uv` closure

kkysen:


randomPoison (approved):
Huh, did this not get caught in CI? How'd you notice it? Just looking at the C sources?

kkysen:
> How'd you notice it? Just looking at the C sources?

Yup!  Luckily.

> Huh, did this not get caught in CI?

I think `iclip_pixel` has a default implementation and so it's actually the same for both bitdepths.

PR #566, opened 2023-11-11
`src/filmgrain.rs`: Setup and use `clippy` so we can use it to incrementally fix things

kkysen:


randomPoison (approved):
Oh dope, it'll be good to have clippy lints now.

PR #568, opened 2023-11-11
`fn generate_grain_{y,uv}_rust`: Replace indexing `ar_lag` loops with iterators to elide bounds checks

kkysen:
This also explicitly truncates `ar_lag` to 2 bits (so `ar_lag <= ar_pad`), which it already is when being set, but the compiler can't see that (a `u2` would be appropriate). `ar_lag <= ar_pad` proves all the bounds are in-bounds, but the compiler only sees this when using slicing and iterators.

There are still some bounds checks for the luma code, but that only happens once per inner loop so it's less bad.

This increases perf by ~4-5% on `donna` and removes 4 bounds checks per `fn`.

### DIFF
--- a/src/filmgrain.rs
+++ b/src/filmgrain.rs
@@ -387,26 +387,30 @@ unsafe fn generate_grain_y_rust<BD: BitDepth>(
     }
 
     let ar_pad = 3;
-    let ar_lag = data.ar_coeff_lag as isize;
+    // `ar_lag` is 2 bits; this tells the compiler it definitely is.
+    // That also means `ar_lag <= ar_pad`.
+    let ar_lag = data.ar_coeff_lag as usize & ((1 << 2) - 1);
 
-    for y in ar_pad..GRAIN_HEIGHT {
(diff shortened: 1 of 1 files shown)


src/filmgrain.rs
```diff
@@ -387,26 +387,30 @@ unsafe fn generate_grain_y_rust<BD: BitDepth>(
     }
 
     let ar_pad = 3;
-    let ar_lag = data.ar_coeff_lag as isize;
+    // `ar_lag` is 2 bits; this tell the compile it definitely is.
```
randomPoison, line 390:
```suggestion
    // `ar_lag` is 2 bits; this tells the compile it definitely is.
```

randomPoison (approved):

fbossen (approved):
LGTM. I am seeing a small reduction of overall run time on aarch64.

src/filmgrain.rs
kkysen: By the way, do you know how to use the "commit suggestion" button but also rebase/amend the previous commit?

PR #570, opened 2023-11-16
CI: Build on self-hosted runner

thedataking:


.github/workflows/build-and-benchmark-x86.yml
```diff
@@ -0,0 +1,47 @@
+name: build and benchmark on x86-64
+on: [push, pull_request]
+jobs:
+    benchmark-on-self-hosted-x86-64:
+        runs-on: [self-hosted, linux, X64]
+        steps:
+            - name: git checkout
+              uses: actions/checkout@v4
+            - name: release build this PR
+              run:  nice cargo build --release
```
kkysen, line 10: Why do we need `nice` for compiling?

.github/workflows/build-and-benchmark-x86.yml
```diff
@@ -0,0 +1,47 @@
...
+            - name: release build main branch
+              run: |
+                git fetch --depth 1 origin/main && git checkout origin/main
+                nice cargo build --release --target-dir target.main
+            - name: benchmark on chimera 8-bit test data
+              run: |
+                mkdir -p `dirname $LOCAL_FILE`
```
kkysen, line 17:
```suggestion
                mkdir -p $(dirname $LOCAL_FILE`)
```

.github/workflows/build-and-benchmark-x86.yml
```diff
@@ -0,0 +1,47 @@
...
+                input=$(basename $LOCAL_FILE)
+                perf record -o perf.data.$input      -- target/release/dav1d -q -i $LOCAL_FILE -o /dev/null
+                perf record -o perf.data.$input.main -- target.main/release/dav1d -q -i $LOCAL_FILE -o /dev/null
+                perf diff perf.data.$input perf.data.$input.main | head -n 100
+              env:
+                REMOTE_URL: http://download.opencontent.netflix.com.s3.amazonaws.com/AV1/Chimera/Old/Chimera-AV1-8bit-1280x720-3363kbps.ivf
+                LOCAL_FILE: /tmp/rav1d/Chimera-AV1-8bit-1280x720-3363kbps.ivf
+            - name: benchmark on chimera 10-bit test data
+              run: |
+                mkdir -p `dirname $LOCAL_FILE`
```
kkysen, line 31:
```suggestion
                mkdir -p $(dirname $LOCAL_FILE)
```

.github/workflows/build-and-benchmark-x86.yml
thedataking: Because this machine has interactive users which we want to bother as little as possible.

.github/workflows/build-and-benchmark-x86.yml
thedataking: It makes no difference which syntax is used here.

kkysen (commented):
Do we also want to benchmark the `summer_nature` videos?  When @fbossen and I've been benchmarking locally, we've been running those, too.

Also, we should fill filesystem caches for these files before running, right?  Locally, the first run is always significantly slower after I haven't benchmarked in a while.  Similarly, should we run a few repetitions instead of just one?

.github/workflows/build-and-benchmark-x86.yml
```diff
@@ -0,0 +1,47 @@
...
+    benchmark-on-self-hosted-x86-64:
+        runs-on: [self-hosted, linux, X64]
+        steps:
+            - name: git checkout
+              uses: actions/checkout@v4
+            - name: release build this PR
+              run:  nice cargo build --release
+            - name: release build main branch
+              run: |
+                git fetch --depth 1 origin/main && git checkout origin/main
```
kkysen, line 13: I don't think `git fetch --depth 1 origin/main` works at all (the `--depth 1` part is fine).

PR #571, opened 2023-11-16
CI: Fix git fetch command

thedataking:
Caught by kkysen during review.


kkysen (approved):
Looks like it's building from `main` correctly now.

PR #574, opened 2023-11-16
`src/ipred.rs`: Deduplicate w/ generics

kkysen:


src/ipred.rs
```diff
@@ -1,11 +1,50 @@
...
 
-#[inline]
-pub unsafe fn get_upsample(wh: c_int, angle: c_int, is_sm: c_int) -> c_int {
-    return (angle < 40 && wh <= 16 >> is_sm) as c_int;
-}
+#[cfg(feature = "asm")]
+use crate::{include::common::bitdepth::bd_fn, src::cpu::rav1d_get_cpu_flags, src::cpu::CpuFlags};
+
+#[cfg(all(feature = "asm", target_arch = "aarch64"))]
+use ::to_method::To;
```
randomPoison, line 41: Why does this path start with `::`? Could it just be `use to_method::To;`?

randomPoison (approved):

src/ipred.rs
kkysen: Yeah, they're equivalent, but for `#[cfg]`ed imports, I usually group them into

```rust
#[cfg(something)]
use crate::{
   // all `crate::` imports
};

#[cfg(something)]
use ::{
   // all absolute imports
};
```

and then that reformats into this.  It's easier if we have to add more non-`crate::` imports to this later on.

PR #576, opened 2023-11-19
`src/loopfilter.rs`: Deduplicate w/ generics

thedataking:


kkysen (commented):
Looks good so far!  One little thing is that I've been using `crate`-relative imports, one per line (except for `#[cfg]`s) since there's often a ton and it helps keep the diffs clean.  I have this in my `.vscode/settings.json` to have rust-analyzer do this automatically; there's probably similar settings for other editors, too:

```json
{
    "rust-analyzer.imports.granularity.group": "item",
    "rust-analyzer.imports.prefix": "crate",
    ...
}
```

src/loopfilter.rs
```diff
@@ -1000,3 +1008,132 @@ unsafe fn loop_filter_v_sb128uv_rust<BD: BitDepth>(
...
+
+    #[cfg(feature = "asm")]
+    cfg_if! {
+        if #[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {
+            loop_filter_dsp_init_x86::<BitDepth16>(c);
+        } else if #[cfg(any(target_arch = "arm", target_arch = "aarch64"))] {
+            loop_filter_dsp_init_arm::<BitDepth16>(c);
+        }
+    }
+}
```
kkysen, line 1139: This is a fine intermediate state, but this doesn't actually deduplicate `fn rav1d_loop_filter_dsp_init`.

src/loopfilter.rs
```diff
@@ -384,3 +395,726 @@ extern "C" {
         bitdepth_max: c_int,
     );
 }
+
+// TODO(perl) Temporarily pub until mod is deduplicated
+#[inline(never)]
+pub(crate) unsafe fn loop_filter<BD: BitDepth>(
```
kkysen, line 401: Still more of these 🤣

src/loopfilter.rs
```diff
@@ -384,3 +395,726 @@ extern "C" {
...
+        x <<= 1;
+        dst = dst.offset(4);
+        l = l.offset(1);
+    }
+}
+
+#[cfg(all(feature = "asm", any(target_arch = "x86", target_arch = "x86_64")))]
+#[inline(always)]
+unsafe fn loop_filter_dsp_init_x86<BD: BitDepth>(c: *mut Rav1dLoopFilterDSPContext) {
+    use crate::include::common::bitdepth::BPC;
```
kkysen, line 1013: I'd put this and the one in the arm `fn` under the `#[cfg(feature = "asm")]` imports.

src/loopfilter.rs
```diff
@@ -384,3 +395,726 @@ extern "C" {
...
+                        128 * ((1 as c_int) << bitdepth_min_8) - 1,
+                    );
+                    f1 = cmp::min(f + 4, ((128 as c_int) << bitdepth_min_8) - 1) >> 3;
+                    f2 = cmp::min(f + 3, ((128 as c_int) << bitdepth_min_8) - 1) >> 3;
+                    *dst.offset(strideb * -(1 as c_int) as isize) =
+                        iclip(p0 + f2, 0 as c_int, bd.bitdepth_max().as_::<c_int>())
+                            .as_::<BD::Pixel>();
+                    *dst.offset((strideb * 0) as isize) =
+                        iclip(q0 - f1, 0 as c_int, bd.bitdepth_max().as_::<c_int>())
+                            .as_::<BD::Pixel>();
```
kkysen, line 709: Should be `bd.iclip_pixel`s.

src/loopfilter.rs
```diff
@@ -384,3 +395,726 @@ extern "C" {
...
+                    *dst.offset((strideb * 0) as isize) =
+                        iclip(q0 - f1_0, 0 as c_int, bd.bitdepth_max().as_::<c_int>())
+                            .as_::<BD::Pixel>();
+                    f_0 = f1_0 + 1 >> 1;
+                    *dst.offset(strideb * -(2 as c_int) as isize) =
+                        iclip(p1 + f_0, 0 as c_int, bd.bitdepth_max().as_::<c_int>())
+                            .as_::<BD::Pixel>();
+                    *dst.offset((strideb * 1) as isize) =
+                        iclip(q1 - f_0, 0 as c_int, bd.bitdepth_max().as_::<c_int>())
+                            .as_::<BD::Pixel>();
```
kkysen, line 732: Should be `bd.iclip_pixel`s.

kkysen (changes requested):

src/loopfilter.rs
thedataking: derp, sorry, I think I did them all now... :grimacing:

src/loopfilter.rs
thedataking: I think I've addressed this suggestion but I'm not 100% sure.

thedataking:
One small thing (can be resolved later): `cargo fmt` doesn't really do a good job on `loop_filter`, do we disable it and format this fn by hand?

src/loopfilter.rs
thedataking: helpful, thanks.

src/loopfilter.rs
```diff
@@ -384,3 +392,702 @@ extern "C" {
...
+                *dst.offset(strideb * -(1 as c_int) as isize) =
+                    (p2 + 2 * p1 + 2 * p0 + 2 * q0 + q1 + 4 >> 3).as_::<BD::Pixel>();
+                *dst.offset((strideb * 0) as isize) =
+                    (p1 + 2 * p0 + 2 * q0 + 2 * q1 + q2 + 4 >> 3).as_::<BD::Pixel>();
+                *dst.offset((strideb * 1) as isize) =
+                    (p0 + 2 * q0 + 2 * q1 + 2 * q2 + q2 + 4 >> 3).as_::<BD::Pixel>();
+            } else {
+                let hev = ((p1 - p0).abs() > H || (q1 - q0).abs() > H) as c_int;
+
+                fn iclip_diff(v: c_int, bitdepth_min_8: u8) -> i32 {
```
kkysen, line 686: If this isn't a closure, you can add `#[inline(always)]` to guarantee inlining.

kkysen (approved):
I made a comment about forcing `iclip_diff` to always be inlined, but other than that, it all lookLGTM now.

kkysen:
> One small thing (can be resolved later): `cargo fmt` doesn't really do a good job on `loop_filter`, do we disable it and format this fn by hand?

I don't think we should format it by hand.  That'd be really annoying.  What about increasing the line length?  Also, sometimes adding parentheses convinced `rustfmt` to format it much nicer, like for the many additions that get put on one line each.

src/loopfilter.rs
thedataking: If we add `#[inline(always)]` here, someone reading this code would be right to assume that inlining that function makes a difference performance-wise. I double checked with `perf stat` and there is no difference between never inlining and always inlining. 

I such cases, I would prefer we let the compiler chose. If someone does a build that is optimized for size (`-s` or `-z`), the compiler can choose not to inline whereas a build optimized for speed (e.g. `--release`) will probably inline this call (which indeed appears to be the case on x86).

kkysen: If that's the case, I think a closure would be better here then, as then we don't need `bitdepth_min_8` as an arg, just like it wasn't in the C `iclip_diff` macro.  But I thought you wanted to ensure inlining to match how C macros ensure inlining; in that case, an inner `fn` like this with `#[inline(always)]` will ensure that.  Overall, I'd lean in the direction of letting the compiler choose (as you said, the end user may want to optimize for different things), and so a closure would be best in that case.

PR #580, opened 2023-11-21
`wrap_fn_ptr!`: Generate a `decl_fn!` macro inside; also document macros

kkysen:
We're already matching the `fn` signature with a macro, so generating the `decl_fn!` macro inside is pretty trivial, and this way we don't have to write all of the `decl_*_fn!` macros ourselves.

This also adds documentation for:
* `trait HasFnPtr`
* `struct WrappedFnPtr`
* `wrap_fn_ptr!`
* `bd_fn!`

### DIFF
--- a/include/common/bitdepth.rs
+++ b/include/common/bitdepth.rs
@@ -422,19 +422,23 @@ where
     }
 }
 
-/// Declare and select a [`BitDepth`]-dependent `extern "C" fn`.
+/// Select and declare a [`BitDepth`]-dependent `extern "C" fn`.
+///
+/// That is, it statically selects which [`BitDepth`] `fn`
+/// (i.e., `bpc8` or `bpc16`) to return based on `$BD:ty`,
+/// declares it inline* with `$decl_fn:path`, and then returns it.
 ///
 /// # Args
 ///
 /// * `$decl_fn:path` (optional):
 ///     A path to a macro that, given a `fn $fn_name:ident`,
-///     declares and evaluates to an `extern "C" fn`
+///     declares and returns an `extern "C" fn`
 ///     with the appropriate signature for this `fn`.
 ///     This should usually be `mod::decl_fn`,
 ///     where the `mod` is defined by [`wrap_fn_ptr!`],
 ///     but it doesn't have to be.
 ///
-///     If omitted, this defaults to [`fn_identity`],
+///     \* If omitted, this defaults to [`fn_identity`],
 ///     which returns the `fn` given without declaring one inline.
 ///     This should be used when the `fn` you are selecting
 ///     is already declared elsewhere.
--- a/include/common/bitdepth.rs
+++ b/include/common/bitdepth.rs
@@ -426,13 +426,18 @@ where
 ///
 /// # Args
 ///
-/// * `$decl_fn:path`:
+/// * `$decl_fn:path` (optional):
 ///     A path to a macro that, given a `fn $fn_name:ident`,
 ///     declares and evaluates to an `extern "C" fn`
 ///     with the appropriate signature for this `fn`.
 ///     This should usually be `mod::decl_fn`,
 ///     where the `mod` is defined by [`wrap_fn_ptr!`],
 ///     but it doesn't have to be.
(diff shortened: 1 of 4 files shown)


rinon (changes requested):
I wish we didn't need the `decl_fn` macros at all, but this is cleaner than writing them all out since we already have all the params in the wrapper macro.

src/wrap_fn_ptr.rs
```diff
@@ -20,23 +20,43 @@ impl<F> WrappedFnPtr<F> {
 }
 
 macro_rules! wrap_fn_ptr {
```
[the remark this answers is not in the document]
kkysen: I'll add docs to explain all of that.  I'll be sure to add docs to any non-trivial/exported macro; I agree that's very important for macros since they're hard to read.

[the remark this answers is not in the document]
kkysen: I've added docs for `wrap_fn_ptr!` now, as well as `trait HasFnPtr`, `struct WrappedFnPtr`, and `bd_fn!`.

include/common/bitdepth.rs
```diff
@@ -422,6 +422,48 @@ where
     }
 }
 
+/// Declare and select a [`BitDepth`]-dependent `extern "C" fn`.
+///
+/// # Args
```
rinon, line 427: Version without the decl_fn is not documented, should it be used?

include/common/bitdepth.rs
```diff
@@ -422,6 +422,48 @@ where
     }
 }
 
+/// Declare and select a [`BitDepth`]-dependent `extern "C" fn`.
```
rinon, line 425: I'm not sure what select you mean in this context? I think this title needs further high-level description in addition to the arguments. Probably want to describe exactly what the declared function name will be here.

include/common/bitdepth.rs
kkysen: Oh, yes, I forgot about it.

include/common/bitdepth.rs
kkysen: Okay, I'll try to make it clearer.  By select I mean choose which function to resolve to.

src/wrap_fn_ptr.rs
```diff
@@ -19,24 +24,71 @@ impl<F> WrappedFnPtr<F> {
...
 }
 
+/// This declares a wrapper for a `fn` ptr
+/// and defines related, useful things for that `fn` ptr.
+///
+/// The API for [`wrap_fn_ptr!`] is a `fn` signature with no body.
+/// This generates a `mod` with the name of the `fn` provided that contains:
+///
+/// * `type Fn`:
+///     A [`WrappedFnPtr`] wrapping the `fn` ptr signature provided.
```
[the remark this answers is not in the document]
kkysen: > Worth mentioning that this is a newtype for the purpose of implementing methods on it.

👍

> What methods does this type provide already?

Not sure if this is worth it.  It's just `new` and `get`, fairly obvious methods, and they're linked in the docs for `WrappedFnPtr`.  And this is getting replaced in the next PR anyways so I don't really want to spend time figuring out the best docs for this if that's okay.

src/wrap_fn_ptr.rs
```diff
@@ -19,24 +24,71 @@ impl<F> WrappedFnPtr<F> {
     }
 }
 
+/// This declares a wrapper for a `fn` ptr
+/// and defines related, useful things for that `fn` ptr.
```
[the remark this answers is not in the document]
kkysen: I've added the newtype part of reworded some of it, but I explain what we use the items it generates for below for each item, and I'm not sure how to do that more succinctly at the beginning.

include/common/bitdepth.rs
kkysen: Fixed in ce61e08.

include/common/bitdepth.rs
kkysen: Fixed in 1dffe4c.

PR #581, opened 2023-11-21
`mod ipred`: Use `wrap_fn_ptr!` to generate `decl_fn!`s and add type-safe wrappers

kkysen:
This also reworks `wrap_fn_ptr!` to declare `Fn` as a newtype `struct` inside the macro rather than a type alias to the generic `WrappedFnPtr`, which needed a type token for disambiguation.  This results in mode code inside the macro vs. out of it, but the code is much simpler and easier to understand this way.

### DIFF
--- a/src/wrap_fn_ptr.rs
+++ b/src/wrap_fn_ptr.rs
@@ -52,7 +52,7 @@ macro_rules! wrap_fn_ptr {
             ///
             /// This allows us to add a safer
             /// (type-safe for sure, and increasingly fully safe)
-            /// wrapper around calling a `fn` ptr.
+            /// interface for calling a `fn` ptr.
             #[derive(Clone, Copy, PartialEq, Eq)]
             #[repr(transparent)]
             pub struct Fn(FnPtr);
--- a/src/wrap_fn_ptr.rs
+++ b/src/wrap_fn_ptr.rs
@@ -1,3 +1,5 @@
+use std::marker::PhantomData;
+
 /// A `trait` to extract the `fn` ptr type of a [`WrappedFnPtr`].
 pub trait HasFnPtr {
     type FnPtr;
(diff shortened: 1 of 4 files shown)


rinon:
What is the `WrappedFnPtr` type actually giving us here? It would be simpler to make `Fn` itself a newtype instead of making it a `pub type`. That would avoid the Token thing entirely, right?

kkysen:
> What is the `WrappedFnPtr` type actually giving us here? It would be simpler to make `Fn` itself a newtype instead of making it a `pub type`. That would avoid the Token thing entirely, right?

Yeah, the alternative to a token type is to just make a newtype, but since most of the rest besides that token type part can be a generic type, I thought it's easier that way as less code has to be inside the macro.

rinon:
The Token type is hard to understand, IMO. It took me a bit to figure out what the actual problem was, because I forgot that we were even using a pub type instead of a fresh newtype. I think the newtype is the way to go here, whether the code is inside or outside the macro doesn't matter here because we don't use `WrappedFnPtr` elsewhere.

kkysen:
> The Token type is hard to understand, IMO. It took me a bit to figure out what the actual problem was, because I forgot that we were even using a pub type instead of a fresh newtype. I think the newtype is the way to go here, whether the code is inside or outside the macro doesn't matter here because we don't use `WrappedFnPtr` elsewhere.

I tried out declaring the newtype inside the macro and it is indeed a lot cleaner and simpler.

src/wrap_fn_ptr.rs
```diff
@@ -1,37 +1,23 @@
...
-///     A [`WrappedFnPtr`] wrapping the `fn` ptr signature provided.
+///     A newtype wrapping `FnPtr`.
+///     It defines `const fn new(FnPtr) -> Self` to construct it
+///     and `const fn get(&self) -> &FnPtr` to read the `FnPtr`.
+///
+///     These methods are marked `pub(super)`
+///     as they are meant to be used in the module calling [`wrap_fn_ptr!`].
+///
+///     It is meant for a `fn call` to also be added
+///     to call the `fn` in a type-safe (e.x. [`BitDepth`]-wise)
```
rinon, line 19:
```suggestion
///     It is meant for a `fn call` method to also be implemented
///     for this type to allow users to call the `fn` in a type-safe
///     (e.x. [`BitDepth`]-wise)
```

rinon: Clarifies a couple things that took me a minute to understand, does this look ok?

src/wrap_fn_ptr.rs
```diff
@@ -56,11 +43,29 @@ macro_rules! wrap_fn_ptr {
...
             use super::*;
 
-            pub type Fn = WrappedFnPtr<unsafe extern "C" fn($($arg_name: $arg_ty),*) -> $return_ty>;
+            pub type FnPtr = unsafe extern "C" fn($($arg_name: $arg_ty),*) -> $return_ty;
+
+            /// A newtype wrapped [`FnPtr`].
+            ///
+            /// This allows us to add a safer
+            /// (type-safe for sure, and increasingly fully safe)
+            /// wrapper around calling a `fn` ptr.
```
rinon, line 55:
```suggestion
            /// method for calling a `fn` ptr.
```

src/wrap_fn_ptr.rs
kkysen: Yeah, that sounds clearer.

src/wrap_fn_ptr.rs
kkysen: This does sound clearer in a way but I meant to imply by wrapper that it's the only public way to call it, whereas method implies more there is also a safer way to call it, but you can still call it the more unsafe way.  If you still think method is clearer though I'll go with it; I don't think it matters too much.

src/wrap_fn_ptr.rs
kkysen: Fixed in 1816aaa.

src/wrap_fn_ptr.rs
kkysen: I changed it in e885327 to `interface for` instead, as I think that better gets to the root purpose of this type: providing a safer interface, whatever that interface may be.  And I think it implies that this safer interface should shadow/wrap the raw, unsafe one.

rinon (approved):

PR #582, opened 2023-11-21
`mod ipred::neon`: Move all `fn *_neon`s to their own `mod`

kkysen:
This is how things are done in C, and it simplifies the `#[cfg]`s a lot, too.

### DIFF
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -39,15 +39,6 @@ use strum::FromRepr;
 #[cfg(feature = "asm")]
 use crate::{include::common::bitdepth::bd_fn, src::cpu::rav1d_get_cpu_flags, src::cpu::CpuFlags};
 
-#[cfg(all(feature = "asm", target_arch = "aarch64"))]
-use ::to_method::To;
-
-#[cfg(all(feature = "bitdepth_8", feature = "asm", target_arch = "aarch64"))]
-use crate::include::common::bitdepth::BitDepth8;
-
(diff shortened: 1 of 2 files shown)


rinon:
Would it make sense to move it into a separate file? I don't have a strong preference, and it's a minor thing, but not having to indent for the submodule might be nice.

kkysen:
> Would it make sense to move it into a separate file? I don't have a strong preference, and it's a minor thing, but not having to indent for the submodule might be nice.

Yeah it definitely could be.  And it is in C.  Should we just file an issue to do it later at some point?

rinon:
Is there any reason not to do it in this PR? You've already spent the time to split it into a separate module, so adding the file should be easy enough, right?

kkysen:
Sure I can do it here, it's pretty simple.

kkysen:
Actually, in C, the `fn *_dsp_init_{x86,arm}`s are also in their own arch-specific file, but that's more unwieldy for us when fully cleaned up (see the final cleaned up version in `mod filmgrain`).  And if I move `mod neon` to its own file, I wouldn't want to include a `use super::*;` in it like I have now, which just adds more noise.  I don't think the benefit of not indenting everything extra is worth it, especially for modules smaller than `ipred`.  So if it's okay with you, I'll keep things as is.

rinon (approved):
It's not a huge deal to me either way, so ok, let's keep it as an inline sub-module for now.

PR #583, opened 2023-11-21
`struct Rav1dIntraPredDspContext`: Fix defaults

kkysen:
This replaces the `Option`s on `.intra_pred` with an `initialized: bool` field as they are always `Some` even if `.is_none()` is checked.

`.is_none()` appears to be checked so that it can lazily initialize DSPs, relying on things being zero-initialized to begin with. 
 Ideally we should move initialization to DSP allocation so its not lazy, but for now (and the way closest to the C), we can use a single initialized flag instead of `Option`s on everything.

This also default initializes `.cfl_pred` as it otherwise isn't fully initialized and is thus leaving uninitialized memory, which is UB if it is made a reference at any point (which it is).  This uses `DefaultValue::DEFAULT` for initialization instead of `None`, though, which moves the check from before the call (`.unwrap()`/`.expect()`) to inside a call that should never happen (`unimplemented!()`).  Normally those `unimplemented!()`s will be dead code eliminated, but here they won't.  Thus, it shouldn't add any perf cost and only a small binary size cost.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -5045,8 +5045,12 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
         }
     }
 
-    if (*f.dsp).ipred.intra_pred[DC_PRED as usize].is_none() {
+    // TODO(kkysen) Rather than lazy initializing this,
+    // we should probably initialize all the fn ptrs
+    // when `c` is allocated during [`rav1d_open`].
+    if !(*f.dsp).initialized {
         let dsp = &mut c.dsp[(*f.seq_hdr).hbd as usize];
+        dsp.initialized = true;
 
         match bpc {
             #[cfg(feature = "bitdepth_8")]
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -155,9 +155,7 @@ impl pal_pred::Fn {
 
 #[repr(C)]
 pub struct Rav1dIntraPredDSPContext {
-    // TODO(legare): Remove `Option` once `dav1d_submit_frame` is no longer checking
-    // this field with `is_none`.
-    pub intra_pred: [Option<angular_ipred::Fn>; 14],
+    pub intra_pred: [angular_ipred::Fn; 14],
     pub cfl_ac: [cfl_ac::Fn; 3],
     pub cfl_pred: [cfl_pred::Fn; 6],
     pub pal_pred: pal_pred::Fn,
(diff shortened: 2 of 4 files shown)


rinon (approved):

src/decode.rs
```diff
@@ -5045,42 +5045,40 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
         }
     }
 
-    if (*f.dsp).ipred.intra_pred[DC_PRED as usize].is_none() {
```
rinon, line 5048: Does this now re-init on each frame? Is that intended?

kkysen: Hmm, no.  I misread that as the opposite.  Good catch.

kkysen: In that case, it's probably relying on it being zero-initialized somewhere before, which is not great.  Let me see if there's somewhere we can non-lazily initialize it.

rinon: Is `Rav1dIntraPredDSPContext` public? We could just add a boolean `initialized` and check that?

rinon (changes requested):
I think we should do something to not reinit on every frame so we don't deviate from the original code.

src/decode.rs
kkysen: No, it's not `pub`.  That'll probably work.  I just want to check where it's actually getting initialized since it's happening implicitly and I want to make sure we avoid UB.

kkysen: I changed it to use your `initialized: bool` field suggestion.

src/decode.rs
```diff
@@ -5045,8 +5045,12 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
...
+    // TODO(kkysen) Rather than lazy initializing this,
+    // we should probably initialize all the fn ptrs
+    // when `c` is allocated during [`rav1d_open`].
+    if !(*f.dsp).ipred.initialized {
```
rinon, line 5051: Is there a reason this should be in the ipred structure rather than the top-level DSP context? We are initializing all of it afaict?

src/ipred.rs
```diff
@@ -155,12 +156,11 @@ impl pal_pred::Fn {
...
 #[repr(C)]
 pub struct Rav1dIntraPredDSPContext {
-    // TODO(legare): Remove `Option` once `dav1d_submit_frame` is no longer checking
-    // this field with `is_none`.
-    pub intra_pred: [Option<angular_ipred::Fn>; 14],
+    pub intra_pred: [angular_ipred::Fn; 14],
     pub cfl_ac: [cfl_ac::Fn; 3],
     pub cfl_pred: [cfl_pred::Fn; 6],
     pub pal_pred: pal_pred::Fn,
+    pub initialized: bool,
```
rinon, line 163: See other comment, but should this move to the `DSPContext`?

src/decode.rs
kkysen: Hmm, yeah, that's probably better.  I was just confused since the C was checking `ipred`, but that was probably just done randomly.  I'll move it.

src/ipred.rs
kkysen: Changed in [7d0c2ce..4609590](https://github.com/memorysafety/rav1d/compare/7d0c2ce6da7544f7a9f026d1b44478f4ac19b5ff..4609590678b588824c7770f5d05ba378e2914709).

src/decode.rs
kkysen: Changed in [7d0c2ce..4609590](https://github.com/memorysafety/rav1d/compare/7d0c2ce6da7544f7a9f026d1b44478f4ac19b5ff..4609590678b588824c7770f5d05ba378e2914709).

randomPoison (approved):
The changes here look fine to me, but make sure you've addressed @rinon's requests before merging.

kkysen:
> The changes here look fine to me, but make sure you've addressed @rinon's requests before merging.

I addressed the last of @rinon's comments and was just waiting on approval, but it was a pretty minimal fix, so I'll go ahead and merge now.

rinon:
I just hadn't gotten a chance to look at it, LGTM thanks 👍

PR #585, opened 2023-11-21
`struct Rav1dIntraPredDspContext::cfl_ac`: Make an `EnumMap` over `Rav1dPixelLayoutSubSampled`

kkysen:
I'll do the same for the other array fields, but I need to make the `enum`s for those first; this one already exists from before.

### DIFF
--- a/include/dav1d/headers.rs
+++ b/include/dav1d/headers.rs
@@ -210,7 +210,7 @@ pub(crate) enum Rav1dPixelLayout {
     I444,
 }
 
-impl EnumKey<4> for Rav1dPixelLayout {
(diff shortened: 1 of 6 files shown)


kkysen:
> However, what does this actually enable with regards to safety improvements? Was this actually blocking anything or is this a readability improvement? We need to focus on only what is blocking memory safety improvements.

This was to remove the bounds check, since `enum`s are restricted to only their values.  Since we're now focusing on getting everything safe first before ensuring we don't have perf regressions, I'll hold off on these types of changes until we get everything safe.  The improvement in readability/type safety is nice here, but was more of a secondary concern.

randomPoison (approved):

PR #586, opened 2023-11-25
`src/lf_apply.rs`: Deduplicate w/ generics

thedataking:


src/lf_apply.rs
```diff
@@ -1,7 +1,10 @@
+use crate::include::common::bitdepth::BitDepth;
 use crate::include::dav1d::headers::Rav1dPixelLayout;
+
```
kkysen, line 3: Can you remove the newlines between imports?

src/lf_apply.rs
```diff
@@ -13,21 +16,12 @@ use std::ffi::c_int;
...
-unsafe fn PXSTRIDE(x: ptrdiff_t) -> ptrdiff_t {
-    if x & 1 != 0 {
-        unreachable!();
-    }
-    return x >> 1;
-}
-
-unsafe fn backup_lpf(
+// TODO(perl) Temporarily pub until mod is deduplicated
+pub(crate) unsafe fn backup_lpf<BD: BitDepth>(
```
kkysen, line 20: This shouldn't be needed anymore, right?  Since the module is entirely deduplicated now.

src/lf_apply.rs
```diff
@@ -45,40 +39,56 @@ unsafe fn backup_lpf(
...
-                    as *const c_void,
-                (dst_w << 1) as usize,
+                &mut *dst.offset(BD::pxstride(dst_stride as usize * top as usize) as isize)
+                    as *mut BD::Pixel as *const c_void,
+                if BD::BITDEPTH == 8 {
+                    dst_w as usize
+                } else {
+                    (dst_w << 1) as usize
+                },
             );
```
kkysen, line 56: I think this is supposed to be a `pixel_copy`.  The

```rust
                if BD::BITDEPTH == 8 {
                    dst_w as usize
                } else {
                    (dst_w << 1) as usize
                }
```

in particular looks like it's probably supposed to be `pixel_copy`.

src/lf_apply.rs
```diff
@@ -45,40 +39,56 @@ unsafe fn backup_lpf(
...
-                &mut *dst.offset(((PXSTRIDE)(dst_stride) * (top + 1) as isize) as isize)
-                    as *mut pixel as *const c_void,
-                (dst_w << 1) as usize,
+                &mut *dst.offset(BD::pxstride(dst_stride as usize * (top + 1) as usize) as isize)
+                    as *mut BD::Pixel as *const c_void,
+                if BD::BITDEPTH == 8 {
+                    dst_w as usize
+                } else {
+                    (dst_w << 1) as usize
+                },
```
kkysen, line 66: Anywhere where these are I think are supposed to be `pixel_copy`s or `pixel_set`s.

src/lf_apply.rs
thedataking: derp, no, good point

src/lf_apply.rs
thedataking: yup, I missed that there was a `pixel_copy` function although you mentioned it in your instructions  😑

kkysen (changes requested):
The `BD::pixel_copy`s look good now (albeit some trivial simplifications which I noted but we can do them later), but there are still some instances of the other comments I left last time on this and the other deduplication PRs.

src/lf_apply.rs
```diff
@@ -6,28 +7,23 @@ use crate::src::lf_mask::Av1Filter;
...
-    return x >> 1;
-}
+use std::slice;
 
-unsafe fn backup_lpf(
+// The loop filter buffer stores 12 rows of pixels. A superblock block will
+// contain at most 2 stripes. Each stripe requires 4 rows pixels (2 above
+// and 2 below) the final 4 rows are used to swap the bottom of the last
+// stripe with the top of the next super block row.
+pub(crate) unsafe fn backup_lpf<BD: BitDepth>(
```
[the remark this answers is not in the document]
thedataking: Yup, I simply forgot, sorry to waste your time on this.

thedataking:
> there are still some instances of the other comments I left last time on this and the other deduplication PRs.

yeah, sorry about that, you shouldn't have to point out the same stuff twice :/

src/lf_apply.rs
```diff
@@ -6,28 +7,21 @@ use crate::src::lf_mask::Av1Filter;
...
+// contain at most 2 stripes. Each stripe requires 4 rows pixels (2 above
+// and 2 below) the final 4 rows are used to swap the bottom of the last
+// stripe with the top of the next super block row.
+unsafe fn backup_lpf<BD: BitDepth>(
     f: *const Rav1dFrameContext,
-    mut dst: *mut pixel,
+    mut dst: *mut BD::Pixel,
     dst_stride: ptrdiff_t,
-    mut src: *const pixel,
+    mut src: *mut BD::Pixel,
```
kkysen, line 24:
```suggestion
    mut src: *const BD::Pixel,
```

kkysen (changes requested):
Everything else looks great now, but I noticed one `mut` that used to be a `const`.  Once that's fixed it's good to merge.

src/lf_apply.rs
thedataking: oops, good catch!

PR #587, opened 2023-11-26
`src/lr_apply.rs`: Deduplicate w/ generics

thedataking:


kkysen (commented):
Thanks for catching this!  I'm not sure how it still compiled with the different (non pointer) types, though.

src/lr_apply.rs
```diff
@@ -1,6 +1,333 @@
+use crate::include::common::bitdepth::BitDepth;
+use crate::include::dav1d::headers::Rav1dPixelLayout;
+use crate::include::dav1d::headers::RAV1D_RESTORATION_NONE;
+
```
kkysen, line 4: No newlines needed (otherwise `rustfmt` won't sort things).

src/lr_apply.rs
```diff
@@ -1,6 +1,333 @@
...
+use std::ffi::c_int;
 use std::ffi::c_uint;
 
 pub type LrRestorePlanes = c_uint;
 pub const LR_RESTORE_V: LrRestorePlanes = 4;
 pub const LR_RESTORE_U: LrRestorePlanes = 2;
 pub const LR_RESTORE_Y: LrRestorePlanes = 1;
+
+// TODO(perl) Temporarily pub until mod is deduplicated
+pub(crate) unsafe fn lr_stripe<BD: BitDepth>(
```
kkysen, line 34: Like #586, we shouldn't need these temporary `pub(crate)`s anymore since now the module is deduplicated.

src/lr_apply.rs
```diff
@@ -1,6 +1,333 @@
...
+// TODO(perl) Temporarily pub until mod is deduplicated
+pub(crate) unsafe fn backup4xU<BD: BitDepth>(
+    mut dst: *mut [BD::Pixel; 4],
+    mut src: *const BD::Pixel,
+    src_stride: ptrdiff_t,
+    mut u: c_int,
+) {
+    while u > 0 {
+        let n = if BD::BITDEPTH == 8 { 4 } else { 4 << 1 };
+        memcpy(dst as *mut c_void, src as *const c_void, n);
```
kkysen, line 163: Should be a `pixel_copy`.

kkysen (changes requested):
Pretty much the same comments about `pixel_copy`, etc. as I had for #586.

src/lr_apply.rs
thedataking: right, thanks

kkysen (changes requested):

kkysen (approved):

PR #595, opened 2023-12-01
`fn backup_lpf`: Fix incorrect pixel_copy

thedataking:
Incorrect slice length introduced in 73e6e6e breaks Argon tests.


thedataking:
> Need to be careful with from_raw_parts(_mut) lengths.

couldn't agree more.

PR #597, opened 2023-12-02
`const DEBUG_{SEQ_HDR,FRAME_HDR,OBU_METADATA}`: Add back debug `#define`s and their debugging/logging code

kkysen:


src/obu.rs
```diff
@@ -137,6 +137,13 @@ unsafe fn parse_seq_hdr(
         return Err(EINVAL);
     }
 
+    const DEBUG_SEQ_HDR: bool = false;
```
randomPoison, line 140: Is this value meant to be edited directly by a programmer when they're debugging, or will it eventually be hooked up to an env var or a compiler flag?

randomPoison (approved):

src/obu.rs
kkysen: I think the programmer is meant to edit it.  It's a #define DEBUG_SEQ_HDR false` in C.

PR #598, opened 2023-12-02
`mod obu`: Cleanup transpilation-specific control flow

kkysen:
Some `continue`s in loops were not translated, and instead turned into a long `if` block with the opposite condition.  This turns them back into `continue`.

And `switch`es with early `break`s and fall-throughs were turned into `current_block_[0-9]+`-style state machines that are very confusing to read.  This keeps them as state machines, but uses this pattern:

```rust
let mut state = scrutinee;
loop {
    match state {
        A => {
            ...
            state = B;
            continue; // fall-through
        }
        B => {
            if early_break_condition {
                break;
            }
        }
    }
    break;
}
```

`break`s at the end of `switch` cases turn into nothing like a normal `match` arm.  Early `break`s turn into `break`s out of the surrounding `loop`.  And fall-throughs turn into switching to the following state and `continue`ing.  This is a lot more readable than the `current_block_[0-9]+`-style state machines and closely matches how C writes it.

There is one early `break` that I left as a long `else` block as I think it's easier to read that way without adding the surrounding `loop`.

### DIFF
--- a/src/obu.rs
+++ b/src/obu.rs
@@ -1435,48 +1435,48 @@ unsafe fn parse_frame_hdr(c: &mut Rav1dContext, gb: &mut GetBits) -> Rav1dResult
             } else {
                 RAV1D_WM_TYPE_AFFINE
             };
-            if !(hdr.gmv[i as usize].r#type == RAV1D_WM_TYPE_IDENTITY) {
-                let ref_gmv;
-                if hdr.primary_ref_frame == RAV1D_PRIMARY_REF_NONE {
-                    ref_gmv = &dav1d_default_wm_params;
-                } else {
-                    let pri_ref = hdr.refidx[hdr.primary_ref_frame as usize];
-                    if (c.refs[pri_ref as usize].p.p.frame_hdr).is_null() {
-                        return error(c);
-                    }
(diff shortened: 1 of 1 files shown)


kkysen:
Okay, I'll get rid of the state machine how you suggested.  Thanks for the patch.  Are you sure this isn't a hot code path, though?  I assumed it definitely could be in certain cases

kkysen:
@thedataking, I put the state machine removal in a separate PR, #600.  Also I applied your patch, but split up the different fall-through removals into their own commits, and then also removed the early `break`s so that we could remove the entire `loop` state machine thingy.

thedataking:
> Are you sure this isn't a hot code path, though? I assumed it definitely could be in certain cases

Yup, I double checked with perf and `rav1d_parse_obus` is among the many functions that account for ~0% of the running time in both 8 and 16 bits modes. 

> I put the state machine removal in a separate PR, https://github.com/memorysafety/rav1d/pull/600. Also I applied your patch, but split up the different fall-through removals into their own commits

Teriffic, let's get this PR merged then 🚀

thedataking (approved):

kkysen:
> Yup, I double checked with perf and `rav1d_parse_obus` is among the many functions that account for ~0% of the running time in both 8 and 16 bits modes.

That's great to know!

randomPoison (approved):

PR #600, opened 2023-12-04
`fn rav1d_parse_obus`: Remove state machine

kkysen:
Starting from @thedataking's suggestion (https://github.com/memorysafety/rav1d/pull/598#issuecomment-1837420285), this first removes the `state` var from the state machine `loop`.  I split up the two fall-through `continue`s in separate commits, and for the `RAV1D_OBU_TILE_GRP` one, I named the inner `fn parse_tile_grp`, since most of the other `fn`s in this `mod` are named `parse_*`.  I also fixed the `gb` references, as we don't need to `&mut` borrow it again in the inner `fn`.

Then I also removed all of the early `break`s, allowing us to fully remove the state machine now, so it's just a plain `match` now if a few `match` guards, a few simple `if`s, and an outlined inner `fn`.

### DIFF
--- a/src/obu.rs
+++ b/src/obu.rs
@@ -1843,368 +1843,362 @@ pub(crate) unsafe fn rav1d_parse_obus(
         Ok(())
     }
 
-    loop {
-        match r#type {
-            RAV1D_OBU_SEQ_HDR => {
-                let mut r#ref = rav1d_ref_create_using_pool(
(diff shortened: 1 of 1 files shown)


thedataking (approved):
Terrific! I think  you strike a nice balance between using Rust's match feature while staying close enough to the C code that any upstream changes will be straightforward to integrate.

PR #601, opened 2023-12-05
`mod include::dav1d::headers`: Deduplicate Section 3 `const`s

kkysen:


thedataking:
Do we need two sets of constants ? Couldn't we just use the `DAVID_` ones or is there some further transformation planned for the `RAVID_` ones?

kkysen:
> Do we need two sets of constants ? Couldn't we just use the `DAVID_` ones or is there some further transformation planned for the `RAVID_` ones?

Yeah, it's for future-proofing.  In the future, the `RAV1D_*` ones may become autogenerated from `enum`s (like a `COUNT` associated `const`), or other things, and since everything else has separate `DAV1D_*` and `RAV1D_*` versions, I want to keep things like this so it's easy to make changes in the future.

PR #611, opened 2023-12-07
`mod obu`: Cleanup error handling

kkysen:


rinon (approved):

src/obu.rs
```diff
@@ -2553,3 +2535,17 @@ pub(crate) unsafe fn rav1d_parse_obus(
...
+    global: c_int,
+) -> Rav1dResult<c_uint> {
+    parse_obus(c, r#in, global).inspect_err(|_| {
+        rav1d_data_props_copy(&mut c.cached_error_props, &mut r#in.m);
+        rav1d_log(
+            c,
+            b"Error parsing OBU data\n\0" as *const u8 as *const c_char,
+        );
+    })
+}
```
rinon, line 2551: nitpick for future - If you put this directly before `parse_obus` the diff will be easier to read and review. Doesn't make much of a difference here because the logic is simple, but for comparing more complex code sequences it's nice to not have to scroll to line things up.

kkysen: Hmm, yeah, but this order makes more logical sense with things higher in the call stack going lower in the file.  I'll keep the diff in mind, though.

PR #613, opened 2023-12-08
Revert "Add custom profile for debug builds with basic optimizations"

kkysen:
Revert #605, as it breaks `cargo build`.  We should fix it, but I'd rather not spend time on it now.

### DIFF
--- a/.github/workflows/build-and-test-aarch64.yml
+++ b/.github/workflows/build-and-test-aarch64.yml
@@ -50,8 +50,8 @@ jobs:
             --platform linux/arm64 \
             ghcr.io/immunant/rav1d/debian-bullseye-aarch64:latest \
             .github/workflows/test.sh -t 2 \
-                -r target/aarch64-unknown-linux-gnu/release/dav1d \
-                -s target/aarch64-unknown-linux-gnu/release/seek_stress
+                -r ../target/aarch64-unknown-linux-gnu/release/dav1d \
+                -s ../target/aarch64-unknown-linux-gnu/release/seek_stress
       - name: upload build artifacts
         if: ${{ !cancelled() }}
         uses: actions/upload-artifact@v3
--- a/.github/workflows/build-and-test-arm7.yml
+++ b/.github/workflows/build-and-test-arm7.yml
@@ -50,8 +50,8 @@ jobs:
             --platform linux/arm/v7 \
             ghcr.io/immunant/rav1d/debian-bullseye-arm7:latest \
             .github/workflows/test.sh -t 2 \
-                -r target/armv7-unknown-linux-gnueabihf/release/dav1d \
-                -s target/armv7-unknown-linux-gnueabihf/release/seek_stress
+                -r ../target/armv7-unknown-linux-gnueabihf/release/dav1d \
+                -s ../target/armv7-unknown-linux-gnueabihf/release/seek_stress
       - name: upload build artifacts
         if: ${{ !cancelled() }}
         uses: actions/upload-artifact@v3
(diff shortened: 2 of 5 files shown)


thedataking:
A program should always be able to build with and without optimization. It appears that #605 surfaced an existing bug. 

As far as I can tell, `cargo build` fails (as you showed here https://github.com/memorysafety/rav1d/pull/605#issuecomment-1846509138) because the `bd_fn!` macro generates references to assembly routines that simply do not exist. For instance, there is no assembly routine called `dav1d_ipred_dc_16bpc_avx512icl` but there is one called `dav1d_ipred_dc_8bpc_avx512icl`. 

We can verify that there are only 8bpc versions by looking at `ipred16avx512.asm` or like this:

```sh
$ nm target/release/dav1d | grep -E  "ipred_dc_(8|16)bpc_avx512icl$"
00000000001d91a0 T dav1d_ipred_dc_8bpc_avx512icl
```

We can also look at `intra_pred_dsp_init_x86`:

```C
#if BITDEPTH == 8
    init_angular_ipred_fn(DC_PRED,       ipred_dc,       avx512icl);
    init_angular_ipred_fn(DC_128_PRED,   ipred_dc_128,   avx512icl);
    init_angular_ipred_fn(TOP_DC_PRED,   ipred_dc_top,   avx512icl);
    init_angular_ipred_fn(LEFT_DC_PRED,  ipred_dc_left,  avx512icl);
    init_angular_ipred_fn(HOR_PRED,      ipred_h,        avx512icl);
    init_angular_ipred_fn(VERT_PRED,     ipred_v,        avx512icl);
#endif
```


The code you added in `fn intra_pred_dsp_init_x86` doesn't look correct to me. The `BPC::BPC8` version will declare functions that don't exist. If that code doesn't get dead code eliminated, the linker is right to complain about references to non-existing functions.

```
        if BD::BPC == BPC::BPC8 {
            c.intra_pred[DC_PRED as usize] =
                bd_fn!(angular_ipred::decl_fn, BD, ipred_dc, avx512icl);
            c.intra_pred[DC_128_PRED as usize] =
                bd_fn!(angular_ipred::decl_fn, BD, ipred_dc_128, avx512icl);
            c.intra_pred[TOP_DC_PRED as usize] =
                bd_fn!(angular_ipred::decl_fn, BD, ipred_dc_top, avx512icl);
            c.intra_pred[LEFT_DC_PRED as usize] =
                bd_fn!(angular_ipred::decl_fn, BD, ipred_dc_left, avx512icl);
            c.intra_pred[HOR_PRED as usize] =
                bd_fn!(angular_ipred::decl_fn, BD, ipred_h, avx512icl);
            c.intra_pred[VERT_PRED as usize] =
                bd_fn!(angular_ipred::decl_fn, BD, ipred_v, avx512icl);
        }
```

As a hack, I added an 8bit only version of `bd_fn` which only declares the 8bpc variants and used that in the above code fragment - then everything builds again at all optimization levels.

```Rust
#[cfg(feature = "asm")]
macro_rules! bd8_fn {
    ($decl_fn:path, $BD:ty, $name:ident, $asm:ident) => {{
        use paste::paste;

        paste! { $decl_fn!(fn [<dav1d_ $name _8bpc_ $asm>]) }
    }};

    ($BD:ty, $name:ident, $asm:ident) => {{
        use $crate::include::common::bitdepth::fn_identity;

        bd8_fn!(fn_identity, $BD, $name, $asm)
    }};
}
```

I hope this makes it clear that PR #605 is not the problem.

kkysen:
I know #605 is not the problem but it's annoying to fix until all the bitdepth deduplication PRs are merged, and I didn't know if you wanted me to spend time fixing it vs. working on safety.  I prototyped a way to fix it, though, which I can make into a PR once the bitdepth deduplication PRs are all done and merged.  It's also more than just those ipred functions, as if you compile only one bitdepth you get tons of these errors everywhere.

PR #615, opened 2023-12-08
`struct Rav1dTileGroupHeader`: Split from `struct Rav1dTileGroup` and return directly from `fn parse_tile_hdr`

kkysen:


rinon (commented):
I'd prefer we returned a tuple from parse_tile_hdr or returned a named struct but then left the start and end fields as is in `Rav1dTileGroup`. Unless we need to do this for borrowing reasons, I'd prefer to stick as closely to the original structure as we can. It's not a big change and not a big deal in this case, but generally I think we should keep things as simple as we can.

kkysen:
> I'd prefer we returned a tuple from parse_tile_hdr or returned a named struct but then left the start and end fields as is in `Rav1dTileGroup`. Unless we need to do this for borrowing reasons, I'd prefer to stick as closely to the original structure as we can. It's not a big change and not a big deal in this case, but generally I think we should keep things as simple as we can.

This is still keeping the original layout, though.  We're already necessarily changing field paths from `dav1d` since these types use a lot of anonymous `struct`s, and this seems pretty similar.  Returning a short-lived type just to assign all of its fields to a different type could work for this one since it's only two fields, but you'll see in some upcoming PRs that there are much larger types like this, and doing that for those is very unwieldy.

rinon:
I wasn't concerned about memory layout, this was more about usability. We should aim for making minimal repetitive changes, because it's not doing anything essential and causes more churn. I'm sure for some cases coming up it will be different but this is just the two fields. Putting the start and end index next to the data instead of in a substructure is the simpler way here.

kkysen:
I'll remove `struct Rav1dTileGroupHeader` and return `RangeInclusive<c_int>` from `fn parse_tile_hdr` instead (so we can do `return start..=end`, for example), but I'd like to merge this as is and do that in a PR after all of the current ones just because there are a ton of merge conflicts if I change this.

rinon (approved):
Ok, this is technically fine, just was more of a general future direction thing. Sounds good.

PR #616, opened 2023-12-08
`struct Rav1d{,Frame}Context::tiles`: Combine `tile`, `n_tile_data_alloc`, `n_tile_data` fields into a `Vec`

kkysen:


src/decode.rs
```diff
@@ -5161,26 +5159,8 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
...
-    }
-    let num_tiles = c.n_tile_data.try_into().unwrap();
-    let f_tiles = slice::from_raw_parts_mut(f.tile, num_tiles);
-    let c_tiles = slice::from_raw_parts_mut(c.tile, num_tiles);
-    f_tiles.clone_from_slice(c_tiles);
-    c_tiles.fill_with(Default::default);
-    f.n_tile_data = c.n_tile_data;
-    c.n_tile_data = 0;
+    f.tiles.clear();
+    mem::swap(&mut f.tiles, &mut c.tiles);
```
kkysen, line 5163: @fbossen, do you know why the C needed to copy from `c` to `f` rather than just swapping the two as I do here?  It seems like it would be semantically equivalent except for slight differences in the capacities, but I don't think that should matter since extra capacity is always allocated elsewhere if it's needed.

src/lib.rs
```diff
@@ -1150,7 +1150,7 @@ unsafe fn close_internal(c_out: &mut *mut Rav1dContext, flush: c_int) {
         rav1d_free_aligned((*f).ts as *mut c_void);
         rav1d_free_aligned((*f).ipred_edge[0] as *mut c_void);
         free((*f).a as *mut c_void);
-        free((*f).tile as *mut c_void);
+        (*f).tiles = Default::default();
```
rinon, line 1153: We should remove this when context is owned and dropped properly, let's mark this with a TODO.

src/lib.rs
```diff
@@ -1178,12 +1178,10 @@ unsafe fn close_internal(c_out: &mut *mut Rav1dContext, flush: c_int) {
...
     }
-    let mut n_3 = 0;
-    while n_3 < (*c).n_tile_data {
-        rav1d_data_unref_internal(&mut (*((*c).tile).offset(n_3 as isize)).data);
-        n_3 += 1;
+    for mut tile in (*c).tiles.drain(..) {
+        rav1d_data_unref_internal(&mut tile.data);
     }
-    free((*c).tile as *mut c_void);
+    (*c).tiles = Default::default();
```
rinon, line 1184: Same as above, we should remove when context is owned.

src/obu.rs
```diff
@@ -1930,63 +1928,43 @@ unsafe fn parse_obus(
...
         }
+
+        if let Err(_) = c.tiles.try_reserve_exact(1) {
+            return Err(EINVAL);
+        }
+        c.tiles.push(Rav1dTileGroup {
+            data: Default::default(),
+            hdr,
+        });
+        // TODO(kkysen) An idiom for push and getting a &mut would be nice and more efficient.
```
rinon, line 1946: Why not move the push to after the initialization? I don't think we need a new api for this.

src/obu.rs
```diff
@@ -1930,63 +1928,43 @@ unsafe fn parse_obus(
...
+            for mut tile in c.tiles.drain(..) {
+                rav1d_data_unref_internal(&mut tile.data);
             }
-            c.n_tile_data = 0;
             c.n_tiles = 0;
             return Err(EINVAL);
         }
-        c.n_tiles += 1 + (*(c.tile).offset(c.n_tile_data as isize)).hdr.end
-            - (*(c.tile).offset(c.n_tile_data as isize)).hdr.start;
-        c.n_tile_data += 1;
```
rinon, line 1989: Here is where the "push" occurs, let's just move the vec push down to here?

src/obu.rs
kkysen: I thought I did this for a reason; I think I thought `rav1d_data_ref` had some internal pointer or something, but I think I was wrong, so I should be able to do it this way.  It's definitely a lot cleaner.

src/decode.rs
```diff
@@ -4886,7 +4885,7 @@ pub(crate) unsafe fn rav1d_decode_frame_exit(f: &mut Rav1dFrameContext, retval:
     rav1d_ref_dec(&mut f.mvs_ref);
     rav1d_ref_dec(&mut f.seq_hdr_ref);
     rav1d_ref_dec(&mut f.frame_hdr_ref);
-    for tile in slice::from_raw_parts_mut(f.tile, f.n_tile_data.try_into().unwrap()) {
+    for tile in &mut f.tiles {
         rav1d_data_unref_internal(&mut tile.data);
```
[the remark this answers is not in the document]
kkysen: Nope!  Should we move the `.clear()` into this `fn`?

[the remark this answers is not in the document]
kkysen: I'll fix this in a later PR.

src/lib.rs
kkysen: Done in 03c4d28.

src/lib.rs
kkysen: Done in 03c4d28.

src/obu.rs
kkysen: Fixed in 349b43c.

src/obu.rs
kkysen: Fixed in 349b43c.

rinon (approved):

PR #619, opened 2023-12-10
`fn rav1d_log`: Replace with safe `write!` calls to a fully safe `Rav1dLogger`

kkysen:
The existing variadic and `unsafe` `rav1d_log` calls are replaced with `write!` calls to an `Option<Rav1dLogger>` field.  `enum Rav1dLogger`'s interface is `fn write_fmt`, which `write!` calls.  These are defined as part of `trait Rav1dLog`, similar to `trait {io,fmt}::Write`.  The difference is that no errors are returned (we don't really need them) and crucially, they are `&self` methods as opposed to `&mut self` methods, which keeps things much simpler for the borrow checker.

`enum Rav1dLogger` can be `Dav1d` or `Std{out,err}` for now.  The `Dav1d` variant is necessary to convert to/from the `Dav1dLogger`, and `Stderr` is the default logger used (and `Stdout` is trivially similar).  There will presumably be other loggers desired, but it's easier to leave those for later once we know their desired behaviors better.

The `Dav1d` variant is called by implementing `fmt::Write`, which lets us just implement `fn write_str` and have `fn write_fmt` auto-implemented.  `fn write_str` is implemented correctly but with poor performance by printing byte by byte.  This is to avoid dealing with intermediary null characters and having to allocate a new null-terminated `CString`.  This is for logging so perf isn't important, and if better perf is desired, users can switch to the Rust API, which has no such limitation.

Then we construct a bijection between `{D,R}av1dLogger` to allow the bidirectional conversions needed in the `DAV1D_API`s.  This is done by storing the `Std{out,err}` variants as special `Dav1dLogger::callback` `fn`s.  During the conversion back, if the callback is one of those special `fn`s, we can decode it back into the same `Rav1dLogger` it originally was.  With use of the `Dav1dLogger::cookie` field, this can be extended to other `Rav1dLogger` variants in the future as well.

This also makes it fairly easy to eliminate `#![feature(c_variadics)]` (it's only used for the special marker `fn`s now, but those can be implemented a bit differently).


CrazyboyQCD:
As https://github.com/rust-lang/rust/pull/117472 was merged, I think cstr crate could be remove sometime.

CrazyboyQCD:
Since https://github.com/rust-lang/rust/pull/117472 has been merged, I think cstr crate could be removed sometime.

kkysen:
> Since rust-lang/rust#117472 has been merged, I think cstr crate could be removed sometime.

Oh, that's awesome!  I saw the RFC got accepted earlier but I didn't realize it would be stabilized so soon.  We'll remove `cstr` once 1.76 is released.

src/log.rs
```diff
@@ -1,47 +1,152 @@
...
+    fn from(logger: Dav1dLogger) -> Self {
+        let Dav1dLogger { cookie, callback } = logger;
+        let callback = callback?;
+        Some(if callback == rav1d_logger_stdout {
+            Rav1dLogger::Stdout
+        } else if callback == rav1d_logger_stderr {
+            Rav1dLogger::Stderr
+        } else {
+            Rav1dLogger::Dav1d(OnlyDav1dLogger { cookie, callback })
+        })
```
rinon, line 135: Match?

src/log.rs
```diff
@@ -1,47 +1,152 @@
...
+    Dav1d(OnlyDav1dLogger),
+    Stdout,
+    #[default]
+    Stderr,
+}
+
+/// Any type implementing [`Rav1dLog`] can be used with [`write!`].
+pub trait Rav1dLog {
+    fn write_fmt(&self, args: fmt::Arguments);
+}
```
rinon, line 89: Why are we making a new Rav1dLog trait instead of implementing fmt::Write for Rav1dLogger? After some inspection, I assume it's so that we can have a immutable borrow `write_fmt` method for `Option<Rav1dLogger>`? That should be documented.

src/log.rs
```diff
@@ -1,47 +1,152 @@
...
+
+impl Rav1dLog for Option<Rav1dLogger> {
+    /// When a logger isn't set, we don't want to have to run any code here,
+    /// so force this to be inlined so a [`None`] can be seen.
+    #[inline(always)]
+    fn write_fmt(&self, args: fmt::Arguments) {
+        if let Some(logger) = self {
+            logger.write_fmt(args);
+        }
+    }
```
rinon, line 116: An alternative I think we should consider: We could replace the `rav1d_log` function with a `rav1d_log` macro that takes a `Rav1dContext` argument and includes the `if let Some(logger) = c.logger`. That way we don't need this extra trait or most of this infrastructure, as the logger can just be something that implements `fmt::Write` then.

src/log.rs
```diff
@@ -1,47 +1,152 @@
...
-pub unsafe extern "C" fn rav1d_log(c: *mut Rav1dContext, format: *const c_char, args: ...) {
-    if validate_input!(!c.is_null()).is_err() {
-        return;
+#[derive(Clone, Default)]
+pub(crate) enum Rav1dLogger {
+    Dav1d(OnlyDav1dLogger),
+    Stdout,
+    #[default]
+    Stderr,
+}
```
rinon, line 84: What do Rust clients who want a customized logger do here? Ideally they could just provide some `dyn fmt::Write` sink as a config. Would it be simpler and more flexible to just use a trait object here with inner mutability and call it good?

src/log.rs
```diff
@@ -1,47 +1,152 @@
...
+
+impl fmt::Write for OnlyDav1dLogger {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        // `s` doesn't have a terminating nul-byte,
+        // and it may have internal nul-bytes,
+        // so it's easiest just to print one byte at a time.
+        // This may be slow, but logging can be disabled if it's slow,
+        // or the Rust API can be used instead.
+        let fmt = cstr!("%c");
```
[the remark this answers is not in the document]
kkysen: Native `c"strings"` were just stabilized in nightly so we'll switch to that once it lands in a couple months.  I think the crate is fine in the meantime as it has an identical API.

src/log.rs
kkysen: You can't match on function pointers.

src/log.rs
kkysen: Yeah, that's why, the `&self`.  I'll document that.

src/log.rs
kkysen: > There will presumably be other loggers desired, but it's easier to leave those for later once we know their desired behaviors better.

Yeah, I was thinking that, but it's a bit tricky because `dyn fmt::Write` needs to be boxed, and for stdout and stderr, the common case, we don't want to impose a `Box`, and those implement `io::Write`, which is `write!`able but a bit different, so it's tricky.  I thought we can save those until later when we know better what exact use case is wanted.

src/log.rs
kkysen: What advantage does that have?  That just seems more complex.  We can't implement `fmt::Write` due to the `&mut self` so we need our own function regardless, plus this way we don't return a `Result` that needs to be unwrapped and that bloats the code a lot for something that might never be run.  So at that point it's just a function vs. a macro and a function seems simpler.

src/log.rs
```diff
@@ -1,47 +1,152 @@
...
-            callback: Some(rav1d_log_default_callback),
+        // Safety: `callback` is [`None`].
+        unsafe { Self::new(ptr::null_mut(), None) }
+    }
+}
+
+/// A [`Dav1dLogger`] from C that's not
+/// just a [`Rav1dLogger`] converted to a [`Dav1dLogger`].
+#[derive(Clone)]
+pub(crate) struct OnlyDav1dLogger {
```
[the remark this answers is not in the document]
kkysen: `Dav1dLogger` has an `Option` for the callback to match the C ABI, but for this the outer `Option` around `Rav1dLogger` takes care of the `None` so we don't have to do the check again.  Also, it means I can decide the special marker `fn` pointers separaty and not have the method get called infinitely recursively.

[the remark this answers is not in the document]
rinon: Not sure I understand, or maybe I wasn't clear what I was asking. Why don't you just stick the Dav1dLogger in the `Rav1dLogger::Dav1d` variant? At that point you know you've checked for the markers. Having to unwrap the Option isn't enough reason to make a second almost identical structure here.

src/log.rs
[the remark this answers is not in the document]
rinon: We can switch to `c""` strings only once we know that potential downstream users are on a recent enough Rust version. I'm not sure I understand what you mean by an identical API, the new literal syntax doesn't use a macro.

That all said, that has no bearing on whether we should add a new dependency just for a single line of convenience. A single line is not enough to justify a new dependency, regardles of how short-lived you expect that to be. Once you add something it often stays longer than you'd think. If the crate was providing something we couldn't have otherwise, then fine, but it's not here. Don't use it.

src/log.rs
rinon: Yeah, fair enough. I was debating the same thing when I was looking at this.

src/log.rs
rinon: We don't need the trait at all, we could just have a method on Rav1dLogger that takes a `fmt::Argmuents` that is called inside the macro after using `format_args!` to construct the arguments. That's exactly what `writeln!` is doing, this would just be a specialized version of `writeln!`.

Thinking more on this, a macro would give us a bit more flexibility in the future if we want to change the structure of logging. I'm not set on it, but it seems more straightforward than the trait to me. This is also one place where I'm a bit concerned about code size and performance. We shouldn't spend time optimizing this right now though, but if we do decide we need to change things in the future, a macro gives us the flexibility to do so.

src/log.rs
rinon: Meant something like this. Minor thing, just a style nitpick, but that if chain was a bit less clear than it could be.

```rust
match callback {
    Some(f) if f == rav1d_logger_stdout => Rav1dLogger::Stdout,
    Some(f) ...
    Some(f) => Rav1dLogger::Dav1d(...),
    None => None,
}
```

src/log.rs
```diff
@@ -1,47 +1,152 @@
...
 
-impl Default for Rav1dLogger {
+impl Default for Dav1dLogger {
     fn default() -> Self {
-        Self {
-            cookie: ptr::null_mut(),
-            callback: Some(rav1d_log_default_callback),
+        // Safety: `callback` is [`None`].
+        unsafe { Self::new(ptr::null_mut(), None) }
+    }
```
rinon, line 45: This changes the default logging functionality to not log, rather than log to stderr?

src/log.rs
```diff
@@ -1,47 +1,152 @@
...
+}
+
+/// Used as a marker for [`Rav1dLogger::Stdout`].  Still a valid (i.e. safe) [`Dav1dLoggerCallback`], though.
+unsafe extern "C" fn rav1d_logger_stdout(_cookie: *mut c_void, _fmt: *const c_char, ...) {}
+
+/// Used as a marker for [`Rav1dLogger::Stderr`].  Still a valid (i.e. safe) [`Dav1dLoggerCallback`], though.
+unsafe extern "C" fn rav1d_logger_stderr(_cookie: *mut c_void, _fmt: *const c_char, ...) {}
+
+impl From<Dav1dLogger> for Option<Rav1dLogger> {
```
rinon, line 125: Thinking a bit more on this in the context of the next PR, do we even need the ability to roundtrip between Rav1dLogger and Dav1dLogger? If we're using the library with the legacy C API we have no way to set up stdout or stderr logging without providing a custom callback. C users can only provide a logger in the configuration struct. Can we translate once into the Rust version and never go back?

src/log.rs
kkysen: Oh sure I can do that.  People have said they don't like that in the past but it does look clean.

src/log.rs
kkysen: Hmm, I may have messed up here.  Let me look at where this default is used and I'll make sure it stays stderr.

src/log.rs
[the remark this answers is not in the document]
kkysen: Isn't `cstr!` the canonical way to write C string literals (before the native way is stabilized)?  I'm just trying to stick to the easiest, simplest way.  I don't think we should worry about these little things but focus more of making things safe fast.  Choices like this are irrelevant until we do and we can revisit things like this later.  I'm fine changing this to `CStr::from_bytes_with_nul(b"%c\0")`, but I don't think we should care either way until everything's safe.

src/log.rs
rinon: Yeah, this is definitely just a style nitpick, I think it would be clean in this case but it's up to you.

src/log.rs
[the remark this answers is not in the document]
rinon: I only care here because it adds a new dependency. That's enough of a reason to not use it unless we plan to make _lots_ of these C literals, and I don't think we need to?

src/log.rs
```diff
@@ -1,47 +1,152 @@
...
 }
 
-impl Default for Rav1dLogger {
+impl Default for Dav1dLogger {
     fn default() -> Self {
-        Self {
-            cookie: ptr::null_mut(),
-            callback: Some(rav1d_log_default_callback),
+        // Safety: `callback` is [`None`].
+        unsafe { Self::new(ptr::null_mut(), None) }
```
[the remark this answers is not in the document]
kkysen: I did it this way to be clear about maintaining the invariant on the `callback` field.  It's meant for `fn new` to be the way to construct it, as it's `unsafe` and clear about the invariants required for it to be safe.  Obviously it's safe to just construct it directly; I just wanted to be clear about safety and the invariants.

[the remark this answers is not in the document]
rinon: Yeah, that's fair. I don't have a strong opinion on this, calling `new` here is fine.

src/log.rs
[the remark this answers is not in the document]
kkysen: > and I don't think we need to?

Probably not?  I didn't look enough to see for sure.

I'll replace this with `CStr::from_bytes_with_nul(b"%c\0").unwrap()` and then `c"%c"` when its stabilization lands on stable.  As for not updating until we know users have a recent enough rust version, that's pretty hard, as we're already on a recent `nightly` and aren't paying any attention to any MSRV things.

src/log.rs
kkysen: Actually, I'll remove this `impl Default` entirely.  It's only used by the binaries and they can stay is )`Dav1dLogger { cookie: 0 as *mut c_void, callback: None }`).

src/log.rs
kkysen: We need to go from `Dav1dLogger` to `Rav1dLogger` when reading a `Dav1dLogger` in `Dav1dSettings` set by the user.  And we need to go from `Rav1dLogger` to `Dav1dLogger` in `fn dav1d_default_settings`, a `DAV1D_API`.  The default is the `stderr` printing.  It's probably not meant to be read, just defaults set and then certain ones overridden, but it's technically part of the ABI.

rinon: Why do we need to go from a Rav1dLogger to a Dav1dLogger for `dav1d_default_settings`? We could just keep the default log callback (or something similar) and completely separate the two. If you build a logger from C, you get the C callback that prints to stderr, if you build one from Rust you don't get any of the C cruft.

kkysen: We could probably do that, but why?  Right now there's a clear way the `DAV1D_API` is handled everywhere: always convert to `Rav1d*` types and back.  Making exceptions to this pattern increases the cognitive burden of reasoning about any of the `DAV1D_API`s a lot, at least for me.  And since I found a solution here anyways, even if it does some unconventional stuff with the marker `fn` ptrs, what's the point?

src/log.rs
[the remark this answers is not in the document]
kkysen: Removed in 881dafe anyways.

src/log.rs
kkysen: Removed in 881dafe.

src/log.rs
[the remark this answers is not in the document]
kkysen: Changed in 0841e50.  No more `OnlyDav1dLogger`, but there's an extra `.callback().unwrap()` that wouldn't otherwise be needed.

src/log.rs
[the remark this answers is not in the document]
kkysen: Changed in 9d47278.

src/log.rs
kkysen: Added in 51bc341.

src/log.rs
kkysen: Changed in b368117.

src/log.rs
kkysen: I don't think it's worth it to reimplement `write!`, `writeln!`, etc.  Remember we'd have to implement the base and `*ln!` variants, and I think it's a lot clearer if we stick to the standard macros for this rather than inventing our own.  A trait isn't very complex and it comes with documentation on why the signature is why it is, which is also useful.

src/log.rs
rinon: We discussed this on a call, makes sense for now, even if we don't stick with this forever.

rinon (approved):

Issue #592, opened 2023-11-29
Switch to stable Rust toolchain


rinon:
Ideally rav1d should build with a stable toolchain at some point. It looks like we ended up with a few nightly-only features from the initial c2rust transpile that we should eventually remove. I currently find three nightly-only features that we use:
* [ ] `core_intrinsics`: 
* [ ] `c_variadic`: #620
* [x] `extern_types`: #593

The intrinsics we can take care of by using stabilized Rust atomic types, the variadic usage appears to just be `vfprintf` which would be easy to remove, and for extern types we can probably do opaque pointers in another way.

kkysen:
The extern types might be all gone now after we deduplicated everything.  At least the `rav1d` library; the few binaries still have a bunch that are hard to unravel due to type erasure generics.

kkysen:
I've been meaning to fix all the atomic intrinsics to use atomic types soon; that'll be part of making the core data types safe.

kkysen:
For the variadics, I'm not sure we can fully get rid of those because there are callbacks the user can set and we need to keep that ABI.  Eventually we can put it behind a legacy C API feature flag, as we shouldn't need it for pure Rust usage.

rinon:
> The extern types might be all gone now after we deduplicated everything. At least the rav1d library; the few binaries still have a bunch that are hard to unravel due to type erasure generics.

Could we do the same thing as bindgen and make the type unconstructable? We don't care what it's purported layout is, only that we can't construct it.

> I've been meaning to fix all the atomic intrinsics to use atomic types soon; that'll be part of making the core data types safe.

Do we have an issue on this? I was just looking at that, will simplify some things nicely.

> For the variadics, I'm not sure we can fully get rid of those because there are callbacks the user can set and we need to keep that ABI. Eventually we can put it behind a legacy C API feature flag, as we shouldn't need it for pure Rust usage.

Eek ok, yeah. This whole issue is super low priority, I just wanted to create an issue on it while I was thinking about it.

kkysen:
> Could we do the same thing as bindgen and make the type unconstructable? We don't care what it's purported layout is, only that we can't construct it.

I'm not sure, but the binaries I don't think are very important compared to the `rav1d` library that others will use.  And there are Rust libraries we can use for the demuxing in those binaries instead if we want to make them safe, sidestepping the issue.  But I don't think people depend on these binaries (except our tests I guess), so it's a very low priority.

[...]

PR #620, opened 2023-12-10
`#![feature(c_variadic)]`: Remove for the `rav1d` library

kkysen:
Remove by replacing the remaining use with a `mem::transmute` from a non-variadic `fn` to a variadic `fn`, as we don't actually use the variadic args, we just need its type.

* Fixes part of #592.

### DIFF
--- a/src/log.rs
+++ b/src/log.rs
@@ -125,22 +125,27 @@ mod marker {
         unsafe { std::mem::transmute(callback) }
     }
 
-    /// Create an empty [`Dav1dLoggerCallback`] for use as a marker `fn`
-    /// for special `fn`s stored in [`Dav1dLogger::callback`].
-    macro_rules! create {
-        ($name:ident) => {{
-            extern "C" fn $name(_cookie: *mut c_void, _fmt: *const c_char) {
-                // The `fn` needs a unique body so that
-                // multiple ones don't get optimized into the same `fn`.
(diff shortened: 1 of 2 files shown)


src/log.rs
```diff
@@ -116,19 +116,42 @@ impl Rav1dLog for Option<Rav1dLogger> {
...
+        ($name:ident) => {{
+            extern "C" fn $name(_cookie: *mut c_void, _fmt: *const c_char) {
+                // The `fn` needs a unique body so that
+                // multiple ones don't get optimized into the same `fn`.
+                unimplemented!(stringify!($name));
+            }
+
+            cast($name)
+        }};
+    }
```
rinon, line 142: Don't add a macro for two uses, just duplicate the code. Same with the cast function, just write this all out explicitly. Macros are only good when they save ongoing effort and are clear, not for a one-off use like this.

rinon (changes requested):

src/log.rs
kkysen: Okay, I can expand it, though keep in mind that if we add new variants to `enum Rav1dLogger` (which we anticipate to write to a `dyn io::Write` or similar once we know what common use cases are), we'll need to create more of these.

kkysen: Expanded in c310b04.

rinon (approved):

PR #621, opened 2023-12-10
`fn parse_frame_size`: Initialize and return `Rav1dFrameSize` directly

kkysen:
`struct Rav1dFrameSize` was extracted from the monolithic `struct Rav1dFrameHeader` to do this.  The `fn` was also renamed from `fn read_frame_size` to match the names and signatures of the other `fn parse_*`s in `mod obu`.

I'd like to double-check perf on this, as due to extracting out `struct Rav1dFrameSize`, the fields were re-arranged, and I just want to make sure that didn't hurt cache perf.

### DIFF
--- a/include/dav1d/headers.rs
+++ b/include/dav1d/headers.rs
@@ -2124,11 +2124,21 @@ pub struct Dav1dFrameHeader {
 
 #[derive(Clone)]
 #[repr(C)]
-pub(crate) struct Rav1dFrameHeader {
-    pub film_grain: Rav1dFrameHeader_film_grain,
-    pub frame_type: Rav1dFrameType,
+pub(crate) struct Rav1dFrameSize {
     pub width: [c_int; 2],
     pub height: c_int,
+    pub render_width: c_int,
+    pub render_height: c_int,
+    pub super_res: Rav1dFrameHeader_super_res,
+    pub have_render_size: c_int,
+}
+
+#[derive(Clone)]
+#[repr(C)]
+pub(crate) struct Rav1dFrameHeader {
+    pub size: Rav1dFrameSize,
+    pub film_grain: Rav1dFrameHeader_film_grain,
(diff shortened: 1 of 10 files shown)


rinon (approved):
What purpose does this change serve for getting to memory safety? Guessing you are aiming to initialize all of `Rav1dFrameContext` at once instead of mutably modifying it for initialization? Worth noting the "why" here in the PR description. Generally fully initializing structs would be preferable, but consider if we can avoid that for most fields for now, because we don't need that for memory safety as long as the fields are primitive or similarly default-able types.

This change by itself looks ok to me. However I want to understand if we _really_ need to be doing this or similar changes, though.

kkysen:
> What purpose does this change serve for getting to memory safety? Guessing you are aiming to initialize all of `Rav1dFrameContext` at once instead of mutably modifying it for initialization? Worth noting the "why" here in the PR description. Generally fully initializing structs would be preferable, but consider if we can avoid that for most fields for now, because we don't need that for memory safety as long as the fields are primitive or similarly default-able types.
> 
> This change by itself looks ok to me. However I want to understand if we _really_ need to be doing this or similar changes, though.

In hindsight, I think you might be right that these direct initialization changes aren't totally necessary for pure memory safety.  I'll avoid these kinds of changes that aren't strictly necessary for pure memory safety next time.  That said, they do have numerous advantages, including safety in some sense:
* Everything is initialized directly, so we don't accidentally zero/default initialize things.  While always zero-initializing things is memory safe, that doesn't mean that it's correctly initialized.
* By making all initializations direct, it's much easier to reason about fields when we need to refactor things.  This can be tremendously helpful.
* There are some `enum` fields, that while they have zero-values, they don't have clear defaults.
* I wanted to detangle as many `&mut`s as possible so as to not cause borrow checker issues.
* This future-proofs future refactors of fields.  For example, `ArrayVec`.

PR #622, opened 2023-12-10
`fn parse_frame_size`: Cleanup and make mostly safe

kkysen:
The only remaining `unsafe`ty is `rav1d_get_bit(s)`, which I'll do soon, and `Rav1dPicture::frame_hdr`, which will be `Arc`ed.

### DIFF
--- a/src/obu.rs
+++ b/src/obu.rs
@@ -614,7 +614,7 @@ unsafe fn parse_frame_size(
     if use_ref != 0 {
         for i in 0..7 {
             if rav1d_get_bit(gb) != 0 {
-                let r#ref = &mut c.refs[hdr.refidx[i as usize] as usize].p;
+                let r#ref = &c.refs[hdr.refidx[i as usize] as usize].p;
                 if (*r#ref).p.frame_hdr.is_null() {
                     return Err(EINVAL);
                 }
--- a/src/obu.rs
+++ b/src/obu.rs
@@ -618,10 +618,11 @@ unsafe fn parse_frame_size(
                 if (*r#ref).p.frame_hdr.is_null() {
                     return Err(EINVAL);
                 }
-                let width1 = (*(*r#ref).p.frame_hdr).size.width[1];
-                let height = (*(*r#ref).p.frame_hdr).size.height;
-                let render_width = (*(*r#ref).p.frame_hdr).size.render_width;
-                let render_height = (*(*r#ref).p.frame_hdr).size.render_height;
+                let ref_size = &(*(*r#ref).p.frame_hdr).size;
+                let width1 = ref_size.width[1];
+                let height = ref_size.height;
+                let render_width = ref_size.render_width;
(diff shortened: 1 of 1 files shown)


rinon (changes requested):
This is too far removed from the original C code. Why?  All you needed here was to replace the pointer usage with references. Most of this PR is changes that reorganize the code, when it was already clean and effectively identical to the original C.

kkysen:
> This is too far removed from the original C code. Why? All you needed here was to replace the pointer usage with references. Most of this PR is changes that reorganize the code, when it was already clean and effectively identical to the original C.

Which commits?  Do you mean 2e9c2e3, 2c5dc3e, or any of the others ones?  I didn't think they were that big changes, as most of the diff was just moving code/indentation around.

rinon:
Most of the commits, but especially https://github.com/memorysafety/rav1d/commit/2e9c2e323d4e5e118be8db571a0859013be1ab85 and https://github.com/memorysafety/rav1d/commit/2c5dc3e9e645a837763bd7697691a0f6df4de0ce, here will make it harder to backport changes from upstream dav1d, and don't improve memory safety. We should be making the minimal edits necessary to get to safe Rust without diverging too much from upstream where we don't need to. Of course we want things to be readable, but we should not be trying to improve on the upstream C code readability or make things more idiomatic unless the original was unreadable.

>  I didn't think they were that big changes, as most of the diff was just moving code/indentation around.
Moving code/indentation are the biggest changes in terms of diffs and divergence from upstream. If it's necessary, that's one thing, but where it's not necessary we should try to avoid it.

kkysen:
> Most of the commits, but especially 2e9c2e3 and 2c5dc3e, here will make it harder to backport changes from upstream dav1d, and don't improve memory safety. We should be making the minimal edits necessary to get to safe Rust without diverging too much from upstream where we don't need to. Of course we want things to be readable, but we should not be trying to improve on the upstream C code readability or make things more idiomatic unless the original was unreadable.

Did our general aims change?  I know we're focusing on more strictly memory safety now, but it's generally been a future goal to make things more idiomatic Rust and to keep things performant (usually the idiomatic way in Rust, e.x. iterators, is more performant).  Even with performance not the top top concern anymore (after safety), it's still quite important, so I've been converting things to iterators where that's simple (like here).  Also, for this part of the code specifically, I don't imagine there'd be many significant changes since the bitstream is frozen.

Darkspirit:
https://www.memorysafety.org/initiative/av1/
> Milestones 1-4 of our [work plan](https://www.memorysafety.org/initiative/av1/av1-work-plan/) are funded.

> 4. Migrate block decoding
> [...]As soon as it is possible, we will change types to be safe and idiomatic.[...]

kkysen:
I've removed the 2e9c2e3 and 2c5dc3e commits; the remaining ones are small and mostly around removing `&mut`s and stuff.

rinon (approved):

PR #624, opened 2023-12-11
`fn parse_frame_hdr`: Return `Rav1dFrameHeader` directly

kkysen:
This passes the args `fn parse_frame_hdr` needs to parse a `Rav1dFrameHeader` directly to it vs. initializing some of them before the call or deriving them from `Rav1dContext` pointers.  And then it uses `MaybeUninit::zeroed()` to do the initialization `memset` was before.  I didn't `#[derive(Default)]` `Rav1dFrameHeader` since we'll soon just initialize it directly and this is simpler for an equivalent intermediary state.

### DIFF
--- a/src/obu.rs
+++ b/src/obu.rs
@@ -711,8 +711,13 @@ unsafe fn parse_frame_hdr(
     c: &Rav1dContext,
     seqhdr: &Rav1dSequenceHeader,
     hdr: &mut Rav1dFrameHeader,
+    temporal_id: c_int,
+    spatial_id: c_int,
     gb: &mut GetBits,
 ) -> Rav1dResult {
+    hdr.temporal_id = temporal_id;
+    hdr.spatial_id = spatial_id;
+
     const DEBUG_FRAME_HDR: bool = false;
     let init_ptr = gb.ptr;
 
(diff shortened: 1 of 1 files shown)


src/obu.rs
```diff
@@ -2053,22 +2065,12 @@ unsafe fn parse_obus(
...
-            let res = parse_frame_hdr(c, &mut gb);
-            if res.is_err() {
-                writeln!(c.logger, "Error parsing frame header");
-                c.frame_hdr = 0 as *mut Rav1dFrameHeader;
-                return Err(EINVAL);
-            }
+            let drav1d_frame_hdr_ptr = (*c.frame_hdr_ref)
+                .data
+                .cast::<DRav1d<Rav1dFrameHeader, Dav1dFrameHeader>>();
+            drav1d_frame_hdr_ptr.write(drav1d_frame_hdr);
```
rinon, line 2071: This implies that `drav1d_frame_hdr_ptr` is aligned correctly for the type, which we don't (yet) enforce. Totall ok with this for now, just a thing to keep in mind in terms of unsafety. This write call is unsafe so we will address this before removing it later after updating the allocation itself.

rinon (approved):

src/obu.rs
kkysen: Well we've been dereferencing and holding references to it before, which is UB if it's not properly aligned.  I used `.write` here vs. a normal write to a dereference just so the previous uninitialized value isn't dropped since it's uninitialized.  But it'll all go away once we make it an `Arc` so it should be fine eventually.

rinon: Yep, just an aside.

PR #626, opened 2023-12-11
`struct Debug`: Simplify `mod obu`'s logging

kkysen:
This also makes it easier to keep the logging as is as I refactor parsing `fn`s to directly initialize things (#627).

I also fixed a few typos.

### DIFF
--- a/src/obu.rs
+++ b/src/obu.rs
@@ -113,7 +113,54 @@ use std::cmp;
 use std::ffi::c_int;
 use std::ffi::c_uint;
 use std::ffi::c_void;
+use std::fmt;
 use std::mem::MaybeUninit;
+
+struct Debug {
+    enabled: bool,
+    name: &'static str,
+    init_ptr: *const u8,
+}
+
+impl Debug {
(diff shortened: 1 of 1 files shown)


src/obu.rs
```diff
@@ -116,8 +116,55 @@ use std::cmp;
...
 use std::mem::MaybeUninit;
 
+struct Debug {
+    enabled: bool,
+    name: &'static str,
+    init_ptr: *const u8,
+}
+
+impl Debug {
+    pub const fn new(enabled: bool, name: &'static str, gb: &GetBits) -> Self {
```
rinon, line 129: Why is this const?

kkysen: Since the `enabled: bool` was previously a `const` as `const DEBUG_*: bool = false`, so I wanted to keep the initialization `const`.

rinon:
I'm debating whether it would be better (or necessary at all) to make this a compile-time switch like it was in C. This seems clean though, we can always do that at a later point.

rinon (approved):

PR #631, opened 2023-12-12
`struct GetBits`: Adjust methods to make fields private

kkysen:


rinon (commented):
I don't think you should use `const` for these methods even though you can. Just because something could be usable in a const context doesn't mean it should be, and we don't want to e.g. initialize a GetBits in a const context. At least I can't think of a context we want to, and it doesn't appear we need to.

Other than the const qualifiers this LGTM.

kkysen:
> I don't think you should use `const` for these methods even though you can. Just because something could be usable in a const context doesn't mean it should be, and we don't want to e.g. initialize a GetBits in a const context. At least I can't think of a context we want to, and it doesn't appear we need to.
> 
> Other than the const qualifiers this LGTM.

I can remove the `const`s, but what's the downside of things like that?  To me, it signals that this `fn` is cheap in that it could be done at compile-time.

rinon (approved):
The consts are fine, I'm just not used to seeing them on methods like this I guess.

PR #633, opened 2023-12-12
`fn GetBits::get_bit`: Make return type a `bool`

kkysen:


rinon (commented):
This is the type of thing I've been talking about that shouldn't be a priority right now. Ok, it's done now, and it looks cleaner. However, why did you choose to do it in the first place?

randomPoison (approved):
Approving to unblock merging, but I would also be interested to hear your thoughts RE: prioritization.

kkysen:
> Approving to unblock merging, but I would also be interested to hear your thoughts RE: prioritization.

Oh, @rinon and I discussed this offline.  I definitely realize things like this aren't a priority as they're not actually making things safer, so I'll stay away from these types of refactors.  I tried rebasing this PR out of the stack, though, and it wasn't pretty, so I'll merge it so we can keep working on the more important safety stuff.

PR #637, opened 2023-12-15
`mod internal`: Add comments to internal structures

rinon:
Copies all comments from internal.h to our Rust version. Preparing to make core internal structures safe Rust types.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -234,7 +234,7 @@ pub struct Rav1dContext {
     pub(crate) cdf_pool: *mut Rav1dMemPool,
     pub(crate) cdf: [CdfThreadContext; 8],
 
-    pub(crate) dsp: [Rav1dDSPContext; 3 /* 8, 10, 12 bits/component */],
+    pub(crate) dsp: [Rav1dDSPContext; 3], /* 8, 10, 12 bits/component */
     pub(crate) refmvs_dsp: Rav1dRefmvsDSPContext,
 
     // tree to keep track of which edges are available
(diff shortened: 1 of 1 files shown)


src/internal.rs
```diff
@@ -322,57 +350,64 @@ impl Rav1dFrameContext_bd_fn {
...
+    // in-loop filter per-frame state keeping
     pub start_of_tile_row: *mut u8,
     pub start_of_tile_row_sz: c_int,
     pub need_cdef_lpf_copy: c_int,
     pub p: [*mut DynPixel; 3],
     pub sr_p: [*mut DynPixel; 3],
     pub mask_ptr: *mut Av1Filter,
     pub prev_mask_ptr: *mut Av1Filter,
-    pub restore_planes: c_int,
+    pub restore_planes: c_int, // enum LrRestorePlanes
```
kkysen, line 410:
```suggestion
    pub restore_planes: LrRestorePlanes,
```

We should be able to just use the `enum` directly.  C forces `enum`s to be `int`-sized, but Rust doesn't (though in this case, it's still `c_int`, but future `dav1d` versions could change that, as they sometimes make the type `uint8_t` to save space).

src/internal.rs
```diff
@@ -160,6 +160,10 @@ pub(crate) struct TaskThreadData {
     pub cond: pthread_cond_t,
     pub first: atomic_uint,
     pub cur: c_uint,
+    // This is used for delayed reset of the task cur pointer when
+    // such operation is needed but the thread doesn't enter a critical
+    // section (typically when executing the next sbrow task locklessly).
+    // See src/thread_task.c:reset_task_cur().
```
kkysen, line 166: Do we want this to be an intra-doc link to the Rust `fn`?

kkysen (commented):
Generally LGMT; it's just comments.  One question, though: do we want the comments on fields, types, etc. to be doc comments?

src/internal.rs
rinon: Sure, but let's do that in a separate change later.

rinon:
I don't think we should spend the time right now to make them proper doc comments while these structures are still changing a lot. At some point when the structures are safe, sure.

src/internal.rs
```diff
@@ -160,6 +160,10 @@ pub(crate) struct TaskThreadData {
     pub cond: pthread_cond_t,
     pub first: atomic_uint,
     pub cur: c_uint,
+    /// This is used for delayed reset of the task cur pointer when
+    /// such operation is needed but the thread doesn't enter a critical
+    /// section (typically when executing the next sbrow task locklessly).
+    /// See [crate::src::thread_task::reset_task_cur].
```
kkysen, line 166: nit:

```suggestion
    /// See [`crate::src::thread_task::reset_task_cur`].
```

kkysen (approved):

PR #646, opened 2023-12-19
Add threaded testing to CI

### LINKED ISSUE #644: Ensure that we are testing multi-threaded decoding
We need to refactor pthreads and C atomics into the Rust threading model. To do this correctly I want to make sure we have tests in place that cover multi-threading.

rinon:
The upstream dav1d project tests with threads=2 and framedelay=1,2 in CI, we should be doing the same.

Closes #644

### DIFF
--- a/.github/workflows/build-and-test-x86.yml
+++ b/.github/workflows/build-and-test-x86.yml
@@ -14,6 +14,12 @@ jobs:
           # debug build with optimizations to catch overflows with optimized assembly routines
           {name: "opt-dev", flags: "--profile opt-dev", timeout_multiplier: 2}
         ]
+        # Test with threads and framedelay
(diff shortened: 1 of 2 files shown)


kkysen (commented):
How do I run it with 2 threads on the command line?

Also, would it be useful to keep testing 1 thread in addition to 2?  So we know if something's a threading bug or not

rinon:
2 threads are run if you pass `-f 1` or `-f 2` to `test.sh`, otherwise it is run single-threaded. I basicall

kkysen:
> And yes, I'm keeping 1 thread and adding 1 or 2 frame delay with 2 threads.

Oh perfect then!

thedataking (approved):

PR #648, opened 2023-12-19
`DAV1D_API`: Don't `ptr::read` and convert `out` args

kkysen:
This way we don't need to do useless `.update_rav1d()`s and updating the API to be safe is simpler later.

For now, this is implemented by using `MaybeUninit::zeroed().assume_init()`, but this is temporary; eventually these types will be returned directly by the `fn rav1d_*`s.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -114,6 +114,7 @@ use std::ffi::c_int;
 use std::ffi::c_uint;
 use std::ffi::c_ulong;
 use std::ffi::c_void;
+use std::mem::MaybeUninit;
 use std::process::abort;
 use std::ptr::NonNull;
 use std::sync::Once;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -930,28 +908,7 @@ pub unsafe extern "C" fn dav1d_apply_grain(
...
-        }
-        if let Some(mut itut_t35_ref) = NonNull::new(out_c.itut_t35_ref) {
-            (*itut_t35_ref
-                .as_mut()
-                .data
-                .cast::<DRav1d<Rav1dITUTT35, Dav1dITUTT35>>())
-            .update_rav1d();
-        }
-        let mut out_rust = out_c.into();
+        let mut out_rust = MaybeUninit::zeroed().assume_init(); // TODO(kkysen) Temporary until we return it directly.
```
rinon, line 911: I assume the `update_ravid()` calls are unnecessary as we never read from `out_c`?

rinon (approved):

src/lib.rs
kkysen: Yeah exactly.

PR #651, opened 2023-12-19
`struct Rav1dPicture`: `impl PartialEq, Eq, Hash` based on what `fn md5_write` hashes

kkysen:
I've been using this to debug the remaining three failing tests for #652.  I'm not sure if we want this, but it could be useful.


rinon:
Looks like this needs rebasing, but b344ece7c710b2934ce612c3a3c09cd794ba0528 LGTM

randomPoison (approved):

kkysen:
Doesn't seem like it'll really be useful for any debugging.

PR #653, opened 2023-12-20
`struct {Rav1dEvent,Picture}Flag`: Make `bitflags!` and replace `unsafe fn rav1d_{picture_,}get_event_flags` with `.into()` and `mem::take`

kkysen:


src/picture.rs
```diff
@@ -42,10 +40,21 @@ use std::ffi::c_void;
...
+        const NEW_SEQUENCE = 1 << 0;
+        const NEW_OP_PARAMS_INFO = 1 << 1;
+        const NEW_TEMPORAL_UNIT = 1 << 2;
+    }
+}
+
+impl From<PictureFlags> for Rav1dEventFlags {
+    fn from(value: PictureFlags) -> Self {
+        // [`Rav1dEventFlags`] just has one extra flag.
+        Self::from_bits_truncate(value.bits())
```
rinon, line 55: What happens here if picture flags has new_temporal_unit?

rinon: Looks like it should be ignored, but I don't know the bitflags API off the top of my head. If so, maybe just add a comment about that?

rinon (approved):

src/picture.rs
kkysen: > What happens here if picture flags has new_temporal_unit?

It'll get truncated off (the same behavior of the code before).

I'll explain it in the comment.  I was trying to but I don't think I did a very good job of it.

kkysen: Added in 928f4a8.

PR #654, opened 2023-12-21
`fn Rav1dFrameType::is_key_or_intra`: Deduplicated missed macro expansions

kkysen:


src/thread_task.rs
```diff
@@ -585,7 +585,7 @@ unsafe fn check_tile(t: *mut Rav1dTask, f: *mut Rav1dFrameContext, frame_mt: c_i
         error = (p2 == TILE_ERROR) as c_int;
         error |= ::core::intrinsics::atomic_or_seqcst(&mut (*f).task_thread.error, error);
     }
-    if error == 0 && frame_mt != 0 && (*(*f).frame_hdr).frame_type as c_uint & 1 as c_uint != 0 {
+    if error == 0 && frame_mt != 0 && !(*(*f).frame_hdr).frame_type.is_key_or_intra() {
```
rinon, line 588: `is_inter_or_switch()`?

rinon (approved):

src/thread_task.rs
kkysen: I thought so too but the C does it this way so I translated it like this.

PR #655, opened 2023-12-21
`struct C{Box,Arc}`: Add C/custom `Box` and `Arc` types that support arbitrary free `fn` ptrs from C

kkysen:


src/c_arc.rs
```diff
@@ -0,0 +1,122 @@
...
+/// never move their data ptr during moves, as it's on the heap.
+/// As long as [`Self::owner`] is never moved
+/// without also re-updating [`Self::stable_ref`], this is sound.
+///
+/// Furthermore, storing this stable ref ptr like this
+/// allows for provenance projections of [`Self::stable_ref`],
+/// such as slicing it for a `CArc<[T]>`.
+#[derive(Debug)]
+pub struct CArc<T: ?Sized> {
+    owner: Arc<CBox<T>>,
```
rinon, line 35: Could this be an `Arc<Pin<CBox<T>>>`? Should it? I think it can be and that would make the assumptions and requirements we're putting on the box explicit. Arc impls Unpin, so it wouldn't change where we can use CArc.

kkysen: I'm not too familiar with `Pin` so I wasn't sure.  I'll try it out and see if it works nicely; it'd be nice to move the unsafely and invariants into `Pin`.

src/c_arc.rs
```diff
@@ -0,0 +1,168 @@
...
+        owner.into()
+    }
+}
+
+impl<T> AddAssign<usize> for CArc<[T]> {
+    /// Slice [`Self::stable_ref`].
+    fn add_assign(&mut self, rhs: usize) {
+        self.stable_ref = self[rhs..].into();
+    }
+}
```
rinon, line 158: I think we should make this our own method, rather than using `+=`. There's not a good way to document this and the behavior wasn't obvious to me. This tripped me up, and we could make it more explicit what we are doing when we shift the start of the slice view.

kkysen: Okay I can do that.  I wanted to make it so we could slice it like a normal slice but that didn't work out with the `Deref` `trait`.

src/c_box.rs
```diff
@@ -0,0 +1,94 @@
...
+/// That is, it is analogous to a [`Box`],
+/// but it lets you set a C-style `free` `fn` for deallocation
+/// instead of the normal [`Box`] (de)allocator.
+/// It can also store a normal [`Box`] as well.
+#[derive(Debug)]
+pub struct CBox<T: ?Sized> {
+    data: NonNull<T>,
+    /// If [`None`], [`Self::data`] should be freed as a [`Box`].
+    free: Option<Free>,
+    _phantom: PhantomData<T>,
```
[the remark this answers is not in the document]
kkysen: Actually, I'm going to leave the `PhantomData`.

https://github.com/rust-lang/rust/blob/2e79c8deb05320c6f2bc35cc0e21b943ec0f29b5/library/core/src/ptr/unique.rs#L36-L44

```rust
pub struct Unique<T: ?Sized> {
    pointer: NonNull<T>,
    // NOTE: this marker has no consequences for variance, but is necessary
    // for dropck to understand that we logically own a `T`.
    //
    // For details, see:
    // https://github.com/rust-lang/rfcs/blob/master/text/0769-sound-generic-drop.md#phantom-data
    _marker: PhantomData<T>,
}
```

says it's necessary for dropck to understand that we logically own a `T`, which we do.

src/c_arc.rs
kkysen: I replaced it with `fn slice_in_place` in 6df10fa, which takes a normal slice range, so we'll be able to do things like `data.slice_in_place(offset..);` or `data.slice_in_place(..len)`.  It's more flexible in that it supports any kind of range.  This necessitates some extra `..`s sometimes, but I think it's worth it because it's similar syntactically to normal slicing.

src/c_box.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/// A C/custom [`Box`].
+///
+/// That is, it is analogous to a [`Box`],
+/// but it lets you set a C-style `free` `fn` for deallocation
+/// instead of the normal [`Box`] (de)allocator.
+/// It can also store a normal [`Box`] as well.
+#[derive(Debug)]
+pub struct CBox<T: ?Sized> {
```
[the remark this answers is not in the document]
kkysen: I checked the codegen/layout on [godbolt](https://godbolt.org/z/cjjYWEsGq):

```rust
use std::ffi::c_void;
use std::marker::PhantomData;
use std::ops::Deref;
use std::ptr::drop_in_place;
use std::ptr::NonNull;

pub fn box_into_raw<T: ?Sized>(r#box: Box<T>) -> NonNull<T> {
    let raw = Box::into_raw(r#box);
    // Safety: [`Box::into_raw`] never returns null.
    unsafe { NonNull::new_unchecked(raw) }
}

pub type FnFree = unsafe extern "C" fn(ptr: *const u8, cookie: *mut c_void);

/// A `free` "closure", i.e. a [`FnFree`] and an enclosed context [`Self::cookie`].
// #[derive(Debug)]
pub struct Free {
    pub free: FnFree,
    pub cookie: *mut c_void,
}

impl Free {
    pub unsafe fn free(&self, ptr: *mut c_void) {
        (self.free)(ptr as *const u8, self.cookie)
    }
}

/// A C/custom [`Box`].
///
/// That is, it is analogous to a [`Box`],
/// but it lets you set a C-style `free` `fn` for deallocation
/// instead of the normal [`Box`] (de)allocator.
/// It can also store a normal [`Box`] as well.
// #[derive(Debug)]
pub struct CBox<T: ?Sized> {
    data: NonNull<T>,
    /// If [`None`], [`Self::data`] should be freed as a [`Box`].
    free: Option<Free>,
    _phantom: PhantomData<T>,
}

impl<T: ?Sized> AsRef<T> for CBox<T> {
    fn as_ref(&self) -> &T {
        unsafe { self.data.as_ref() }
    }
}

impl<T: ?Sized> Deref for CBox<T> {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        self.as_ref()
    }
}

impl<T: ?Sized> Drop for CBox<T> {
    fn drop(&mut self) {
        let ptr = self.data.as_ptr();
        match &self.free {
            None => {
                // Safety: If [`Self::free`] is [`None`],
                // then [`Self::data`] is a [`Box`].
                // See [`Self::from_box`].
                let _ = unsafe { Box::from_raw(ptr) };
            }
            Some(free) => {
                // Safety: See below.
                // The [`FnFree`] won't run Rust's `fn drop`,
                // so we have to do this ourselves first.
                unsafe { drop_in_place(ptr) };
                let ptr = ptr.cast();
                // Safety: If [`Self::free`] is [`Some`],
                // then that [`FnFree`] should be used for `free`ing.
                // See [`Self::from_c`].
                unsafe { free.free(ptr) }
            }
        }
    }
}

impl<T: ?Sized> CBox<T> {
    pub fn from_c(data: NonNull<T>, free: Free) -> Self {
        Self {
            data,
            free: Some(free),
            _phantom: PhantomData,
        }
    }

    pub fn from_box(data: Box<T>) -> Self {
        Self {
            data: box_into_raw(data),
            free: None,
            _phantom: PhantomData,
        }
    }
}

pub enum CBoxEnum<T: ?Sized> {
    Rust(Box<T>),
    C {
        data: NonNull<T>,
        free: Free,
        _phantom: PhantomData<T>,
    },
}

impl<T: ?Sized> AsRef<T> for CBoxEnum<T> {
    fn as_ref(&self) -> &T {
        match self {
            Self::Rust(r#box) => r#box.as_ref(),
            Self::C { data, .. } => unsafe { data.as_ref() },
        }
    }
}

impl<T: ?Sized> Deref for CBoxEnum<T> {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        self.as_ref()
    }
}

impl<T: ?Sized> Drop for CBoxEnum<T> {
    fn drop(&mut self) {
        match self {
            Self::Rust(_) => {}, // Drop normally.
            Self::C { data, free, .. } => {
                let ptr = data.as_ptr();
                unsafe { drop_in_place(ptr) };
                let ptr = ptr.cast();
                unsafe { free.free(ptr) }
            }
        }
    }
}

impl<T: ?Sized> CBoxEnum<T> {
    pub fn from_c(data: NonNull<T>, free: Free) -> Self {
        Self::C {
            data,
            free,
            _phantom: PhantomData,
        }
    }

    pub fn from_box(data: Box<T>) -> Self {
        Self::Rust(data)
    }
}

pub fn f(data: &Box<[u8]>, i: usize) -> u8 {
    data[i]
}

pub fn g(data: &CBox<[u8]>, i: usize) -> u8 {
    data[i]
}

pub fn h(data: &CBoxEnum<[u8]>, i: usize) -> u8 {
    data[i]
}
```

`f`, `g`, and `h` are all identical, so I'll switch to the `enum` version as it definitely is simpler.  I didn't realize the optimizer would be smart enough to ensure data ptr is in the same place for both variants.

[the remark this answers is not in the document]
kkysen: Done in fc43ddc.

src/c_arc.rs
kkysen: I `Pin`ned it in b0b6cdf.  Is that worth it?  It doesn't really improve the `unsafe`ty, I don't think, though it makes the safety invariants a little bit clearer.

randomPoison (approved):

PR #656, opened 2023-12-21
`struct Rav1dUserData`: Replace `Rav1dRef` with `Option<CArc<u8>>`

kkysen:
This replaces `struct Rav1dUserData`, with its `Rav1dRef`, with `Option<CArc<u8>>`.

This lets us do ref-counting automatically and safely with a `#[derive(Clone)]`.  Thus, this replaces other `fn`s on `Rav1dDataProps`:

* `fn *_unref` s are `fn drop`s that leave the source in a valid, generally zeroed/`Default::default()` state, similar to C++ move constructors.  Thus, I've replaced these with `let _ = mem::take(...)`s.

* `fn *_ref`s and `fn *_copy`s are `fn clone`s that do ref-counting, so I've replaced them with `.clone()`s, where the ref-counting happens automatically with `#[derive(Clone)]` from the `Arc`.

* `fn *_move_ref`s are moves that leave the source in a valid state like `fn *_unref`s.  Thus, I've replaced these with `mem::take`s.

Furthermore, using `memset(0)` to clear droppable types like `Rav1dUserData` is unsound as it may leave the type in an invalid state that is then dropped.  Thus, we `impl Default` types that contain `Rav1dUserData` and use `mem::take` instead.

Also, if you could review this somewhat carefully, that'd be great, as I don't think any of the tests test non-null/`None` user data.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -519,8 +519,8 @@ pub(crate) unsafe fn rav1d_parse_sequence_header(
                 offset: 0,
                 size: 0,
                 user_data: Rav1dUserData {
-                    data: 0 as *const u8,
-                    r#ref: 0 as *mut Rav1dRef,
+                    data: None,
+                    r#ref: None,
                 },
             },
         };
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,7 +5,6 @@ use crate::include::common::bitdepth::DynCoef;
 use crate::include::common::validate::validate_input;
 use crate::include::dav1d::common::Dav1dDataProps;
 use crate::include::dav1d::common::Rav1dDataProps;
-use crate::include::dav1d::common::Rav1dUserData;
 use crate::include::dav1d::data::Dav1dData;
 use crate::include::dav1d::data::Rav1dData;
 use crate::include::dav1d::dav1d::Dav1dContext;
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -1248,26 +1230,23 @@ pub unsafe extern "C" fn dav1d_data_wrap(
...
-    result.into()
+    || -> Rav1dResult {
+        let buf = validate_input!(NonNull::new(buf).ok_or(EINVAL))?;
+        // Note that `dav1d` doesn't do this check, but they do for the similar [`dav1d_data_wrap`].
+        let user_data = validate_input!(NonNull::new(user_data.cast_mut()).ok_or(EINVAL))?;
+        let mut data = buf.as_ptr().read().to::<Rav1dData>();
+        data.wrap_user_data(user_data, free_callback, cookie)?;
+        buf.as_ptr().write(data.into());
+        Ok(())
+    }()
```
randomPoison, line 1248: im begging, screaming, crying for a `try` block to save us from this javascript-ass shit

randomPoison: (this is not a criticism, just a jumping-off point for future discussion)

src/picture.rs
```diff
@@ -202,7 +215,10 @@ unsafe fn picture_alloc_with_edges(
         return res;
     }
     (*pic_ctx).allocator = (*p_allocator).clone();
-    (*pic_ctx).pic = (*p).clone();
+    // TODO(kkysen) A normal assignment here as it used to be
+    // calls `fn drop` on `(*pic_ctx).pic`, which segfaults as it is uninitialized.
+    // We need to figure out the right thing to do here.
+    addr_of_mut!((*pic_ctx).pic).write((*p).clone());
```
randomPoison, line 221: What's the `addr_of_mut!` doing here?

randomPoison (approved):

src/picture.rs
kkysen: Getting a pointer to the field to do a write without creating a reference to uninitialized memory.

PR #652, opened 2023-12-19
(WIP) `struct Rav1dData{,Props}`: Replace `Rav1dRef`s with `CArc`s

kkysen:
There are still 3 remaining failing tests that I've been trying to debug for a while but not having much success (I left my current debugging code in).  I'll split up the PR more carefully once I get it working.

PR #658, opened 2023-12-21
`struct Rav1dData`: Replace `Rav1dRef` with `Option<CArc<[u8]>>`

kkysen:
~~This is still buggy on those 3 tests (`00000063`, `redundant_frame_header`, `issue_48`) failing in #652.  The output video is off by a small number of bits for each of them, but the picture parameters and headers and OBU metadata are all the same.  I've now narrowed down the PR to this smaller one (specifically the 2nd commit, 6afc41b), which still evidently contains the bug.  The divergence in the data happens somewhere in `fn rav1d_decode_tile_sbrow`, but the dataflow gets very confusing and I haven't been able to narrow it down further yet.~~

~~Among the changes in 6afc41b, I also tried not splitting up `Rav1dData` into `CArc<[u8]>` and `Rav1dDataProps` within `fn parse_obus`, but that doesn't change anything.  `fn gen_picture` is also changed in a more straightforward manner that is less concise and efficient as in #652, but the bug is still there.~~

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1033,7 +1033,7 @@ unsafe fn close_internal(c_out: &mut *mut Rav1dContext, flush: c_int) {
         rav1d_free_aligned((*f).ts as *mut c_void);
         rav1d_free_aligned((*f).ipred_edge[0] as *mut c_void);
         free((*f).a as *mut c_void);
-        (*f).tiles = Default::default(); // TODO(kkysen) Remove when dropped properly.
+        let _ = mem::take(&mut (*f).tiles);
         free((*f).lf.mask as *mut c_void);
         free((*f).lf.lr_mask as *mut c_void);
         free((*f).lf.level as *mut c_void);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -501,7 +501,9 @@ pub(crate) unsafe fn rav1d_parse_sequence_header(
             let len = rav1d_parse_obus(&mut *c, &mut buf, true)?;
             assert!(len <= buf.sz);
             buf.sz -= len;
-            buf.data = buf.data.add(len);
+            buf.data = buf
+                .data
+                .and_then(|data| NonNull::new(data.as_ptr().add(len)));
         }
 
         if (*c).seq_hdr.is_none() {
@@ -692,7 +694,10 @@ unsafe fn gen_picture(c: &mut Rav1dContext) -> Rav1dResult {
             Ok(len) => {
                 assert!(len <= c.in_0.sz);
                 c.in_0.sz -= len;
-                c.in_0.data = c.in_0.data.add(len);
+                c.in_0.data = c
+                    .in_0
+                    .data
+                    .and_then(|data| NonNull::new(data.as_ptr().add(len)));
                 if c.in_0.sz == 0 {
                     rav1d_data_unref_internal(&mut c.in_0);
                 }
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -502,15 +499,22 @@ pub(crate) unsafe fn rav1d_parse_sequence_header(
...
+        } = match NonNull::new(ptr.cast_mut()) {
+            None => Default::default(),
+            Some(ptr) => Rav1dData::wrap(
+                slice::from_raw_parts(ptr.as_ptr(), sz).into(),
+                Some(dummy_free),
+                ptr::null_mut(),
+            )?,
+        };
+        if let Some(data) = &mut data {
+            while !data.is_empty() {
```
rinon, line 514: ~~Am I correct in reading this as a deref to the inner slice and calling is_empty() on that? It's hard to figure out types here. If so, how is that ever possible? We don't change the slice size, just the stable_ptr field, right?~~

Ok, this is pretty confusing when we're dealing with slices. Probably worth rethinking some of this or making it more explicit. I see now that we are effectively keeping a slice in the stable_ref and that's what is deref'd, not the actual underlying boxed slice. That's kind of what we want, but it isn't obvious that's what will happen.

src/obu.rs
```diff
@@ -2203,23 +2208,27 @@ unsafe fn parse_obus(c: &mut Rav1dContext, r#in: &Rav1dData, global: bool) -> Ra
         let bit_pos = gb.pos();
         assert!(bit_pos & 7 == 0);
         assert!(pkt_bytelen >= bit_pos >> 3);
-        let mut data = Default::default();
-        rav1d_data_ref(&mut data, r#in);
-        data.data = data.data.offset((bit_pos >> 3) as isize);
-        data.sz = (pkt_bytelen - (bit_pos >> 3)) as usize;
```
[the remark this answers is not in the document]
rinon: I added `assert_eq!(data.len(), pkt_bytelen)` and a lot more tests started failing. `pkt_bytelen` is smaller than `data.len()` in some cases and so this line was shrinking the data slice, in addition to moving the cursor forward.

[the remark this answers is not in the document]
rinon: A hacky "shrink" method I added to `CArc<[T]>` and shrinking the slice view to `pkt_bytelen` before moving the start fixed the failing tests.

src/lib.rs
rinon: What about making RavidData not a `type` but instead a newtype with a narrower interface? `RavidData` is apparently a ref-counted slice view, which isn't exactly a type Rust already has, so we should be explicit what that type is.

kkysen: Wdym exactly?  It's already a `struct` right now.

rinon: I must have been looking at a different PR, but the point about all the refs and derefs being confusing is still here. It's confusing to have a type that derefs to an inner type except when it's a slice, when it behaves more like a view into that thing. I think I put a comment on the CArc PR that explains that better.

src/obu.rs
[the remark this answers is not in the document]
kkysen: Thanks for finding the bug!  It's all fixed now, and we don't need a hacky `shrink` method either now that `impl AddAssign` has been replaced with the more flexible `fn slice_in_place`.

src/lib.rs
kkysen: I tried to separate the two, but it proved fairly difficult and very unwieldy.  Can I just improve the docs (I might've added some stuff by now)?

randomPoison: @rinon Checking in with you: Is this something to block the PR on? The current version of things seems fairly understandable to me, I'm not sure what has changed between when you last commented and now.

src/lib.rs
```diff
@@ -690,22 +693,26 @@ unsafe fn gen_picture(c: &mut Rav1dContext) -> Rav1dResult {
...
+        data: r#in,
+        m: props,
+    } = mem::take(&mut c.in_0);
+    let Some(mut r#in) = r#in else { return Ok(()) };
+    while !r#in.is_empty() {
+        let len = rav1d_parse_obus(c, &r#in, &props, false);
+        if let Ok(len) = len {
+            r#in.slice_in_place(len..);
         }
+        // Note that [`output_picture_ready`] doesn't read [`Rav1dContext::in_0`].
```
randomPoison, line 707: general nit (not something to necessarily fix here, just a note for the future): I wouldn't bother using the `[markdown link syntax]` in non-doc comments. Keep using \`backticks\` around item names, but the [square brackets] just add noise when they're not actually being used to create a link in the generated docs.

randomPoison (approved):
Looks good in it's current state but make sure you address Stephen's outstanding comment before merging.

src/lib.rs
kkysen: @rinon, does this look okay now?

rinon:
Sorry missed the pings on this, LGTM for now, yeah.

PR #659, opened 2024-01-03
`struct Rav1d{ContentLightLevel,MasteringDisplay`: Replace `Rav1dRef`s with `Option<Arc<_>>`s

kkysen:
These are the non-pooled `Rav1dRef`s that don't have separate `{D,R}av1d*` types, so the conversion is very straightforward.


randomPoison (approved):
So nice to see all this custom C logic turn into Rust std types 😁

PR #660, opened 2024-01-03
`struct Rav1dITUTT35`: Make `payload{,_size}` fields a `Box<[u8]>`

kkysen:
This makes `Rav1dITUTT35` no longer a DST, which is very tricky (it makes `Arc`ification much more difficult, for example), at the cost of an extra layer of indirection, but since the payload probably won't be accessed through the `Rav1dITUTT35` excessively, I don't think this should be a problem.  Note that having a separate `Rav1dITUTT35` from `Dav1dITUTT35` allows us to do this easily.

### DIFF
--- a/src/obu.rs
+++ b/src/obu.rs
@@ -2462,17 +2462,7 @@ unsafe fn parse_obus(c: &mut Rav1dContext, r#in: &Rav1dData, global: bool) -> Ra
 
                         let country_code = country_code as u8;
                         let country_code_extension_byte = country_code_extension_byte as u8;
-                        // We need our public headers to be C++ compatible, so payload can't be
-                        // a flexible array member
-                        let payload = (*r#ref)
-                            .data
-                            .cast::<u8>()
-                            .offset(::core::mem::size_of::<DRav1d<Rav1dITUTT35, Dav1dITUTT35>>()
-                                as isize);
-                        let payload_size = payload_size as usize;
-                        for i in 0..payload_size {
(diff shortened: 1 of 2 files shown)


src/obu.rs
```diff
@@ -2462,25 +2462,14 @@ unsafe fn parse_obus(c: &mut Rav1dContext, r#in: &Rav1dData, global: bool) -> Ra
...
-                        let payload = (*r#ref)
-                            .data
-                            .cast::<u8>()
-                            .offset(::core::mem::size_of::<DRav1d<Rav1dITUTT35, Dav1dITUTT35>>()
-                                as isize);
-                        let payload_size = payload_size as usize;
-                        for i in 0..payload_size {
-                            *payload.offset(i as isize) = gb.get_bits(8) as u8;
-                        }
+                        let payload = (0..payload_size).map(|_| gb.get_bits(8) as u8).collect(); // TODO(kkysen) fallible allocation
```
randomPoison, line 2465: Huh, I didn't realize you could collect into a `Box<[T]>`. I'm guessing it's just using a `Vec` in the middle, but it's nice that the stdlib APIs are there to make that work automatically

randomPoison (approved):

src/obu.rs
kkysen: I'm not sure what it's using in the middle; it could conceivably skip the overhead of an intermediary `Vec` with this API.  `.collect()` is so nice.

PR #664, opened 2024-01-04
`mod picture`: Cleanup

kkysen:
This makes a bunch of pointers into refs and other simple safety cleanups.  This is preparing to make `Rav1dPicture` fully `Arc`ified and safe.


src/picture.rs
```diff
@@ -171,8 +172,8 @@ unsafe extern "C" fn free_buffer(_data: *const u8, user_data: *mut c_void) {
 }
 
 unsafe fn picture_alloc_with_edges(
-    c: *mut Rav1dContext,
-    p: *mut Rav1dPicture,
+    logger: &Option<Rav1dLogger>,
```
randomPoison, line 175: `&Option<T>` is weird to see in a Rust API, wouldn't `Option<&T>` make more sense?

randomPoison (approved):

src/picture.rs
kkysen: `write!` is set up to work on `Option<Rav1dLogger>` so I'm not sure if that'd work.

Issue #614, opened 2023-12-08
`enum BPC` depends on `opt-level >= 1` to dead-code eliminate asm `fn`s


kkysen:
This is why #605 caused `cargo build` to break because it uses `opt-level = 0`.  While this is being fixed (not the highest priority, but important), #605 is being reverted in #613.

kkysen:
Fixed by #891.

PR #665, opened 2024-01-04
 Add arm/aarch64 assembly code for loop restoration from dav1d 1.3.0

fbossen:
Backport changes from dav1d 1.3.0 and provide Rust equivalents to C functions.

### DIFF
--- a/src/looprestoration.rs
+++ b/src/looprestoration.rs
@@ -2212,7 +2212,9 @@ unsafe fn sgr_filter_3x3_neon<BD: BitDepth>(
         }
     }
 
-    while track == Track::main && h > 0 {
+    // h > 0 can be true only if track == Track::main
+    // The original C code uses goto statements and skips over this loop when h <= 0
+    while h > 0 {
         sgr_box3_hv_neon(
             &mut sumsq_ptrs,
             &mut sum_ptrs,
@@ -2552,7 +2554,9 @@ unsafe fn sgr_filter_5x5_neon<BD: BitDepth>(
         }
     }
 
-    while track == Track::main && h > 0 {
+    // h > 0 can be true only if track == Track::main
+    // The original C code uses goto statements and skips over this loop when h <= 0
+    while h > 0 {
         rav1d_sgr_box5_row_h_neon(sumsq_ptrs[3], sum_ptrs[3], left, src, w, edges, bd);
         left = left.offset(1);
         src = src.add(stride);
@@ -3073,7 +3077,9 @@ unsafe fn sgr_filter_mix_neon<BD: BitDepth>(
         }
     }
 
-    while track == Track::main && h > 0 {
+    // h > 0 can be true only if track == Track::main
+    // The original C code uses goto statements and skips over this loop when h <= 0
+    while h > 0 {
         rav1d_sgr_box35_row_h_neon(
             sumsq3_ptrs[2],
             sum3_ptrs[2],
--- a/src/looprestoration.rs
+++ b/src/looprestoration.rs
@@ -2897,12 +2897,12 @@ unsafe fn sgr_filter_mix_neon<BD: BitDepth>(
             h -= 1;
             if h <= 0 {
                 track = Track::vert2;
+            } else {
+                // ptrs are rotated by 2; both [3] and [4] now point at rows[0]; set
+                // one of them to point at the previously unused rows[4].
+                sumsq5_ptrs[3] = sumsq5_rows[4];
+                sum5_ptrs[3] = sum5_rows[4];
             }
-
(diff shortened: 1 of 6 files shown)


src/looprestoration.rs
```diff
@@ -1556,32 +1556,1851 @@ unsafe fn sgr_filter_3x3_neon<BD: BitDepth>(
...
+                    bitdepth_max: c_int,
+                );
+            }
+            $name
+        }};
+    }
+    (match BD::BPC {
+        BPC::BPC8 => asm_fn!(dav1d_sgr_box3_row_h_8bpc_neon),
+        BPC::BPC16 => asm_fn!(dav1d_sgr_box3_row_h_16bpc_neon),
+    })(sumsq, sum, left.cast(), src.cast(), w, edges, bd.into_c())
```
kkysen, line 1612:
```suggestion
    bd_fn!(asm_fn, BD, sgr_box3_row_h, neon)(sumsq, sum, left.cast(), src.cast(), w, edges, bd.into_c())
```

You can use `bd_fn!` for these (same applies for all the other patterns like this), and doing so makes fixing #614 simpler.

src/looprestoration.rs
```diff
@@ -1556,32 +1556,1851 @@ unsafe fn sgr_filter_3x3_neon<BD: BitDepth>(
...
-        p.cast(),
-        stride,
+    macro_rules! asm_fn {
+        ($name:ident) => {{
+            extern "C" {
+                fn $name(
+                    sumsq: *mut i32,
+                    sum: *mut i16,
+                    left: *const c_void,
+                    src: *const c_void,
```
kkysen, line 1600:
```suggestion
                    left: *const [DynPixel; 4],
                    src: *const DynPixel,
```

kkysen (changes requested):
I made a few comments on one of the functions that applies to a lot of them.  I'm still reviewing the rest of the PR, though.

fbossen:
> I made a few comments on one of the functions that applies to a lot of them. I'm still reviewing the rest of the PR, though.

Thank you for the feedback. I made changes as suggested

kkysen:
@fbossen, sorry I haven't gotten around to fully reviewing this in a while.  I'll try to at some point soonish, though I think it's a lower priority as safetifying is higher priority than backporting at this point.

src/looprestoration.rs
```diff
@@ -1556,32 +1556,1845 @@ unsafe fn sgr_filter_3x3_neon<BD: BitDepth>(
...
+            if h <= 0 {
+                track = Track::vert2;
+            } else {
+                sumsq_ptrs[2] = sumsq_rows[2];
+                sum_ptrs[2] = sum_rows[2];
+            }
+        }
+    }
+
+    while track == Track::main && h > 0 {
```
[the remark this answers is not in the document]
randomPoison: This also applies to the other places where you translated the the loops similarly, since it looks like the structure of the other functions is similar.

src/looprestoration.rs
```diff
@@ -1556,32 +1556,1845 @@ unsafe fn sgr_filter_3x3_neon<BD: BitDepth>(
...
+
+            h -= 1;
+            if h <= 0 {
+                track = Track::vert2;
+            }
+
+            // ptrs are rotated by 2; both [3] and [4] now point at rows[0]; set
+            // one of them to point at the previously unused rows[4].
+            sumsq5_ptrs[3] = sumsq5_rows[4];
+            sum5_ptrs[3] = sum5_rows[4];
```
randomPoison, line 2901: Should these lines be in an `else` on the previous `if`? That's the structure you've been using to translate the `goto`s, and these lines were after the `goto` in the original example.

fbossen: Good catch! I fixed it.

src/looprestoration.rs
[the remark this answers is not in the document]
fbossen: `h > 0` will be true if `track == Track::main` and the `h` input to the function is `> 0` (which should always be the case). I modified the condition to remove the `track == Track::main` part which is redundant and added a comment in the code.

randomPoison (approved):

kkysen (commented):
The strides in here are mostly `usize`s not `isize`s.  These aren't always positive, right?  I think you've might've written this before we realized we need to account for negative strides everywhere, unless I'm mistaken about where this code is being called from since I didn't look at it very closely.

fbossen:
> The strides in here are mostly `usize`s not `isize`s. These aren't always positive, right? I think you've might've written this before we realized we need to account for negative strides everywhere, unless I'm mistaken about where this code is being called from since I didn't look at it very closely.

Indeed. I made changes to have strides as `isize`.

PR #666, opened 2024-01-04
`struct Rav1dThreadPicture::progress`: Make a `struct ThreadPictureProgress` of [`AtomicU32; 2]` that is always allocated inline since it's so small

kkysen:
Previously, `progress` was either null or `*const [AtomicU32; 2]`, allocated by the `extra` and `extra_ptr` in `fn picture_alloc_with_edges` as part of the `pic_ctx` allocation.  But since `extra_ptr` is only ever the `*const [AtomicU32; 2]` or null, we can just allocate it inline and avoid this tricky dynamically sized allocation.  This also lets us avoid extra branches and checks that we would have to do if we used an `Option` here.

I made `ThreadPictureProgress` a newtype to `impl Clone` on it, as this has to be done manually for atomics.  It also lets us name the `[0]` and `[1]` indices as `{block,pixel}_data` as the comments say.

@rinon, this may interfere with some of your atomic work, but it wasn't too large of a change to potentially duplicate.  I wasn't intending on touching that part of the codebase, but it's entangled with `Rav1dPicture`, its `Rav1dRef`, and the `pic_ctx`.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4987,7 +4987,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
             c.cached_error_props = out_delayed.p.m.clone();
             rav1d_thread_picture_unref(out_delayed);
         } else if !out_delayed.p.data[0].is_null() {
-            let progress = (*out_delayed.progress.add(1)).load(Ordering::Relaxed);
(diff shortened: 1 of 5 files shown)


kkysen:
This was the wrong approach.  Replaced by:
* #669
* #670

PR #667, opened 2024-01-04
`fn dav1d_default_picture_alloc`: Cleanup

kkysen:


src/picture.rs
```diff
@@ -84,58 +84,47 @@ pub(crate) struct pic_ctx_context {
...
-        ((*p).p.layout as c_uint != Rav1dPixelLayout::I400 as c_int as c_uint) as c_int;
-    let ss_ver = ((*p).p.layout as c_uint == Rav1dPixelLayout::I420 as c_int as c_uint) as c_int;
-    let ss_hor = ((*p).p.layout as c_uint != Rav1dPixelLayout::I444 as c_int as c_uint) as c_int;
-    let mut y_stride: ptrdiff_t = (aligned_w << hbd) as ptrdiff_t;
-    let mut uv_stride: ptrdiff_t = if has_chroma != 0 {
-        y_stride >> ss_hor
-    } else {
-        0
-    };
+    assert!(::core::mem::size_of::<Rav1dMemPoolBuffer>() <= 64);
```
randomPoison, line 90: `std::mem` is already imported at the top of the file, so could we do just `mem::size_of` here?

src/picture.rs
```diff
@@ -84,58 +84,47 @@ pub(crate) struct pic_ctx_context {
...
+    p.stride[1] = uv_stride;
+    let y_sz = (y_stride * aligned_h as isize) as usize;
+    let uv_sz = (uv_stride * (aligned_h >> ss_ver) as isize) as usize;
+    let pic_size = y_sz + 2 * uv_sz;
+    let buf = rav1d_mem_pool_pop(
         cookie as *mut Rav1dMemPool,
-        pic_size
-            .wrapping_add(64)
-            .wrapping_sub(::core::mem::size_of::<Rav1dMemPoolBuffer>()),
+        pic_size + 64 - ::core::mem::size_of::<Rav1dMemPoolBuffer>(),
```
randomPoison, line 113: We can also simplify this to `mem::size_of`, I think.

randomPoison (approved):

src/picture.rs
kkysen: Yeah, we could, but I may clean this up more and remove these entirely so I didn't bother changing it here.

PR #671, opened 2024-01-10
`fn rav1d_worker_task`: Refactor state machine

rinon:
`rav1d_worker_task` involves complex control flow with gotos, and therefore was transpiled into a state machine. This refactoring restores the original structure with minimal additional structured control flow to emulate gotos.

### DIFF
--- a/src/thread_task.rs
+++ b/src/thread_task.rs
@@ -813,869 +813,738 @@ unsafe fn delayed_fg_task(c: *const Rav1dContext, ttd: *mut TaskThreadData) {
 }
 
 pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
-    let mut flush;
-    let mut error_0;
-    let mut sby;
-    let mut f: *mut Rav1dFrameContext;
-    let mut t: *mut Rav1dTask;
-    let mut prev_t: *mut Rav1dTask;
-    let mut current_block: u64;
     let tc: *mut Rav1dTaskContext = data as *mut Rav1dTaskContext;
     let c: *const Rav1dContext = (*tc).c;
     let ttd: *mut TaskThreadData = (*tc).task_thread.ttd;
+
     rav1d_set_thread_name(b"dav1d-worker\0" as *const u8 as *const c_char);
-    pthread_mutex_lock(&mut (*ttd).lock);
-    's_18: while !(*tc).task_thread.die {
-        if !(::core::intrinsics::atomic_load_seqcst((*c).flush) != 0) {
-            merge_pending(c);
-            if (*ttd).delayed_fg.exec != 0 {
-                delayed_fg_task(c, ttd);
-                continue;
-            } else {
-                f = 0 as *mut Rav1dFrameContext;
-                t = 0 as *mut Rav1dTask;
-                prev_t = 0 as *mut Rav1dTask;
-                if (*c).n_fc > 1 as c_uint {
-                    let mut i: c_uint = 0 as c_int as c_uint;
-                    loop {
-                        if !(i < (*c).n_fc) {
-                            current_block = 5601891728916014340;
-                            break;
-                        }
-                        let first: c_uint =
-                            ::core::intrinsics::atomic_load_seqcst(&mut (*ttd).first);
-                        f = &mut *((*c).fc)
-                            .offset(first.wrapping_add(i).wrapping_rem((*c).n_fc) as isize)
-                            as *mut Rav1dFrameContext;
-                        if !(::core::intrinsics::atomic_load_seqcst(
-                            &mut (*f).task_thread.init_done as *mut atomic_int,
-                        ) != 0)
-                        {
-                            t = (*f).task_thread.task_head;
-                            if !t.is_null() {
-                                if (*t).type_0 as c_uint == RAV1D_TASK_TYPE_INIT as c_int as c_uint
-                                {
-                                    current_block = 7012560550443761033;
-                                    break;
-                                }
-                                if (*t).type_0 as c_uint
-                                    == RAV1D_TASK_TYPE_INIT_CDF as c_int as c_uint
-                                {
-                                    let p1 = (if !((*f).in_cdf.progress).is_null() {
-                                        ::core::intrinsics::atomic_load_seqcst((*f).in_cdf.progress)
-                                    } else {
-                                        1 as c_int as c_uint
-                                    }) as c_int;
-                                    if p1 != 0 {
-                                        ::core::intrinsics::atomic_or_seqcst(
-                                            &mut (*f).task_thread.error,
-                                            (p1 == TILE_ERROR) as c_int,
-                                        );
-                                        current_block = 7012560550443761033;
-                                        break;
-                                    }
-                                }
-                            }
-                        }
-                        i = i.wrapping_add(1);
-                    }
-                } else {
-                    current_block = 5601891728916014340;
-                }
-                's_107: loop {
-                    match current_block {
-                        5601891728916014340 => {
-                            if (*ttd).cur < (*c).n_fc {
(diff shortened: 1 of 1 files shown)


src/thread_task.rs
```diff
@@ -813,869 +813,746 @@ unsafe fn delayed_fg_task(c: *const Rav1dContext, ttd: *mut TaskThreadData) {
...
+                } else {
+                    (*f).task_thread.task_head
+                };
+                while !t.is_null() {
+                    if (*t).type_0 as c_uint == RAV1D_TASK_TYPE_INIT_CDF as c_int as c_uint {
+                        // goto next;
+                        prev_t = t;
+                        t = (*t).next;
+                        (*f).task_thread.task_cur_prev = prev_t;
+                        continue;
```
kkysen, line 912: I think it'd be helpful to have every `continue` and `break` labeled in this `fn` due to the sheer complexity.

src/thread_task.rs
```diff
@@ -813,869 +813,746 @@ unsafe fn delayed_fg_task(c: *const Rav1dContext, ttd: *mut TaskThreadData) {
...
+                    }
+                    RAV1D_TASK_TYPE_ENTROPY_PROGRESS => {
+                        // dummy to convert tile progress to frame
+                    }
+                    _ => {
+                        abort();
+                    }
+                }
+                // exit fallthrough loop
+                break;
```
kkysen, line 1432:
```suggestion
                break 'fallthrough;
```

src/thread_task.rs
```diff
@@ -813,869 +813,746 @@ unsafe fn delayed_fg_task(c: *const Rav1dContext, ttd: *mut TaskThreadData) {
...
+                    } else {
+                        Ok(())
+                    },
+                );
+                pthread_cond_signal(&mut (*f).task_thread.cond);
             }
+            reset_task_cur(c, ttd, (*t).frame_idx);
+
+            // exit found_unlocked
+            break;
```
kkysen, line 1552:
```suggestion
            break 'found_unlocked;
```

src/thread_task.rs
```diff
@@ -813,869 +813,746 @@ unsafe fn delayed_fg_task(c: *const Rav1dContext, ttd: *mut TaskThreadData) {
...
+                            (p1 == TILE_ERROR) as c_int,
+                        );
+                        found = true;
+                        break;
+                    }
+                }
+            }
+        }
+        if !found {
+            'decoding: while (*ttd).cur < (*c).n_fc {
```
kkysen, line 894: If you add back some of the comments in a `fn`, type, etc., could you add back all of them?  E.x. here there's a `// run decoding tasks last`.  That way it's easy to tell if all the comments have been added back or if we should look in the C for them.

src/thread_task.rs
```diff
@@ -813,869 +813,746 @@ unsafe fn delayed_fg_task(c: *const Rav1dContext, ttd: *mut TaskThreadData) {
...
+        {
+            (*ttd).cur = ((*ttd).cur).wrapping_add(1);
+        }
+        (*t).next = 0 as *mut Rav1dTask;
+        // we don't need to check cond_signaled here, since we found a task
+        // after the last signal so we want to re-signal the next waiting thread
+        // and again won't need to signal after that
+        ::core::intrinsics::atomic_store_seqcst(&mut (*ttd).cond_signaled, 1 as c_int);
+        pthread_cond_signal(&mut (*ttd).cond);
+        pthread_mutex_unlock(&mut (*ttd).lock);
```
kkysen, line 1052:
```suggestion
        '_found: {
            // remove t from list
            if !prev_t.is_null() {
                (*prev_t).next = (*t).next;
            } else {
                (*f).task_thread.task_head = (*t).next;
            }
            if ((*t).next).is_null() {
                (*f).task_thread.task_tail = prev_t;
            }
            if (*t).type_0 as c_uint > RAV1D_TASK_TYPE_INIT_CDF as c_int as c_uint
                && ((*f).task_thread.task_head).is_null()
            {
                (*ttd).cur = ((*ttd).cur).wrapping_add(1);
            }
            (*t).next = 0 as *mut Rav1dTask;
            // we don't need to check cond_signaled here, since we found a task
            // after the last signal so we want to re-signal the next waiting thread
            // and again won't need to signal after that
            ::core::intrinsics::atomic_store_seqcst(&mut (*ttd).cond_signaled, 1 as c_int);
            pthread_cond_signal(&mut (*ttd).cond);
            pthread_mutex_unlock(&mut (*ttd).lock);
        }
```

Writing `goto` labels as labeled Rust blocks helps with clarity I think and matching it with the C (vs. just a comment).

src/thread_task.rs
```diff
@@ -813,869 +813,746 @@ unsafe fn delayed_fg_task(c: *const Rav1dContext, ttd: *mut TaskThreadData) {
...
+                            let p2 = ::core::intrinsics::atomic_load_seqcst(
+                                &mut *((*ts).progress).as_mut_ptr().offset(p as isize)
+                                    as *mut atomic_int,
+                            );
+                            if p2 < (*t).recon_progress {
+                                // goto next;
+                                prev_t = t;
+                                t = (*t).next;
+                                (*f).task_thread.task_cur_prev = prev_t;
+                                continue 'decoding;
```
kkysen, line 969: I think this should be a `continue` for the inner `while !t.is_null()` (yet to be labeled) loop, as the other `goto next`s are.

kkysen: Also, instead of inlining `goto next` everywhere, I think you can have a labeled block inside the `while !t.is_null()` loop, e.x. `'next`, use `break 'next` for the `goto next`s, and have the `next:` code following the `'next` block.  Or maybe name the `'next` block `'before_next`, and then have a `'_next'` block after for the `next:` code.

kkysen (changes requested):
I think https://github.com/memorysafety/rav1d/pull/671/files#r1449731224 is the only thing that's wrong, though I made some other suggestions that would IMO greatly help readability.

src/thread_task.rs
rinon: Meant to, missed that one.

src/thread_task.rs
rinon: Good catch. Surprised this didn't actually make a difference in testing. And yeah, I like the idea of a block, I'll add that.

src/thread_task.rs
rinon: I'd prefer not to add labeled blocks unless we're actually using them. A comment is just as readable but doesn't require indenting the inner code. For the 'next block where we have shared code, that's great, but here we wouldn't actually break out of the block early.

rinon:
Thanks for the useful feedback, I've applied everything except for the labeled block for `found`, just want to avoid extra indentation if we can. I forgot that labelled blocks were added, but that cleans up the `goto next` code perfectly, thanks.

src/thread_task.rs
```diff
@@ -813,869 +813,737 @@ unsafe fn delayed_fg_task(c: *const Rav1dContext, ttd: *mut TaskThreadData) {
...
-                                &mut (*ttd).cond_signaled,
-                                1 as c_int,
-                            );
-                            pthread_cond_signal(&mut (*ttd).cond);
-                            pthread_mutex_unlock(&mut (*ttd).lock);
-                            current_block = 8464383504555462953;
-                            break;
                         }
                     }
+                    prev_t = t;
```
kkysen, line 1006:
```suggestion
                    // next:
                    prev_t = t;
```

I think the `next:` comment you had before is still helpful for mapping back to the C.

kkysen (approved):
There's that one `// next:` comment I think you should add back, but otherwise everything LGTM now.  I slightly prefer actual labeled blocks vs. a `// label:` comment, but the indentation isn't great either, so it's fine as is.

PR #672, opened 2024-01-10
`struct Rav1dFrameContext_task_thread`: Replace atomic intrinsics

rinon:


src/internal.rs
```diff
@@ -430,12 +430,12 @@ pub(crate) struct Rav1dFrameContext_task_thread {
...
     pub num_tile_tasks: c_int,
-    pub init_done: atomic_int,
-    pub done: [atomic_int; 2],
+    pub init_done: AtomicI32,
+    pub done: [AtomicI32; 2],
     pub retval: Rav1dResult,
     pub update_set: bool, // whether we need to update CDF reference
-    pub error: atomic_int,
-    pub task_counter: atomic_int,
+    pub error: AtomicI32,
```
kkysen, line 437: I think we want this to be an `Atomic<Rav1dError>`.  We can `#[derive(Atom)]` on `enum Rav1dError` (using `atomig`).

kkysen: It might be fine to just do that later, though.

src/internal.rs
```diff
@@ -516,7 +516,7 @@ pub(crate) struct Rav1dFrameContext {
     pub lf: Rav1dFrameContext_lf,
     pub task_thread: Rav1dFrameContext_task_thread,
     pub tile_thread: FrameTileThreadData,
-}
+    }
```
kkysen, line 519: Typo?

kkysen (approved):
LGTM besides that indentation typo.

src/internal.rs
rinon: Definitely at some point, but that isn't the priority now or in the near future.

rinon:
Added another atomic_i32 -> AtomicI32 commit, so worth a quick additional look.

kkysen (approved):
Still looks good with the `pending_tasks_merge` change now, but what was the hoist from `Rav1dFrameContext_task_thread_pending_tasks` to `Rav1dFrameContext_task_thread` for, (for some mutability reasons?)?  Can you add that to the PR description or commit message?

rinon:
Updated the commit message. I moved it out in preparation for making the `pending_tasks` protected by a Mutex. The merge field was accessed outside the critical section (and it's atomic, so that's fine), so I needed to move it out of the struct that I planned to put into a Mutex.

PR #673, opened 2024-01-11
`struct TaskThreadData`: Replace pthread mutexes with Rust equivalents

rinon:
Replaces the locking in `Rav1dFrameContext_task_thread.pending_tasks` and `TaskThreadData` with Rust equivalents. `TaskThreadData` can now be accessed as read-only with interior mutability as needed.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -220,10 +220,6 @@ use crate::src::warpmv::rav1d_get_shear_params;
 use crate::src::warpmv::rav1d_set_affine_mv2d;
 use libc::free;
 use libc::malloc;
-use libc::pthread_cond_signal;
-use libc::pthread_cond_wait;
-use libc::pthread_mutex_lock;
-use libc::pthread_mutex_unlock;
 use libc::ptrdiff_t;
 use libc::uintptr_t;
 use std::array;
@@ -237,6 +233,7 @@ use std::ptr;
 use std::ptr::addr_of_mut;
 use std::slice;
 use std::sync::atomic::AtomicI32;
+use std::sync::atomic::AtomicU32;
 use std::sync::atomic::Ordering;
 
 #[cfg(feature = "bitdepth_8")]
@@ -4902,8 +4899,8 @@ pub(crate) unsafe fn rav1d_decode_frame(f: &mut Rav1dFrameContext) -> Rav1dResul
     if res.is_ok() {
         if (*f.c).n_tc > 1 {
             res = rav1d_task_create_tile_sbrow(f, 0, 1);
-            pthread_mutex_lock(&mut (*f.task_thread.ttd).lock);
-            pthread_cond_signal(&mut (*f.task_thread.ttd).cond);
+            let mut task_thread_lock = (*f.task_thread.ttd).delayed_fg.lock().unwrap();
+            (*f.task_thread.ttd).cond.notify_one();
             if res.is_ok() {
                 while f.task_thread.done[0].load(Ordering::Relaxed) == 0
                 // TODO(kkysen) Make `.task_counter` an `AtomicI32`, but that requires recursively removing `impl Copy`s.
@@ -4911,10 +4908,10 @@ pub(crate) unsafe fn rav1d_decode_frame(f: &mut Rav1dFrameContext) -> Rav1dResul
                         .load(Ordering::SeqCst)
                         > 0
                 {
-                    pthread_cond_wait(&mut f.task_thread.cond, &mut (*f.task_thread.ttd).lock);
+                    task_thread_lock = f.task_thread.cond.wait(task_thread_lock).unwrap();
                 }
             }
-            pthread_mutex_unlock(&mut (*f.task_thread.ttd).lock);
+            drop(task_thread_lock);
             res = f.task_thread.retval;
         } else {
             res = rav1d_decode_frame_main(f);
@@ -4940,8 +4937,8 @@ fn get_upscale_x0(in_w: c_int, out_w: c_int, step: c_int) -> c_int {
 
 pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
     // wait for c->out_delayed[next] and move into c->out if visible
-    let (f, out) = if c.n_fc > 1 {
-        pthread_mutex_lock(&mut c.task_thread.lock);
+    let (f, out, _task_thread_lock) = if c.n_fc > 1 {
+        let mut task_thread_lock = c.task_thread.delayed_fg.lock().unwrap();
         let next = c.frame_thread.next;
         c.frame_thread.next += 1;
         if c.frame_thread.next == c.n_fc {
@@ -4950,7 +4947,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
 
         let f = &mut *c.fc.offset(next as isize);
         while !f.tiles.is_empty() {
-            pthread_cond_wait(&mut f.task_thread.cond, &mut c.task_thread.lock);
+            task_thread_lock = f.task_thread.cond.wait(task_thread_lock).unwrap();
         }
         let out_delayed = &mut *c.frame_thread.out_delayed.offset(next as isize);
         if !out_delayed.p.data.data[0].is_null() || f.task_thread.error.load(Ordering::SeqCst) != 0
@@ -4967,8 +4964,10 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
                 Ordering::SeqCst,
                 Ordering::SeqCst,
             );
-            if c.task_thread.cur != 0 && c.task_thread.cur < c.n_fc {
-                c.task_thread.cur -= 1;
+            // `cur` is not actually mutated from multiple threads concurrently
+            let cur = c.task_thread.cur.load(Ordering::Relaxed);
+            if cur != 0 && cur < c.n_fc {
+                c.task_thread.cur = AtomicU32::new(cur - 1);
             }
         }
         let error = f.task_thread.retval;
@@ -4985,9 +4984,9 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
             }
             rav1d_thread_picture_unref(out_delayed);
         }
-        (f, out_delayed as *mut _)
+        (f, out_delayed as *mut _, Some(task_thread_lock))
     } else {
-        (&mut *c.fc, &mut c.out as *mut _)
+        (&mut *c.fc, &mut c.out as *mut _, None)
     };
 
     f.seq_hdr = c.seq_hdr.clone();
@@ -4997,11 +4996,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
 
     let bpc = 8 + 2 * seq_hdr.hbd;
 
-    unsafe fn on_error(
-        f: &mut Rav1dFrameContext,
-        c: &mut Rav1dContext,
-        out: *mut Rav1dThreadPicture,
-    ) {
+    unsafe fn on_error(f: &mut Rav1dFrameContext, c: &Rav1dContext, out: *mut Rav1dThreadPicture) {
         f.task_thread.error = AtomicI32::new(1);
         rav1d_cdf_thread_unref(&mut f.in_cdf);
         if f.frame_hdr.as_ref().unwrap().refresh_context != 0 {
@@ -5019,13 +5014,9 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
         rav1d_ref_dec(&mut f.mvs_ref);
         let _ = mem::take(&mut f.seq_hdr);
         let _ = mem::take(&mut f.frame_hdr);
-        *c.cached_error_props.get_mut().unwrap() = c.in_0.m.clone();
+        *c.cached_error_props.lock().unwrap() = c.in_0.m.clone();
 
         f.tiles.clear();
-
-        if c.n_fc > 1 {
-            pthread_mutex_unlock(&mut c.task_thread.lock);
-        }
     }
 
     // TODO(kkysen) Rather than lazy initializing this,
@@ -5149,7 +5140,11 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
     mem::swap(&mut f.tiles, &mut c.tiles);
 
     // allocate frame
-    let res = rav1d_thread_picture_alloc(c, f, bpc);
+
+    // We must take itut_t35 out of the context before the call so borrowck can
+    // see we mutably borrow `c.itut_t35` disjointly from the task thread lock.
+    let itut_t35 = c.itut_t35.take();
+    let res = rav1d_thread_picture_alloc(c, f, bpc, itut_t35);
     if res.is_err() {
         on_error(f, c, out);
         return res;
@@ -5363,7 +5358,6 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
         }
     } else {
         rav1d_task_frame_init(f);
-        pthread_mutex_unlock(&mut c.task_thread.lock);
     }
 
     Ok(())
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4940,7 +4940,7 @@ fn get_upscale_x0(in_w: c_int, out_w: c_int, step: c_int) -> c_int {
 
 pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
     // wait for c->out_delayed[next] and move into c->out if visible
-    let (f, out_delayed) = if c.n_fc > 1 {
+    let (f, out) = if c.n_fc > 1 {
         pthread_mutex_lock(&mut c.task_thread.lock);
         let next = c.frame_thread.next;
         c.frame_thread.next += 1;
@@ -4987,7 +4987,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
         }
         (f, out_delayed as *mut _)
     } else {
-        (&mut *c.fc, ptr::null_mut())
+        (&mut *c.fc, &mut c.out as *mut _)
     };
 
     f.seq_hdr = c.seq_hdr.clone();
@@ -5000,7 +5000,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
     unsafe fn on_error(
         f: &mut Rav1dFrameContext,
         c: &mut Rav1dContext,
-        out_delayed: *mut Rav1dThreadPicture,
+        out: *mut Rav1dThreadPicture,
     ) {
         f.task_thread.error = AtomicI32::new(1);
         rav1d_cdf_thread_unref(&mut f.in_cdf);
@@ -5013,11 +5013,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
             }
             rav1d_ref_dec(&mut f.ref_mvs_ref[i]);
         }
-        if c.n_fc == 1 {
-            rav1d_thread_picture_unref(&mut c.out);
-        } else {
-            rav1d_thread_picture_unref(out_delayed);
-        }
+        rav1d_thread_picture_unref(out);
         rav1d_picture_unref_internal(&mut f.cur);
         rav1d_thread_picture_unref(&mut f.sr_cur);
         rav1d_ref_dec(&mut f.mvs_ref);
@@ -5066,7 +5062,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
                     "Compiled without support for {}-bit decoding",
                     8 + 2 * seq_hdr.hbd
                 );
-                on_error(f, c, out_delayed);
+                on_error(f, c, out);
                 return Err(ENOPROTOOPT);
             }
         }
@@ -5093,7 +5089,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
         if frame_hdr.primary_ref_frame != RAV1D_PRIMARY_REF_NONE {
             let pri_ref = frame_hdr.refidx[frame_hdr.primary_ref_frame as usize] as usize;
             if c.refs[pri_ref].p.p.data.data[0].is_null() {
-                on_error(f, c, out_delayed);
+                on_error(f, c, out);
                 return Err(EINVAL);
             }
         }
@@ -5110,7 +5106,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
                 for j in 0..i {
                     rav1d_thread_picture_unref(&mut f.refp[j]);
                 }
-                on_error(f, c, out_delayed);
+                on_error(f, c, out);
                 return Err(EINVAL);
             }
             rav1d_thread_picture_ref(&mut f.refp[i], &mut c.refs[refidx].p);
@@ -5143,7 +5139,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
     if frame_hdr.refresh_context != 0 {
         let res = rav1d_cdf_thread_alloc(c, &mut f.out_cdf, (c.n_fc > 1) as c_int);
         if res.is_err() {
-            on_error(f, c, out_delayed);
+            on_error(f, c, out);
             return res;
         }
     }
@@ -5155,7 +5151,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
     // allocate frame
     let res = rav1d_thread_picture_alloc(c, f, bpc);
     if res.is_err() {
-        on_error(f, c, out_delayed);
+        on_error(f, c, out);
         return res;
     }
 
@@ -5165,7 +5161,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
     if frame_hdr.size.width[0] != frame_hdr.size.width[1] {
         let res = rav1d_picture_alloc_copy(c, &mut f.cur, frame_hdr.size.width[0], &mut f.sr_cur.p);
         if res.is_err() {
-            on_error(f, c, out_delayed);
+            on_error(f, c, out);
             return res;
         }
     } else {
@@ -5188,7 +5184,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
             c.event_flags |= f.sr_cur.flags.into();
         }
     } else {
-        rav1d_thread_picture_ref(out_delayed, &mut f.sr_cur);
+        rav1d_thread_picture_ref(out, &mut f.sr_cur);
     }
 
     f.w4 = frame_hdr.size.width[0] + 3 >> 2;
@@ -5220,7 +5216,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
                 * (f.b4_stride >> 1) as usize,
         );
         if f.mvs_ref.is_null() {
-            on_error(f, c, out_delayed);
+            on_error(f, c, out);
             return Err(ENOMEM);
         }
         f.mvs = (*f.mvs_ref).data.cast::<refmvs_temporal_block>();
@@ -5288,7 +5284,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
             );
             if f.cur_segmap_ref.is_null() {
                 rav1d_ref_dec(&mut f.prev_segmap_ref);
-                on_error(f, c, out_delayed);
+                on_error(f, c, out);
                 return Err(ENOMEM);
             }
             f.cur_segmap = (*f.cur_segmap_ref).data.cast::<u8>();
@@ -5304,7 +5300,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
                 ::core::mem::size_of::<u8>() * f.b4_stride as usize * 32 * f.sb128h as usize;
             f.cur_segmap_ref = rav1d_ref_create_using_pool(c.segmap_pool, segmap_size);
             if f.cur_segmap_ref.is_null() {
-                on_error(f, c, out_delayed);
+                on_error(f, c, out);
                 return Err(ENOMEM);
             }
             f.cur_segmap = (*f.cur_segmap_ref).data.cast::<u8>();
@@ -5362,7 +5358,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
                     rav1d_ref_dec(&mut c.refs[i].refmvs);
                 }
             }
-            on_error(f, c, out_delayed);
+            on_error(f, c, out);
             return res;
         }
     } else {
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4975,7 +4975,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
         if error.is_err() {
             f.task_thread.retval = Ok(());
             c.cached_error = error;
-            c.cached_error_props = out_delayed.p.m.clone();
+            *c.cached_error_props.get_mut().unwrap() = out_delayed.p.m.clone();
             rav1d_thread_picture_unref(out_delayed);
         } else if !out_delayed.p.data.data[0].is_null() {
             let progress = out_delayed.progress.as_ref().unwrap()[1].load(Ordering::Relaxed);
@@ -5023,7 +5023,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
         rav1d_ref_dec(&mut f.mvs_ref);
         let _ = mem::take(&mut f.seq_hdr);
         let _ = mem::take(&mut f.frame_hdr);
-        c.cached_error_props = c.in_0.m.clone();
+        *c.cached_error_props.get_mut().unwrap() = c.in_0.m.clone();
 
         f.tiles.clear();
 
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -253,7 +253,7 @@ pub struct Rav1dContext {
     pub(crate) drain: c_int,
     pub(crate) frame_flags: Atomic<PictureFlags>,
     pub(crate) event_flags: Rav1dEventFlags,
-    pub(crate) cached_error_props: Rav1dDataProps,
+    pub(crate) cached_error_props: Mutex<Rav1dDataProps>,
     pub(crate) cached_error: Rav1dResult,
 
     pub(crate) logger: Option<Rav1dLogger>,
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -73,7 +73,6 @@ use crate::src::refmvs::refmvs_tile;
 use crate::src::refmvs::Rav1dRefmvsDSPContext;
 use crate::src::thread_data::thread_data;
 use atomig::Atomic;
-use libc::pthread_cond_t;
 use libc::pthread_mutex_t;
 use libc::ptrdiff_t;
 use std::ffi::c_int;
@@ -82,6 +81,7 @@ use std::ptr;
 use std::sync::atomic::AtomicI32;
 use std::sync::atomic::AtomicU32;
 use std::sync::Arc;
+use std::sync::Condvar;
 use std::sync::Mutex;
 
 #[repr(C)]
@@ -158,27 +158,32 @@ impl BitDepthDependentType for Grain {
 #[repr(C)]
 pub(crate) struct TaskThreadData_delayed_fg {
     pub exec: c_int,
-    pub cond: pthread_cond_t,
     pub in_0: *const Rav1dPicture,
     pub out: *mut Rav1dPicture,
     pub type_0: TaskType,
-    pub progress: [AtomicI32; 2], /* [0]=started, [1]=completed */
     pub grain: BitDepthUnion<Grain>,
 }
 
 #[repr(C)]
 pub(crate) struct TaskThreadData {
-    pub lock: pthread_mutex_t,
-    pub cond: pthread_cond_t,
+    pub cond: Condvar,
     pub first: AtomicU32,
-    pub cur: c_uint,
+    pub cur: AtomicU32,
     /// This is used for delayed reset of the task cur pointer when
     /// such operation is needed but the thread doesn't enter a critical
     /// section (typically when executing the next sbrow task locklessly).
     /// See [`crate::src::thread_task::reset_task_cur`].
     pub reset_task_cur: AtomicU32,
     pub cond_signaled: AtomicI32,
-    pub delayed_fg: TaskThreadData_delayed_fg,
+    pub delayed_fg_progress: [AtomicI32; 2], /* [0]=started, [1]=completed */
+    pub delayed_fg_cond: Condvar,
+    /// This lock has a dual purpose - protecting the delayed_fg structure, as
+    /// well as synchronizing tasks across threads. Many cases do not use the
+    /// inner data when holding the lock but instead use it to sequence
+    /// operations. Rather than disentagle these related uses in the original C
+    /// code, we have kept a single mutex and put the delayed_fg structure into
+    /// it.
+    pub delayed_fg: Mutex<TaskThreadData_delayed_fg>,
     pub inited: c_int,
 }
 
@@ -426,7 +431,7 @@ impl Default for Rav1dFrameContext_task_thread_pending_tasks {
 #[repr(C)]
 pub(crate) struct Rav1dFrameContext_task_thread {
     pub lock: pthread_mutex_t,
-    pub cond: pthread_cond_t,
+    pub cond: Condvar,
     pub ttd: *mut TaskThreadData,
     pub tasks: *mut Rav1dTask,
     pub tile_tasks: [*mut Rav1dTask; 2],
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -78,9 +78,11 @@ use libc::pthread_mutex_t;
 use libc::ptrdiff_t;
 use std::ffi::c_int;
 use std::ffi::c_uint;
+use std::ptr;
 use std::sync::atomic::AtomicI32;
 use std::sync::atomic::AtomicU32;
 use std::sync::Arc;
+use std::sync::Mutex;
 
 #[repr(C)]
 pub(crate) struct Rav1dDSPContext {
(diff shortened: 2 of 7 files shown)


src/cdf.rs
```diff
@@ -5653,7 +5653,7 @@ pub unsafe fn rav1d_cdf_thread_copy(dst: *mut CdfContext, src: *const CdfThreadC
 }
 
 pub unsafe fn rav1d_cdf_thread_alloc(
-    c: *mut Rav1dContext,
+    c: &Rav1dContext,
```
kkysen, line 5656: When you make these things into references, can you also update all of the `(*c)`s in the `fn` to just `c`s?  It's not strictly necessary, but it makes reading the `fn` and cleaning it up further a lot easier.

src/decode.rs
```diff
@@ -4939,8 +4935,8 @@ fn get_upscale_x0(in_w: c_int, out_w: c_int, step: c_int) -> c_int {
 
 pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
     // wait for c->out_delayed[next] and move into c->out if visible
-    let (f, out_delayed) = if c.n_fc > 1 {
-        pthread_mutex_lock(&mut c.task_thread.lock);
+    let (f, out_delayed, guard) = if c.n_fc > 1 {
+        let mut task_thread = c.task_thread.lock.lock().unwrap();
```
kkysen, line 4939: Why is this one named `task_thread` vs. `guard`?

src/decode.rs
```diff
@@ -4964,8 +4960,10 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
                 first,
                 u32::MAX,
             );
-            if c.task_thread.cur != 0 && c.task_thread.cur < c.n_fc {
-                c.task_thread.cur -= 1;
+            if c.task_thread.cur.load(Ordering::Relaxed) != 0
+                && c.task_thread.cur.load(Ordering::Relaxed) < c.n_fc
```
kkysen, line 4964: Can we just do one load here?

src/decode.rs
```diff
@@ -4964,8 +4960,10 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
...
+            {
+                c.task_thread.cur.fetch_sub(1, Ordering::Relaxed);
```
kkysen, line 4966: Is the `Ordering::Relaxed` because it previously wasn't atomic and so shouldn't really need atomic synchronization, and `Ordering::Relaxed` is the least strict `Ordering`?  Previously (e.x. for `Abcd` in #662) I just used `Ordering::SeqCst` as the default for all atomics.  Should I do `Ordering::Relaxed` for these kinds on previously not atomic operations instead?

src/decode.rs
```diff
@@ -5071,6 +5065,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
                     "Compiled without support for {}-bit decoding",
                     8 + 2 * (*f.seq_hdr).hbd
                 );
+                drop(guard);
                 on_error(f, c, out_delayed);
                 return Err(ENOPROTOOPT);
```
kkysen, line 5070: I don't think any of these `drop(guard)`s immediately before `on_error(f, c, out_delayed)` and `return ...` are necessary, since `pthread_mutex_unlock` was called as the last statement in `on_error` and so that corresponds to the automatic drop after `return`ing.

src/internal.rs
```diff
@@ -155,27 +158,30 @@ impl BitDepthDependentType for Grain {
...
     pub in_0: *const Rav1dPicture,
     pub out: *mut Rav1dPicture,
     pub type_0: TaskType,
-    pub progress: [atomic_int; 2], /* [0]=started, [1]=completed */
     pub grain: BitDepthUnion<Grain>,
 }
 
+pub(crate) struct TaskThreadData_inner {
+    pub delayed_fg: TaskThreadData_delayed_fg,
+}
```
kkysen, line 169: For new types we can use `TitleCase`.

kkysen: Also, what's the point of `TaskThreadData_inner` vs. just using `TaskThreadData_delayed_fg`?

src/internal.rs
```diff
@@ -155,27 +158,30 @@ impl BitDepthDependentType for Grain {
...
-    pub cur: c_uint,
+    pub cur: AtomicU32,
     /// This is used for delayed reset of the task cur pointer when
     /// such operation is needed but the thread doesn't enter a critical
     /// section (typically when executing the next sbrow task locklessly).
     /// See [`crate::src::thread_task::reset_task_cur`].
     pub reset_task_cur: atomic_uint,
     pub cond_signaled: atomic_int,
-    pub delayed_fg: TaskThreadData_delayed_fg,
+    pub delayed_fg_progress: [atomic_int; 2], /* [0]=started, [1]=completed */
```
kkysen, line 183: As I've refactored things, I've moved these kinds of comments to doc comments.  Not too important, but helpful and it seemed small and fine when I was already changing the field in the first place.

src/lib.rs
```diff
@@ -350,23 +347,13 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
         ::core::mem::size_of::<Rav1dTaskContext>().wrapping_mul((*c).n_tc as usize),
     );
     if (*c).n_tc > 1 as c_uint {
-        if pthread_mutex_init(&mut (*c).task_thread.lock, 0 as *const pthread_mutexattr_t) != 0 {
-            return error(c, c_out, &mut thread_attr);
```
kkysen, line 354: In Rust, `Mutex`, `Condvar`, etc. can never fail on creation, right?

src/lib.rs
```diff
@@ -637,19 +602,21 @@ unsafe fn drain_picture(c: &mut Rav1dContext, out: &mut Rav1dPicture) -> Rav1dRe
...
+            if c.task_thread.cur.load(Ordering::Relaxed) != 0
+                && c.task_thread.cur.load(Ordering::Relaxed) < c.n_fc
+            {
+                c.task_thread.cur.fetch_sub(1, Ordering::Relaxed);
             }
             drained = 1 as c_int;
         } else if drained != 0 {
-            pthread_mutex_unlock(&mut c.task_thread.lock);
+            drop(task_thread);
             break;
```
kkysen, line 613: The `break` will `drop` it automatically.

src/lib.rs
```diff
@@ -995,13 +956,9 @@ unsafe fn close_internal(c_out: &mut *mut Rav1dContext, flush: c_int) {
                     break;
                 }
                 pthread_join((*pf).task_thread.td.thread, 0 as *mut *mut c_void);
-                pthread_cond_destroy(&mut (*pf).task_thread.td.cond);
                 pthread_mutex_destroy(&mut (*pf).task_thread.td.lock);
                 n_0 = n_0.wrapping_add(1);
             }
-            pthread_cond_destroy(&mut (*ttd).delayed_fg.cond);
-            pthread_cond_destroy(&mut (*ttd).cond);
-            pthread_mutex_destroy(&mut (*ttd).lock);
```
kkysen, line 1004: Where are these dropped?

src/lib.rs
```diff
@@ -1023,8 +980,7 @@ unsafe fn close_internal(c_out: &mut *mut Rav1dContext, flush: c_int) {
             freep(&mut (*f).frame_thread.cbi as *mut *mut CodedBlockInfo as *mut c_void);
         }
         if (*c).n_tc > 1 as c_uint {
-            pthread_mutex_destroy(&mut (*f).task_thread.pending_tasks.lock);
-            pthread_cond_destroy(&mut (*f).task_thread.cond);
+            mem::take(&mut (*f).task_thread.pending_tasks); // TODO: remove when context is owned
```
kkysen, line 983:
```suggestion
            let _ = mem::take(&mut (*f).task_thread.pending_tasks); // TODO: remove when context is owned
```

I've been using `let _ = mem::take` as a searchable idiom for explicitly dropping a `Default`able field in a type that's not yet correctly dropped.

src/lib.rs
kkysen: Or does it not matter at all since `{Mutex,Condvar}::new` are `const fn`s, which I think implies they don't have `Drop` `impl`s?

src/decode.rs
kkysen: And the same for the few other place this is done.

src/thread_task.rs
```diff
@@ -55,13 +53,16 @@ use std::process::abort;
...
 use std::sync::atomic::Ordering;
+use std::sync::MutexGuard;
 
 #[cfg(target_os = "linux")]
 use libc::prctl;
 
 #[cfg(target_os = "macos")]
 use libc::pthread_setname_np;
 
+use super::internal::TaskThreadData_inner;
```
kkysen, line 64: Can you use `"rust-analyzer.imports.prefix": "crate"` (and `"rust-analyzer.imports.granularity.group": "item"`) for imports?  I've been avoiding `use super` (except for a few `use super::*`s for inline `mod`s) and also grouping all of the imports together in one block so they're all sorted together.

src/thread_task.rs
```diff
@@ -78,6 +79,18 @@ unsafe fn rav1d_set_thread_name(name: *const c_char) {
...
+// insert some tasks in the queue).
+// When frame_idx is set, it can be either from a completed task, or from tasks
+// inserted in the queue, in which case we have to make sure the cur pointer
+// isn't past this insert.
+// The special case where frame_idx is UINT_MAX is to handle the reset after
+// completing a task and locklessly signaling progress. In this case we don't
+// enter a critical section, which is needed for this function, so we set an
+// atomic for a delayed handling, happening here. Meaning we can call this
+// function without any actual update other than what's in the atomic, hence
+// this special case.
```
kkysen, line 93:
```suggestion
/// This function resets the cur pointer to the first frame theoretically
/// executable after a task completed (ie. each time we update some progress or
/// insert some tasks in the queue).
/// When frame_idx is set, it can be either from a completed task, or from tasks
/// inserted in the queue, in which case we have to make sure the cur pointer
/// isn't past this insert.
/// The special case where frame_idx is UINT_MAX is to handle the reset after
/// completing a task and locklessly signaling progress. In this case we don't
/// enter a critical section, which is needed for this function, so we set an
/// atomic for a delayed handling, happening here. Meaning we can call this
/// function without any actual update other than what's in the atomic, hence
/// this special case.
```

src/thread_task.rs
```diff
@@ -310,32 +330,33 @@ unsafe fn insert_task(f: *mut Rav1dFrameContext, t: *mut Rav1dTask, cond_signal:
...
-        (*f).task_thread.pending_tasks.tail = 0 as *mut Rav1dTask;
-        (*f).task_thread
-            .pending_tasks_merge
-            .store(0, Ordering::SeqCst);
-        pthread_mutex_unlock(&mut (*f).task_thread.pending_tasks.lock);
+        let mut pending_tasks = (*f).task_thread.pending_tasks.lock().unwrap();
+        let mut t = {
+            let old_head = pending_tasks.head;
+            pending_tasks.head = 0 as *mut Rav1dTask;
+            pending_tasks.tail = 0 as *mut Rav1dTask;
```
kkysen, line 354:
```suggestion
            *pending_tasks = Default::default();
```

src/thread_task.rs
```diff
@@ -310,32 +330,33 @@ unsafe fn insert_task(f: *mut Rav1dFrameContext, t: *mut Rav1dTask, cond_signal:
...
     let merge = (*f).task_thread.pending_tasks_merge.load(Ordering::SeqCst);
     if merge != 0 {
-        pthread_mutex_lock(&mut (*f).task_thread.pending_tasks.lock);
-        let mut t: *mut Rav1dTask = (*f).task_thread.pending_tasks.head;
-        (*f).task_thread.pending_tasks.head = 0 as *mut Rav1dTask;
-        (*f).task_thread.pending_tasks.tail = 0 as *mut Rav1dTask;
-        (*f).task_thread
-            .pending_tasks_merge
-            .store(0, Ordering::SeqCst);
-        pthread_mutex_unlock(&mut (*f).task_thread.pending_tasks.lock);
```
kkysen, line 338: `pending_tasks` was previously `unlock`ed/`drop`ped earlier, immediately after `t` is fetched from `pending_tasks`.  We should match that behavior, right?  This does access `t` through a `*mut` after the lock is dropped, though, which I'm not sure will work once `pending_tasks` is made safe, though I think it will since it's removed from the linked list (?) and thus isn't accessible through the shared `pending_tasks` anymore.

src/thread_task.rs
```diff
@@ -478,21 +499,22 @@ pub(crate) unsafe fn rav1d_task_create_tile_sbrow(
...
-    if ((*f).task_thread.pending_tasks.head).is_null() {
-        (*f).task_thread.pending_tasks.head = &mut *tasks.offset(0) as *mut Rav1dTask;
-    } else {
-        (*(*f).task_thread.pending_tasks.tail).next = &mut *tasks.offset(0) as *mut Rav1dTask;
+    {
+        let mut pending_tasks = (*f).task_thread.pending_tasks.lock().unwrap();
+        if !(pending_tasks.head.is_null() || pass == 2) {
+            unreachable!();
+        }
+        if (pending_tasks.head).is_null() {
```
kkysen, line 507:
```suggestion
        if pending_tasks.head.is_null() {
```

src/thread_task.rs
```diff
@@ -478,21 +499,22 @@ pub(crate) unsafe fn rav1d_task_create_tile_sbrow(
...
     (*f).task_thread.done[(pass & 1) as usize].store(0, Ordering::SeqCst);
-    pthread_mutex_lock(&mut (*f).task_thread.pending_tasks.lock);
-    if !(((*f).task_thread.pending_tasks.head).is_null() || pass == 2) {
-        unreachable!();
-    }
-    if ((*f).task_thread.pending_tasks.head).is_null() {
-        (*f).task_thread.pending_tasks.head = &mut *tasks.offset(0) as *mut Rav1dTask;
-    } else {
-        (*(*f).task_thread.pending_tasks.tail).next = &mut *tasks.offset(0) as *mut Rav1dTask;
+    {
```
kkysen, line 502: Why do we need this block when the function ends right after the end of the block?

src/thread_task.rs
```diff
@@ -514,16 +536,15 @@ pub(crate) unsafe fn rav1d_task_delayed_fg(
     in_0: *const Rav1dPicture,
 ) {
     let ttd: *mut TaskThreadData = &mut (*c).task_thread;
```
kkysen, line 538: You should make this a ref first, since mutability isn't being checked by the compiler right now.

kkysen, line 538: You should make this a ref first, since I don't think mutability through it is checked yet.

src/thread_task.rs
```diff
@@ -514,16 +536,15 @@ pub(crate) unsafe fn rav1d_task_delayed_fg(
...
-    pthread_mutex_unlock(&mut (*ttd).lock);
+    let mut ttd_locked = (*ttd).lock.lock().unwrap();
+    ttd_locked.delayed_fg.in_0 = in_0;
+    ttd_locked.delayed_fg.out = out;
+    ttd_locked.delayed_fg.type_0 = RAV1D_TASK_TYPE_FG_PREP;
+    *(&mut *((*ttd).delayed_fg_progress).as_mut_ptr().offset(0) as *mut atomic_int) = 0 as c_int;
+    *(&mut *((*ttd).delayed_fg_progress).as_mut_ptr().offset(1) as *mut atomic_int) = 0 as c_int;
+    ttd_locked.delayed_fg.exec = 1 as c_int;
+    (*ttd).cond.notify_one();
+    drop((*ttd).delayed_fg_cond.wait(ttd_locked).unwrap());
```
kkysen, line 547: We don't need the `drop` since it's at the end of the `fn`, right?

src/picture.rs
```diff
@@ -294,14 +294,17 @@ pub(crate) unsafe fn rav1d_thread_picture_alloc(
     if res.is_err() {
         return res;
     }
-    let _ = mem::take(&mut (*c).itut_t35);
+    let _ = mem::take(&mut *(*c).itut_t35.write().unwrap());
```
kkysen, line 297: If this is the only place we write to `itut_t35`, dropping it, we may be able to avoid the `RwLock` once ownership is better.  Can you add a comment on the `itut_t35` field for that?

src/internal.rs
```diff
@@ -255,7 +261,7 @@ pub struct Rav1dContext {
     pub(crate) inloop_filters: Rav1dInloopFilterType,
     pub(crate) decode_frame_type: Rav1dDecodeFrameType,
     pub(crate) drain: c_int,
-    pub(crate) frame_flags: PictureFlags,
+    pub(crate) frame_flags: RwLock<PictureFlags>,
```
kkysen, line 264: This can be an atomic instead I think.

If you do `cargo add atomig --features derive` and

```rust
#[derive(Clone, Copy, PartialEq, Eq, Hash, Default, Atom, AtomLogic)]
pub struct PictureFlags(u8);

bitflags! {
    impl PictureFlags: u8 {
        const NEW_SEQUENCE = 1 << 0;
        const NEW_OP_PARAMS_INFO = 1 << 1;
        const NEW_TEMPORAL_UNIT = 1 << 2;
    }
}
```

then you can do `Atomic<PictureFlags>`.

src/thread_task.rs
```diff
@@ -679,37 +701,42 @@ unsafe fn abort_frame(f: *mut Rav1dFrameContext, error: Rav1dResult) {
...
+    (*f).task_thread.cond.notify_one();
 }
 
 #[inline]
-unsafe fn delayed_fg_task(c: *const Rav1dContext, ttd: *mut TaskThreadData) {
-    let in_0 = (*ttd).delayed_fg.in_0;
-    let out = (*ttd).delayed_fg.out;
+unsafe fn delayed_fg_task(
+    c: *const Rav1dContext,
+    ttd: *mut TaskThreadData,
```
kkysen, line 710: Can these be made refs?  All the mutability is unchecked right now.  A `&mut TaskThreadData` can't exist alongside a `MutexGuard<TaskThreadData_inner>`.

src/thread_task.rs
```diff
@@ -718,92 +745,94 @@ unsafe fn delayed_fg_task(c: *const Rav1dContext, ttd: *mut TaskThreadData) {
...
-                    &mut *out,
-                    &*in_0,
-                    BitDepth16::select_mut(&mut (*ttd).delayed_fg.grain),
-                    row as usize,
-                );
-            }
-            _ => {
-                abort();
+        {
+            let thread_data = (*ttd).lock.lock().unwrap();
```
kkysen, line 781: Why are we using a different lock here vs. `thread_lock`?

src/obu.rs
```diff
@@ -2537,7 +2536,7 @@ unsafe fn parse_obus(c: &mut Rav1dContext, r#in: &Rav1dData, global: bool) -> Ra
                     &mut (*c).out.p,
                     &c.content_light,
                     &c.mastering_display,
-                    &c.itut_t35,
+                    &*c.itut_t35.read().unwrap(),
```
kkysen, line 2539: For these `.read()`s of `itut_t35`, I'm making the `&` into `.clone()`s, which I think will fix this issue?  I'll put up a PR for it and let me know if it helps after rebasing.  It was one small commit in a larger change I'm working on, but I didn't realize it could be helpful elsewhere.

include/dav1d/picture.rs
```diff
@@ -320,7 +320,7 @@ impl From<Rav1dPicAllocator> for Dav1dPicAllocator {
 }
 
 impl Rav1dPicAllocator {
-    pub unsafe fn alloc_picture(&mut self, p: *mut Rav1dPicture) -> Rav1dResult {
+    pub unsafe fn alloc_picture(&self, p: *mut Rav1dPicture) -> Rav1dResult {
```
kkysen, line 323: This should come along with a doc comment saying internal access to `Rav1dPicAllocator::cookie` (a `*mut c_void`) must be thread-safe.  Otherwise the `&self` here would be UB.

kkysen (changes requested):
I still have a bit more to review at the end of `src/thread_task.rs`, but overall, it'd be helpful if changes that can be made independently are split into separate PRs that come first.  For example:
* the `RwLock`s/`Atomic` for `itut_t35` and `picture_flags`
  * I can do this; I have a change to one of the `copy_props` `fn`s that might obviate some of this, and I know `atomig`
* replacing all of the `atomic_{u,}int`s with `Atomic{U,I}32`s
  * especially `cur` and `delayed_fg_progress` here, but we should just do them all up front first
* the `pending_tasks` `Mutex` (that's a separate commit, which is great, but a separate PR would be more helpful given the complexity of things)
* making `Rav1dPicAllocator` use `&self` instead of `&mut self`, which is more complex than it seems
  * I can do this, too, as I'm working on `Rav1dPicAllocator`
* symbolication of `delayed_fg.type_0` `match`es
* making some ptr args into refs first, if possible without immediately running into borrowck errors, as mixing ptrs and refs into them results in a lot of unchecked mutability

src/cdf.rs
rinon: I was thinking we should do that in a separate commit to avoid extra noise for the more important review. Not sure it matters a ton whether it's in this change or not, I don't have a strong preference for future changes. Maybe it's easy enough to just read the deref changes if we're making enough other changes in the function body (but we aren't here).

include/dav1d/picture.rs
rinon: A raw pointer is not Send or Sync, so we're going to have to deal with the thread safety of this structure later anyway. I'll have to look at what the existing code does, it may be a moot point and not actually accessed from multiple threads.

What is true even without thread safety is that the callbacks must not be re-entrant, or if they are they must use the cookie in a manner that is reentrant safe.

src/cdf.rs
kkysen: I think if we can make the ptr to ref change first, that's ideal (e.x. `*mut` => `&mut` and `(*c)` => `c`, which we can here).  If the `*mut` => `&` change is only possible (and not the `*mut => &mut` change) because of the rest of the PR, then it makes sense to do it together in the same PR, or save the `(*c)` => `c` change to a follow-up commit.  That way it's also clear which changes are specifically only possible due to the other changes in a PR (e.x. the ptr to ref change here is not).

src/decode.rs
rinon: They're necessary to end the guard lifetime before mutably borrowing `c`. Ideally `on_error` wouldn't need to mutate the context but it looks like it does need to for now. We might be able to remove these later.

src/decode.rs
rinon: Exactly. If the field isn't actually concurrently accessed it doesn't need any ordering and Relaxed should be the fastest lowering. We shouldn't be adding an ordering requirement where there was previously not one.

src/internal.rs
rinon: It originally had the `cur` field as well and I had to remove it. I'll clean this up, good point.

src/picture.rs
rinon: I looked at that, and I think we need an immutable context at this point. Not certain, I'll put a comment.

src/thread_task.rs
rinon: Wanted a short local lock that wasn't present in the original code. That's cleaner here IMO so we don't have to move the lock in and out of the top-level `Option`.

src/thread_task.rs
rinon: Good point, we don't.

src/thread_task.rs
rinon: Thanks, didn't notice this. Those are helpful settings.

src/lib.rs
rinon: Right

src/lib.rs
rinon: It doesn't matter yet, there's no drop impl and we don't drop anything in the inner structs. Once we have proper ownership here it won't matter, so I don't want to add extra code that is a noop and will be removed.

src/decode.rs
kkysen: I opened #678 to make `Abcd` do the same.

src/picture.rs
kkysen: `fn rav1d_thread_picture_alloc` calls `fn picture_alloc_with_edges`, which calls `alloc_picture_callback`, which as #677 notes, is always called on the main thread.  Maybe all of `Rav1dContext` should be in a `Mutex`/`RwLock` here?

src/internal.rs
```diff
@@ -155,27 +158,26 @@ impl BitDepthDependentType for Grain {
...
     pub type_0: TaskType,
-    pub progress: [atomic_int; 2], /* [0]=started, [1]=completed */
     pub grain: BitDepthUnion<Grain>,
 }
 
 #[repr(C)]
 pub(crate) struct TaskThreadData {
-    pub lock: pthread_mutex_t,
-    pub cond: pthread_cond_t,
+    pub lock: Mutex<TaskThreadData_delayed_fg>,
```
kkysen, line 169:
```suggestion
    pub delayed_fg: Mutex<TaskThreadData_delayed_fg>,
```

And for the `ttd_locked`-like variables, can you just name them `delayed_fg`?

src/decode.rs
kkysen: Ahh.  But is it valid to drop the lock earlier than it was previously?  I assume they were dropped in an order on purpose.

To keep the lock dropping in the same order as before, you can make `on_error` a macro instead, like

```rust
macro_rules! on_error {
        () => {
            // current on_error body
        }
}

on_error!();
```

kkysen: Or more ideally, we could do

```rust
impl Rav1dFrameContext {
    pub unsafe fn on_error(&mut self) {
        self.task_thread.error = AtomicI32::new(1);
        rav1d_cdf_thread_unref(&mut self.in_cdf);
        if (*self.frame_hdr).refresh_context != 0 {
            rav1d_cdf_thread_unref(&mut self.out_cdf);
        }
        for i in 0..7 {
            if !self.refp[i].p.frame_hdr.is_null() {
                rav1d_thread_picture_unref(&mut self.refp[i]);
            }
            rav1d_ref_dec(&mut self.ref_mvs_ref[i]);
        }
        rav1d_picture_unref_internal(&mut self.cur);
        rav1d_thread_picture_unref(&mut self.sr_cur);
        rav1d_ref_dec(&mut self.mvs_ref);
        rav1d_ref_dec(&mut self.seq_hdr_ref);
        rav1d_ref_dec(&mut self.frame_hdr_ref);
        for mut tile in self.tiles.drain(..) {
            rav1d_data_unref_internal(&mut tile.data);
        }
    }
}

...

which keeps most of the `Rav1dFrameContext` logic of dropping things separate and out of the macro.  Also, a bunch of that code has changed so it needs a rebase.
    macro_rules! on_error {
        () => {
            f.on_error();
            if c.n_fc == 1 {
                rav1d_thread_picture_unref(&mut c.out);
            } else {
                rav1d_thread_picture_unref(out_delayed);
            }
            c.cached_error_props = c.in_0.m.clone();
        };
    }
```

rinon: It's still a bit of a mess as the lock outlives the mutation (although dynamically it doesn't). Fair point on moving the unlock, I'll fix this somehow.

src/internal.rs
kkysen: See #679.

src/decode.rs
kkysen: The lock only outlives the `&mut` for the whole `c`, but if it's in the same `fn` (the macro way) and can see which fields we access, it knows different ones are accessed so there's no conflict.

src/cdf.rs
rinon: Done in #687.

include/dav1d/picture.rs
rinon: Handled in #680

src/thread_task.rs
rinon: We need to use the return value of `wait`, and I figure explicitly dropping it makes that clear we are immediately unlocking (which is an unusual pattern).

src/picture.rs
rinon: We don't want to take a lock around the entire structure, as we might be doing something else with the context on a different thread, no? Also, we write to itut_t35 in several places (moving it from the context to the frame).

src/obu.rs
rinon: I don't think that removes the need to lock the field, we immediately take it out which requires mutability.

rinon:
I've rebased this change on top of a bunch of other PRs (#679, #680, #686, #687, #688) which we need to land first. Only look at the top two commits of this PR for the real change.

src/thread_task.rs
kkysen: Can you use `let _ = ` instead?  That's the usual pattern for drop immediately.

kkysen:
> I've rebased this change on top of a bunch of other PRs (#679, #680, #686, #687, #688) which we need to land first. Only look at the top two commits of this PR for the real change.

Can you make the PRs stack?  It works pretty seamlessly if you enable

```
[rebase]
    updateRefs = true
```

in your `~/.gitconfig`.  And then after merging the head PR to `main`, if you delete the branch, the next PR changes to target `main`.

src/thread_task.rs
```diff
@@ -292,28 +292,32 @@ unsafe fn insert_task(f: &mut Rav1dFrameContext, t: *mut Rav1dTask, cond_signal:
...
-        pthread_mutex_lock(&mut f.task_thread.pending_tasks.lock);
-        let mut t: *mut Rav1dTask = f.task_thread.pending_tasks.head;
-        f.task_thread.pending_tasks.head = 0 as *mut Rav1dTask;
-        f.task_thread.pending_tasks.tail = 0 as *mut Rav1dTask;
-        f.task_thread.pending_tasks_merge.store(0, Ordering::SeqCst);
-        pthread_mutex_unlock(&mut f.task_thread.pending_tasks.lock);
+        let mut t = {
+            let mut pending_tasks = f.task_thread.pending_tasks.lock().unwrap();
+            let old_head = pending_tasks.head;
+            *pending_tasks = Default::default();
```
kkysen, line 329:
```suggestion
            let old_head = mem::take(&mut *pending_tasks).head;
```

src/thread_task.rs
```diff
@@ -458,19 +462,20 @@ pub(crate) unsafe fn rav1d_task_create_tile_sbrow(
...
-    pthread_mutex_lock(&mut f.task_thread.pending_tasks.lock);
-    if !((f.task_thread.pending_tasks.head).is_null() || pass == 2) {
+    let mut pending_tasks = f.task_thread.pending_tasks.lock().unwrap();
+    if !(pending_tasks.head.is_null() || pass == 2) {
         unreachable!();
     }
-    if (f.task_thread.pending_tasks.head).is_null() {
-        f.task_thread.pending_tasks.head = &mut *tasks.offset(0) as *mut Rav1dTask;
+    if pending_tasks.head.is_null() {
+        pending_tasks.head = &mut *tasks.offset(0) as *mut Rav1dTask;
```
kkysen, line 482:
```suggestion
        pending_tasks.head = tasks.offset(0);
```

src/thread_task.rs
```diff
@@ -458,19 +462,20 @@ pub(crate) unsafe fn rav1d_task_create_tile_sbrow(
...
     } else {
-        (*f.task_thread.pending_tasks.tail).next = &mut *tasks.offset(0) as *mut Rav1dTask;
+        (*pending_tasks.tail).next = &mut *tasks.offset(0) as *mut Rav1dTask;
```
kkysen, line 484:
```suggestion
        (*pending_tasks.tail).next = tasks.offset(0);
```

kkysen (commented):
fa83f57 LGTM.

src/thread_task.rs
```diff
@@ -935,15 +960,15 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
                     prev_t = Some(t);
                     f.task_thread.task_cur_prev = prev_t.as_deref_mut().unwrap() as *mut Rav1dTask;
                 }
-                ttd.cur = (ttd.cur).wrapping_add(1);
+                ttd.cur.fetch_add(1, Ordering::Relaxed);
```
kkysen, line 963:
```suggestion
                ttd.cur.store(cur, Ordering::Relaxed);
```

and if `fn park`, etc. pass and return `delayed_fg` by value so borrowck can reason about disjoint field borrows, I think this could just be `ttd.cur = AtomicU32::new(cur);`.

kkysen (changes requested):
I still have a bunch of comments for fb747f, so if you want to split out that first simpler commit and merge it, you could.

rinon:
> Can you make the PRs stack? It works pretty seamlessly if you enable

No, because they do not need to depend on each other and can land in any order (plus some are mine and some are yours). We shouldn't be blocking PRs unnecessarily.

src/decode.rs
```diff
@@ -5065,6 +5059,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
                     "Compiled without support for {}-bit decoding",
                     8 + 2 * seq_hdr.hbd
                 );
+                drop(task_thread_lock);
                 on_error(f, c, out_delayed);
                 return Err(ENOPROTOOPT);
```
randomPoison, line 5064: General Q: Why do we explicitly drop the thread lock guard right before early returns? Wouldn't the guard's destructor get run on the return anyway? Is there something I'm missing or misinterpreting here?

kkysen:
> > Can you make the PRs stack? It works pretty seamlessly if you enable
> 
> No, because they do not need to depend on each other and can land in any order (plus some are mine and some are yours). We shouldn't be blocking PRs unnecessarily.

That's true, but from my experience stacking them is usually overall easier unless you anticipate one PR will take significantly longer to merge.

src/decode.rs
kkysen: https://github.com/memorysafety/rav1d/pull/673#discussion_r1449506993

src/decode.rs
```diff
@@ -4902,19 +4898,19 @@ pub(crate) unsafe fn rav1d_decode_frame(f: &mut Rav1dFrameContext) -> Rav1dResul
     if res.is_ok() {
         if (*f.c).n_tc > 1 {
             res = rav1d_task_create_tile_sbrow(f, 0, 1);
-            pthread_mutex_lock(&mut (*f.task_thread.ttd).lock);
-            pthread_cond_signal(&mut (*f.task_thread.ttd).cond);
+            let mut thread_lock = (*f.task_thread.ttd).delayed_fg.lock().unwrap();
```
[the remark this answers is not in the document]
rinon: I don't think we should exactly do that. I have made all the names consistent now (either `task_thread_lock` or `delayed_fg`), but I don't want everything to be `delayed_fg`. There's actually two uses of this lock in the original C code that we can't really do in Rust (and might have chosen to use two different mutexes for, but that would be a very invasive and risky change now). There is access to `delayed_fg`, which needs to be locked for mutation. But there is also usage, such as here, where we are using the lock as a generic task thread lock for synchronization and not actually accessing any shared mutable data. Where we are doing this, and not accessing the field contents, I've named the variable `task_thread_lock` to indicate its use, rather than `delayed_fg`, which it has nothing to do with.

src/picture.rs
```diff
@@ -249,10 +249,10 @@ pub(crate) unsafe fn rav1d_thread_picture_alloc(
         f.frame_hdr.clone(),
         c.content_light.clone(),
         c.mastering_display.clone(),
-        c.itut_t35.take(),
+        c.itut_t35.lock().unwrap().take(),
```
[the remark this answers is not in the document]
rinon: This is a really good point, thank you. I've implemented this (will upload shortly).

src/thread_task.rs
```diff
@@ -771,27 +788,32 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
 
     rav1d_set_thread_name(b"dav1d-worker\0" as *const u8 as *const c_char);
 
-    let park = |tc: &mut Rav1dTaskContext, ttd: &mut TaskThreadData| {
+    unsafe fn park(
+        c: &Rav1dContext,
+        tc: &mut Rav1dTaskContext,
+        ttd: &TaskThreadData,
+        task_thread_lock: &mut Option<MutexGuard<TaskThreadData_delayed_fg>>,
```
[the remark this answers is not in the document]
rinon: Great point, thanks. Yeah that works beautifully.

src/thread_task.rs
```diff
@@ -836,10 +858,13 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
                 }
             }
             // run decoding tasks last
-            while ttd.cur < c.n_fc {
+            while ttd.cur.load(Ordering::Relaxed) < c.n_fc {
```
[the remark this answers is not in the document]
rinon: I think we should match the C code here, it's not un-ergonomic and we need to update ttd.cur anyway. There may be other updates to cur, so we might not actually go sequentially. I haven't looked into this but I'd prefer not to and we don't have to if we leave it as is.

src/thread_task.rs
rinon: Should be fixed now.

src/internal.rs
```diff
@@ -158,27 +158,26 @@ impl BitDepthDependentType for Grain {
...
+    pub cur: AtomicU32,
     /// This is used for delayed reset of the task cur pointer when
     /// such operation is needed but the thread doesn't enter a critical
     /// section (typically when executing the next sbrow task locklessly).
     /// See [`crate::src::thread_task::reset_task_cur`].
     pub reset_task_cur: AtomicU32,
     pub cond_signaled: AtomicI32,
-    pub delayed_fg: TaskThreadData_delayed_fg,
+    pub delayed_fg_progress: [AtomicI32; 2], /* [0]=started, [1]=completed */
+    pub delayed_fg_cond: Condvar,
```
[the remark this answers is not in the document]
rinon: Yeah, probably best to do that, moved it.

src/thread_task.rs
```diff
@@ -94,15 +102,18 @@ unsafe fn reset_task_cur(
         }
         reset_frame_idx = u32::MAX;
     }
-    if ttd.cur == 0 && ((*(c.fc).offset(first as isize)).task_thread.task_cur_prev).is_null() {
+    if ttd.cur.load(Ordering::Relaxed) == 0
+        && ((*(c.fc).offset(first as isize)).task_thread.task_cur_prev).is_null()
+    {
```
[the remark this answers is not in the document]
rinon: I think there's an early return that would also need a store. We should not prematurely optimize here. If you're suggesting this for readability, I'm first worrying about correctness and doing the simplest possible change. I'm still not convinced that there aren't interleaved access to the cur field that requires re-loading it from ttd and until we can say that for sure it's safer to load and store where the original code did so.

src/picture.rs
rinon: I've removed the lock for itut_t35 based on your suggestions elsewhere.

src/thread_task.rs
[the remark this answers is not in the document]
kkysen: Awesome!  I wasn't sure if that'd work.

src/thread_task.rs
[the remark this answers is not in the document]
kkysen: Oh, I had thought it was a for loop in C.  Not sure why it isn't.  It's fine to leave as is I guess then.  Just extra atomics.

src/thread_task.rs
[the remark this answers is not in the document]
kkysen: > I think there's an early return that would also need a store.

I'm pretty sure there aren't.  I did check.

> We should not prematurely optimize here. If you're suggesting this for readability, I'm first worrying about correctness and doing the simplest possible change. I'm still not convinced that there aren't interleaved access to the cur field that requires re-loading it from ttd and until we can say that for sure it's safer to load and store where the original code did so.

It's fine to leave this for later as an optimization, I just thought you were looking to avoid atomics where there weren't any previously.

src/decode.rs
rinon: I've replaced the cached_error_props field with a Mutex. Looking at the other uses, I think we're going to want interior mutability as we tighten mutable borrows of the top-level context (i.e. this field is modified after the context structure is initially populated). I also worked around the borrow of `c.out` by borrowing it as a pointer for now (as we do with `out_delayed`). We're going to have to do something else with that eventually, but we can deal with that when dealing with Rav1dPicture ownership.

I'm not adding a macro here. We shouldn't be complicating the code structure on a slow error path to work around a borrowing issue. It's better to pay the (tiny) bit of runtime work since this is on a slow path (error handling).

src/decode.rs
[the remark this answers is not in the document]
rinon: I just noticed there is actually a `Rav1dFrameContext_task_thread.lock` field, so maybe these should be called `task_thread_data_lock` instead, but we definitely shouldn't just call them all `delayed_fg` because the locking pattern sometimes has nothing to do with that data.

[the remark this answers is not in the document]
kkysen: Okay, that makes sense then.  I had assumed all of its uses were to access `delayed_fg`.  Your naming is much better then, since it's clear when `task_thread_lock` is just used for synchronization of external data.

> There's actually two uses of this lock in the original C code that we can't really do in Rust (and might have chosen to use two different mutexes for, but that would be a very invasive and risky change now).

This is referring to moving the other data protected by `task_thread_lock` actually inside the `Mutex`, right?  That's unfortunate but okay for now.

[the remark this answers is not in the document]
kkysen: Could you add some of this explanation as docs for the `delayed_fg: Mutex<_>` field though?

src/decode.rs
```diff
@@ -5105,6 +5096,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
                 for j in 0..i {
                     rav1d_thread_picture_unref(&mut f.refp[j]);
                 }
+                drop(task_thread_lock);
```
kkysen, line 5099: This isn't needed anymore now that you've fixed `fn on_error` to not take `c: &mut` anymore, right?

src/lib.rs
```diff
@@ -1087,7 +1087,7 @@ pub unsafe extern "C" fn dav1d_get_decode_error_data_props(
     (|| {
         validate_input!((!c.is_null(), EINVAL))?;
         validate_input!((!out.is_null(), EINVAL))?;
-        out.write(mem::take(&mut (*c).cached_error_props).into());
+        out.write(mem::take(&mut *((*c).cached_error_props).lock().unwrap()).into());
```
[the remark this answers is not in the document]
rinon: I intend on basically making all uses of `Rav1dContext` immutable, as much as we can after it is constructed. It's a context, not an actively owned and mutated thing. We can always remove inner mutability if we determine later that it's not necessary, but I expect the opposite to be true in most of these cases.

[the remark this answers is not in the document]
kkysen: Hmm, okay.  I'm fine putting the `Mutex` proactively, but I think it's useful to use `.get_mut()` as long as we can.  Then when it's switched to `.lock()` it's clear what necessitated that change.  It IMO makes those future PRs easier to understand, and is a small change.

kkysen (approved):
I still think it's better to leave `.get_mut()`s until we actually need them to be `.lock()`s (https://github.com/memorysafety/rav1d/pull/673#discussion_r1467010857), but if you disagree, that's okay.  Otherwise everything else LGTM!

PR #678, opened 2024-01-11
`struct Abcd`: Fix outdated comment and use `Ordering::Relaxed` instead of `SeqCst` since C uses nothing

kkysen:


rinon (approved):
At some point I'd like to remove this atomic if possible, but this is good for now.

kkysen:
> At some point I'd like to remove this atomic if possible, but this is good for now.

Yeah, if possible, that'd be good.  I'm not sure how we would though since it's in an `Arc`, and these atomics aren't accessed much so it seems pretty cheap.

PR #679, opened 2024-01-12
`struct Rav1dContext::frame_flags`: Make `PictureFlags` `Atomic<PictureFlags>`

kkysen:
A cheaper `Atomic` alternative to the `RwLock` in #673 (see https://github.com/memorysafety/rav1d/pull/673#discussion_r1449325925).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,26 @@
 version = 3
 
 [[package]]
+name = "atomig"
+version = "0.4.0"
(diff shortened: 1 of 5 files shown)


rinon (approved):
I don't want to add atomig just for this, we have other places we need it?

kkysen:
> I don't want to add atomig just for this, we have other places we need it?

Yeah, we do.  Other similar places where we have something we want to be atomic but it's not a primitive type (e.x. enums, `Option<&T>`, etc.).

PR #680, opened 2024-01-12
`struct Rav1dPicAllocator`: Make methods `&self` and require callbacks to access `cookie` thread-safely

kkysen:
Note that this adds to the requirements of `Dav1dPicAllocator::alloc_picture_callback`, part of the `DAV1D_API`.  It requires all accesses to `cookie` to be thread-safe (if threading is enabled), whereas previously it said it will always be called on the main thread.

In practice, I've found 3 implementations of `alloc_picture_callback` (which account for all of the uses in chromium and its dependencies) and they are all unconditionally thread-safe:
* `rav1d`'s default `fn dav1d_default_picture_alloc`: `cookie` is a `Rav1dMemPool`, which is protected by a mutex and thus thread-safe
* `dav1d`'s `fn picture_alloc` (for negative strides): the `cookie` is not uses
* [`ffmpeg`'s `fn libdav1d_picture_allocator`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/ffmpeg/libavcodec/libdav1d.c;l=76?q=libdav1d_picture_allocator&ss=chromium%2Fchromium%2Fsrc): `cookie`'s internal mutability is through [`AVBufferPool`](https://ffmpeg.org/doxygen/4.0/structAVBufferPool.html), which is thread-safe

### DIFF
--- a/include/dav1d/picture.rs
+++ b/include/dav1d/picture.rs
@@ -234,8 +234,9 @@ pub struct Dav1dPicAllocator {
     /// Any allocated memory area should also be padded by [`DAV1D_PICTURE_ALIGNMENT`] bytes.
     /// [`data`]`[1]` and [`data`]`[2]` must share the same [`stride`]`[1]`.
     ///
-    /// This function will be called on the main thread
-    /// (the thread which calls [`dav1d_get_picture`]).
+    /// # Safety
+    ///
+    /// If frame threading is used, accesses to [`Self::cookie`] must be thread-safe.
     ///
     /// # Args
     ///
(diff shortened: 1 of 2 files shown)


include/dav1d/picture.rs
```diff
@@ -290,10 +294,24 @@ pub(crate) struct Rav1dPicAllocator {
     pub cookie: *mut c_void,
 
     /// See [`Dav1dPicAllocator::alloc_picture_callback`].
+    ///
+    /// # Safety
+    ///
+    /// If frame threading is used, accesses to [`Self::cookie`] must be thread-safe.
+    /// If [`Self::cookie`] is accessed in Rust and has interior mutability, it must go through [`UnsafeCell`].
```
rinon, line 301: I don't think mentioning UnsafeCell here is necessary or helpful (most Rust usage of UnsafeCell is encapsulated in other types anyway).

rinon (approved):
LGTM but I think we shouldn't mention `UnsafeCell`

include/dav1d/picture.rs
kkysen: As discussed, I'll replace this with saying the cookie needs to be `Send + Sync`.

PR #681, opened 2024-01-14
`struct Rav1dPicture::data`: Make private and access through `fn data{,_mut,_mut_unchecked}` to guarantee aliased XOR mutable

kkysen:
Before more heavily refactoring `Rav1dPicture`, the `.data` self-referential slices, the `.allocator_data` data owner, and `.r#ref` ref count (which is very complex and entangled), this ensures we access `.data` in a safe aliased XOR mutable way.

For immutable access (`fn data`), there are no limits.

For mutable access (`fn data_mut`), we check that the ref count is 1 (similar to `{Arc,Mutex}::get_mut`). This is used for accesses through `out: &mut Rav1dPicture` in `mod fg_apply`.

In `mod recon` in `fn rav1d_recon_b_{intra,inter}`, mutable accesses are through `f.cur`, but the ref count is not 1.  For these, we use `fn data_mut_unchecked`, which will eventually have to be dynamically protected, probably through a `Mutex`.

In `mod decode` in `fn rav1d_decode_frame_init`, mutable accesses are through `f.cur` and `f.sr_cur.p`. The accesses here are more complex since they are stored in `Rav1dFrameContext_lf`, and because the luma pixels are re-used for the chroma pixels when there is no chroma. This mutably aliases the luma pixels, which would be UB, but they are not actually accessed, as they are only there to avoid extra branches.  Thus, we use `fn data_mut_unchecked` for these and will make them safe later.

### DIFF
--- a/include/dav1d/picture.rs
+++ b/include/dav1d/picture.rs
@@ -234,8 +234,9 @@ pub struct Dav1dPicAllocator {
     /// Any allocated memory area should also be padded by [`DAV1D_PICTURE_ALIGNMENT`] bytes.
     /// [`data`]`[1]` and [`data`]`[2]` must share the same [`stride`]`[1]`.
     ///
-    /// This function will be called on the main thread
-    /// (the thread which calls [`dav1d_get_picture`]).
+    /// # Safety
+    ///
+    /// If frame threading is used, accesses to [`Self::cookie`] must be thread-safe.
     ///
     /// # Args
     ///
@@ -259,13 +260,16 @@ pub struct Dav1dPicAllocator {
     ///
     /// [`data`]: Dav1dPicture::data
     /// [`stride`]: Dav1dPicture::data
-    /// [`dav1d_get_picture`]: crate::src::lib::dav1d_get_picture
     /// [`allocator_data`]: Dav1dPicture::allocator_data
     /// [`release_picture_callback`]: Self::release_picture_callback
     pub alloc_picture_callback:
         Option<unsafe extern "C" fn(pic: *mut Dav1dPicture, cookie: *mut c_void) -> Dav1dResult>,
 
     /// Release the picture buffer.
+    ///
+    /// # Safety
+    ///
+    /// If frame threading is used, accesses to `cookie` must be thread-safe.
     ///
     /// If frame threading is used, this function may be called by the main thread
     /// (the thread which calls [`dav1d_get_picture`]),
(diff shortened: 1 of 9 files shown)


include/dav1d/picture.rs
```diff
@@ -220,6 +224,58 @@ impl Default for Rav1dPicture {
...
+        mut r#ref,
+        ..
+    } = mem::take(p);
+    if !r#ref.is_null() {
+        if validate_input!(!data[0].is_null()).is_err() {
+            return;
+        }
+        rav1d_ref_dec(&mut r#ref);
+    }
+}
```
kkysen, line 277: I moved this to have access to the now private `data` field.  Right now things are split between `include/` and `src/`; eventually we'll want to consolidate the related ones into single `mod`ules, though that's not very important yet.  But since some fields are private now, we need to for some methods.  Though eventually this `fn` will be replaced by an auto-generated `impl Drop`, so it won't matter.

rinon:
Looks like this needs rebasing, it's just the single commit, right?

include/dav1d/picture.rs
rinon: Why not make this a method of `Rav1dPicture`?

include/dav1d/picture.rs
```diff
@@ -220,6 +224,53 @@ impl Default for Rav1dPicture {
...
+    ///
+    /// This must be the only mutable access to the pixel data.
+    ///
+    /// TODO(kkysen) Replace this with a [`Mutex`].
+    pub unsafe fn data_mut_unchecked(&self) -> [*mut c_void; 3] {
+        self.data
+    }
+
+    // TODO(kkysen) Temporary, but for now guarantees aliased XOR mutable.
+    pub unsafe fn data_mut(&self) -> [*mut c_void; 3] {
```
rinon, line 247: Shouldn't this take `&mut self`, like `Arc::get_mut()`?

kkysen:
> Looks like this needs rebasing, it's just the single commit, right?

This is semi-abandoned.  I'm doing it in a different order, after `Arc`ification.

include/dav1d/picture.rs
kkysen: It will be deleted eventually, replaced with just a `mem::take`.

rinon: ok, makes sense

randomPoison:
@kkysen I'm going to close this PR under the assumption it's not going to be updated further. You can always re-open it if we end up having a use for it.

kkysen:
> @kkysen I'm going to close this PR under the assumption it's not going to be updated further. You can always re-open it if we end up having a use for it.

Sure, that's fine.  Just don't delete the branch yet cause I'll reference it in future work probably.

PR #686, opened 2024-01-17
`fn rav1d_worker_task`: Refactor found control flow into a labelled block with a return value.

rinon:


src/thread_task.rs
```diff
@@ -1090,26 +1077,26 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
...
-                                    (*f).task_thread.done[(2 - p_0) as usize]
+                                    f.task_thread.done[(2 - p_0) as usize]
                                         .store(1 as c_int, Ordering::SeqCst);
-                                    (*f).task_thread
-                                        .error
-                                        .store(-(1 as c_int), Ordering::SeqCst);
-                                    (*f).task_thread.task_counter.fetch_sub(
-                                        frame_hdr.tiling.cols * frame_hdr.tiling.rows + (*f).sbh,
+                                    f.task_thread.error.store(-(1 as c_int), Ordering::SeqCst);
+                                    let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
```
kkysen, line 1083: Can you hoist these unwraps as high as possible so we do them as few times as possible?  I had put them close to their uses previously since the control flow was very complex, but now it's a lot simpler and clear.

src/thread_task.rs
```diff
@@ -834,163 +835,151 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
...
                                     .error
                                     .fetch_or((p2 == TILE_ERROR) as c_int, Ordering::SeqCst);
                             }
-                            if ((*t).sby + 1) < (*f).sbh {
-                                let next_t: *mut Rav1dTask = &mut *t.offset(1) as *mut Rav1dTask;
-                                *next_t = (*t).clone();
+                            if (t.sby + 1) < f.sbh {
+                                // add sby+1 to list to replace this one
+                                let next_t: *mut Rav1dTask =
+                                    &mut *(t as *mut Rav1dTask).offset(1) as *mut Rav1dTask;
```
kkysen, line 945: Offsetting/slicing a reference seems kind of wrong?  Is it?

src/thread_task.rs
```diff
@@ -834,163 +835,151 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
...
+            if (*c).n_fc > 1 as c_uint {
+                // run init tasks second
+                'init_tasks: for i in 0..(*c).n_fc {
+                    let first: c_uint = ::core::intrinsics::atomic_load_seqcst(&mut (*ttd).first);
+                    let f = &mut *((*c).fc)
+                        .offset(first.wrapping_add(i).wrapping_rem((*c).n_fc) as isize);
+                    if f.task_thread.init_done.load(Ordering::SeqCst) != 0 {
+                        continue 'init_tasks;
+                    }
+                    let t = if let Some(t) = f.task_thread.task_head.as_mut() {
```
kkysen, line 848:
```suggestion
                    let Some(t) = f.task_thread.task_head.as_mut() {
```

kkysen (commented):
This mostly LGTM but I left a few comments, and I'm not sure if you're allowed to offset from a reference (it might be okay but it feels like it shouldn't).

src/thread_task.rs
```diff
@@ -834,163 +835,151 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
...
                             }
                         }
                     }
                     // next:
-                    prev_t = t;
-                    t = (*t).next;
-                    (*f).task_thread.task_cur_prev = prev_t;
+                    next_t = t.next.as_mut();
+                    prev_t = Some(t);
+                    f.task_thread.task_cur_prev = prev_t.as_deref_mut().unwrap() as *mut Rav1dTask;
```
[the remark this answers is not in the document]
rinon: Can't reborrow it because we moved it into prev_t before that. Could work around this but I don't think it's worth it.

src/thread_task.rs
```diff
@@ -999,24 +988,25 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
...
         } else {
-            (*f).task_thread.task_head = (*t).next;
+            f.task_thread.task_head = t.next;
         }
-        if ((*t).next).is_null() {
-            (*f).task_thread.task_tail = prev_t;
+        if (t.next).is_null() {
+            f.task_thread.task_tail = prev_t
+                .as_deref_mut()
+                .map_or(ptr::null_mut(), |p| p as *mut Rav1dTask);
```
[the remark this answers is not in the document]
rinon: Why? Same difference here.

src/thread_task.rs
```diff
@@ -1078,9 +1068,6 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
                             );
                         }
                         if res_0.is_ok() {
-                            // Note that `progress.is_some() == (*c).n_fc > 1`.
-                            let progress = &**(*f).sr_cur.progress.as_ref().unwrap();
```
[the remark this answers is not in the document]
rinon: Had to move it to avoid overlapping borrows.

src/thread_task.rs
rinon: Nope, overlapping borrows. I had to move them down to their uses in some places.

src/thread_task.rs
rinon: It is, and we can't be leaving this forever. The reference here points into an array. We need to replace this with indexing later, but this isn't the place to do that.

src/thread_task.rs
rinon: Would need an extra indented block, and I wanted to preserve the C control flow here. I don't really think we should change that right now.

src/thread_task.rs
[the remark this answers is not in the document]
kkysen: It'll be easy to search and replace in the future since we use `.map_or_else(ptr::null_mut,` lots of other places.  Also I think clippy will suggest this but that's not really important.

src/thread_task.rs
kkysen: Are the overlapping borrows from other borrows of these specific fields or some function that takes `&mut f`?

src/thread_task.rs
kkysen: Okay, we don't need to fix it rn. But can you add a comment here and where `t` is declared saying it needs to be a slice eventually?

src/thread_task.rs
kkysen: This shouldn't change control flow or add indentation?  At least that's what I meant to suggest.  It's just using let else blocks to simplify things.

rinon: Sorry, my bad, I didn't understand this new let ... else syntax. Tried it but I didn't get it right. Huh. Did that now.

src/thread_task.rs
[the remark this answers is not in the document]
kkysen: Ohh.  But the only reason it works this way is because `f.task_thread.task_cur_prev` is a `*mut`.  You'll have multiple `&mut`s to `prev_t` once that's cleaned up.  I think you can just replace uses of `prev_t` with `f.task_thread.task_cur_prev.as_mut()`.  The problem with that is then you're `break 'found (f, t, f.task_thread.task_cur_prev.as_mut()`, and the `f` can also access `f.task_thread.task_cur_prev`.  That's still a problem with your current approach; it's just hidden by the raw pointer.  One way of getting around that is only doing `break 'found (f, t)` everywhere and then `let mut prev_t = f.task_thread.task_cur_prev.as_mut();` after.  And where you did `break 'found (f, t, None)` before, you can just `assert!(f.task_thread.task_cur_prev.is_null())` beforehand, which works in the tests.

If you'd like, you could fix this in a follow-up PR, but it seems like it's gonna need fixing either way.

src/thread_task.rs
```diff
@@ -802,160 +803,149 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
...
+                                if p1_0 < t.sby {
                                     break 'next;
                                 }
-                                (*f).task_thread
+                                f.task_thread
                                     .error
                                     .fetch_or((p1_0 == TILE_ERROR) as c_int, Ordering::SeqCst);
                             }
-                            let frame_hdr = &***(*f).frame_hdr.as_ref().unwrap();
+                            let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
```
kkysen, line 892:
```suggestion
```

This one seems unnecessary; the one above is still in scope.  I understand for the others, though, that they can't be hoisted because of `&mut f`s.

src/thread_task.rs
```diff
@@ -991,30 +982,30 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
...
-            let mut task_type = (*t).type_0 as c_uint;
+            let mut sby = t.sby;
+            let mut task_type = t.type_0 as c_uint;
             'fallthrough: loop {
                 match task_type {
                     RAV1D_TASK_TYPE_INIT => {
                         if !((*c).n_fc > 1 as c_uint) {
                             unreachable!();
                         }
                         let res = rav1d_decode_frame_init(&mut *f);
```
kkysen, line 997:
```suggestion
                        let res = rav1d_decode_frame_init(f);
```

I noticed these; these can become `f`s like `(*f)`s became `f`s.

kkysen (changes requested):
I had a couple of other minor comments (`&mut *f` => `f`, one `let frame_hdr = ` is redundant), and one larger one for `prev_t` having a second `&mut`.  For the latter, feel free to fix it in a follow-up PR if that seems better.

src/thread_task.rs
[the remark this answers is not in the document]
rinon: I actually want to replace the entire task list with something else, because it's a linked list right now and that's a pain in Rust. We might be able to keep the linked list with Arcs, but that needs further thought and investigation. I'm planning to look into this. We shouldn't do anything extra here until that gets figured out.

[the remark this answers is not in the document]
kkysen: Okay that's fine then if you plan to replace it and make it safe some other way.

src/thread_task.rs
rinon: Good point, fixed all of them in this function.

kkysen (approved):

PR #687, opened 2024-01-19
`mod thread_task`: Replace raw pointers with references

rinon:
Replaces raw pointers to the following types with references where possible in `thread_task.rs`:
- `Rav1dContext`
- `TaskThreadData`
- `Rav1dFrameContext`
- `Rav1dTaskContext`

### DIFF
--- a/src/thread_task.rs
+++ b/src/thread_task.rs
@@ -97,12 +97,7 @@ unsafe fn reset_task_cur(
         }
         reset_frame_idx = u32::MAX;
     }
-    if ttd.cur == 0
-        && ((*(c.fc).offset(first as isize))
-            .task_thread
-            .task_cur_prev)
-            .is_null()
-    {
+    if ttd.cur == 0 && ((*(c.fc).offset(first as isize)).task_thread.task_cur_prev).is_null() {
         return 0 as c_int;
     }
     if reset_frame_idx != u32::MAX {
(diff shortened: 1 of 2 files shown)


src/thread_task.rs
```diff
@@ -774,74 +761,74 @@ unsafe fn delayed_fg_task(c: *const Rav1dContext, ttd: *mut TaskThreadData) {
...
-        pthread_cond_signal(&mut (*ttd).delayed_fg.cond);
+        pthread_cond_signal(&mut ttd.delayed_fg.cond);
     }
 }
 
 pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
-    let tc: *mut Rav1dTaskContext = data as *mut Rav1dTaskContext;
-    let c: *const Rav1dContext = (*tc).c;
-    let ttd: *mut TaskThreadData = (*tc).task_thread.ttd;
+    let tc: &mut Rav1dTaskContext = (data as *mut Rav1dTaskContext).as_mut().unwrap();
```
kkysen, line 795: Can this just be a `&mut *` like we do elsewhere?  Once we replace the thread spawning with the Rust version we can pass the data directly so I don't think the unwrap is helpful.

kkysen (approved):

kkysen:
Umm, I think you merged in the wrong order. You need to merge the head PR into main first.

rinon:
Yeah, you're right, I did, sorry.

PR #688, opened 2024-01-19
Replace remaining C atomic intrinsics with native Rust atomics

### LINKED ISSUE #638: Replace C atomics with Rust equivalents
In trying to make Rav1dFrameContext safe I noticed that much of the mutable access we need is actually for fields that are atomic and therefore have inner mutability. Replacing these atomics will give us a better view into what parts of the higher-level context structures can be immutable.

rinon:
Each commit in this PR replaces the atomic field(s) in an individual struct.

Closes #638.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -28,7 +28,6 @@ use crate::include::dav1d::headers::RAV1D_TX_SWITCHABLE;
 use crate::include::dav1d::headers::RAV1D_WM_TYPE_AFFINE;
 use crate::include::dav1d::headers::RAV1D_WM_TYPE_IDENTITY;
 use crate::include::dav1d::headers::RAV1D_WM_TYPE_TRANSLATION;
-use crate::include::stdatomic::atomic_int;
 use crate::src::align::Align16;
 use crate::src::cdef::rav1d_cdef_dsp_init;
 use crate::src::cdf::rav1d_cdf_thread_alloc;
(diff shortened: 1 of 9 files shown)


src/decode.rs
```diff
@@ -4076,7 +4075,7 @@ pub(crate) unsafe fn rav1d_decode_tile_sbrow(t: &mut Rav1dTaskContext) -> Result
             f.a.offset((off_2pass + col_sb128_start + tile_row * f.sb128w) as isize);
         for bx in (ts.tiling.col_start..ts.tiling.col_end).step_by(sb_step as usize) {
             t.bx = bx;
-            if ::core::intrinsics::atomic_load_acquire(c.flush) != 0 {
+            if (*c).flush.load(Ordering::Acquire) != 0 {
```
kkysen, line 4078:
```suggestion
            if c.flush.load(Ordering::Acquire) != 0 {
```

src/ref.rs
```diff
@@ -48,7 +49,7 @@ pub unsafe fn rav1d_ref_create(mut size: usize) -> *mut Rav1dRef {
     (*res).data = data as *mut c_void;
     (*res).user_data = (*res).data;
     (*res).const_data = (*res).user_data;
-    *&mut (*res).ref_cnt = 1 as c_int;
+    (*res).ref_cnt = AtomicI32::new(0);
```
kkysen, line 52:
```suggestion
    (*res).ref_cnt = AtomicI32::new(1);
```

src/thread_task.rs
```diff
@@ -772,17 +751,11 @@ unsafe fn delayed_fg_task(c: *const Rav1dContext, ttd: *mut TaskThreadData) {
                 abort();
             }
         }
-        row = ::core::intrinsics::atomic_xadd_seqcst(
-            &mut *((*ttd).delayed_fg.progress).as_mut_ptr().offset(0) as *mut atomic_int,
-            1 as c_int,
-        );
+        row = (*ttd).delayed_fg.progress[0].fetch_add(1, Ordering::SeqCst);
         #[allow(unused_assignments)]
```
kkysen, line 755: Can you get rid of this and my comment now?

kkysen (changes requested):
There's a typo in the ref count, 0 instead of 1.  Otherwise it LGTM so I'll approve after that small fix.

src/thread_task.rs
rinon: It never needed to exist afaict. The compiler is right, this value is never read. I think at one time it was but the way the C code is structured now it is not used. I think this should be a separate PR though, just in case I'm wrong and it turns out to break something.

kkysen: It can't break anything; it's just disabling warnings.

kkysen (approved):
I'm pretty sure we can just remove that warning allowance, but it's fine if you'd rather do it later. IMO you're touching that code now so it makes sense now.  Either way the PR LGTM now.

rinon:
We can't remove the warning allowance, the warning is still valid. I'd have to remove the assignment, and I couldn't figure out why the original code even bothers with the assignment so I didn't.

kkysen:
> We can't remove the warning allowance, the warning is still valid. I'd have to remove the assignment, and I couldn't figure out why the original code even bothers with the assignment so I didn't.

Oh, okay, I didn't realize.  Then leaving it as is is all good.

PR #690, opened 2024-01-23
`fn rav1d_prepare_intra_edges`: clean up and make safe

fbossen:


src/recon.rs
```diff
@@ -2648,13 +2647,16 @@ pub(crate) unsafe fn rav1d_recon_b_intra<BD: BitDepth>(
...
-                        if t.by & (*f).sb_step - 1 == 0 {
-                            top_sb_edge = (*f).ipred_edge[0] as *mut BD::Pixel;
+                        let top_sb_edge_slice = if t.by & (*f).sb_step - 1 == 0 {
+                            let mut top_sb_edge: *const BD::Pixel =
+                                (*f).ipred_edge[0] as *mut BD::Pixel;
                             let sby = t.by >> (*f).sb_shift;
                             top_sb_edge =
                                 top_sb_edge.offset(((*f).sb128w * 128 * (sby - 1)) as isize);
-                        }
+                            slice::from_raw_parts(top_sb_edge, (*f).ipred_edge_sz as usize * 128)
```
kkysen, line 2656: Is `ipred_edge_sz` the full length of the slice `ipred_edge[0]`, or the length left after offsetting by `sb128w * 128 * (sby - 1)`?

src/ipred_prepare.rs
```diff
@@ -131,7 +131,7 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
     edge_flags: EdgeFlags,
     dst: *const BD::Pixel,
     stride: ptrdiff_t,
-    prefilter_toplevel_sb_edge: *const BD::Pixel,
+    prefilter_toplevel_sb_edge: &[BD::Pixel],
```
kkysen, line 131: This can be an `Option<&[BD::Pixel]>` since the only use of it is checked with `!prefilter_toplevel_sb_edge.is_null()`, so that can be an `if let Some(prefilter_toplevel_sb_edge) = prefilter_toplevel_sb_edge`.  Then you don't need to conflate empty slices with `None`.

src/recon.rs
fbossen: `ipred_edge_sz * 128` is the full length in bytes. Some fixing is definitely needed here.

src/ipred_prepare.rs
```diff
@@ -262,27 +250,23 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
...
         if have_top {
             let px_have_1 = cmp::min(sz_0, w - x << 2);
-            memcpy(
-                top as *mut c_void,
-                dst_top as *const c_void,
-                (px_have_1 << 1) as usize,
-            );
+            for i in 0..px_have_1 {
+                top[i as usize] = *dst_top.offset(i as isize);
+            }
```
kkysen: We mistranslated this before.  It should be a `BD::pixel_copy`.  Could you also fix that?

src/ipred_prepare.rs
```diff
@@ -301,53 +285,48 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
...
+                let top_right = &mut top[sz_0 as usize..];
                 let px_have_2 = cmp::min(sz_0, w - x - tw << 2);
-                memcpy(
-                    top.offset(sz_0 as isize) as *mut c_void,
-                    dst_top.offset(sz_0 as isize) as *const c_void,
-                    (px_have_2 << 1) as usize,
-                );
+                for i in 0..px_have_2 {
+                    top_right[i as usize] = *dst_top.offset((sz_0 + i) as isize);
+                }
```
kkysen: This one should also be a `BD::pixel_copy`.

src/ipred_prepare.rs
```diff
@@ -301,53 +285,48 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
...
+                BD::pixel_set(top_right, fill_value, sz_0 as usize);
             }
         }
     }
     if av1_intra_prediction_edges[mode as usize]
         .needs
         .contains(Needs::TOP_LEFT)
     {
+        // top-left sample and immediate neighbours
+        let corner = &mut topleft_out[(topleft_origin - 1)..(topleft_origin + 2)];
```
kkysen, line 309: Your way works and is good, but if we make `corner` an array we eliminate bounds checking, like this:

```suggestion
        let corner = <&mut [_; 3]>::try_from(&mut topleft_out[topleft_origin - 1..][..3]).unwrap();
```

src/ipred_prepare.rs
fbossen: This should be fixed in a later commit

src/ipred_prepare.rs
fbossen: If one defines a sub-slice of size 3 such as `&mut data[first_idx..][..3]`, can't the compiler figure out that bounds checking should be done only once (i.e. the sub-slice fits within the slice) and not every time the sub-slice is accessed if the only indices used are 0, 1 and 2?

kkysen: It can but it can be tripped up sometimes.  With an array that optimization is guaranteed.

fbossen: Thanks for the clarification. I made the change as suggested

src/ipred_prepare.rs
```diff
@@ -139,6 +139,21 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
...
+    let first_pos = if have_left { -1 } else { 0 } + if have_top { -stride } else { 0 };
+    let last_pos = if have_left {
+        (4 * h - 1) as isize * stride
+    } else if have_top {
+        4 * w as isize
+    } else {
+        0
+    };
+    let dst = if last_pos > first_pos {
+        slice::from_raw_parts(dst.offset(first_pos), (last_pos - first_pos) as usize)
```
kkysen:
```suggestion
        &slice::from_raw_parts(dst, last_pos as usize)[first_pos as usize..]
```

kkysen: Oh, nevermind.

src/ipred_prepare.rs
```diff
@@ -139,6 +139,21 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
...
+    } else {
+        &[]
+    };
+    let dst_origin = (-first_pos) as usize;
```
kkysen, line 156: What if `first_pos` is positive, like if `stride` is negative?  I know I have to re-design the `BD::pxstride` API, but assuming that's fixed.

src/ipred_prepare.rs
```diff
@@ -149,7 +149,7 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
         0
     };
     let dst = if last_pos > first_pos {
-        slice::from_raw_parts(dst.offset(first_pos), (last_pos - first_pos) as usize)
+        unsafe { slice::from_raw_parts(dst.offset(first_pos), (last_pos - first_pos) as usize) }
```
kkysen, line 152: You can't do this yet and mark the `fn` safe, since to be safe, since calling a safe `fn` must never cause UB.  If it does, then it's unsound.  We can't guarantee that the `dst` pointer is valid in this way, so we have to leave the `fn` as `unsafe`.  And generally, when you add an `unsafe` block like this, you want to add a `// Safety: reason` comment explaining why this is safe, i.e. why this can never be UB, no matter how the `fn` is called.  We're very close, though; once `dst` is made into a slice (or potentially some other type that safely encapsulates the raw pointer), we can mark this `fn` safe.

src/ipred_prepare.rs
```diff
@@ -120,59 +118,72 @@ static av1_intra_prediction_edges: [av1_intra_prediction_edge; N_IMPL_INTRA_PRED
...
+    let first_pos = if have_left { -1 } else { 0 } + if have_top { -stride } else { 0 };
+    let last_pos = if have_left {
+        (4 * (h - y) - 1) as isize * stride
+    } else if have_top {
+        4 * (w - x) as isize
+    } else {
+        0
+    };
+    let dst = if last_pos > first_pos {
+        unsafe { slice::from_raw_parts(dst.offset(first_pos), (last_pos - first_pos) as usize) }
```
randomPoison, line 152: I don't think this function can be marked safe with how the `dst` parameter is being used. For this function to be safely called, it relies on `dst` being a valid pointer to an allocation of at least N elements for us to convert it to a slice safely. This function will have to be left unsafe until that can be made into a slice (or some other safe reference type), I think.

src/ipred_prepare.rs
```diff
@@ -182,177 +193,131 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
...
-                && have_left == 0)
+                && !have_left)
     {
-        if !prefilter_toplevel_sb_edge.is_null() {
-            dst_top = &*prefilter_toplevel_sb_edge.offset((x * 4) as isize) as *const BD::Pixel;
+        let px_have = cmp::min(8 * tw, 4 * (w - x)) as usize;
+        let n = px_have + have_left as usize;
+        if let Some(prefilter_toplevel_sb_edge) = prefilter_toplevel_sb_edge {
+            let offset = (x * 4) as usize - have_left as usize;
+            &prefilter_toplevel_sb_edge[offset..offset + n]
```
kkysen, line 202:
```suggestion
            &prefilter_toplevel_sb_edge[offset..][..n]
```

src/ipred_prepare.rs
```diff
@@ -182,177 +193,131 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
...
                 .contains(Needs::LEFT)
-                && have_left == 0)
+                && !have_left)
     {
-        if !prefilter_toplevel_sb_edge.is_null() {
-            dst_top = &*prefilter_toplevel_sb_edge.offset((x * 4) as isize) as *const BD::Pixel;
+        let px_have = cmp::min(8 * tw, 4 * (w - x)) as usize;
+        let n = px_have + have_left as usize;
+        if let Some(prefilter_toplevel_sb_edge) = prefilter_toplevel_sb_edge {
+            let offset = (x * 4) as usize - have_left as usize;
```
kkysen, line 201: Is the `- have_left as usize` done here to avoid a `.offset(-1)` later on?

src/ipred_prepare.rs
```diff
@@ -182,177 +193,131 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
...
-        let left: *mut BD::Pixel = &mut *topleft_out.offset(-sz as isize) as *mut BD::Pixel;
-        if have_left != 0 {
-            let px_have = cmp::min(sz, h - y << 2);
-            let mut i = 0;
-            while i < px_have {
-                *left.offset((sz - 1 - i) as isize) =
-                    *dst.offset(BD::pxstride(stride as usize) as isize * i as isize - 1);
-                i += 1;
+        let sz = 4 * th as usize;
+        let left = &mut topleft_out[(topleft_origin - sz)..];
```
kkysen, line 215:
```suggestion
        let left = &mut topleft_out[topleft_origin - sz..];
```

You don't need parentheses next to `..` since `..` has a really late binding preference.  You would need parentheses to call a method on a range, though, like `(0..n).step_by(step)`.

src/ipred_prepare.rs
```diff
@@ -182,177 +193,131 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
...
+                sz,
             );
         }
         if av1_intra_prediction_edges[mode as usize]
             .needs
             .contains(Needs::BOTTOM_LEFT)
         {
-            let have_bottomleft = (if have_left == 0 || y + th >= h {
-                0 as c_int as c_uint
+            let bottom_left = &mut topleft_out[(topleft_origin - 2 * sz)..];
```
kkysen, line 239:
```suggestion
            let bottom_left = &mut topleft_out[topleft_origin - 2 * sz..];
```

src/ipred_prepare.rs
```diff
@@ -182,177 +193,131 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
...
-                BD::pixel_set(
-                    slice::from_raw_parts_mut(
-                        top.offset(px_have_1 as isize),
-                        (sz_0 - px_have_1).try_into().unwrap(),
-                    ),
-                    *top.offset((px_have_1 - 1) as isize),
-                    (sz_0 - px_have_1).try_into().unwrap(),
-                );
+        let sz = 4 * tw as usize;
+        let top = &mut topleft_out[(topleft_origin + 1)..];
```
kkysen, line 263:
```suggestion
        let top = &mut topleft_out[topleft_origin + 1..];
```

src/ipred_prepare.rs
```diff
@@ -182,177 +193,131 @@ pub unsafe fn rav1d_prepare_intra_edges<BD: BitDepth>(
...
-                        ),
-                        *top.offset((sz_0 + px_have_2 - 1) as isize),
-                        (sz_0 - px_have_2).try_into().unwrap(),
-                    );
+                (edge_flags & EDGE_I444_TOP_HAS_RIGHT) != 0
+            };
+            if have_topright {
+                let top_right = &mut top[sz..];
+                let px_have = cmp::min(sz, (w - x - tw << 2) as usize);
+                BD::pixel_copy(top_right, &dst_top[(sz + have_left as usize)..], px_have);
```
kkysen, line 294:
```suggestion
                BD::pixel_copy(top_right, &dst_top[sz + have_left as usize..], px_have);
```

src/recon.rs
```diff
@@ -3441,29 +3469,37 @@ pub(crate) unsafe fn rav1d_recon_b_inter<BD: BitDepth>(
...
                 m,
                 &mut angle,
                 bw4,
                 bh4,
                 0 as c_int,
-                tl_edge,
+                &mut BD::select_mut(&mut t.scratch.c2rust_unnamed_0.interintra_edge)
+                    .0
+                    .edge,
+                32,
```
kkysen, line 3502: I think it would be better to, where `tl_edge` is defined, rewrite that section as:

```rust
let tl_edge_array = BD::select_mut(&mut t.scratch.c2rust_unnamed_0.interintra_edge).0.edge;
let tl_edge_offset = 32;
let tl_edge = tl_edge_array[tl_edge_offset..].as_mut_ptr();
```

and then make the call here with `tl_edge_array, tl_edge_offset`.

Otherwise, it's hard to realize they're using the same offset and starting from the same array, and eventually we'll also need to fix the negative indices in other uses of `tl_edge`, so `tl_edge_offset` will be useful there.

And this same comment also applies to `edge`, the other version of this.

src/ipred_prepare.rs
```diff
@@ -120,59 +118,72 @@ static av1_intra_prediction_edges: [av1_intra_prediction_edge; N_IMPL_INTRA_PRED
...
+    prefilter_toplevel_sb_edge: Option<&[BD::Pixel]>,
     mut mode: IntraPredMode,
-    angle: *mut c_int,
+    angle: &mut c_int,
     tw: c_int,
     th: c_int,
     filter_edge: c_int,
-    topleft_out: *mut BD::Pixel,
+    topleft_out: &mut [BD::Pixel],
+    topleft_origin: usize, // position of top-left sample in `topleft_out`
```
kkysen, line 138:
```suggestion
    topleft_out_offset: usize,
```

I think `offset` makes more sense here than `origin`, since `offset` is often used in this way.  We do have a `Cursor` type @randomPoison added, but this is fine, too, and the logic is more explicit so it's easy to follow the changes.

src/ipred_prepare.rs
kkysen: Also, similar to `topleft_out_offset`, I think `dst_offset` makes more sense here.  Also, I think it may be clearer to calculate

```rust
let dst_offset = if have_left { 1 } else { 0 } + if have_top { stride } else { 0 };
```

instead of a negative `first_pos`.  Then you'd do `dst.sub(dst_offset)` instead of `dst.offset(first_pos)` and `dst_offset + last_pos` instead of `last_pos - first_pos`.

src/ipred_prepare.rs
```diff
@@ -120,59 +118,72 @@ static av1_intra_prediction_edges: [av1_intra_prediction_edge; N_IMPL_INTRA_PRED
...
     bd: BD,
 ) -> IntraPredMode {
     let bitdepth = bd.bitdepth();
+    let stride = BD::pxstride(stride as usize) as isize;
+    let first_pos = if have_left { -1 } else { 0 } + if have_top { -stride } else { 0 };
+    let last_pos = if have_left {
```
kkysen, line 144: I think `last` is a bit confusing here, as last usually refers to the last index (`.len() - 1`).  I don't think that's what's meant here?  Maybe `end`, or `len`?  I'm not sure.

src/ipred_prepare.rs
```diff
@@ -120,59 +118,72 @@ static av1_intra_prediction_edges: [av1_intra_prediction_edge; N_IMPL_INTRA_PRED
...
+        (4 * (h - y) - 1) as isize * stride
```
kkysen, line 145: Where does this `- 1` come from?  I'm not super familiar with how these calculations are done, so I'm just curious.

src/ipred_prepare.rs
```diff
@@ -120,59 +118,72 @@ static av1_intra_prediction_edges: [av1_intra_prediction_edge; N_IMPL_INTRA_PRED
...
+    };
+    let dst = if last_pos > first_pos {
+        unsafe { slice::from_raw_parts(dst.offset(first_pos), (last_pos - first_pos) as usize) }
+    } else {
+        &[]
+    };
+    let dst_origin = (-first_pos) as usize;
     if !(y < h && x < w) {
         unreachable!();
     }
```
kkysen, line 159: Could you hoist this `assert!(y < h && x < w)` to the top of the `fn`?  I think it's helpful in reasoning about some of the offsets and lengths above (for readability).

kkysen (changes requested):
Overall this looks really great!  Great job untangling the negative indices and tricky stuff like that.  I left a few minor comments, some questions, and some suggestions re. naming and making things clearer.  Also, I think some comments for the `dst` calculations at the beginning of the `fn` would be very helpful so it's easier to understand how things are being calculated.

src/ipred_prepare.rs
fbossen: Yes

src/ipred_prepare.rs
fbossen: done

src/ipred_prepare.rs
fbossen: `height - 1` to point to the last row in the picture.

kkysen (commented):
This `dst` slice calculation is very clear and useful.  Thanks!  It'll also be very helpful when I make `Rav1dPicture::data`, the origin of these slices, safe, too.

src/recon.rs
fbossen: I made modifications to avoid duplication of offset constants and array/ptr derivation. However, I have one question regarding your suggestion: does `let tl_edge_array = BD::select_mut(&mut t.scratch.c2rust_unnamed_0.interintra_edge).0.edge;` result in the array `edge` being copied into a new array `tl_edge_array`?

src/ipred_prepare.rs
CrazyboyQCD: > Just a neater, more readable pattern I think.

Actually, I found this is hard for compiler to get [optimized](https://godbolt.org/z/Mzd9a6E3b), maybe we should keep it.

src/ipred_prepare.rs
fbossen: I made the `dst` argument a slice which is now created before calling this function.

src/ipred_prepare.rs
kkysen: Hi @CrazyboyQCD!  Thanks for pointing this out!  I wasn't aware this optimized poorly sometimes, as there are other cases I've checked where `[offset..][..len]` was equivalent or better than `[offset..offset + len]`, though I don't remember the specifics.

I'm surprised rustc/LLVM is worse at optimizing `[offset..][..len]`, which I narrowed down a bit more to these two nearly identical versions:

https://godbolt.org/z/n76YcdEP5

```rust
pub fn f1(a: &[u8], b: usize, c: usize) -> &[u8] {
    if a.len() > b + c && a.len() > b && b + c > b {
        &a[b..b + c]
        // &a[b..][..c]
    } else {
        &[]
    }
}

pub fn f2(a: &[u8], b: usize, c: usize) -> &[u8] {
    if a.len() > b + c && a.len() > b && b + c > b {
        &a[b..b + c]
    } else {
        &[]
    }
}
```

Have you considered opening this as a rustc bug, as I don't see why one version should be optimized much worse?

Also, although in this case `[offset..offset + len]` optimized better, that's not always the case, and we can't generalize.  For example, for b861557, `[offset..][..len]` is optimized better instead:

https://godbolt.org/z/MjPEjbnsW

```rust
pub fn corner1(topleft_out: &mut [u8], topleft_out_offset: usize) -> &mut [u8; 3] {
    <&mut [_; 3]>::try_from(&mut topleft_out[topleft_out_offset - 1..=topleft_out_offset + 1]).unwrap()
}

pub fn corner2(topleft_out: &mut [u8], topleft_out_offset: usize) -> &[u8] {
    <&mut [_; 3]>::try_from(&mut topleft_out[topleft_out_offset - 1..][..3]).unwrap()
}
```

`corner1` contains an extra check and call to `slice_end_index_overflow_fail` that's not there for `corner2`.  Though this might just be due to `..=` vs. `..`, as `..=` usually seems to optimize worse.  So for this case, @fbossen, could you rebase and drop b861557?

Either way, I don't think we should generalize this difference in optimizations.  It may well change in future versions of rustc and LLVM.

src/recon.rs
kkysen: No, as you can see from `unsafe fn select_mut<T>(bd: &mut BitDepthUnion<T>) -> &mut T::T<Self> where ...`'s signature, it returns a `&mut`, so the array isn't being copied.

src/recon.rs
```diff
@@ -2668,7 +2672,11 @@ pub(crate) unsafe fn rav1d_recon_b_intra<BD: BitDepth>(
                             (*ts).tiling.col_end,
                             (*ts).tiling.row_end,
                             edge_flags,
-                            dst,
+                            slice::from_raw_parts(
+                                ((*f).cur.data[0] as *const BD::Pixel)
+                                    .offset(cmp::min(data_diff, 0)),
+                                data_diff.abs() as usize + data_width as usize,
```
kkysen, line 2678:
```suggestion
                                data_diff.unsigned_abs() + data_width as usize,
```

src/recon.rs
```diff
@@ -3906,7 +3942,11 @@ pub(crate) unsafe fn rav1d_recon_b_inter<BD: BitDepth>(
                             (*ts).tiling.col_end >> ss_hor,
                             (*ts).tiling.row_end >> ss_ver,
                             0 as EdgeFlags,
-                            uvdst,
+                            slice::from_raw_parts(
+                                ((*f).cur.data[1 + pl as usize] as *const BD::Pixel)
+                                    .offset(cmp::min(data_diff, 0)),
+                                data_diff.abs() as usize + data_width as usize,
+                            ),
```
kkysen, line 3949: To keep the "distance" from the existing code smaller, could you put this in a `uvdst_slice` variable or something, and then pass `uvdst_slice` as the argument?  And the same for the other similar cases in this commit.  I think it will help when moving the other uses of `uvdst`, etc. to the safe, slice version.

Overall, though, this is a really great change, and is very helpful in understanding the layout and index/length calculations on the pixel `data` better.  And it's exactly the right way to make `fn rav1d_prepare_intra_edges` fully safe.  Thanks!

kkysen (commented):
Is the reason a lot of these variables/their calculations had to be repeated due to avoiding overlapping/multiple `&mut`s?  If so, things are fine as is.

src/recon.rs
```diff
@@ -2660,6 +2656,10 @@ pub(crate) unsafe fn rav1d_recon_b_intra<BD: BitDepth>(
                         } else {
                             None
                         };
+                        let interintra_edge =
+                            BD::select_mut(&mut t.scratch.c2rust_unnamed_0.interintra_edge);
+                        let edge_array = &mut interintra_edge.0.edge;
+                        let edge_offset = 128 as usize;
```
kkysen, line 2662:
```suggestion
                        let edge_offset = 128;
```

If it's a literal, rustc should be able to figure out it's a `usize`.  If it's ambiguous, though, things are fine as is, though `128usize` can also work.

src/recon.rs
```diff
@@ -2676,12 +2676,11 @@ pub(crate) unsafe fn rav1d_recon_b_intra<BD: BitDepth>(
...
                             seq_hdr.intra_edge_filter,
-                            &mut BD::select_mut(&mut t.scratch.c2rust_unnamed_0.interintra_edge)
-                                .0
-                                .edge,
-                            128,
+                            edge_array,
+                            edge_offset,
                             BD::from_c((*f).bitdepth_max),
                         );
+                        let edge = edge_array.as_ptr().add(edge_offset);
```
kkysen, line 2691:
```suggestion
                        let edge = edge_array[edge_offset..].as_ptr();
```

Although if this doesn't work due to multiple `&mut`s, your way is okay.

fbossen: Both ways work. My thinking was that `edge` is a pointer to the middle of a slice and will be later dereferenced with negative offsets. It doesn't seem "right" to first construct a slice `[edge_offset..]` that doesn't contain elements that will be referenced by the `edge` pointer.

fbossen:
> Is the reason a lot of these (in [6b18e4d](https://github.com/memorysafety/rav1d/commit/6b18e4d407f17cef01ca6583791d1086292e157c)) variables/their calculations had to be repeated due to avoiding overlapping/multiple `&mut`s? If so, things are fine as is.

I was running into issues with borrowing. In particular, `t` is borrowed in several places.

src/recon.rs
kkysen: Oh, that's a very good point.  I didn't realize `edge` will be accessed with negative offsets.  Your way is better in this case then.

kkysen (approved):
LGTM!

src/ipred_prepare.rs
CrazyboyQCD: @kkysen I'm glad to see further exploration about this.
> https://godbolt.org/z/s5bEzhaTa
> ```rust
> pub fn corner1(topleft_out: &mut [u8], topleft_out_offset: usize) -> &mut [u8; 3] {
>     <&mut [_; 3]>::try_from(&mut topleft_out[topleft_out_offset - 1..topleft_out_offset + 2]).unwrap()
> }
>
> pub fn corner2(topleft_out: &mut [u8], topleft_out_offset: usize) -> &mut [u8; 3] {
>     <&mut [_; 3]>::try_from(&mut topleft_out[topleft_out_offset - 1..][..3]).unwrap()
> }
> ```
In the link, I didn't see the output you have, what I saw is that they both have two calls to `index_overflow_fail`, my output is on [the clipboard](https://goonlinetools.com/online-clipboard/#3920). And after I added same [asserts](https://godbolt.org/z/nEv939zxs) in them, the `corner1` seems to be better optimized(I use `unreachable_unchecked` to get cleaner output).
Anyway, I agree that this shoud be reported to rust and we could keep the changes.

kkysen: > In the link, I didn't see the output you have, what I saw is that they both have two calls to `index_overflow_fail`, my output is on [the clipboard](https://goonlinetools.com/online-clipboard/#3920).

Whoops, I deleted the `..=` that was in b861557, which made them the same again.

CrazyboyQCD: Alright, I see the diffrence now.

kkysen: I also found that if you re-arrange the precondition checks/assumes to more closely follow the `[offset..][..len]` version, you can get even more optimized code:

https://godbolt.org/z/ve83f8Gjb

```rust
pub fn f3(a: &[u8], b: usize, c: usize) -> &[u8] {
    if a.len() > b {
        let a = &a[b..];
        if a.len() > c {
            return &a[..c];
        }
    }
    &[]
}
```

PR #692, opened 2024-01-23
`fn rav1d_worker_task`: Add missing null check

rinon:
In refactoring the state machine of `fn rav1d_worker_task` I missed a
null check that was necessary. This change adds that check back in.

### DIFF
--- a/src/thread_task.rs
+++ b/src/thread_task.rs
@@ -1353,10 +1353,13 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
             };
             // Note that `progress.is_some() == (*c).n_fc > 1`.
             if let Some(progress) = &(*f).sr_cur.progress {
-                progress[1].store(
-                    if error_0 != 0 { FRAME_ERROR } else { y_0 },
-                    Ordering::SeqCst,
-                );
+                // upon flush, this can be free'ed already
+                if !((*f).sr_cur.p.data[0]).is_null() {
+                    progress[1].store(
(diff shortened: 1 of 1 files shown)


src/thread_task.rs
```diff
@@ -1353,10 +1353,13 @@ pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
...
-                    if error_0 != 0 { FRAME_ERROR } else { y_0 },
-                    Ordering::SeqCst,
-                );
+                // upon flush, this can be free'ed already
+                if !((*f).sr_cur.p.data[0]).is_null() {
+                    progress[1].store(
+                        if error_0 != 0 { FRAME_ERROR } else { y_0 },
+                        Ordering::SeqCst,
+                    );
+                }
```
randomPoison, line 1362: Can you explain how the value that you're checking for null (`(*f).sr_cur.p.data[0]`) is tied to the operation being guarded (`progress[1].store()`)? I'm not familiar with this part of the code and the connection isn't clear to me.

rinon: I don't entirely know, this is just what the C code does: https://github.com/memorysafety/rav1d/blob/30582c7ad08e454fecc09fc92c6159f7e43f3a45/src/thread_task.c#L909

kkysen: Let me check, but I think `progress` and `data`'s nullability are correlated.

kkysen: I think this check may be unnecessary, but it's hard to tell for sure, so the check is good.

kkysen (approved):

PR #693, opened 2024-01-23
`struct Rav1dFrameContext_task_thread::lock`: Make a Rust `Mutex`

rinon:
Replaces `Rav1dFrameContext_task_thread.lock` with a Rust instead of pthreads mutex. This lock does not cover any data but appears to be an alternative lock to the lock in TaskThreadData to reduce unnecessary contention.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -73,7 +73,6 @@ use crate::src::refmvs::refmvs_tile;
 use crate::src::refmvs::Rav1dRefmvsDSPContext;
 use crate::src::thread_data::thread_data;
 use atomig::Atomic;
-use libc::pthread_mutex_t;
 use libc::ptrdiff_t;
 use std::ffi::c_int;
 use std::ffi::c_uint;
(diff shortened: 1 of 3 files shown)


randomPoison (approved):

src/internal.rs
```diff
@@ -424,7 +423,7 @@ impl Default for Rav1dFrameContext_task_thread_pending_tasks {
 
 #[repr(C)]
 pub(crate) struct Rav1dFrameContext_task_thread {
-    pub lock: pthread_mutex_t,
+    pub lock: Mutex<()>,
```
kkysen, line 426: What is this `Mutex` supposed to protect?  The change in this PR seems fine as an intermediate state, but we should figure out exactly what data this lock is protecting and put it inside.

rinon: As far as I can tell there isn't data being locked here, it's to keep multiple threads from entering the critical section between `get_frame_progress` and updating the `f.sr_cur.progress` atomic. These accesses are all atomic but in this case need to be synchronized between separate atomic fields.

kkysen: Okay.  I guess I'm still not sure why exactly those accesses need to be synchronized together.  It's all good here, though, I'm just confused.

PR #697, opened 2024-01-25
`fn rav1d_ref_create`: Remove as all uses are gone (forgot to remove the final use even though it was unused)

### LINKED ISSUE #640: `fn rav1d_ref_create`: Replace with `Arc::new`
This is for the non-pooled ones.  We definitely want to do this.

kkysen:
* Fixes #640.

#640 has already been fixed, but the last use of `rav1d_ref_create` I forgot to remove even though it was really unused.  #640 was fixed mostly by:
* #659
* #661

### DIFF
--- a/src/obu.rs
+++ b/src/obu.rs
@@ -5,7 +5,6 @@ use crate::include::dav1d::data::Rav1dData;
 use crate::include::dav1d::dav1d::RAV1D_DECODEFRAMETYPE_INTRA;
 use crate::include::dav1d::dav1d::RAV1D_DECODEFRAMETYPE_REFERENCE;
 use crate::include::dav1d::headers::DRav1d;
-use crate::include::dav1d::headers::Dav1dITUTT35;
 use crate::include::dav1d::headers::Rav1dAdaptiveBoolean;
 use crate::include::dav1d::headers::Rav1dChromaSamplePosition;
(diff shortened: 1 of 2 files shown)


src/obu.rs
```diff
@@ -2442,14 +2439,6 @@ unsafe fn parse_obus(
                     if payload_size <= 0 {
                         writeln!(c.logger, "Malformed ITU-T T.35 metadata message format");
                     } else {
-                        let r#ref = rav1d_ref_create(
-                            ::core::mem::size_of::<DRav1d<Rav1dITUTT35, Dav1dITUTT35>>()
-                                + payload_size as usize * ::core::mem::size_of::<u8>(),
-                        );
-                        if r#ref.is_null() {
-                            return Err(ENOMEM);
-                        }
```
randomPoison, line 2451: Was this code just creating a ref and then doing nothing with it? Do you know if there was some case where that `ref` var was being used originally?

randomPoison (approved):

src/obu.rs
kkysen: Yup, doing nothing.  It didn't get flagged as dead code due to the error checking.  It was originally being used but I replaced it with an `Arc`.  So actually this is a memory leak I missed.

PR #717, opened 2024-02-01
`struct Rav1dContext`: Move destruction into `impl Drop`

rinon:
Moves `Rav1dContext` destruction from `fn close_internal` to the `Drop` impl for the struct.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -907,109 +907,127 @@ unsafe fn close_internal(c_out: &mut *mut Rav1dContext, flush: c_int) {
     if flush != 0 {
         rav1d_flush(c);
     }
-    if !((*c).tc).is_null() {
-        let ttd: *mut TaskThreadData = &mut (*c).task_thread;
-        if (*ttd).inited != 0 {
-            let task_thread_lock = (*ttd).delayed_fg.lock().unwrap();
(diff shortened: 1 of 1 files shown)


randomPoison (approved):

src/lib.rs
```diff
@@ -907,109 +907,124 @@ unsafe fn close_internal(c_out: &mut *mut Rav1dContext, flush: c_int) {
...
-            while n < (*c).n_tc && (*((*c).tc).offset(n as isize)).task_thread.td.inited != 0 {
-                (*((*c).tc).offset(n as isize)).task_thread.die = true;
-                n = n.wrapping_add(1);
+    c.drop_in_place();
+    rav1d_freep_aligned(c_out as *mut _ as *mut c_void);
+}
+
+impl Drop for Rav1dContext {
+    fn drop(&mut self) {
+        unsafe {
```
kkysen, line 916: Can you leave a safety comment saying this isn't probably safe yet (which we'll fix eventually), but we need it here since it's in a trait?

kkysen (approved):
LGTM besides that one safety comment.  Previously I also ran into trouble with manual `impl Drop` vs. automatically derived ones and moving out of a type with a manual `impl Drop`, but it looks like that's not an issue here so it should be all fine.

PR #718, opened 2024-02-01
`struct Rav1dFrameContext`: Remove `*const Rav1dContext` field

rinon:


randomPoison (approved):

kkysen (approved):
This looks really great and clean!  Better than what the `Arc` idea would've turned out to be I think.

PR #719, opened 2024-02-01
`struct TaskThreadData`: `Arc`ify

rinon:


src/lib.rs
```diff
@@ -360,7 +363,8 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
             (*f).task_thread.cond = Condvar::new();
             (*f).task_thread.pending_tasks = Default::default();
         }
-        (*f).task_thread.ttd = &mut (*c).task_thread;
+        (&mut (*f).task_thread.ttd as *mut Arc<TaskThreadData>)
+            .write(Arc::clone(&(*c).task_thread));
```
kkysen, line 367: Can you just write `.clone()` instead of `Arc::clone`?

src/thread_task.rs
```diff
@@ -791,14 +792,18 @@ unsafe fn delayed_fg_task<'l, 'ttd: 'l>(
 pub unsafe extern "C" fn rav1d_worker_task(data: *mut c_void) -> *mut c_void {
     let tc: &mut Rav1dTaskContext = &mut *(data as *mut Rav1dTaskContext);
     let c: &Rav1dContext = &*tc.c;
-    let ttd: &mut TaskThreadData = &mut *tc.task_thread.ttd;
+
+    // We clone the Arc here to avoid an immutable borrow of tc across the call
```
kkysen, line 796: This clone is immediately dropped after the function ends, right?  Can you add that to the comment, that this Arc clone is just temporary to get around borrowck?

src/thread_task.rs
```diff
@@ -791,14 +792,18 @@ unsafe fn delayed_fg_task<'l, 'ttd: 'l>(
...
+    // to park
+    let ttd_clone = Arc::clone(&tc.task_thread.ttd);
```
kkysen, line 798: I might be forgetting something from a previous PR, but why can't we just access `ttd` through `tc` in `fn park`?  So we don't need a separate `&TaskThreadData` and extra the `Arc` clone and drop.

kkysen (commented):
Mostly looks really good, I just had a couple small suggestions/questions.

src/lib.rs
```diff
@@ -333,15 +333,18 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
...
+        first: AtomicU32::new(0),
+        cur: AtomicU32::new((*c).n_fc),
+        reset_task_cur: AtomicU32::new(u32::MAX),
+        cond_signaled: AtomicI32::new(0),
+        delayed_fg_progress: [AtomicI32::new(0), AtomicI32::new(0)],
+        delayed_fg_cond: Condvar::new(),
+        delayed_fg: Mutex::new(mem::zeroed()),
+        inited: 1,
+    };
+    (&mut (*c).task_thread as *mut Arc<TaskThreadData>).write(Arc::new(ttd));
```
randomPoison, line 347: Why are these writes to `Arc` fields being done via this awkward pointer write operation (if I'm understanding this correctly)? Is this because the fields are uninitialized and we can't have the destructor run when we overwrite the field?

randomPoison (approved):

src/lib.rs
kkysen: I believe so; that's what I assumed.

src/lib.rs
rinon: https://doc.rust-lang.org/std/rc/index.html#cloning-references

src/thread_task.rs
rinon: Sure, done. This function has the same lifetime as the thread, but that's fine.

src/thread_task.rs
rinon: I suppose we could. But I think we still need the extra clone, otherwise we have an immutable borrow (of ttd) across calls to `park`, even if we don't pass it as a parameter.

src/lib.rs
rinon: Correct, I don't want the destructor to run here. That will be removed when `Rav1dContext` is not unsafely zero initialized.

src/thread_task.rs
kkysen: Ah, okay, it's fine then

src/lib.rs
kkysen: Hmm, I didn't realize that.  I'm not sure I totally agree with the reasoning (we don't do the same thing for `drop`s of `Arc`s, for example, and we still `#[derive(Clone)]` types containing `Arc`s), but it's okay if it's already a convention.  I haven't been doing that in my existing `Arc` clones, though.

rinon: No need to go back through and fix anything but I think it's worth doing in the future.

PR #721, opened 2024-02-01
`struct Rav1dTaskContext`: Remove `*const Rav1dContext` field

rinon:
`Rav1dTaskContext` doesn't need to hold a pointer to its parent if we pass this borrow into the thread start function explicitly.


kkysen (approved):
I assume that temporary `struct` is being removed shortly once we replace `pthread_create` with the Rust version.

PR #722, opened 2024-02-01
Share thread communication between main and worker threads

rinon:
This change shares an altered version of Rav1dTaskContext_task_thread with the main context structure because the main thread needs to be able to shut down and flush worker threads.


src/lib.rs
```diff
@@ -924,26 +932,17 @@ impl Drop for Rav1dContext {
...
                     drop(task_thread_lock);
-                    let mut n_0: c_uint = 0 as c_int as c_uint;
-                    while n_0 < self.n_tc {
-                        let pf: *mut Rav1dTaskContext =
-                            &mut *(self.tc).offset(n_0 as isize) as *mut Rav1dTaskContext;
-                        if (*pf).task_thread.td.inited == 0 {
-                            break;
+                    for tc in &self.tc_shared {
+                        if let Some(thread_id) = tc.handle {
+                            pthread_join(thread_id, 0 as *mut *mut c_void);
```
kkysen, line 944: Can you name `thread_id` `handle` instead?  Or something like `thread`.  Thread IDs are something else.

src/internal.rs
```diff
@@ -203,12 +204,37 @@ pub struct Rav1dContext_intra_edge {
...
+        Self {
+            handle,
+            thread_data,
+        }
+    }
+
+    pub(crate) fn flushed(&self) -> bool {
+        self.thread_data.flushed.load(Ordering::Relaxed)
+    }
+}
```
kkysen, line 229: I think you can make all the methods and fields `pub` as long as the type itself is `pub(crate)`.

src/internal.rs
```diff
@@ -203,12 +204,37 @@ pub struct Rav1dContext_intra_edge {
...
+impl Rav1dContextTaskThread {
+    pub(crate) fn new(
+        handle: Option<pthread_t>,
+        thread_data: Arc<Rav1dTaskContext_task_thread>,
+    ) -> Self {
+        Self {
+            handle,
+            thread_data,
+        }
+    }
```
kkysen, line 224: If the constructor isn't doing anything and the fields aren't private, I think just directly constructing the type is better than calling an `fn new`, since you get things like field names, too.

rinon: I prefer a method here, personally, but it's not a big deal. I kind of wanted to make these fields private but we don't really need to here.

randomPoison (approved):

src/internal.rs
```diff
@@ -203,12 +204,27 @@ pub struct Rav1dContext_intra_edge {
...
+pub(crate) struct Rav1dContextTaskThread {
+    /// Thread join handle, if this task is on a worker thread. The main thread
+    /// task does not contain a handle.
+    pub handle: Option<pthread_t>,
+    /// Data shared between the main thread and a worker thread.
+    pub thread_data: Arc<Rav1dTaskContext_task_thread>,
+}
+
+impl Rav1dContextTaskThread {
+    pub(crate) fn flushed(&self) -> bool {
```
kkysen, line 216:
```suggestion
    pub fn flushed(&self) -> bool {
```

src/internal.rs
```diff
@@ -656,11 +672,21 @@ pub struct Rav1dTaskContext_frame_thread {
 
 #[repr(C)]
 pub(crate) struct Rav1dTaskContext_task_thread {
-    pub td: thread_data,
+    pub cond: Condvar,
     pub ttd: Arc<TaskThreadData>,
-    pub fttd: *mut FrameTileThreadData,
```
kkysen, line 661: Was this just not used before?

kkysen (approved):

src/internal.rs
rinon: Yeah, totally unused.

PR #723, opened 2024-02-01
Spawn Rust threads instead of pthreads

### LINKED ISSUE #696: Make thread pool implementation safe
The thread pool implementation in `thread_task.rs` needs to be cleaned up and made safe. The generated code uses pthread threading primitives, so we need to convert those to use safe primitives from [`std::sync`](https://doc.rust-lang.org/stable/std/sync/index.html). For now the goal is just to cleanup the existing thread pool implementation, rather than trying to move to a different Rust thread pool implementation. This includes cleaning up the last mutex, split off from #691.

This task also covers any remaining work needed in `thread_task.rs` to cleanup the c2rust-generated code. The bulk of the cleanup will be done as part of cleaning up the thread pool, but we should make sure `thread_task.rs` is completely cleaned up before closing this.

* [ ] Replace pthread synchronization primitives with `std::sync` types.
* [ ] Cleanup c2rust output in `thread_task.rs`.
* [ ] Make `TaskThreadData_delayed_fg` fields safe
* [ ] Make `Rav1dFrameContext_task_thread_pending_tasks` fields safe
* [ ] Make `Rav1dFrameContext_task_thread` fields safe

rinon:
This change spawns threads as Rust threads rather than C pthread
threads. Each thread owns its own `Rav1dTaskContext` and communicates
with the main thread via a shared, reference-counted
`Rav1dTaskContext_task_thread` structure.

Closes #696


src/internal.rs
```diff
@@ -728,11 +739,35 @@ pub(crate) struct Rav1dTaskContext {
...
+            by: 0,
+            l: mem::zeroed(),
+            a: ptr::null_mut(),
+            rt: mem::zeroed(),
+            cf,
+            al_pal: Default::default(),
+            pal_sz_uv: Default::default(),
+            txtp_map: [0u8; 1024],
+            scratch: mem::zeroed(),
+            warpmv: mem::zeroed(),
```
kkysen, line 762: We might want to `#[derive(Zeroable)]` (using `bytemuck` or something similar) these `union`s to more safely initialize them.  This is clearly sound for now, but if in the future some sub-field becomes an `enum`, it becomes UB, for example.  Not important to do right here, but we should in the future.

src/lib.rs
```diff
@@ -371,38 +346,23 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
...
-            None
-        };
-        (*c).tc_shared
-            .push(Rav1dContextTaskThread::new(thread, task_thread));
-        m = m.wrapping_add(1);
+            (*c).main_tc = Some(Mutex::new(Box::new(Rav1dTaskContext::new(
+                &*((*c).fc).offset(0),
+                Arc::clone(&task_thread),
+            ))));
+            (*c).tc.push(Rav1dContextTaskThread::new(None, task_thread));
```
kkysen, line 364: The C allocated all up front, so this should either be a `(0..(*c).n_fc).map(|_| ...).collect()` or there should be a `Vec::with_capacity` at the beginning.

src/internal.rs
```diff
@@ -233,9 +237,13 @@ pub struct Rav1dContext {
     pub(crate) fc: *mut Rav1dFrameContext,
     pub(crate) n_fc: c_uint,
 
-    pub(crate) tc: *mut Rav1dTaskContext,
-    pub(crate) tc_shared: Vec<Rav1dContextTaskThread>,
+    /// Task context if we aren't using worker threads
+    pub(crate) main_tc: Option<Mutex<Box<Rav1dTaskContext>>>,
```
kkysen, line 241: Why do you need this `Box`?

src/internal.rs
```diff
@@ -291,6 +299,9 @@ pub struct Rav1dContext {
     pub(crate) picture_pool: *mut Rav1dMemPool,
 }
 
+unsafe impl Send for Rav1dContext {}
+unsafe impl Sync for Rav1dContext {}
```
kkysen, line 303: For `unsafe impl`s, can you add `TODO`s to remove them or add a safety comment?

src/internal.rs
rinon: Yeah, I think I'd prefer we impl Default or provide initializers rather than adding bytemuck for this.

src/lib.rs
rinon: I don't think we care, this is in a small loop and will get initialized together. This structure didn't even exist in the original, it's new. Trying to optimize its allocation seems like premature optimization.

src/internal.rs
rinon: I wasn't sure about adding significant size to the `Rav1dContext` struct. On second thought, I don't think it matters here, I'll remove the Box.

kkysen: Ahhh.  I don't think `Rav1dContext` is ever really cloned or moved much.  Good consideration, though.

src/internal.rs
kkysen: I meant `impl Default` through `impl Zeroable` using `bytemuck`'s auto derive, since otherwise it's easy to get wrong and not initialize the largest `union` variant or something.

src/lib.rs
rinon: I dropped it into a boxed slice just to make it clear the size doesn't change after initialization.

src/internal.rs
rinon: And on third thought, I'm keeping it. The TaskContext is large and I don't want to pay extra memory in the threaded case where we don't use it.

src/lib.rs
kkysen: Oh perfect, I was just gonna suggest that, too.

src/internal.rs
kkysen: Okay, that's fine.  Could you leave a comment that that's why the `Box` is there?  Since it's not very clear otherwise.

randomPoison (approved):

src/internal.rs
```diff
@@ -718,11 +732,33 @@ pub(crate) struct Rav1dTaskContext {
...
-pub(crate) struct Rav1dTaskContext_borrow<'c> {
-    pub c: &'c Rav1dContext,
-    pub tc: &'c mut Rav1dTaskContext,
+impl Rav1dTaskContext {
+    pub(crate) unsafe fn new(
+        f: *const Rav1dFrameContext,
+        task_thread: Arc<Rav1dTaskContext_task_thread>,
+    ) -> Self {
+        let mut cf = mem::zeroed();
+        *BitDepth16::select_mut(&mut cf) = Align64([0; 32 * 32]);
```
kkysen, line 741: Now that #711 is merged, can you use `Default::default()` for this?

src/lib.rs
```diff
@@ -371,41 +346,31 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
...
+            if (*c).n_tc > 1 {
+                let context_borrow = &*c;
+                let thread_data_copy = Arc::clone(&thread_data);
+                let handle = thread::spawn(|| rav1d_worker_task(context_borrow, thread_data_copy));
+                Rav1dContextTaskThread {
+                    handle: Some(handle),
+                    thread_data,
+                }
+            } else {
+                (*c).main_tc = Some(Mutex::new(Box::new(Rav1dTaskContext::new(
```
kkysen, line 363: When is `main_tc` ever `None`?  I only see it set to `Some`, and the only use of it `.unwrap()`s it.

src/lib.rs
```diff
@@ -935,22 +896,19 @@ impl Drop for Rav1dContext {
...
+                    tc.thread_data.die.store(true, Ordering::Relaxed);
+                }
+                ttd.cond.notify_all();
+                drop(task_thread_lock);
+                for task_thread in self.tc.iter_mut() {
+                    if let Some(handle) = task_thread.handle.take() {
+                        handle.join().expect("Could not join task thread");
                     }
                 }
-                rav1d_free_aligned(self.tc as *mut c_void);
```
kkysen, line 953: What's this replaced by?

src/lib.rs
```diff
@@ -371,41 +346,31 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
...
-            ) != 0
-            {
-                return error(c, c_out, &mut thread_attr);
+    (*c).tc = (0..(*c).n_tc)
+        .map(|_| {
+            let thread_data = Arc::new(Rav1dTaskContext_task_thread::new(Arc::clone(
+                &(*c).task_thread,
+            )));
+            if (*c).n_tc > 1 {
+                let context_borrow = &*c;
```
kkysen, line 355: This is something that can be removed once `c` is no longer a raw pointer, right?  If so, could you leave a `TODO`?

src/lib.rs
```diff
@@ -935,22 +896,19 @@ impl Drop for Rav1dContext {
...
+                drop(task_thread_lock);
+                for task_thread in self.tc.iter_mut() {
+                    if let Some(handle) = task_thread.handle.take() {
+                        handle.join().expect("Could not join task thread");
                     }
                 }
-                rav1d_free_aligned(self.tc as *mut c_void);
             }
             let mut n_1: c_uint = 0 as c_int as c_uint;
             while !(self.fc).is_null() && n_1 < self.n_fc {
```
kkysen, line 914: I realized this too late (not on #717), and can't comment on what's below since it's too far from the diff, but all of the `mem::take`s below can be removed now that this is in an `impl Drop` since they'll be done automatically.

kkysen (changes requested):
It mostly looks very good, but I still have some questions about things.

src/internal.rs
rinon: Understood. Yeah, I still don't want to add a dependency for that right now.

src/lib.rs
rinon: If n_tc > 1, we never get to this path.

src/lib.rs
rinon: We can rewrite Drop in a separate PR, I figured we'd do that when it's completely safe, as those `take`s aren't doing anything bad and just cost a tiny amount.

src/internal.rs
```diff
@@ -207,7 +212,7 @@ pub struct Rav1dContext_intra_edge {
 pub(crate) struct Rav1dContextTaskThread {
     /// Thread join handle, if this task is on a worker thread. The main thread
     /// task does not contain a handle.
-    pub handle: Option<pthread_t>,
+    pub handle: Option<JoinHandle<()>>,
```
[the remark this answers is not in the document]
rinon: Hmmmmm I think I like this idea, I'll try it.

src/lib.rs
kkysen: So I assume it's zero-initialized somewhere else?  It might be helpful to just set it explicitly in both cases here.

kkysen (commented):
It LGMT now, though I was wondering if we can remove the `n_tc` field now and replace it with `tc.len()`.

src/internal.rs
```diff
@@ -227,12 +240,8 @@ pub struct Rav1dContext {
...
-    // This Rav1dTaskContext is heap-allocated because we don't want to bloat
-    // the size of Rav1dContext, especially when it isn't used when we have
-    // worker threads.
-    pub(crate) main_tc: Option<Mutex<Box<Rav1dTaskContext>>>,
-    /// Worker thread join handles and communication
+    /// Worker thread join handles and communication, or single thread task
+    /// context if n_tc == 1
     pub(crate) tc: Box<[Rav1dContextTaskThread]>,
     /// Number of worker threads
     pub(crate) n_tc: c_uint,
```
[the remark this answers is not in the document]
rinon: Yes, but I'm going to make a separate PR for that.

Issue #701, opened 2024-01-30
Cleanup c2rust output in cdf.rs


randomPoison:
* [ ] Cleanup helper functions to take safe arguments where possible (may take raw  pointers as arguments currently).
* [ ] General cleanup of c2rust generated code in functions.
* [ ] Restore comments from original C.
* [ ] Replace calls to `memcpy` with safe Rust copy operations.

folkertdev:
I'm exploring this, and thought I'd just convert the memcpys to safe operations. Of course, it's never that simple. Most memcpys in this file originate in a bunch of C macros.

https://code.videolan.org/videolan/dav1d/-/blob/master/src/cdf.c?ref_type=heads#L3952

    #define update_cdf_1d(n1d, name) \
        do { \
            memcpy(dst->name, src->name, sizeof(dst->name)); \
            dst->name[n1d] = 0; \
        } while (0)

I assume that the rust code would, in the best case, mirror the C code as much as possible. Macros seem the way to go here. Unfortunately that doesn't quite work for the more complex invocations like  

    update_cdf_1d(10, dmv.comp[k].classes);

Is there some way to parse the `dmv.comp[k].classes` bit that I'm missing? You can parse record access just fine but array index makes it much more challenging. 

```rust
macro_rules! update_cdf_1d {
    ($n1d:expr, $($name:ident).+) => {
        (*dst).$($name).+.0.copy_from_slice(&src.$($name).+.0);
        (*dst).$($name).+[$n1d] = 0;
    };
}
```

As an alternative, maybe a function-based approach works better?

```rust
fn update_cdf_1d<const N: usize>(src: &[u16; N], dst: &mut [u16; N], n1d: usize) {
    dst.copy_from_slice(src);
    dst[n1d] = 0;
}

...

update_cdf_1d(&src.m.filter_intra.0, &mut dst.m.filter_intra.0, 4);
```
maybe with a `#[inline(always)]` on there? 

Any thoughts on the best way to clean this up?

kkysen:
In cleaning up similar macros, I've usually turned them into Rust `fn`s or macros.  We want to keep the code similar to the C so that backporting is easy, but it doesn't have to be super close always.  So your function based approach looks like a good idea.  And I don't think there's a way to pass a field path (including array indexing) as a macro argument like C is doing, as you noted.  Also, an `#[inline(always)]` is probably a good idea, too, to match the macro always being inlined, though it'll probably be inlined anyways.

[...]

PR #724, opened 2024-02-02
`src/cdf.rs`: use macros to clean up code

folkertdev:
use macros to match the original C source as closely as possible 

https://code.videolan.org/videolan/dav1d/-/blob/master/src/cdf.c?ref_type=heads#L3952

kind of a messy diff, best reviewed with the "split" view I think.

@kkysen turns out a macro is possible! It works in the majority of cases here, sometimes a variable bound in the macro is required, but then the macro can just be inlined one level and you still get most of the succinctness of the macros.

linking https://github.com/memorysafety/rav1d/issues/701

### DIFF
--- a/src/cdf.rs
+++ b/src/cdf.rs
@@ -4884,9 +4884,9 @@ static av1_default_coef_cdf: [CdfCoefContext; 4] = [
 ];
 
 pub(crate) unsafe fn rav1d_cdf_thread_update(
-    hdr: *const Rav1dFrameHeader,
+    hdr: &Rav1dFrameHeader,
     dst: *mut CdfContext,
-    src: *const CdfContext,
+    src: &CdfContext,
 ) {
     let mut i = 0;
     while i < N_BS_SIZES as c_int {
--- a/src/cdf.rs
+++ b/src/cdf.rs
@@ -5620,7 +5620,7 @@ pub unsafe fn rav1d_cdf_thread_init_static(cdf: *mut CdfThreadContext, qidx: c_i
     (*cdf).data.qcat = get_qcat_idx(qidx) as c_uint;
 }
 
-pub unsafe fn rav1d_cdf_thread_copy(dst: *mut CdfContext, src: *const CdfThreadContext) {
+pub unsafe fn rav1d_cdf_thread_copy(dst: *mut CdfContext, src: &CdfThreadContext) {
     if !((*src).r#ref).is_null() {
         memcpy(
             dst as *mut c_void,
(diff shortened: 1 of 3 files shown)


src/cdf.rs
```diff
@@ -4884,720 +4884,174 @@ static av1_default_coef_cdf: [CdfCoefContext; 4] = [
 ];
 
 pub(crate) unsafe fn rav1d_cdf_thread_update(
-    hdr: *const Rav1dFrameHeader,
+    hdr: &Rav1dFrameHeader,
     dst: *mut CdfContext,
```
folkertdev, line 4888: this raw pointer comes from a data field ultimately

    #[derive(Clone, Copy)]
    #[repr(C)]
    pub union CdfThreadContext_data {
        pub cdf: *mut CdfContext,
        pub qcat: c_uint,
    }

what is the policy on turning that into a mutable reference? It seems to happen quite freely in the codebase.

src/decode.rs
```diff
@@ -4924,7 +4924,7 @@ pub(crate) unsafe fn rav1d_decode_frame(
                 rav1d_cdf_thread_update(
                     frame_hdr,
                     f.out_cdf.data.cdf,
```
folkertdev, line 4926: similarly here, can we just `&mut *f.out_cdf.data.cdf` here? or should other code be refactored so that this is in fact a mutable reference and not a raw mut pointer.

src/cdf.rs
rinon: We've been preemptively turning non-null pointers into references where we can and are pretty sure they will be references once we've removed all raw pointers.

src/cdf.rs
```diff
@@ -5629,22 +5083,10 @@ pub unsafe fn rav1d_cdf_thread_copy(dst: *mut CdfContext, src: *const CdfThreadC
         );
     } else {
         (*dst).m = av1_default_cdf.clone();
-        memcpy(
-            ((*dst).kfym.0).as_mut_ptr() as *mut c_void,
-            default_kf_y_mode_cdf.0.as_ptr() as *const c_void,
-            ::core::mem::size_of::<[[[u16; 16]; 5]; 5]>(),
-        );
+        (*dst).kfym.0.copy_from_slice(&default_kf_y_mode_cdf.0);
```
kkysen, line 5086:
```suggestion
        (*dst).kfym = default_kf_y_mode_cdf;
```

src/cdf.rs
```diff
@@ -5629,22 +5083,10 @@ pub unsafe fn rav1d_cdf_thread_copy(dst: *mut CdfContext, src: *const CdfThreadC
...
-            default_mv_joint_cdf.0.as_ptr() as *const c_void,
-            ::core::mem::size_of::<[u16; 4]>(),
-        );
-        memcpy(
-            ((*dst).dmv.joint.0).as_mut_ptr() as *mut c_void,
-            default_mv_joint_cdf.0.as_ptr() as *const c_void,
-            ::core::mem::size_of::<[u16; 4]>(),
-        );
+        (*dst).mv.joint.0.copy_from_slice(&default_mv_joint_cdf.0);
+        (*dst).dmv.joint.0.copy_from_slice(&default_mv_joint_cdf.0);
```
kkysen, line 5089:
```suggestion
        (*dst).mv.joint = default_mv_joint_cdf;
        (*dst).dmv.joint = default_mv_joint_cdf;
```

src/cdf.rs
```diff
@@ -4883,721 +4883,175 @@ static av1_default_coef_cdf: [CdfCoefContext; 4] = [
...
-                ((*dst).m.partition[k_3 as usize][j_4 as usize]).as_mut_ptr() as *mut c_void,
-                ((*src).m.partition[k_3 as usize][j_4 as usize]).as_ptr() as *const c_void,
-                ::core::mem::size_of::<[u16; 16]>(),
-            );
-            (*dst).m.partition[k_3 as usize][j_4 as usize]
-                [dav1d_partition_type_count[k_3 as usize] as usize] = 0 as c_int as u16;
-            j_4 += 1;
+    macro_rules! update_cdf_1d {
+        ($n1d:expr, $($name:tt)+) => {
+            dst.$($name)+.copy_from_slice(&src.$($name)+);
```
kkysen, line 4893:
```suggestion
            dst.$($name)+ = src.$($name)+;
```

kkysen (changes requested):
I'm still reviewing the macros, but I realized all of the `.copy_from_slice`s can be just array assignments (with an implicit `Copy` I believe?).

src/cdf.rs
```diff
@@ -4883,721 +4883,175 @@ static av1_default_coef_cdf: [CdfCoefContext; 4] = [
...
-            (*dst).m.comp_uni_ref[j_33 as usize][i_23 as usize][0] =
-                (*src).m.comp_uni_ref[j_33 as usize][i_23 as usize][0];
-            (*dst).m.comp_uni_ref[j_33 as usize][i_23 as usize][1] = 0 as c_int as u16;
-            i_23 += 1;
+    update_bit_2d!(7, 3, m.txpart);
+    update_cdf_2d!(2, 15, m.txtp_inter1);
+    update_cdf_1d!(11, m.txtp_inter2.0);
+    update_bit_1d!(4, m.txtp_inter3);
+
+    if (*hdr).frame_type.is_key_or_intra() {
```
kkysen, line 5006: nit:

```suggestion
    if hdr.frame_type.is_key_or_intra() {
```

kkysen (approved):
LGTM!  The macros match the C really closely, which is really great.  I think a version with normal `fn`s is may still be slightly better for readability on the Rust side, but I'm not sure if that's feasible while keeping the calls as simple as they are now, so this way is very good.

Just one little nit, though.

Issue #707, opened 2024-01-31
`Rav1dContext_frame_thread::out_delayed`: Convert to a `Vec`


randomPoison:
Convert the field `out_delayed` in the struct `Rav1dContext_frame_thread` to be a `Vec<Rav1dThreadPicture>`. This overlaps with #705 since the bulk of the logic that touches `out_delayed` is in `lib.rs`.

PR #725, opened 2024-02-02
`Rav1dContext_frame_thread::out_delayed`: Make a `Box<[_]>`

folkertdev:
* fixes https://github.com/memorysafety/rav1d/issues/707

the issue suggests this could be a `Vec<_>`, but because this allocation cannot grow a boxed slice seems more natural to me.

I tried to take this a bit further, but the code in `lib.rs` runs into serious borrow checker issues.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,7 +71,6 @@ use crate::src::thread_task::rav1d_task_delayed_fg;
 use crate::src::thread_task::rav1d_worker_task;
 use crate::src::thread_task::FRAME_ERROR;
 use cfg_if::cfg_if;
-use libc::calloc;
 use libc::free;
 use libc::memset;
 use libc::pthread_attr_destroy;
@@ -346,14 +345,10 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
         inited: 1,
     };
     (&mut (*c).task_thread as *mut Arc<TaskThreadData>).write(Arc::new(ttd));
-    if (*c).n_fc > 1 as c_uint {
-        (*c).frame_thread.out_delayed = calloc(
-            (*c).n_fc as usize,
-            ::core::mem::size_of::<Rav1dThreadPicture>(),
-        ) as *mut Rav1dThreadPicture;
-        if ((*c).frame_thread.out_delayed).is_null() {
-            return error(c, c_out, &mut thread_attr);
-        }
+    if (*c).n_fc > 1 {
+        (*c).frame_thread.out_delayed = std::iter::repeat_with(Rav1dThreadPicture::default)
+            .take((*c).n_fc as usize)
+            .collect();
     }
     let mut n: c_uint = 0 as c_int as c_uint;
     while n < (*c).n_fc {
@@ -523,7 +518,7 @@ unsafe fn output_image(c: &mut Rav1dContext, out: &mut Rav1dPicture) -> Rav1dRes
     } else {
         res = rav1d_apply_grain(c, out, &(*r#in).p);
     }
-    rav1d_thread_picture_unref(r#in);
+    rav1d_thread_picture_unref(&mut *r#in);
 
     if !c.all_layers && c.max_spatial_id && !(c.out.p.data.data[0]).is_null() {
         rav1d_thread_picture_move_ref(r#in, &mut c.out);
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -561,9 +556,8 @@ unsafe fn drain_picture(c: &mut Rav1dContext, out: &mut Rav1dPicture) -> Rav1dRe
         while !(*f).tiles.is_empty() {
             task_thread_lock = (*f).task_thread.cond.wait(task_thread_lock).unwrap();
         }
-        let out_delayed: *mut Rav1dThreadPicture =
-            &mut *(c.frame_thread.out_delayed).offset(next as isize) as *mut Rav1dThreadPicture;
-        if !((*out_delayed).p.data.data[0]).is_null()
+        let out_delayed: &mut Rav1dThreadPicture = &mut (c.frame_thread.out_delayed)[next as usize];
```
kkysen, line 559:
```suggestion
        let out_delayed = &mut c.frame_thread.out_delayed[next as usize];
```

src/lib.rs
```diff
@@ -864,10 +858,8 @@ pub(crate) unsafe fn rav1d_flush(c: *mut Rav1dContext) {
                 &mut *((*c).fc).offset(next as isize) as *mut Rav1dFrameContext;
             rav1d_decode_frame_exit(&*c, &mut *f, Err(EGeneric));
             (*f).task_thread.retval = Ok(());
-            let out_delayed: *mut Rav1dThreadPicture = &mut *((*c).frame_thread.out_delayed)
-                .offset(next as isize)
-                as *mut Rav1dThreadPicture;
-            if (*out_delayed).p.frame_hdr.is_some() {
+            let out_delayed = &mut ((*c).frame_thread.out_delayed)[next as usize];
```
kkysen, line 861:
```suggestion
            let out_delayed = &mut (*c).frame_thread.out_delayed[next as usize];
```

src/lib.rs
```diff
@@ -987,21 +979,21 @@ impl Drop for Rav1dContext {
                 n_1 = n_1.wrapping_add(1);
             }
             rav1d_free_aligned(self.fc as *mut c_void);
-            if self.n_fc > 1 as c_uint && !(self.frame_thread.out_delayed).is_null() {
+            if self.n_fc > 1 as c_uint && !(self.frame_thread.out_delayed).is_empty() {
                 let mut n_2: c_uint = 0 as c_int as c_uint;
                 while n_2 < self.n_fc {
-                    if (*(self.frame_thread.out_delayed).offset(n_2 as isize))
+                    if (self.frame_thread.out_delayed[n_2 as usize])
```
kkysen, line 985:
```suggestion
                    if self.frame_thread.out_delayed[n_2 as usize]
```

src/lib.rs
```diff
@@ -987,21 +979,21 @@ impl Drop for Rav1dContext {
...
                     {
                         rav1d_thread_picture_unref(
-                            &mut *(self.frame_thread.out_delayed).offset(n_2 as isize),
+                            &mut self.frame_thread.out_delayed[n_2 as usize],
                         );
                     }
                     n_2 = n_2.wrapping_add(1);
                 }
-                free(self.frame_thread.out_delayed as *mut c_void);
+                let _ = std::mem::take(&mut self.frame_thread.out_delayed);
```
kkysen, line 996:
```suggestion
                let _ = mem::take(&mut self.frame_thread.out_delayed);
```

We've generally been importing things like `std::mem`, `std::iter`, etc.

src/lib.rs
```diff
@@ -342,14 +341,10 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
...
-            ::core::mem::size_of::<Rav1dThreadPicture>(),
-        ) as *mut Rav1dThreadPicture;
-        if ((*c).frame_thread.out_delayed).is_null() {
-            return error(c, c_out, &mut thread_attr);
-        }
+    if (*c).n_fc > 1 {
+        (*c).frame_thread.out_delayed = std::iter::repeat_with(Rav1dThreadPicture::default)
+            .take((*c).n_fc as usize)
+            .collect();
     }
```
kkysen, line 348: `c: *mut Rav1dContext` is currently zero-initialized (see the `memset` above in `fn rav1d_open`).  So either `out_delayed` needs to be valid for zeroed bits, or we need to initialize it explicitly in an `else` case as well.  And until we remove the `memset` and initialize things directly, we should use `addr_of_mut!((*c).frame_thread.out_delayed).write(...)` to write to `out_delayed` to avoid dropping the zero-initialized `out_delayed` if that's not a valid representation.  Thus, I'd recommend something like this:

```suggestion
    addr_of_mut!((*c).frame_thread.out_delayed).write(if (*c).n_fc > 1 {
        (0..(*c).n_fc).map(|_| Default::default()).collect()
    } else {
        Box::new([])
    });
```

src/lib.rs
```diff
@@ -987,21 +979,21 @@ impl Drop for Rav1dContext {
                 n_1 = n_1.wrapping_add(1);
             }
             rav1d_free_aligned(self.fc as *mut c_void);
-            if self.n_fc > 1 as c_uint && !(self.frame_thread.out_delayed).is_null() {
+            if self.n_fc > 1 as c_uint && !(self.frame_thread.out_delayed).is_empty() {
```
kkysen, line 982:
```suggestion
            if self.n_fc > 1 as c_uint && !self.frame_thread.out_delayed.is_empty() {
```

kkysen (changes requested):
I had one important comment about the zero-initialized state for the `Box<[]>` and properly initializing it, and then a bunch of nitpicky comments about parentheses, etc.

src/lib.rs
folkertdev: makes sense! fixed

kkysen (approved):

PR #726, opened 2024-02-03
`Rav1dFrameContext_lf::level`: make a `Box<[_]>`

folkertdev:
there is one tricky bit of logic here.

the `filter_plane_rows_y` and `filter_plane_rows_uv` functions are called in a loop that increments the level_ptr by some chunk size, but the helper functions use more than that chunk size number of elements.

The functions asserted an actual length, but the comments there indicate (to me) that that is not really needed, and just passing in the whole remaining slice should be fine.


src/internal.rs
```diff
@@ -665,7 +665,7 @@ pub(crate) struct Rav1dTaskContext_task_thread {
 
 #[repr(C)]
 pub(crate) struct Rav1dTaskContext {
-    pub f: *const Rav1dFrameContext,
+    pub f: *mut Rav1dFrameContext,
```
folkertdev, line 668: in practice mutation happens through this pointer. I'm assuming c2rust makes a best effort when assigning const/mut to a pointer and it just couldn't figure it out in this case?

kkysen: `c2rust transpile` just copies how C declared things, so this must've been declared as a `const Dav1dFrameContext*` in C.  `const` and `mut` don't really matter for raw pointers in Rust since you can cast between them like in C.  I think it only matters if it originally comes from a `const` or `mut` source.

src/lf_apply.rs
```diff
@@ -786,35 +781,30 @@ pub(crate) unsafe fn rav1d_loopfilter_sbrow_rows<BD: BitDepth>(
...
     if frame_hdr.loopfilter.level_u == 0 && frame_hdr.loopfilter.level_v == 0 {
         return;
     }
     let mut uv_off: ptrdiff_t;
-    level_ptr = ((*f).lf.level).offset((*f).b4_stride * (sby * sbsz >> ss_ver) as isize);
+    level_ptr = &(*f).lf.level[((*f).b4_stride * (sby * sbsz >> ss_ver) as isize) as usize..];
     uv_off = 0 as c_int as ptrdiff_t;
-    x = 0 as c_int;
-    while x < (*f).sb128w {
+    for (x, level_ptr) in (0..(*f).sb128w).zip(suffixes(level_ptr, (32 >> ss_hor) as usize)) {
```
kkysen, line 793:
```suggestion
    for (x, level_ptr) in (0..(*f).sb128w).zip(suffixes(level_ptr, 32 >> ss_hor) {
```

src/decode.rs
```diff
@@ -4515,14 +4515,13 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
...
-        freep(&mut f.lf.level as *mut *mut [u8; 4] as *mut c_void);
+        let _ = mem::take(&mut f.lf.level);
         f.lf.mask =
             malloc(::core::mem::size_of::<Av1Filter>() * num_sb128 as usize) as *mut Av1Filter;
         // over-allocate by 3 bytes since some of the SIMD implementations
         // index this from the level type and can thus over-read by up to 3
-        f.lf.level = malloc(::core::mem::size_of::<[u8; 4]>() * num_sb128 as usize * 32 * 32 + 3)
-            as *mut [u8; 4];
-        if f.lf.mask.is_null() || f.lf.level.is_null() {
+        f.lf.level = vec![[0u8; 4]; num_sb128 as usize * 32 * 32 + 3].into();
```
kkysen, line 4523:
```suggestion
        f.lf.level = vec![[0u8; 4]; num_sb128 as usize * 32 * 32 + 3].into(); // TODO fallible allocation
```

We're ignoring allocation failures now, but in case we need to handle them later as `dav1d` did, we want to mark where they're happening.

src/decode.rs
```diff
@@ -1452,7 +1452,7 @@ unsafe fn decode_b(
     }
 
     let ts = &mut *t.ts;
-    let f = &*t.f;
+    let f = &mut *t.f;
```
kkysen, line 1455: Ask @rinon, but I think getting a `&mut Rav1dFrameContext` here might be difficult when we get rid of the raw pointer.  This is probably okay for now, though.

src/lf_apply.rs
```diff
@@ -715,69 +704,75 @@ pub(crate) unsafe fn rav1d_loopfilter_sbrow_cols<BD: BitDepth>(
...
     let mut ptr: *mut BD::Pixel;
-    let mut level_ptr: *mut [u8; 4] =
-        ((*f).lf.level).offset((*f).b4_stride * sby as isize * sbsz as isize);
-    ptr = *p.offset(0);
-    x = 0 as c_int;
-    while x < (*f).sb128w {
+    let mut level_ptr: &[[u8; 4]] =
+        &(*f).lf.level[((*f).b4_stride * sby as isize * sbsz as isize) as usize..];
+    ptr = p[0];
+    for (x, level_ptr) in (0..(*f).sb128w).zip(suffixes(level_ptr, 32)) {
```
kkysen, line 771: I think this can be done more simply without `suffixes`.  Can you just do `&level_ptr[32 * x as usize..]` with no `zip` or `suffixes`?

kkysen: And the same for the other `suffixes` call.

kkysen: Or you can do `level_ptr = &level_ptr[32..]`.

kkysen (changes requested):
I had some minor comments/suggestions, and for the `suffixes` thing, I think we can do it more simply without the helper function and just normal slicing.  Overall this looks great, though!

kkysen (approved):

PR #727, opened 2024-02-04
`src/lf_apply.rs`: clean up usage of masks

folkertdev:
the commits are self-contained so reviewing per-commit should work well.

### DIFF
--- a/src/lf_apply.rs
+++ b/src/lf_apply.rs
@@ -143,8 +143,7 @@ unsafe fn backup_lpf<BD: BitDepth>(
     } else {
         while row + stripe_h <= row_h {
             let n_lines_0 = 4 - (row + stripe_h + 1 == h) as c_int;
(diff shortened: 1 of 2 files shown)


randomPoison:
Ah, it looks like there's been some confusion about tasking! @fbossen is already working on cleaning up `lf_apply.rs` in #702. Would you be able to clean up one of the other files instead? Any of the ["Cleanup c2rust output"](https://github.com/memorysafety/rav1d/issues?q=is%3Aopen+is%3Aissue+c2rust+output) tasks that aren't already assigned to someone would be good tasks to start with. Really sorry I didn't communicate that earlier.

fbossen:
@folkertdev Unfortunately, looks like there isn't anything left to steal. I pushed my current changes to #728. Let me know if I missed something.

folkertdev:
nothing significant no

PR #728, opened 2024-02-05
`mod lf_apply`: clean up code and use slices instead of raw pointers

### LINKED ISSUE #702: Cleanup c2rust output in lf_apply.rs
* [ ] Cleanup helper functions to take safe arguments where possible (may take raw  pointers as arguments currently).

fbossen:
* Fixes #702

### DIFF
--- a/src/lf_apply.rs
+++ b/src/lf_apply.rs
@@ -612,20 +612,20 @@ pub(crate) unsafe fn rav1d_loopfilter_sbrow_cols<BD: BitDepth>(
             &mut (*lflvl.offset(x as isize)).filter_y[0][bx4 as usize];
         for y in starty4 as u32..endy4 as u32 {
             let mask: u32 = 1 << y;
-            let sidx = (mask >= 0x10000) as c_int;
+            let sidx = (mask >= 0x10000) as usize;
             let smask: c_uint = mask >> (sidx << 4);
-            let idx = 2 as c_int * (y_hmask[2][sidx as usize] as c_uint & smask != 0) as c_int
-                + (y_hmask[1][sidx as usize] as c_uint & smask != 0) as c_int;
-            let ref mut fresh0 = y_hmask[2][sidx as usize];
+            let idx = 2 as c_int * (y_hmask[2][sidx] as c_uint & smask != 0) as c_int
+                + (y_hmask[1][sidx] as c_uint & smask != 0) as c_int;
+            let ref mut fresh0 = y_hmask[2][sidx];
             *fresh0 = (*fresh0 as c_uint & !smask) as u16;
-            let ref mut fresh1 = y_hmask[1][sidx as usize];
+            let ref mut fresh1 = y_hmask[1][sidx];
             *fresh1 = (*fresh1 as c_uint & !smask) as u16;
-            let ref mut fresh2 = y_hmask[0][sidx as usize];
+            let ref mut fresh2 = y_hmask[0][sidx];
             *fresh2 = (*fresh2 as c_uint & !smask) as u16;
             let ref mut fresh3 = y_hmask[cmp::min(
                 idx,
                 lpf_y[y.wrapping_sub(starty4 as c_uint) as usize] as c_int,
-            ) as usize][sidx as usize];
+            ) as usize][sidx];
             *fresh3 = (*fresh3 as c_uint | smask) as u16;
         }
         if f.cur.p.layout != Rav1dPixelLayout::I400 {
@@ -633,17 +633,17 @@ pub(crate) unsafe fn rav1d_loopfilter_sbrow_cols<BD: BitDepth>(
                 &mut (*lflvl.offset(x as isize)).filter_uv[0][cbx4 as usize];
             for y in (starty4 >> ss_ver) as u32..uv_endy4 {
                 let uv_mask: u32 = 1 << y;
-                let sidx_0 = (uv_mask >= vmax) as c_int;
-                let smask_0: c_uint = uv_mask >> (sidx_0 << 4 - ss_ver);
-                let idx_0 = (uv_hmask[1][sidx_0 as usize] as c_uint & smask_0 != 0) as c_int;
-                let ref mut fresh4 = uv_hmask[1][sidx_0 as usize];
+                let sidx = (uv_mask >= vmax) as usize;
+                let smask_0: c_uint = uv_mask >> (sidx << 4 - ss_ver);
+                let idx_0 = (uv_hmask[1][sidx] as c_uint & smask_0 != 0) as c_int;
+                let ref mut fresh4 = uv_hmask[1][sidx];
                 *fresh4 = (*fresh4 as c_uint & !smask_0) as u16;
-                let ref mut fresh5 = uv_hmask[0][sidx_0 as usize];
+                let ref mut fresh5 = uv_hmask[0][sidx];
                 *fresh5 = (*fresh5 as c_uint & !smask_0) as u16;
                 let ref mut fresh6 = uv_hmask[cmp::min(
                     idx_0,
                     lpf_uv[y.wrapping_sub((starty4 >> ss_ver) as c_uint) as usize] as c_int,
-                ) as usize][sidx_0 as usize];
+                ) as usize][sidx];
                 *fresh6 = (*fresh6 as c_uint | smask_0) as u16;
             }
         }
@@ -660,20 +660,19 @@ pub(crate) unsafe fn rav1d_loopfilter_sbrow_cols<BD: BitDepth>(
             let w: c_uint = cmp::min(32 as c_int, f.w4 - (x << 5)) as c_uint;
             for i in 0..w {
                 let mask: u32 = 1 << i;
-                let sidx_1 = (mask >= 0x10000 as c_uint) as c_int;
-                let smask_1: c_uint = mask >> (sidx_1 << 4);
+                let sidx = (mask >= 0x10000) as usize;
+                let smask_1: c_uint = mask >> (sidx << 4);
                 let idx_1 = 2 as c_int
-                    * ((*y_vmask.offset(2))[sidx_1 as usize] as c_uint & smask_1 != 0) as c_int
-                    + ((*y_vmask.offset(1))[sidx_1 as usize] as c_uint & smask_1 != 0) as c_int;
-                let ref mut fresh7 = (*y_vmask.offset(2))[sidx_1 as usize];
+                    * ((*y_vmask.offset(2))[sidx] as c_uint & smask_1 != 0) as c_int
+                    + ((*y_vmask.offset(1))[sidx] as c_uint & smask_1 != 0) as c_int;
+                let ref mut fresh7 = (*y_vmask.offset(2))[sidx];
                 *fresh7 = (*fresh7 as c_uint & !smask_1) as u16;
-                let ref mut fresh8 = (*y_vmask.offset(1))[sidx_1 as usize];
+                let ref mut fresh8 = (*y_vmask.offset(1))[sidx];
                 *fresh8 = (*fresh8 as c_uint & !smask_1) as u16;
-                let ref mut fresh9 = (*y_vmask.offset(0))[sidx_1 as usize];
+                let ref mut fresh9 = (*y_vmask.offset(0))[sidx];
                 *fresh9 = (*fresh9 as c_uint & !smask_1) as u16;
                 let ref mut fresh10 = (*y_vmask
-                    .offset(cmp::min(idx_1, (*a).tx_lpf_y[i as usize] as c_int) as isize))
-                    [sidx_1 as usize];
+                    .offset(cmp::min(idx_1, (*a).tx_lpf_y[i as usize] as c_int) as isize))[sidx];
                 *fresh10 = (*fresh10 as c_uint | smask_1) as u16;
             }
             if f.cur.p.layout != Rav1dPixelLayout::I400 {
@@ -683,17 +682,16 @@ pub(crate) unsafe fn rav1d_loopfilter_sbrow_cols<BD: BitDepth>(
                     .as_mut_ptr();
                 for i in 0..cw {
                     let uv_mask: u32 = 1 << i;
-                    let sidx_2 = (uv_mask >= hmax) as c_int;
-                    let smask_2: c_uint = uv_mask >> (sidx_2 << 4 - ss_hor);
-                    let idx_2 =
-                        ((*uv_vmask.offset(1))[sidx_2 as usize] as c_uint & smask_2 != 0) as c_int;
-                    let ref mut fresh11 = (*uv_vmask.offset(1))[sidx_2 as usize];
+                    let sidx = (uv_mask >= hmax) as usize;
+                    let smask_2: c_uint = uv_mask >> (sidx << 4 - ss_hor);
+                    let idx_2 = ((*uv_vmask.offset(1))[sidx] as c_uint & smask_2 != 0) as c_int;
+                    let ref mut fresh11 = (*uv_vmask.offset(1))[sidx];
                     *fresh11 = (*fresh11 as c_uint & !smask_2) as u16;
-                    let ref mut fresh12 = (*uv_vmask.offset(0))[sidx_2 as usize];
+                    let ref mut fresh12 = (*uv_vmask.offset(0))[sidx];
                     *fresh12 = (*fresh12 as c_uint & !smask_2) as u16;
                     let ref mut fresh13 = (*uv_vmask
                         .offset(cmp::min(idx_2, (*a).tx_lpf_uv[i as usize] as c_int) as isize))
-                        [sidx_2 as usize];
+                        [sidx];
                     *fresh13 = (*fresh13 as c_uint | smask_2) as u16;
                 }
             }
--- a/src/lf_apply.rs
+++ b/src/lf_apply.rs
@@ -392,7 +392,8 @@ unsafe fn filter_plane_cols_y<BD: BitDepth>(
     lvl: &[[u8; 4]],
     b4_stride: ptrdiff_t,
     mask: &[[[u16; 2]; 3]; 32],
-    dst: *mut BD::Pixel,
+    dst: &mut [BD::Pixel],
+    dst_offset: usize,
     ls: ptrdiff_t,
     w: c_int,
     starty4: c_int,
@@ -418,7 +419,7 @@ unsafe fn filter_plane_cols_y<BD: BitDepth>(
             }
             // hmask[3] = 0; already initialized above
             (*dsp).lf.loop_filter_sb[0][0](
-                dst.add(x * 4).cast(),
+                dst.as_mut_ptr().add(dst_offset + x * 4).cast(),
                 ls,
                 hmask.as_mut_ptr(),
                 lvl[x..].as_ptr(),
@@ -475,8 +476,9 @@ unsafe fn filter_plane_cols_uv<BD: BitDepth>(
     lvl: &[[u8; 4]],
     b4_stride: ptrdiff_t,
     mask: &[[[u16; 2]; 2]; 32],
-    u: *mut BD::Pixel,
-    v: *mut BD::Pixel,
+    u: &mut [BD::Pixel],
+    v: &mut [BD::Pixel],
+    uv_offset: usize,
     ls: ptrdiff_t,
     w: c_int,
     starty4: c_int,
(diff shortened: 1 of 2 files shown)


kkysen (commented):
By the way, feel free to split this PR up more if it's taking a while, and it's easier to review it in smaller chunks as well.

src/lf_apply.rs
```diff
@@ -43,8 +44,9 @@ unsafe fn backup_lpf<BD: BitDepth>(
     };
     // The first stripe of the frame is shorter by 8 luma pixel rows.
     let mut stripe_h = ((64 as c_int) << (cdef_backup & sb128)) - 8 * (row == 0) as c_int >> ss_ver;
-    src = src.offset((stripe_h - 2) as isize * BD::pxstride(src_stride as usize) as isize);
-    if c.n_tc == 1 as c_uint {
+    src_offset = src_offset
+        .wrapping_add_signed((stripe_h - 2) as isize * BD::pxstride(src_stride as usize) as isize);
```
kkysen, line 48: Why do we need wrapping here?  These are offsets; they shouldn't get anywhere near `usize::MAX`, right?

src/lf_apply.rs
```diff
@@ -429,50 +437,51 @@ unsafe fn filter_plane_rows_y<BD: BitDepth>(
...
                 endy4 - starty4,
                 (*f).bitdepth_max,
             );
             (*dsp).lf.loop_filter_sb[1][0](
-                v.offset((x * 4) as isize).cast(),
+                v.as_mut_ptr().add(uv_offset + x * 4).cast(),
                 ls,
                 hmask.as_mut_ptr(),
-                lvl[x as usize][3..].as_ptr() as *const [u8; 4],
+                &lvl[x as usize][3] as *const u8 as *const [u8; 4],
```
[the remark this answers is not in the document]
kkysen: A better abstraction:

```rust
#[inline(always)]
fn unaligned_lvl(lvl: &[[u8; 4]], y: usize) -> [u8; 4] {
    // Safety: `[u8; 4]` is transmutable to `u8`s.
    unsafe { lvl.align_to::<u8>() }.1[..y + 4][y..].try_into().unwrap()
}
```

used here as

```suggestion
                unaligned_lvl(&lvl[x..], 3),
```

src/lf_apply.rs
```diff
@@ -534,213 +541,171 @@ unsafe fn filter_plane_rows_uv<BD: BitDepth>(
...
             f,
             have_left,
             level_ptr,
             (*f).b4_stride,
-            ((*lflvl.offset(x as isize)).filter_uv[0]).as_mut_ptr() as *const [[u16; 2]; 2],
-            &mut *p[1].offset(uv_off as isize),
-            &mut *p[2].offset(uv_off as isize),
+            &(*lflvl.offset(x as isize)).filter_uv[0],
+            pu[0],
+            pv[0],
```
kkysen, line 687:
```suggestion
            p[1],
            p[2],
```

Why can't you just do this?

src/lf_apply.rs
```diff
@@ -750,23 +715,23 @@ pub(crate) unsafe fn rav1d_loopfilter_sbrow_rows<BD: BitDepth>(
...
     for x in 0..(*f).sb128w {
         filter_plane_rows_y::<BD>(
             f,
             have_top,
             level_ptr,
             (*f).b4_stride,
-            ((*lflvl.offset(x as isize)).filter_y[1]).as_mut_ptr() as *const [[u16; 2]; 3],
-            ptr,
+            &(*lflvl.offset(x as isize)).filter_y[1],
+            slice.as_mut_ptr().offset(p_offset[0] as isize),
```
kkysen, line 728: If `p_offset` is a `usize` and thus positive, we can just do this, right?

```suggestion
            slice[p_offset[0]..].as_mut_ptr(),
```

src/lf_apply.rs
```diff
@@ -775,19 +740,18 @@ pub(crate) unsafe fn rav1d_loopfilter_sbrow_rows<BD: BitDepth>(
...
             f,
             have_top,
             level_ptr,
             (*f).b4_stride,
-            ((*lflvl.offset(x as isize)).filter_uv[1]).as_mut_ptr() as *const [[u16; 2]; 2],
-            &mut *p[1].offset(uv_off as isize),
-            &mut *p[2].offset(uv_off as isize),
+            &(*lflvl.offset(x as isize)).filter_uv[1],
+            p[1][uv_off as usize..].as_mut_ptr().add(p_offset[1]),
+            p[2][uv_off as usize..].as_mut_ptr().add(p_offset[1]),
```
kkysen, line 754:
```suggestion
            p[1][uv_off as usize + p_offset[1]..].as_mut_ptr(),
            p[2][uv_off as usize + p_offset[1]..].as_mut_ptr(),
```

src/recon.rs
```diff
@@ -4469,22 +4470,42 @@ pub(crate) unsafe fn rav1d_filter_sbrow_deblock_cols<BD: BitDepth>(
...
+    let ss_hor = (f.cur.p.layout as c_uint != Rav1dPixelLayout::I444 as c_int as c_uint) as c_int;
+
+    let datay_stride = BD::pxstride((*f).cur.stride[0] as usize) as isize;
+    let datay_width = (*f).cur.p.w;
+    let datay_height = (*f).cur.p.h;
+    let datay_diff = (datay_height - 1) as isize * datay_stride;
+    let datauv_stride = BD::pxstride((*f).cur.stride[1] as usize) as isize;
+    let datauv_width = datay_width >> ss_hor;
+    let datauv_height = datay_height >> ss_ver;
+    let datauv_diff = (datauv_height - 1) as isize * datauv_stride;
```
kkysen, line 4482: Could you add `_`s after the `data` for these variables?

kkysen: You could also just do all these calculations for `p` and `p_offset` inside an inner block and then drop the `data*` prefixes of all these variables since the variables have a tight scope.

kkysen (changes requested):
This mostly looks good.  I had some small comments, as well as a bigger one concerning the "unaligned" reads of `lvl`, which is UB the way we're currently doing it.

src/lf_apply.rs
```diff
@@ -389,34 +400,31 @@ unsafe fn filter_plane_cols_y<BD: BitDepth>(
...
+                mask[y as usize][1][0] as u32 | (mask[y as usize][1][1] as u32) << 16,
+                mask[y as usize][2][0] as u32 | (mask[y as usize][2][1] as u32) << 16,
+                0,
             ];
             (*dsp).lf.loop_filter_sb[0][1](
                 dst.cast(),
                 ls,
                 vmask.as_ptr(),
-                lvl[0][1..].as_ptr() as *const [u8; 4],
+                &lvl[0][1] as *const u8 as *const [u8; 4],
```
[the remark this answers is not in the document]
kkysen: See #731 for a fix of this.  It's also done as an "unaligned" slice rather than a single "unaligned" read now, as the `fn` ptr reads multiple levels not just one.

src/lf_apply.rs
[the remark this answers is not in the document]
fbossen: We're not going anywhere near `usize::MAX`. To do `a += b` where `a` has type `usize` and `b` has type `isize`, is there a simpler/cleaner way than `a = (a as isize + b) as usize`?

src/lf_apply.rs
```diff
@@ -114,7 +116,7 @@ unsafe fn backup_lpf<BD: BitDepth>(
             ((*(*f).dsp).mc.resize)(
                 dst.cast(),
                 dst_stride,
-                src.cast(),
+                src.as_ptr().add(src_offset).cast(),
```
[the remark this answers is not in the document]
fbossen: The pointer may be indexed with a negative offset in `mc.resize`. I was thus avoiding to use `src[src_offset..]` which can be read to imply that only elements with offset `src_offset` or larger will be accessed.

src/lf_apply.rs
```diff
@@ -429,50 +437,51 @@ unsafe fn filter_plane_rows_y<BD: BitDepth>(
...
-                hmask[0] = (*mask.offset(x as isize))[0][1] as u32;
-                hmask[1] = (*mask.offset(x as isize))[1][1] as u32;
+                hmask[0] = mask[x as usize][0][1] as u32;
+                hmask[1] = mask[x as usize][1][1] as u32;
             }
-            hmask[2] = 0 as c_int as u32;
+            // hmask[2] = 0; Already initialized to 0 above
             (*dsp).lf.loop_filter_sb[1][0](
-                u.offset((x * 4) as isize).cast(),
+                u.as_mut_ptr().add(uv_offset + x * 4).cast(),
```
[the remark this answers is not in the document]
fbossen: If the goal is bounds check, I suppose we should also add a check on the slice end.

src/lf_apply.rs
fbossen: Can't borrow `p[1]` and `p[2]`.

src/recon.rs
fbossen: Moved it to an inner block

fbossen:
> By the way, feel free to split this PR up more if it's taking a while, and it's easier to review it in smaller chunks as well.

The first few commits are now in #739

src/lf_apply.rs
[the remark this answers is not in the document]
kkysen: Ahh, that makes sense.  I'll keep that in mind when I see patterns like this that it means it may be negatively offset in a call.

src/lf_apply.rs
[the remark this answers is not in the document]
kkysen: Following the discussion from https://github.com/memorysafety/rav1d/pull/731#discussion_r1482792829, we definitely want the bounds check on the slice beginning (if negative offsets aren't used in the callee), and subsequent bounds checks (like the end) are up to the callee to do safely.  To do so as safely as possible, we'll pass an extra `.len()` arg in future PRs and re-create the slice in Rust (non-asm) callees.

src/lf_apply.rs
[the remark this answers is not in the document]
kkysen: Fixed in #731.

src/lf_apply.rs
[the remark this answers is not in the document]
kkysen: Fixed in #731.

src/lf_apply.rs
kkysen: It seemed to work fine when I tried changing it to just use `p` directly.

src/lf_apply.rs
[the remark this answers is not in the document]
kkysen: Hmm, they're both kinda ugly.  `a = (a as isize + b) as usize` is done because `b` can be negative, right?  I wish there was just a `.add_signed`.  I think, as unwieldy as `a = (a as isize + b) as usize` is, it might be a little bit better?  Since the `.wrapping` isn't there to confuse a reader as to why it's being done this way.  I'm not totally sure, though.

rinon:
Is this good to go in now? Seems it needs a rebase but anything else beyond that?

kkysen:
> Is this good to go in now? Seems it needs a rebase but anything else beyond that?

I need to re-review it.

kkysen:
> Addresses #702

Does this fix and close #702, or just address part of it?  If it does, can you use one of the [special linking keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) like "fix", "close", or "resolve"?

src/lf_apply.rs
kkysen: My bad, you're right.  You can do it more elegantly, though:

```suggestion
            pu,
            pv,
```

with a `let [_, pu, pv] = p;` array destructure above instead.

kkysen (changes requested):
Just a few little things left.  Also, for the previous unresolved comments, if those were `.add()`s where the resulting pointer is negatively offset later, you can dismiss my comments.  I wasn't sure if that was the case for all of them, though it may.

src/lf_apply.rs
fbossen: That's much more elegant, indeed.

src/lf_apply.rs
fbossen: Not applicable anymore since slice is used instead of raw pointer.

src/lf_apply.rs
fbossen: Not applicable anymore since slice is used instead of raw pointer.

src/lf_apply.rs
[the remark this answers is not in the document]
fbossen: I'm reverting to not using `.wrapping` to avoid unnecessary head scratching.

fbossen:
I ran into some issues with borrowing `f.lf.mask` when rebasing on top of #768 which borrows `f.lf`. I removed the call to `tx_lpf_right_edge()` in 1727e4ee @randomPoison can you have a look?

kkysen:
> I ran into some issues with borrowing `f.lf.mask` when rebasing on top of #768 which borrows `f.lf`. I removed the call to `tx_lpf_right_edge()` in [1727e4e](https://github.com/memorysafety/rav1d/commit/1727e4ee26131ea9c3b835235f3bd200862baa7a) @randomPoison can you have a look?

That looks right.  That said, we could put that `Vec` in its own new type struct and put the method on that.  Then the borrowing should work.  We should probably do the same for other similar `Vec`s and their offsets if they have splitting methods.

src/lf_apply.rs
```diff
@@ -786,17 +786,19 @@ pub(crate) unsafe fn rav1d_loopfilter_sbrow_rows<BD: BitDepth>(
         return;
     }
 
-    let mut uv_off: ptrdiff_t = 0;
+    let mut uv_off: usize = 0;
     let mut level_ptr = &f.lf.level[(f.b4_stride * (sby * sbsz >> ss_ver) as isize) as usize..];
+    let uv = p[1..].split_at_mut(1);
```
kkysen: I think you can array destructure this like we did in another PR.

src/internal.rs
```diff
@@ -451,6 +451,7 @@ pub struct Rav1dFrameContext_lf {
 }
 
 impl Rav1dFrameContext_lf {
+    #[allow(dead_code)]
```
kkysen, line 454: Can we instead do #782?

kkysen (changes requested):

src/lf_apply.rs
```diff
@@ -52,70 +53,60 @@ unsafe fn backup_lpf<BD: BitDepth>(
             let top = (4 as c_int) << sb128;
```
kkysen, line 53:
```suggestion
            let top = 4 << sb128;
```

This should let you avoid some of the extra casts below, I think.

src/lf_apply.rs
```diff
@@ -52,70 +53,60 @@ unsafe fn backup_lpf<BD: BitDepth>(
...
+                dst_offset_delta1 -= BD::pxstride(-dst_stride as usize);
+                dst_tmp.1[dst_offset_delta1..][..dst_w as usize]
+                    .copy_from_slice(&dst_tmp.0[dst_offset_delta0..][..dst_w as usize]);
+                dst_offset_delta0 -= BD::pxstride(-dst_stride as usize);
+                dst_offset_delta1 -= BD::pxstride(-dst_stride as usize);
+                dst_tmp.1[dst_offset_delta1..][..dst_w as usize]
+                    .copy_from_slice(&dst_tmp.0[dst_offset_delta0..][..dst_w as usize]);
+            } else {
+                let mut dst_offset_delta = 0;
+                let dst_tmp = dst.split_at_mut(
```
kkysen, line 77:
```suggestion
                let (dst, top) = dst.split_at_mut(
```

I think this would make it a bit clearer to read.

src/lf_apply.rs
```diff
@@ -52,70 +53,60 @@ unsafe fn backup_lpf<BD: BitDepth>(
...
+                    (dst_offset as isize
+                        + top as isize * BD::pxstride(dst_stride as usize) as isize)
+                        as usize,
+                );
+                dst_tmp.0[dst_offset + dst_offset_delta..][..dst_w as usize]
+                    .copy_from_slice(&dst_tmp.1[dst_offset_delta..][..dst_w as usize]);
```
kkysen, line 81: For these `.copy_from_slice`s, can we still use `BD::pixel_copy`?  `fn pixel_copy` does `dst[..n].copy_from_slice(&src[..n])`, which is what you're doing with `dst_w` anyways.

src/lf_apply.rs
```diff
@@ -44,78 +46,60 @@ unsafe fn backup_lpf<BD: BitDepth>(
...
+                dst_tmp.1[dst_offset_delta1..][..dst_w as usize]
+                    .copy_from_slice(&dst_tmp.0[dst_offset_delta0..][..dst_w as usize]);
+            } else {
+                let mut dst_offset_delta = 0;
+                let dst_tmp = dst.split_at_mut(
+                    (dst_offset as isize + top as isize * BD::pxstride(dst_stride)) as usize,
+                );
+                dst_tmp.0[dst_offset + dst_offset_delta..][..dst_w as usize]
+                    .copy_from_slice(&dst_tmp.1[dst_offset_delta..][..dst_w as usize]);
+                dst_offset_delta = (dst_offset_delta as isize + BD::pxstride(dst_stride)) as usize;
```
kkysen, line 82: Why do we have to do the addition through `isize` here?  `dst_stride >= 0` in this case, so it could be casted to `usize` and we can do everything through `usize`.

src/internal.rs
fbossen: sure

src/lf_apply.rs
fbossen: I rewrote this section of code. Should be cleaner now and address this comment and others

src/lf_apply.rs
```diff
@@ -44,78 +46,55 @@ unsafe fn backup_lpf<BD: BitDepth>(
...
+            let (dst, dst_top) = if dst_stride < 0 {
+                let dst = &mut dst[dst_offset - top_size - 3 * px_abs_stride..];
+                let (dst_top, dst) = dst.split_at_mut(top_size);
+                (dst, dst_top)
+            } else {
+                let dst = &mut dst[dst_offset..];
+                dst.split_at_mut(top_size)
+            };
+
             BD::pixel_copy(
```
kkysen, line 67:
```rust
            for row in 0..4 {
               BD::pixel_copy(
                    &mut dst[i * px_abs_stride..],
                    &dst_top[i * px_abs_stride..],
                    dst_w as usize,
                );
            }
```

kkysen (commented):
Besides integrating #782 and that one loop I suggested, I think it all looks good now!

fbossen:
> Besides integrating #782 and that one loop I suggested, I think it all looks good now!

Also had to deal with rebasing on top of #776 but should be good now.

src/lf_apply.rs
```diff
@@ -126,70 +92,93 @@ unsafe fn backup_lpf<BD: BitDepth>(
...
+                    let dst_tmp = dst.split_at_mut(dst_offset + BD::pxstride(-dst_stride) as usize);
+                    dst_tmp.0[dst_offset..][..dst_w as usize]
+                        .copy_from_slice(&dst_tmp.1[..dst_w as usize]);
+                } else {
+                    let dst_tmp = dst.split_at_mut(dst_offset);
+                    dst_tmp.1[..dst_w as usize].copy_from_slice(
+                        &dst_tmp.0[dst_offset - BD::pxstride(dst_stride) as usize..]
+                            [..dst_w as usize],
+                    );
+                }
```
kkysen, line 111: Could you rewrite this in a similar way you did the last one?  Like using `BD::pixel_copy`, and unifying some of the branches if possible (it was very clean how you did that in the last one).  And in general, for methods like `.split_at_mut`, I think it's clearer to destructure the tuple and give the halves names rather than using `.0` and `.1`?

src/lf_apply.rs
```diff
@@ -126,70 +92,93 @@ unsafe fn backup_lpf<BD: BitDepth>(
...
-                        src_w as usize,
-                    ),
+                    dst_tmp.1,
+                    if i == n_lines_0 {
+                        &dst_tmp.0[dst_offset - BD::pxstride(dst_stride) as usize..]
+                    } else {
+                        &src[src_offset..]
+                    },
                     src_w as usize,
                 );
```
kkysen, line 128: Same here about the `.0` and `.1`.

kkysen (changes requested):
Found a few more things.

randomPoison (approved):
LGTM! Unfortunately there are now merge conflicts from #779. That change was converting pointers to offsets, so hopefully merging them is smooth.

kkysen (approved):

fbossen:
> LGTM! Unfortunately there are now merge conflicts from #779. That change was converting pointers to offsets, so hopefully merging them is smooth.

Had to resolve some conflicts but nothing tedious

PR #729, opened 2024-02-05
`Rav1dFrameContext_frame_thread::b`: Convert to `Vec`

randomPoison:
Converting `b` to a `Box<[]>` means that in `decode_b` we now have a proper borrow of `f.frame_thread.b`, whereas before we had a raw pointer to the borrowed element. This triggered a number of borrowcheck errors since throughout the function we also pass a reference to `f` to different functions, causing a conflict with the mutable borrow of `f.frame_thread.b`. I was able to make some relatively minor changes to `decode_b` to resolve the borrow conflicts, but I'm not sure if the change to how `DEBUG_BLOCK_INFO` is called (i.e. calling it once at the beginning of the function and caching the result) preserves the original behavior. Since `DEBUG_BLOCK_INFO` is currently hard-coded to return false I figured it was more important to fix the borrow issue than to preserve how debug logging was being done, but let me know if there's a better solution.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4529,7 +4529,6 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
         }
         if c.n_fc > 1 {
             freep(&mut f.frame_thread.cbi as *mut *mut CodedBlockInfo as *mut c_void);
-            f.frame_thread.b.clear();
             f.frame_thread
                 .b
                 .resize_with(num_sb128 as usize * 32 * 32, Default::default);
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -1453,7 +1453,7 @@ unsafe fn decode_b(
 
     let ts = &mut *t.ts;
     let f = &mut *t.f;
-    let decode_block_info = DEBUG_BLOCK_INFO(f, t);
+    let dbg = DEBUG_BLOCK_INFO(f, t);
     let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
     let mut b_mem = Default::default();
     let b = if t.frame_thread.pass != 0 {
@@ -1539,7 +1539,7 @@ unsafe fn decode_b(
                     t.warpmv.matrix[5] = b.matrix()[3] as i32 + 0x10000;
                     rav1d_set_affine_mv2d(bw4, bh4, *b.mv2d(), &mut t.warpmv, t.bx, t.by);
                     rav1d_get_shear_params(&mut t.warpmv);
-                    if decode_block_info {
+                    if dbg {
                         println!(
                             "[ {} {} {}\n  {} {} {} ]\n\
                             alpha={}, beta={}, gamma={}, deta={}, mv=y:{},x:{}",
@@ -1683,7 +1683,7 @@ unsafe fn decode_b(
                 }
             }
 
-            if decode_block_info {
+            if dbg {
                 println!("Post-segid[preskip;{}]: r={}", b.seg_id, ts.msac.rng);
             }
 
@@ -1704,7 +1704,7 @@ unsafe fn decode_b(
         b.skip_mode =
             rav1d_msac_decode_bool_adapt(&mut ts.msac, &mut ts.cdf.m.skip_mode.0[smctx as usize])
                 as u8;
-        if decode_block_info {
+        if dbg {
             println!("Post-skipmode[{}]: r={}", b.skip_mode, ts.msac.rng);
         }
     } else {
@@ -1718,7 +1718,7 @@ unsafe fn decode_b(
         let sctx = (*t.a).skip[bx4 as usize] + t.l.skip[by4 as usize];
         b.skip =
             rav1d_msac_decode_bool_adapt(&mut ts.msac, &mut ts.cdf.m.skip[sctx as usize]) as u8;
-        if decode_block_info {
+        if dbg {
             println!("Post-skip[{}]: r={}", b.skip, ts.msac.rng);
         }
     }
@@ -1780,7 +1780,7 @@ unsafe fn decode_b(
 
         seg = Some(&frame_hdr.segmentation.seg_data.d[b.seg_id as usize]);
 
-        if decode_block_info {
+        if dbg {
             println!("Post-segid[postskip;{}]: r={}", b.seg_id, ts.msac.rng);
         }
     }
@@ -1806,7 +1806,7 @@ unsafe fn decode_b(
                 *(t.cur_sb_cdef_idx_ptr).offset(idx + 3) = v;
             }
 
-            if decode_block_info {
+            if dbg {
                 println!(
                     "Post-cdef_idx[{}]: r={}",
                     *t.cur_sb_cdef_idx_ptr, ts.msac.rng
@@ -1845,7 +1845,7 @@ unsafe fn decode_b(
                 delta_q *= 1 << frame_hdr.delta.q.res_log2;
             }
             ts.last_qidx = iclip(ts.last_qidx + delta_q, 1, 255);
-            if have_delta_q && decode_block_info {
+            if have_delta_q && dbg {
                 println!(
                     "Post-delta_q[{}->{}]: r={}",
                     delta_q, ts.last_qidx, ts.msac.rng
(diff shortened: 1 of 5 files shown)


randomPoison:
As I continue to work on the fields of `Rav1dFrameContext_frame_thread`, it occurs to me that we could simplify this PR (and likely the ones that follow it) by not fully making references to the field safe. The borrow checking issues I ran into in `decode_b` were caused by changing the local `b` variable to be a `&mut Av1Block`. I could have left that variable a raw pointer and put off further cleanup in `decode_b` for later.

I'm not sure that's the right call to make for `decode_b`, since we've already largely cleaned up that function and so part of the point of making these fields safe is to allow us to remove the unsafe pointer operations in functions like `decode_b`. However, it looks like there are similar borrowing issues in `recon.rs`, which largely hasn't been cleaned up yet. So maybe it would make sense to leave the raw pointer operations in `recon.rs` until we can do a cleanup pass on those functions?

src/levels.rs
```diff
@@ -308,7 +309,7 @@ impl Default for Av1Block_intra_inter {
     }
 }
 
-#[derive(Default)]
+#[derive(Default, Clone, Copy)]
```
kkysen, line 312: This is a bit big to be `Copy`.  Why do we need it to be `Copy`?  The `union` fields are only `Copy` because they have to be.  If it's because of `vec![]`, we can write it as `.map` and `.collect` instead.

src/lib.rs
```diff
@@ -952,7 +951,7 @@ impl Drop for Rav1dContext {
                         &mut (*f).tile_thread.lowest_pixel_mem as *mut *mut [[c_int; 2]; 7]
                             as *mut c_void,
                     );
-                    freep(&mut (*f).frame_thread.b as *mut *mut Av1Block as *mut c_void);
+                    mem::take(&mut (*f).frame_thread.b); // TODO: remove when context is owned
```
kkysen, line 954:
```suggestion
                    let _ = mem::take(&mut (*f).frame_thread.b); // TODO: remove when context is owned
```

Not strictly needed but we've been writing it like this for `take`s that should be done automatically in `impl Drop`s, which helps with searchability.

src/decode.rs
```diff
@@ -1453,13 +1453,11 @@ unsafe fn decode_b(
 
     let ts = &mut *t.ts;
     let f = &mut *t.f;
+    let decode_block_info = DEBUG_BLOCK_INFO(f, t);
```
kkysen, line 1456: Can you name this just `dbg`?  I named a similarly hoisted variable the same before.

kkysen: Also, to answer your question, it should be safe to precompute like this if `f.bx` and `f.by` don't change at all in the function.

src/decode.rs
```diff
@@ -4526,15 +4545,12 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
             return Err(ENOMEM);
         }
         if c.n_fc > 1 {
-            freep(&mut f.frame_thread.b as *mut *mut Av1Block as *mut c_void);
             freep(&mut f.frame_thread.cbi as *mut *mut CodedBlockInfo as *mut c_void);
-            f.frame_thread.b =
-                malloc(::core::mem::size_of::<Av1Block>() * num_sb128 as usize * 32 * 32)
-                    as *mut Av1Block;
+            f.frame_thread.b = vec![Default::default(); num_sb128 as usize * 32 * 32].into(); // TODO: fallible allocation
```
kkysen, line 4549:
```suggestion
            f.frame_thread.b.resize_with(num_sb128 as usize * 32 * 32, Default::default);
```

If `b` is a `Vec<Av1Block>` instead, this cuts down on some of the extra work we have to do here in default initializing all of the elements for safety, as we only need to default initialize new elements.  Also, this reuses the allocation, something the C didn't do, but is very reasonable to do.  Also, by doing it this way, we can remove the `#[derive(Clone, Copy)]` on `Av1Block` (32 bytes) and `Av1Block_intra_inter`.

kkysen (changes requested):

kkysen:
> As I continue to work on the fields of `Rav1dFrameContext_frame_thread`, it occurs to me that we could simplify this PR (and likely the ones that follow it) by not fully making references to the field safe. The borrow checking issues I ran into in `decode_b` were caused by changing the local `b` variable to be a `&mut Av1Block`. I could have left that variable a raw pointer and put off further cleanup in `decode_b` for later.
> 
> I'm not sure that's the right call to make for `decode_b`, since we've already largely cleaned up that function and so part of the point of making these fields safe is to allow us to remove the unsafe pointer operations in functions like `decode_b`. However, it looks like there are similar borrowing issues in `recon.rs`, which largely hasn't been cleaned up yet. So maybe it would make sense to leave the raw pointer operations in `recon.rs` until we can do a cleanup pass on those functions?

I think it's still best to make the fields safe now.  We need to make them safe eventually one way or another, and we've cleaned up enough to understand most of the access patterns, I think.  Once we make the fields safe, cleaning up `mod recon`, for example, will become much easier I imagine.

src/decode.rs
```diff
@@ -1621,8 +1619,15 @@ unsafe fn decode_b(
     if frame_hdr.segmentation.enabled != 0 {
         if frame_hdr.segmentation.update_map == 0 {
             if !(f.prev_segmap).is_null() {
-                let seg_id =
-                    get_prev_frame_segid(f, t.by, t.bx, w4, h4, f.prev_segmap, f.b4_stride);
+                let seg_id = get_prev_frame_segid(
+                    f.frame_hdr.as_ref().unwrap(),
```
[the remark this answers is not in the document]
randomPoison: Heh, turns out we're already doing that at the top of the function, I just missed that there was already a `frame_hdr` variable.

src/levels.rs
randomPoison: Fair point. I needed the type to be `Clone` so that I could use `vec!` to initialize the field, but I don't need it to be copy. Will remove the derive.

kkysen: You don't need to use `vec![]`, though, that was my point.

src/decode.rs
randomPoison: `dbg` works for me 👍

And I did check `decode_b` to see if `f.bx`/`by` change, and it doesn't seem like it? I see it being modified in `decode_sb` and a few other places, but not `decode_b`.

src/decode.rs
randomPoison: Hrm, that wouldn't reset the existing elements of the array though right? The original version was resetting all the elements, whereas using `resize_with` would only give you default values for new elements that were added. But we could do `clear` first and then `resize_with`, which should have the same effect. I'm also in favor of reusing the allocation, so seems like a good change to make 👍

kkysen: The C code didn't initialize the new elements when it allocates.  But in Rust, we have to initialize them for safety.  But it doesn't matter what value we initialize them to.  Your way default initializes them, but we could instead only default initialize new elements.  For existing elements, they can be initialized to whatever old values were there, no matter what they are.  So we save on the cost of having to default initialize everything, and only do so for new elements, which is more similar to how C doesn't initialize anything.

src/levels.rs
randomPoison: Yup, I realized that after writing the comment, sorry. I forgot to remove the derive on `Av1Block_intra_inter` but that's fixed now.

src/decode.rs
randomPoison: I removed the `clear` call and left just `resize_with` 👍

src/decode.rs
```diff
@@ -4526,15 +4528,14 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
...
         }
         if c.n_fc > 1 {
-            freep(&mut f.frame_thread.b as *mut *mut Av1Block as *mut c_void);
             freep(&mut f.frame_thread.cbi as *mut *mut CodedBlockInfo as *mut c_void);
-            f.frame_thread.b =
-                malloc(::core::mem::size_of::<Av1Block>() * num_sb128 as usize * 32 * 32)
-                    as *mut Av1Block;
+            f.frame_thread
+                .b
+                .resize_with(num_sb128 as usize * 32 * 32, Default::default);
```
rinon, line 4534: Does this size ever change? Should this be a boxed slice rather than a vector?

rinon: Nevermind, answered my own question, it can.

rinon:
Looks good to me, unless Khyber has any further comments. You should update the title, as it's now a Vec.

src/decode.rs
kkysen, line 4534:
```suggestion
                .resize_with(num_sb128 as usize * 32 * 32, Default::default); // TODO fallible allocation
```

kkysen (approved):
You accidentally deleted the `// TODO fallible allocation` comment but otherwise it all LGTM!

src/decode.rs
kkysen: I see you added it in #730 instead; that's fine.

rinon (approved):

PR #730, opened 2024-02-06
`Rav1dFrameContext_frame_thread::cbi`: Convert to a `Vec<_>`

randomPoison:


src/recon.rs
```diff
@@ -1708,7 +1707,7 @@ unsafe fn read_coef_tree<BD: BitDepth>(
         let mut cf_ctx: u8 = 0;
         let eob;
         let cf: *mut BD::Coef;
-        let mut cbi: *mut CodedBlockInfo = 0 as *mut CodedBlockInfo;
+        let mut cbi = None;
```
kkysen, line 1710:
```suggestion
        let mut cbi = &mut Default::default();
```

`CodedBlockInfo` is small (10 bytes), so creating a `Default::default()` instead of a `None` has minimal overhead.  This way no `.unwrap()`s are needed.  A more complex solution tying together `frame_thread.pass` as an `enum` to the `CodedBlockInfo` works, too, but is more complex and probably equally efficient.

kkysen (changes requested):

src/internal.rs
```diff
@@ -351,6 +351,7 @@ impl Rav1dFrameContext_bd_fn {
     }
 }
 
+#[derive(Default, Clone, Copy)]
```
kkysen, line 354:
```suggestion
#[derive(Default)]
```

Both the `Box<[_]>` and `Vec<_>` can be written without `Clone` or `Copy`, so I wouldn't add them for that.

rinon:
LGTM, again unless Khyber has anything additional. I'd like to get these landed ASAP as I will need to rebase my current out-of-tree work on these.

kkysen (approved):

PR #732, opened 2024-02-07
`Rav1dFrameContext_frame_thread::pal`: Make into an aligned `Vec`

randomPoison:
This also adds `AlignedVec64` to have the `Vec` aligned, since a `Vec<Align64<T>>` would add extra padding between each element since it would align each element, and we only want to align the whole `Vec`/slice.

### DIFF
--- a/src/align.rs
+++ b/src/align.rs
@@ -5,8 +5,15 @@
 //! make them easier to use in common cases, e.g. [`From`] and
 //! [`Index`]/[`IndexMut`] (since it's usually array fields that require
 //! specific aligment for use with SIMD instructions).
+
+use std::marker::PhantomData;
+use std::mem;
+use std::mem::MaybeUninit;
+use std::ops::Deref;
+use std::ops::DerefMut;
 use std::ops::Index;
 use std::ops::IndexMut;
+use std::slice;
 
 /// [`Default`] isn't `impl`emented for all arrays `[T; N]`
 /// because they were implemented before `const` generics
@@ -87,3 +94,99 @@ def_align!(8, Align8);
 def_align!(16, Align16);
 def_align!(32, Align32);
 def_align!(64, Align64);
+
+/// A [`Vec`] that uses a 64-byte aligned allocation.
+///
+/// Only works with [`Copy`] types so that we don't have to handle drop logic.
+pub struct AlignedVec64<T: Copy> {
+    inner: Vec<MaybeUninit<Align64<[u8; 64]>>>,
+
+    /// The number of `T`s in [`Self::inner`] currently initialized.
+    len: usize,
+    _phantom: PhantomData<T>,
+}
+
+impl<T: Copy> AlignedVec64<T> {
+    pub const fn new() -> Self {
+        Self {
+            inner: Vec::new(),
+            len: 0,
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Returns the number of elements in the vector.
+    pub fn len(&self) -> usize {
+        self.len
+    }
+
+    pub fn as_ptr(&self) -> *const T {
+        self.inner.as_ptr().cast()
+    }
+
+    pub fn as_mut_ptr(&mut self) -> *mut T {
+        self.inner.as_mut_ptr().cast()
+    }
+
+    /// Extracts a slice containing the entire vector.
+    pub fn as_slice(&self) -> &[T] {
+        // Safety: The first `len` elements have been initialized to `T`s in
+        // `Self::resize_with`.
+        unsafe { slice::from_raw_parts(self.as_ptr(), self.len) }
+    }
+
+    /// Extracts a mutable slice of the entire vector.
+    pub fn as_mut_slice(&mut self) -> &mut [T] {
+        // Safety: The first `len` elements have been initialized to `T`s in
+        // `Self::resize_with`.
+        unsafe { slice::from_raw_parts_mut(self.as_mut_ptr(), self.len) }
+    }
+
+    pub fn resize(&mut self, new_len: usize, value: T) {
+        let old_len = self.len();
+
+        // Resize the underlying vector to have enough chunks for the new length.
+        //
+        // NOTE: We don't need to `drop` any elements if the `Vec` is truncated since
+        // `T: Copy`.
+        let new_bytes = mem::size_of::<T>() * new_len;
+        let new_chunks = if (new_bytes % 64) == 0 {
+            new_bytes / 64
+        } else {
+            (new_bytes / 64) + 1
+        };
+        self.inner.resize_with(new_chunks, MaybeUninit::uninit);
+
+        // If we grew the vector, initialize the new elements past `len`.
+        for offset in old_len..new_len {
+            // SAFETY: We've allocated enough space to write up to `new_len` elements into
+            // the buffer.
+            unsafe {
+                self.as_mut_ptr().add(offset).write(value);
+            }
+        }
+
+        self.len = new_len;
+    }
+}
+
+impl<T: Copy> Deref for AlignedVec64<T> {
+    type Target = [T];
+
+    fn deref(&self) -> &Self::Target {
+        self.as_slice()
+    }
+}
+
+impl<T: Copy> DerefMut for AlignedVec64<T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.as_mut_slice()
+    }
+}
+
+// NOTE: Custom impl so that we don't require `T: Default`.
+impl<T: Copy> Default for AlignedVec64<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
--- a/src/align.rs
+++ b/src/align.rs
@@ -38,6 +38,7 @@ impl_ArrayDefault!(u8);
 impl_ArrayDefault!(i8);
 impl_ArrayDefault!(i16);
 impl_ArrayDefault!(i32);
+impl_ArrayDefault!(u16);
 
 macro_rules! def_align {
     ($align:literal, $name:ident) => {
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -2262,7 +2262,7 @@ unsafe fn decode_b_inner(
                 let pal = if t.frame_thread.pass != 0 {
                     let index = ((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
                         + ((t.bx >> 1) + (t.by & 1)) as isize;
-                    &f.frame_thread.pal[index as usize]
+                    &f.frame_thread.pal[index as usize].0
                 } else {
                     &t.scratch.c2rust_unnamed_0.pal
                 };
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -2262,7 +2262,7 @@ unsafe fn decode_b_inner(
                 let pal = if t.frame_thread.pass != 0 {
                     let index = ((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
                         + ((t.bx >> 1) + (t.by & 1)) as isize;
-                    &f.frame_thread.pal[index as usize].0
+                    &f.frame_thread.pal[index as usize]
                 } else {
                     &t.scratch.c2rust_unnamed_0.pal
                 };
@@ -4383,12 +4383,10 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
         }
 
         if frame_hdr.allow_screen_content_tools != 0 {
-            if num_sb128 as usize != f.frame_thread.pal_sz() {
-                // TODO: Fallible allocation
-                f.frame_thread
-                    .pal
-                    .resize_with(num_sb128 as usize * 16 * 16, Default::default);
-            }
+            // TODO: Fallible allocation
+            f.frame_thread
+                .pal
+                .resize(num_sb128 as usize * 16 * 16, Default::default());
 
             let pal_idx_sz = num_sb128 * size_mul[1] as c_int;
             if pal_idx_sz != f.frame_thread.pal_idx_sz {
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4383,12 +4383,11 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
         }
 
         if frame_hdr.allow_screen_content_tools != 0 {
-            if num_sb128 != f.frame_thread.pal_sz {
+            if num_sb128 as usize != f.frame_thread.pal_sz() {
                 // TODO: Fallible allocation
                 f.frame_thread
                     .pal
                     .resize_with(num_sb128 as usize * 16 * 16, Default::default);
-                f.frame_thread.pal_sz = num_sb128;
             }
 
             let pal_idx_sz = num_sb128 * size_mul[1] as c_int;
@@ -4408,7 +4407,6 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
             let _ = mem::take(&mut f.frame_thread.pal);
             rav1d_freep_aligned(&mut f.frame_thread.pal_idx as *mut *mut u8 as *mut c_void);
             f.frame_thread.pal_idx_sz = 0;
-            f.frame_thread.pal_sz = f.frame_thread.pal_idx_sz;
         }
     }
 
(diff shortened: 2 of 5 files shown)


src/internal.rs
```diff
@@ -369,7 +369,7 @@ pub struct Rav1dFrameContext_frame_thread {
     pub b: Vec<Av1Block>,
     pub cbi: Vec<CodedBlockInfo>,
     // indexed using (t->by >> 1) * (f->b4_stride >> 1) + (t->bx >> 1)
-    pub pal: *mut [[u16; 8]; 3], /* [3 plane][8 idx] */
+    pub pal: Vec<[[u16; 8]; 3]>, /* [3 plane][8 idx] */
```
rinon, line 372: Do we need an explicit alignment on this? We're replacing `alloc_aligned` with the vector allocation.

kkysen:
```suggestion
    pub pal: Vec<Align64<[[u16; 8]; 3]>>, /* [3 plane][8 idx] */
```

src/decode.rs
```diff
@@ -4388,8 +4377,8 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
                 }
                 f.frame_thread.pal_idx_sz = pal_idx_sz;
             }
-        } else if !f.frame_thread.pal.is_null() {
-            rav1d_freep_aligned(&mut f.frame_thread.pal as *mut *mut [[u16; 8]; 3] as *mut c_void);
+        } else if !f.frame_thread.pal.is_empty() {
+            let _ = mem::take(&mut f.frame_thread.pal);
             rav1d_freep_aligned(&mut f.frame_thread.pal_idx as *mut *mut u8 as *mut c_void);
             f.frame_thread.pal_idx_sz = 0;
             f.frame_thread.pal_sz = f.frame_thread.pal_idx_sz;
```
kkysen, line 4384: `.pal_sz` can be replaced with a method,

```rust
impl Rav1dFrameContext_frame_thread {
    pub fn pal_sz(&self) -> usize {
        self.pal.len() / (16 * 16)
    }
}
```

kkysen (changes requested):
@rinon made a good point about the alignment; we should use `Align64`.  Also, the `pal_sz` field can now be folded into `pal.len()`, after adjusting by `16 * 16`.

src/decode.rs
```diff
@@ -4388,8 +4377,8 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
                 }
                 f.frame_thread.pal_idx_sz = pal_idx_sz;
             }
-        } else if !f.frame_thread.pal.is_null() {
-            rav1d_freep_aligned(&mut f.frame_thread.pal as *mut *mut [[u16; 8]; 3] as *mut c_void);
+        } else if !f.frame_thread.pal.is_empty() {
+            let _ = mem::take(&mut f.frame_thread.pal);
             rav1d_freep_aligned(&mut f.frame_thread.pal_idx as *mut *mut u8 as *mut c_void);
```
rinon, line 4382: Is `pal_idx` a pointer into `pal`? If so, we shouldn't be freeing it.

rinon: Nevermind, I see that it is allocated separately. Weird naming. I guess we need to have another PR to switch it over too.

src/internal.rs
randomPoison: Ah good catch! Fixed.

src/decode.rs

kkysen (commented):
nit: I think you meant `pal_sz` instead of `pal_idx` in the commit message.

kkysen (approved):

src/align.rs
```diff
@@ -86,3 +90,79 @@ def_align!(8, Align8);
...
+    }
+
+    pub fn as_mut_ptr(&mut self) -> *mut T {
+        self.inner.as_mut_ptr().cast()
+    }
+
+    /// Extracts a slice containing the entire vector.
+    pub fn as_slice(&self) -> &[T] {
+        // SAFETY: The first `len` elements of the buffer have been initialized.
+        unsafe { slice::from_raw_parts(self.as_ptr(), self.len) }
```
kkysen, line 121: We can't know `T` is transmutable from `u8`s.  There'd have to be some sort of bound, like `T: bytemuck::AnyBitPattern`.

kkysen (changes requested):
See also #744 for a discussion of how to approach this for all aligned allocs.

randomPoison:
@kkysen @rinon I've changed to the custom allocator approach proposed in #744. Please give this another look!

src/align.rs
```diff
@@ -87,3 +90,79 @@ def_align!(8, Align8);
 def_align!(16, Align16);
 def_align!(32, Align32);
 def_align!(64, Align64);
+
+/// A `Vec` that uses a 64 byte aligned allocation.
+///
+/// Only works with `Copy` types so that we don't have to handle drop logic.
+#[derive(Default)]
```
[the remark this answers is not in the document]
kkysen: Also, we actually want a manual `impl Default` here so that we don't require `T: Default`.

src/align.rs
```diff
@@ -87,3 +90,79 @@ def_align!(8, Align8);
...
+pub struct AlignedVec64<T: Copy> {
+    inner: Vec<MaybeUninit<Align64<[u8; 64]>>>,
```
kkysen: I thought of a way to make this generic over the alignment by implementing a sealed trait (generic `C` for chunk) for `AlignN<[u8; N]>` and then using `mem::size_of::<C>()` instead of `64`.  I can open a PR for that if you think that's a good idea; I'm not sure how to explain it super well.

kkysen (changes requested):
Taking a look at this again, I realized we don't actually need bounds on `T` for safe transmutability, since we leave the `u8`s uninitialized and only ever initialize things as `T`s.  I had a bunch of other comments, though, and I thought of a way to make it generic over the alignment.

src/align.rs
```diff
@@ -87,3 +90,79 @@ def_align!(8, Align8);
...
+
+    pub fn resize_with<F>(&mut self, new_len: usize, mut f: F)
+    where
+        F: FnMut() -> T,
+    {
+        let len = self.len();
+
+        if new_len > len {
+            // Resize the underlying vector to have enough capacity for the new length.
+            let new_chunks = (mem::size_of::<T>() * new_len / 64) + 1;
```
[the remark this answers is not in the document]
randomPoison: That's correct. I did the unconditional `+1` for simplicity, but I've changed it to an `if` block only adding the extra chunk of allocation if it's needed.

src/align.rs
```diff
@@ -87,3 +90,79 @@ def_align!(8, Align8);
 def_align!(16, Align16);
 def_align!(32, Align32);
 def_align!(64, Align64);
+
+/// A `Vec` that uses a 64 byte aligned allocation.
+///
+/// Only works with `Copy` types so that we don't have to handle drop logic.
```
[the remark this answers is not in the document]
randomPoison: As discussed in the sync meeting, we're going to hold off on `Drop` support since we don't seem to need it for any of the types we expect to put in an `AlignedVec64`. We can revisit this if we end up needing drop support.

src/align.rs
```diff
@@ -87,3 +90,79 @@ def_align!(8, Align8);
...
+            for offset in len..new_len {
+                // SAFETY: We've allocated enough space to write up to `new_len` elements into the buffer.
+                unsafe {
+                    self.as_mut_ptr().add(offset).write(f());
+                }
+
+                self.len += 1;
+            }
+        } else {
+            self.len = new_len;
```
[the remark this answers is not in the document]
randomPoison: I moved the update to `self.len` to the end of the function. The original logic for updating by 1 each loop came from the original `Vec` logic in order to be panic safe when running destructors, but we don't need support for that and so can do the simpler thing.

Likewise I've updated the function to always call `resize_with` on the underlying vector to make sure it has a chance to reduce its size.

src/align.rs
```diff
@@ -87,3 +90,79 @@ def_align!(8, Align8);
...
+
+            // Initialize the new elements past `len`.
+            for offset in len..new_len {
+                // SAFETY: We've allocated enough space to write up to `new_len` elements into the buffer.
+                unsafe {
+                    self.as_mut_ptr().add(offset).write(f());
+                }
+
+                self.len += 1;
+            }
```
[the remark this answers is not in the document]
randomPoison: I'd prefer to stick with the approach I already have using a `for` loop and `ptr::write` to initialize the elements. I think it's a more straightforward and comprehensible approach. Converting the allocation to a `&[MaybeUninit<T>]` seems more verbose and less obviously correct to me.

src/align.rs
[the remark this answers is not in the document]
randomPoison: Good point about the `Default` impl, I've moved that to a manual impl. I've also added a const `new` fn, and implemented `Default` using it.

src/align.rs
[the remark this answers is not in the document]
kkysen: Ah, that was the panic safety stuff you were talking about.  I didn't realize it came from `Vec`'s sources?  Is it still an issue if `f` can panic, though, even though the destructor can't?  We can similarly avoid that issue by changing `fn resize_with` to `fn resize` with a `value: T` arg that is copied, right?

src/align.rs
```diff
@@ -86,3 +91,98 @@ def_align!(8, Align8);
 def_align!(16, Align16);
 def_align!(32, Align32);
 def_align!(64, Align64);
+
+/// A [`Vec`] that uses a 64-byte aligned allocation.
+///
+/// Only works with `Copy` types so that we don't have to handle drop logic.
```
kkysen, line 97: nit:

```suggestion
/// Only works with [`Copy`] types so that we don't have to handle drop logic.
```

src/align.rs
```diff
@@ -87,3 +90,79 @@ def_align!(8, Align8);
...
 def_align!(32, Align32);
 def_align!(64, Align64);
+
+/// A `Vec` that uses a 64 byte aligned allocation.
+///
+/// Only works with `Copy` types so that we don't have to handle drop logic.
+#[derive(Default)]
+pub struct AlignedVec64<T: Copy> {
+    inner: Vec<MaybeUninit<Align64<[u8; 64]>>>,
+    len: usize,
```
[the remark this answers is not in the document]
kkysen: In a discussion, we realized this reverse conversion can't be done, as information is lost.

src/align.rs
```diff
@@ -86,3 +91,98 @@ def_align!(8, Align8);
 def_align!(16, Align16);
 def_align!(32, Align32);
 def_align!(64, Align64);
+
+/// A [`Vec`] that uses a 64-byte aligned allocation.
+///
+/// Only works with `Copy` types so that we don't have to handle drop logic.
+pub struct AlignedVec64<T: Copy> {
+    inner: Vec<MaybeUninit<Align64<[u8; 64]>>>,
+    len: usize,
```
kkysen, line 100:
```suggestion
    
    /// The number of `T`s in [`Self::inner`] currently initialized.
    len: usize,
```

src/align.rs
```diff
@@ -86,3 +91,98 @@ def_align!(8, Align8);
...
+    {
+        let len = self.len();
+
+        // Resize the underlying vector to have enough chunks for the new length.
+        let new_bytes = mem::size_of::<T>() * new_len;
+        let new_chunks = if (new_bytes % 64) == 0 {
+            new_bytes / 64
+        } else {
+            (new_bytes / 64) + 1
+        };
```
kkysen, line 152:
```suggestion
        let new_chunks = new_bytes / 64 + if (new_bytes % 64) == 0 { 0 } else { 1 };
```

src/decode.rs
```diff
@@ -4392,19 +4388,11 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
...
-                    f.frame_thread.pal_sz = 0;
-                    return Err(ENOMEM);
-                }
-                f.frame_thread.pal_sz = num_sb128;
+            if num_sb128 as usize != f.frame_thread.pal_sz() {
+                // TODO: Fallible allocation
+                f.frame_thread
+                    .pal
+                    .resize_with(num_sb128 as usize * 16 * 16, Default::default);
             }
```
kkysen, line 4396:
```suggestion
            // TODO: Fallible allocation
            f.frame_thread
                .pal
                .resize_with(num_sb128 as usize * 16 * 16, Default::default);
```

`.resize_with` will do the check.

src/align.rs
```diff
@@ -86,3 +91,98 @@ def_align!(8, Align8);
...
+        self.inner.resize_with(new_chunks, MaybeUninit::uninit);
+
+        // If we grew the vector, initialize the new elements past `len`.
+        if new_len > len {
+            for offset in len..new_len {
```
kkysen, line 157: The loop does the `new_len > len` check already.  Though maybe it's still useful to be more clear about what's meant?

src/align.rs
```diff
@@ -86,3 +91,98 @@ def_align!(8, Align8);
...
+        let len = self.len();
+
+        // Resize the underlying vector to have enough chunks for the new length.
+        let new_bytes = mem::size_of::<T>() * new_len;
+        let new_chunks = if (new_bytes % 64) == 0 {
+            new_bytes / 64
+        } else {
+            (new_bytes / 64) + 1
+        };
+        self.inner.resize_with(new_chunks, MaybeUninit::uninit);
```
kkysen, line 153:
```suggestion
        
        // We don't need to `drop` any elements if the `Vec` is truncated since `T: Copy`.
        
        self.inner.resize_with(new_chunks, MaybeUninit::uninit);
```

I think it's useful to leave this comment to make our assumptions/optimizations clearer and so we don't forget if we remove `T: Copy`.

src/align.rs
```diff
@@ -86,3 +91,98 @@ def_align!(8, Align8);
...
+
+    pub fn as_ptr(&self) -> *const T {
+        self.inner.as_ptr().cast()
+    }
+
+    pub fn as_mut_ptr(&mut self) -> *mut T {
+        self.inner.as_mut_ptr().cast()
+    }
+
+    /// Extracts a slice containing the entire vector.
```
kkysen, line 126: What do you mean by "Extracts" here?  IMO, we don't really need a comment since it's copying `Vec`s API and "extract" makes it a bit more confusing.

src/align.rs
kkysen: We decided I'll add this in a follow-up PR instead.

src/align.rs
rinon: I personally find the current version much easier to read than adding 0 or 1.

src/align.rs
```diff
@@ -86,3 +91,98 @@ def_align!(8, Align8);
...
+        // Safety: The first `len` elements have been initialized to `T`s in
+        // `Self::resize_with`.
+        unsafe { slice::from_raw_parts_mut(self.as_mut_ptr(), self.len) }
+    }
+
+    pub fn resize_with<F>(&mut self, new_len: usize, mut f: F)
+    where
+        F: FnMut() -> T,
+    {
+        let len = self.len();
```
rinon, line 144: nit: `old_len`, perhaps? Just makes a few things clearer.

rinon:
LGTM overall, minor nitpicks aside.

src/align.rs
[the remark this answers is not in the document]
randomPoison: I don't think panic safety is an issue here because `T` doesn't need to be dropped, and I think the issues around panic safety all relate to making sure invalid data isn't dropped (`Vec` generally leaks destructors in the case of a panic). But you're right that `resize` makes more sense than `resize_with` since `T: Copy`. I'll make that change.

src/align.rs
randomPoison: This is the wording that the corresponding `Vec` function uses: https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.as_slice

src/align.rs
randomPoison: Ah, I didn't realize you could do a backwards range, e.g. `5..1`, I assumed that would panic or something. I'll remove the redundant `if` check, I think the comment makes it clear enough what's happening.

src/align.rs
kkysen: Oh that is weird wording for `Vec` then.  It's fine for us then.

kkysen (approved):

rinon (approved):

kkysen:
* Fixes `pal` field for #744.

PR #735, opened 2024-02-09
`Rav1dFrameContext_lf::level`: Make into a `Vec` and fix padding size

randomPoison:
Change `level` into a `Vec` (instead of a `Box<[]>`) and use `resize` to reallocate. This makes the field more consistent with how we've been translating similar fields.

I also fixed a minor error in how the array was being over-allocated, i.e. the original C over-allocated 3 **bytes**, whereas we were allocating an extra 3 **elements** (12 bytes). We now over-allocate a single element (4 bytes).

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4524,9 +4524,10 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
         let _ = mem::take(&mut f.lf.level);
         f.lf.mask =
             malloc(::core::mem::size_of::<Av1Filter>() * num_sb128 as usize) as *mut Av1Filter;
-        // over-allocate by 3 bytes since some of the SIMD implementations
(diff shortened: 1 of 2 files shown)


src/decode.rs
```diff
@@ -4524,9 +4524,10 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
         let _ = mem::take(&mut f.lf.level);
```
kkysen, line 4524: We also need to remove this (the `.clear()` effectively).

kkysen (changes requested):
Good catch on the extra padding!  I had one comment about removing the `mem::take` of `.level`; otherwise it LGTM.

src/decode.rs

kkysen (approved):

PR #736, opened 2024-02-09
`Rav1dFrameContext_frame_thread::pal_idx`: Make into `AlignedVec`

randomPoison:


src/decode.rs
```diff
@@ -4364,22 +4362,15 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
...
-                    f.frame_thread.pal_idx_sz = 0;
-                    return Err(ENOMEM);
-                }
-                f.frame_thread.pal_idx_sz = pal_idx_sz;
+            if pal_idx_sz as usize != f.frame_thread.pal_idx_sz() {
+                // TODO: Fallible allocation
+                f.frame_thread
+                    .pal_idx
+                    .resize_with(pal_idx_sz as usize * 128 * 128 / 4, Default::default);
             }
```
kkysen, line 4370:
```suggestion
            // TODO: Fallible allocation
            f.frame_thread
                .pal_idx
                .resize_with(pal_idx_sz as usize * 128 * 128 / 4, Default::default);
```

`.resize_with` will be a no-op when the sizes are equal.  I think what the C was trying to do is similar to what we're doing here with `.resize_with`, so don't have to duplicate that part.

kkysen (changes requested):
LGTM, though I think we can remove the `if pal_idx_sz as usize != f.frame_thread.pal_idx_sz()` check.

kkysen (approved):

src/internal.rs
```diff
@@ -410,9 +410,8 @@ pub struct Rav1dFrameContext_frame_thread {
     // indexed using (t->by >> 1) * (f->b4_stride >> 1) + (t->bx >> 1)
     pub pal: AlignedVec64<[[u16; 8]; 3]>, /* [3 plane][8 idx] */
     // iterated over inside tile state
-    pub pal_idx: *mut u8,
+    pub pal_idx: AlignedVec64<u8>,
```
kkysen, line 413: Looking into this more, I think we can store a `Vec<Align64<[u8; 64]>>`.  We allocate in multiples of `128 * 128 / 4 = 4096`, and `pal_idx` also sometimes comes from `[u8; 8192]`.  So both are divisible by `64`.  And when `pal_idx` is sliced in `fn rav1d_recon_b_intra`, it's done by `cbw4 * cbh4 * 16` and `bw4 * bh4 * 16`.  I believe the `4` in those variables may mean it's a multiple of 4 (@fbossen, is that right at all?).  Regardless, running the tests shows that all of those offsets are divisible by `64`.  So we can do the slice on the `Vec<Align64<[u8; 64]>>`, and then pass `&[Align64<[u8; 64]>]` all the way to the asm boundary (which we can upgrade to `*const Align64<[u8; 64]>`).  And `fn pal_pred_rust` can be rewritten to work with `idx: &[Align64<[u8; 64]>]` instead, such as by using the `fn flatten` I added recently.

kkysen: Let me take a look at how this would work out.

kkysen (approved):

kkysen:
* Fixes `pal_idx` field for #744.

PR #737, opened 2024-02-09
`Rav1dFrameContext_frame_thread::cf`: Make into `AlignedVec`

randomPoison:


src/decode.rs
```diff
@@ -4364,20 +4361,10 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
...
-                f.frame_thread.cf.cast::<u8>(),
-                usize::try_from(cf_sz).unwrap() * 128 * 128 / 2,
-            )
-            .fill(0);
-            f.frame_thread.cf_sz = cf_sz;
+        if cf_sz as usize != f.frame_thread.cf_sz() {
+            f.frame_thread.cf.clear();
+            // TODO: Fallible allocation
+            f.frame_thread.cf.resize(cf_sz as usize * 128 * 128 / 2, 0);
         }
```
kkysen, line 4368:
```suggestion
        // TODO: Fallible allocation
        f.frame_thread.cf.resize(cf_sz as usize * 128 * 128 / 2, 0);
```

src/align.rs
```diff
@@ -168,6 +168,11 @@ impl<T: Copy> AlignedVec64<T> {
 
         self.len = new_len;
     }
+
+    pub fn clear(&mut self) {
+        self.inner.clear();
+        self.len = 0;
```
kkysen, line 174:
```suggestion
        // Note: Dropping `T`s not needed because `T: Copy`.
        self.len = 0;
        self.inner.clear();
```

That said, I don't think we need `.clear()` yet (see below).

Also, I'm not really sure if it matters, but setting `self.len = 0` first seems safer in case `.clear()` panics or something.

kkysen (changes requested):

src/decode.rs
randomPoison: This doesn't have the same semantics as the original C. The original logic was filling the entire buffer with 0 after resizing, not just the new elements. Are you sure we can change the semantics here without causing issues? I can try it and at least see if it causes any test failures.

src/decode.rs
```diff
@@ -3838,14 +3838,11 @@ unsafe fn setup_tile(
...
-                .cast::<u8>()
-                .offset(
-                    (tile_start_off * size_mul[0] as usize >> (seq_hdr.hbd == 0) as c_int) as isize,
-                )
-                .cast::<DynCoef>()
+        ts.frame_thread[p].cf = if !f.frame_thread.cf.is_empty() {
+            f.frame_thread.cf
+                [(tile_start_off * size_mul[0] as usize >> (seq_hdr.hbd == 0) as c_int) as usize..]
+                .as_ptr()
+                .cast::<DynCoef>() as *mut _
```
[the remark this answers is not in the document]
randomPoison: This doesn't work because `f` is a `&Rav1dFrameData`, so we can't get mutable access to `cf` in order to call `as_mut_ptr()`. I tried reworking the call site to pass a `&mut Rav1dFrameData` instead, but there are borrowing issues in the calling function (`rav1d_decode_frame_init_cdf`) that prevent that from working.

src/decode.rs
randomPoison: Tests seem to pass so I've made the change.

kkysen: Hmm, I *think* it's just because C made an entirely new allocation so they have to, but I'm not sure.  Though if the tests pass it's probably okay?

src/decode.rs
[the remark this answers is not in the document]
kkysen: Ah, okay.  We can fix them later then.

kkysen (approved):

kkysen:
* Fixes `cf` field for #744.

PR #738, opened 2024-02-11
`src/cdef_apply.rs`: more cleanup

folkertdev:
builds on https://github.com/memorysafety/rav1d/pull/733

### DIFF
--- a/src/cdef_apply.rs
+++ b/src/cdef_apply.rs
@@ -194,7 +194,7 @@ pub(crate) unsafe fn rav1d_cdef_brow<BD: BitDepth>(
 
     let mut bit = false;
     for by in (by_start..by_end).step_by(2) {
-        let tf = tc.top_pre_cdef_toggle;
+        let tf = tc.top_pre_cdef_toggle != 0;
         let by_idx = (by & 30) >> 1;
         if by + 2 >= f.bh {
             edges.remove(CdefEdgeFlags::HAVE_BOTTOM);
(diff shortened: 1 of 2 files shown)


src/cdef_apply.rs
```diff
@@ -107,24 +105,28 @@ unsafe fn backup2x8<BD: BitDepth>(
...
 ) {
-    let mut y_off: ptrdiff_t = 0 as c_int as ptrdiff_t;
+    let mut y_off = 0;
+
     if flag.contains(Backup2x8Flags::Y) {
         for y in 0..8 {
             BD::pixel_copy(
                 &mut dst[0][y],
-                slice::from_raw_parts(&mut *src[0].offset(y_off + x_off as isize - 2), 2),
+                slice::from_raw_parts(src[0].offset(y_off + x_off as isize - 2), 2),
```
folkertdev, line 114: is that `&mut` something that `c2rust` inserts? it's not really needed and also unsound when that pointer is dangling (a dangling pointer would be sound if the slice length were 0). Not a problem in practice here but in general it seems odd to me.

src/cdef_apply.rs
```diff
@@ -161,181 +163,180 @@ pub(crate) unsafe fn rav1d_cdef_brow<BD: BitDepth>(
...
 ) {
-    let f: *mut Rav1dFrameContext = tc.f as *mut Rav1dFrameContext;
+    let f = &*tc.f;
     let bitdepth_min_8 = match BD::BPC {
         BPC::BPC8 => 0,
-        BPC::BPC16 => (*f).cur.p.bpc - 8,
+        BPC::BPC16 => f.cur.p.bpc - 8,
     };
-    let dsp: *const Rav1dDSPContext = (*f).dsp;
+    let dsp = &*f.dsp;
```
folkertdev, line 174: `f` and `dsp` are now asserted to be valid references. That seems OK, but it is an assumption.

folkertdev (commented):
I think this is about as far as the cleanup of this file will go at the moment. More can be done when data structure fields become safe, but then cleaning up the code here should follow naturally from those changes.

src/cdef_apply.rs
kkysen: Somewhat (it also looks extra weird since we didn't delete it as we partially cleaned it up/bitdepth deduplicated it).  `c2rust transpile` sometimes has redundant `&mut *`s, and it also generates that kind of code when it really should generate `addr_of_mut!` to be sound (see https://github.com/immunant/c2rust/issues/301).

src/cdef_apply.rs
kkysen: Yes, that should be totally fine.  They should be fully initialized long before this I'm pretty sure.

Also, nice that `f` is no longer `&mut`.

kkysen (approved):

PR #740, opened 2024-02-12
`Rav1dFrameContext_frame_thread::tile_start_off`: Make into `Vec`

### LINKED ISSUE #709: `Rav1dFrameContext_frame_thread`: Make fields safe
* [x] `b: *mut Av1Block` - Probably `Box<[T]>` or `Vec`
* [x] `cbi: *mut CodedBlockInfo` - Probably `Box<[T]>` or `Vec`
* [x] `pal: *mut [[u16; 8]; 3]` - Probably `Box<T>`
* [x] `pal_idx: *mut u8` - Probably `Box<[T]>` or `Vec`
* [x] `cf: *mut DynCoef` - Unclear how to make this safe, since it can't directly be a `Vec<DynCoef>` or similar. Might have to be a `Box<[u8]>` that gets cast to a `*mut DynCoef` where it's used.
* [x] `tile_start_off: *mut c_int` - Probably `Box<[T]>` or `Vec`

randomPoison:
* Closes #709

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4286,13 +4286,8 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
     let n_ts = frame_hdr.tiling.cols * frame_hdr.tiling.rows;
     if n_ts != f.n_ts {
         if c.n_fc > 1 {
-            freep(&mut f.frame_thread.tile_start_off as *mut *mut u32 as *mut c_void);
-            f.frame_thread.tile_start_off =
-                malloc(::core::mem::size_of::<u32>() * n_ts as usize) as *mut u32;
-            if f.frame_thread.tile_start_off.is_null() {
-                f.n_ts = 0;
-                return Err(ENOMEM);
-            }
+            // TODO: Fallible allocation
+            f.frame_thread.tile_start_off.resize(n_ts as usize, 0);
         }
         rav1d_free_aligned(f.ts as *mut c_void);
         f.ts = rav1d_alloc_aligned(::core::mem::size_of::<Rav1dTileState>() * n_ts as usize, 32)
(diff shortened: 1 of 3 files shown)


src/decode.rs
```diff
@@ -4679,7 +4674,7 @@ pub(crate) unsafe fn rav1d_decode_frame_init_cdf(
         for (j, (ts, tile_start_off)) in iter::zip(
             slice::from_raw_parts_mut(f.ts, end + 1),
             if uses_2pass {
-                slice::from_raw_parts(f.frame_thread.tile_start_off, end + 1)
+                &f.frame_thread.tile_start_off[end + 1..]
```
kkysen, line 4677:
```suggestion
                &f.frame_thread.tile_start_off[..end + 1]
```

That said, I don't think we need the slice at all here.  The `zip` takes care of that.

src/lib.rs
```diff
@@ -908,12 +908,7 @@ impl Drop for Rav1dContext {
                         &mut (*f).tile_thread.lowest_pixel_mem as *mut *mut [[c_int; 2]; 7]
                             as *mut c_void,
                     );
-                    let _ = mem::take(&mut (*f).frame_thread.b); // TODO: remove when context is owned
-                    let _ = mem::take(&mut (*f).frame_thread.pal_idx); // TODO: remove when context is owned
-                    let _ = mem::take(&mut (*f).frame_thread.cf); // TODO: remove when context is owned
-                    freep(&mut (*f).frame_thread.tile_start_off as *mut *mut c_int as *mut c_void);
-                    let _ = mem::take(&mut (*f).frame_thread.pal); // TODO: remove when context is owned
-                    let _ = mem::take(&mut (*f).frame_thread.cbi); // TODO: remove when context is owned
+                    let _ = mem::take(&mut (*f).frame_thread); // TODO: remove when context is owned
```
kkysen, line 911: The two lines below can also be removed (I can't comment on them directly):

```rust
                mem::take(&mut (*f).frame_thread.frame_progress); // TODO: remove when context is owned
                mem::take(&mut (*f).frame_thread.copy_lpf_progress); // TODO: remove when context is owned
```

kkysen (changes requested):
See above comments.

kkysen (approved):

PR #747, opened 2024-02-15
`fn DEBUG_BLOCK_INFO`: Make a Rust macro to let the compiler see which fields are borrowed

rinon:
DEBUG_BLOCK_INFO was a macro in the original dav1d sources and we keep running into borrow check issues with it being a function in Rust. Rather than work around these issues by adding complexity, we should just make this a Rust macro.

### DIFF
--- a/src/recon.rs
+++ b/src/recon.rs
@@ -470,7 +470,7 @@ unsafe fn decode_coefs<BD: BitDepth>(
     let frame_hdr = &***(*f).frame_hdr.as_ref().unwrap();
     let lossless = frame_hdr.segmentation.lossless[b.seg_id as usize];
     let t_dim = &dav1d_txfm_dimensions[tx as usize];
-    let dbg = debug_block_info!(&*f, &*t) as c_int;
+    let dbg = (debug_block_info!(&*f, &*t) && plane != 0 && false) as c_int;
     if dbg != 0 {
         printf(
             b"Start: r=%d\n\0" as *const u8 as *const c_char,
--- a/src/recon.rs
+++ b/src/recon.rs
@@ -116,15 +116,25 @@ use std::ffi::c_void;
 use std::ops::BitOr;
 use std::slice;
 
-/// TODO: add feature and compile-time guard around this code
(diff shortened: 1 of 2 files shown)


thedataking (approved):

src/recon.rs
```diff
@@ -116,15 +116,18 @@ use std::ffi::c_void;
...
 
-/// TODO: add feature and compile-time guard around this code
-pub(crate) fn DEBUG_BLOCK_INFO(f: &Rav1dFrameContext, t: &Rav1dTaskContext) -> bool {
-    false
-        && f.frame_hdr.as_ref().unwrap().frame_offset == 2
-        && t.by >= 0
-        && t.by < 4
-        && t.bx >= 8
-        && t.bx < 12
+// TODO: add feature and compile-time guard around this code
```
kkysen, line 119:
```suggestion
// This a macro not a `fn` so that the compiler can see which specific fields are used to avoid borrowck errors.
```

The previous comment also isn't necessary since C doesn't have any sort of feature or compile time guard around this besides a `0 &&`.

src/recon.rs
```diff
@@ -460,7 +463,7 @@ unsafe fn decode_coefs<BD: BitDepth>(
     let frame_hdr = &***(*f).frame_hdr.as_ref().unwrap();
     let lossless = frame_hdr.segmentation.lossless[b.seg_id as usize];
     let t_dim = &dav1d_txfm_dimensions[tx as usize];
-    let dbg = DEBUG_BLOCK_INFO(&*f, &*t) as c_int;
+    let dbg = debug_block_info!(&*f, &*t) as c_int;
```
kkysen, line 466:
```suggestion
    let dbg = (debug_block_info!(&*f, &*t) && plane != 0 && false) as c_int;
```

Separate issue, but this seems mistranslated looking at the C source.

kkysen (approved):
After some discussion, LGTM, but I'd appreciate that comment explaining that we're doing this as a macro so that the compiler can see which fields we use.

src/recon.rs
rinon: Fixed in f3eae29b224a8a86629bc78f020c1746895f2eed

PR #748, opened 2024-02-16
`fn decode_b`: Make `Rav1dFrameData` a reference

rinon:
Resolves the borrow check errors by pulling the current block out of `Rav1dFrameData::frame_thread::b` and operating on it, then replacing the block back into the vector.  We pass the current block around so much, along with the frame context, that this version of inner mutability is necessary.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -1429,6 +1429,36 @@ unsafe fn decode_b(
     bp: BlockPartition,
     intra_edge_flags: EdgeFlags,
 ) -> Result<(), ()> {
+    let f = &mut *t.f;
+    // Pull out the current block from Rav1dFrameData so that we can operate on
+    // it without borrow check errors.
+    let (mut b_mem, b_idx) = if t.frame_thread.pass != 0 {
+        let b_idx = (t.by as isize * f.b4_stride + t.bx as isize) as usize;
+        (mem::take(&mut f.frame_thread.b[b_idx]), Some(b_idx))
+    } else {
+        (Default::default(), None)
+    };
+    let b = &mut b_mem;
+    let res = decode_b_inner(c, t, f, bl, bs, bp, intra_edge_flags, b);
+    if let Some(i) = b_idx {
+        let _old_b = mem::replace(&mut f.frame_thread.b[i], b_mem);
+        // TODO(SJC): We should be able to compare Av1Blocks, but there are C
+        // unions in them.
+        // assert_eq!(old_b, Default::default());
+    }
+    res
+}
+
+unsafe fn decode_b_inner(
+    c: &Rav1dContext,
+    t: &mut Rav1dTaskContext,
+    f: &mut Rav1dFrameData,
+    bl: BlockLevel,
+    bs: BlockSize,
+    bp: BlockPartition,
+    intra_edge_flags: EdgeFlags,
+    b: &mut Av1Block,
+) -> Result<(), ()> {
     use std::fmt;
 
     /// Helper struct for printing a number as a signed hexidecimal value.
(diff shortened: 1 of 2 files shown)


src/decode.rs
```diff
@@ -1442,14 +1472,7 @@ unsafe fn decode_b(
     }
 
     let ts = &mut *t.ts;
-    let f = &mut *t.f;
-    let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
```
kkysen, line 1446: Why was this removed in favor of the `f.frame_hdr()` method?  The method requires an `.unwrap()` and `Arc` deref on every access, and also hides this cost.  I checked and there don't appear to be any borrowck issues keeping the old way.

src/decode.rs
```diff
@@ -1442,14 +1472,7 @@ unsafe fn decode_b(
...
-    let mut b_mem = Default::default();
-    let b = if t.frame_thread.pass != 0 {
-        &mut f.frame_thread.b[(t.by as isize * f.b4_stride + t.bx as isize) as usize]
-    } else {
-        &mut b_mem
-    };
+    let bd_fn = f.bd_fn.clone();
```
kkysen, line 1475:
```suggestion
    let bd_fn = &f.bd_fn;
```

Doesn't appear to cause any borrowck issues.  And if it did, it could be derived from `f.bitdepth_max` and `const` `Rav1dFrameContext_bd_fn`s, which can be created as `const`s.

kkysen: Actually, since `&f.bd_fn` works, `f.bd_fn` can be used everywhere in this `fn`, so no change is needed there at all.

src/decode.rs
```diff
@@ -2036,6 +2060,7 @@ unsafe fn decode_b(
             }
         }
 
+        let seq_hdr = f.seq_hdr();
```
kkysen, line 2063: Similar to `f.frame_hdr()`, this can be done at the beginning of `fn decode_b_inner` with a `let seq_hdr = &***f.seq_hdr.as_ref().unwrap();`, so we only ever pay the cost once and it isn't hidden.

src/decode.rs
kkysen: New version, `bd_fn` not needed:

```suggestion
    let seq_hdr = &***f.seq_hdr.as_ref().unwrap();
    let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
```

kkysen (changes requested):
A lot of these changes (maybe almost all of them) don't appear to be necessary, namely
* the `f.bd_fn.clone()`, `f.bd_fn` works fine
* `f.frame_hdr()`, the pre-`.unwrap()`ed `frame_hdr` works fine
* `f.seq_hdr()`, the pre-`.unwrap()`ed `seq_hdr` works fine
* the `f.frame_thread.b` take, it works fine the old way

I'm confused what this PR is for.  Is it in anticipation of further changes that make these things have borrow checker issues?

In general, I'm also wary of the `fn frame_hdr()` and `fn seq_hdr()` methods, since they hide the cost of the `.unwrap()` and `Arc` deref, and for this PR we don't need them at all.

kkysen (commented):
I hadn't realized the changes in this PR unlock the changes in #750 without causing borrowck errors, so I'll re-review in light of that.

rinon:
All of the changed frame_hdr and bd_fn access here is required by future PRs.

src/internal.rs
```diff
@@ -567,6 +568,16 @@ pub(crate) struct Rav1dFrameData {
...
 
+impl Rav1dFrameData {
+    pub fn frame_hdr(&self) -> &Rav1dFrameHeader {
+        self.frame_hdr.as_ref().unwrap()
+    }
+
+    pub fn seq_hdr(&self) -> &Rav1dSequenceHeader {
+        self.seq_hdr.as_ref().unwrap()
+    }
+}
```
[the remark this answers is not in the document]
kkysen: While I still prefer the overhead of these methods to be removed, we can save that for later.  I'll open an issue for this: #751.

kkysen (approved):
I left some comments, but I think they can be saved for later.

src/internal.rs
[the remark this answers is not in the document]
rinon: Without a clear performance impact to point to, I don't think that the cost here is actually meaningful. The Arc deref is the same cost as a pointer deref, and the unwrap potentially adds code size but the single branch shouldn't add much unless that code size impact accidentally breaks some caching optimality (assuming these method calls are even inlined, they might not be). I checked the benchmark results from CI and this PR looks performance neutral, within margin of error.

PR #749, opened 2024-02-16
`mod recon`: Use `cbi` index instead of mutable reference

rinon:
We can't hold a mutable reference to the `cbi` field while passing a reference to the frame data to children. Instead we can store an index and do the dereference when needed.

### DIFF
--- a/src/recon.rs
+++ b/src/recon.rs
@@ -1720,7 +1720,7 @@ unsafe fn read_coef_tree<BD: BitDepth>(
         let mut cf_ctx: u8 = 0;
         let eob;
         let cf: *mut BD::Coef;
-        let mut cbi = &mut Default::default();
+        let mut cbi_idx = 0;
         if (*t).frame_thread.pass != 0 {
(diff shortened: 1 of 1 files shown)


src/recon.rs
```diff
@@ -1945,7 +1945,7 @@ pub(crate) unsafe fn rav1d_read_coef_blocks<BD: BitDepth>(
                                 (*ts).msac.rng,
                             );
                         }
-                        cbi[t.bx as usize].txtp[0] = txtp as u8;
+                        f.frame_thread.cbi[cbi_idx..][t.bx as usize].txtp[0] = txtp as u8;
```
kkysen, line 1948: Can this and the above index be combined into one lookup?  With `debug_block_info!` as a macro now, I don't think it'll cause any issues.

Same for the other pair below.

kkysen (approved):
LGTM, though I'd prefer if the two pairs of the same lookups I commented on could be combined into one each.

PR #750, opened 2024-02-16
`struct Rav1dTaskContext`: Remove `Rav1dFrameData` pointer

rinon:
This PR adds `Rav1dFrameData` as an explicit argument everywhere it is needed and removes the pointer to it from `Rav1dTaskContext`.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -1424,12 +1424,12 @@ unsafe fn obmc_lowest_px(
 unsafe fn decode_b(
     c: &Rav1dContext,
     t: &mut Rav1dTaskContext,
+    f: &mut Rav1dFrameData,
     bl: BlockLevel,
     bs: BlockSize,
     bp: BlockPartition,
     intra_edge_flags: EdgeFlags,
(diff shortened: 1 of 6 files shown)


src/decode.rs
```diff
@@ -858,18 +859,18 @@ unsafe fn read_pal_plane(
     }
 }
 
-unsafe fn read_pal_uv(
+unsafe fn read_pal_uv<'a>(
```
kkysen, line 862: What's this lifetime for?

src/decode.rs
```diff
@@ -1037,14 +1038,14 @@ unsafe fn read_pal_indices(
     }
 }
 
-unsafe fn read_vartx_tree(
+unsafe fn read_vartx_tree<'a>(
```
kkysen, line 1041:
```suggestion
unsafe fn read_vartx_tree(
```

kkysen (approved):
LGTM but I think you might've left in unused lifetimes since they don't seem to be used anywhere.

PR #752, opened 2024-02-16
`fn Rav1dFrameContext_bd_fn::get`: Replace field with lookups to static versions

kkysen:
`fn Rav1dFrameContext_bd_fn::new` is `const` and pure, so we can have a pure `bitdepth_max: c_int` => `&'static Rav1dFrameContext_bd_fn` `fn`, thereby avoiding `.clone()`s and saving on memory.

See https://github.com/memorysafety/rav1d/pull/748#discussion_r1491936115 and https://github.com/memorysafety/rav1d/pull/748#discussion_r1491949677.

We can also do a similar thing for `Rav1dDSPContext`, but with it cached based on runtime CPU features.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -331,8 +331,6 @@ pub struct ScalableMotionParams {
     pub step: c_int,
 }
 
-#[repr(C)]
-#[derive(Clone)]
 pub(crate) struct Rav1dFrameContext_bd_fn {
     pub recon_b_intra: recon_b_intra_fn,
     pub recon_b_inter: recon_b_inter_fn,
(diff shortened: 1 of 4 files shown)


src/internal.rs
```diff
@@ -573,6 +582,11 @@ pub(crate) struct Rav1dFrameData {
 }
 
 impl Rav1dFrameData {
+    pub fn bd_fn(&self) -> &'static Rav1dFrameContext_bd_fn {
+        let bpc = BPC::from_bitdepth_max(self.bitdepth_max);
```
rinon, line 586: Constructing a bpc seems like it will do unnecessary work here. Can we just have a check against a constant here?

rinon:
I like this idea of a static function pointer table, just a small comment. Needs rebasing now too.

src/internal.rs
kkysen: That's what's happening.  `BPC::from_bitdepth_max` checks against a `const` (`BitDepth8::new(())`) to construct `BPC`, and `BPC` is basically an index.  It compiles down to a `cmp` and `jne` to the two statically known `fn`s.

rinon (approved):

Issue #753, opened 2024-02-20
Backport security fix from dav1d 1.4.0


randomPoison:
A CVE was filed against dav1d 1.3 and earlier due to an integer overflow error: https://www.cvedetails.com/cve/CVE-2024-1580/. We should backport the fixes in dav1d 1.4 to rav1d. This appears to be the PR that fixes the issue: https://code.videolan.org/videolan/dav1d/-/merge_requests/1593

kkysen:
Fixed by #754.

PR #754, opened 2024-02-21
Backport security fix from dav1d 1.4.0

fbossen:
Addresses #753 by using unsigned integer values for `tile_start_off` instead of signed ones to avoid possible overflow.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -2637,7 +2637,7 @@ static void setup_tile(Dav1dTileState *const ts,
                        const Dav1dFrameContext *const f,
                        const uint8_t *const data, const size_t sz,
                        const int tile_row, const int tile_col,
-                       const int tile_start_off)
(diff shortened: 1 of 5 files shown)


src/decode.c
```diff
@@ -2637,7 +2637,7 @@ static void setup_tile(Dav1dTileState *const ts,
                        const Dav1dFrameContext *const f,
                        const uint8_t *const data, const size_t sz,
                        const int tile_row, const int tile_col,
-                       const int tile_start_off)
+                       const unsigned tile_start_off)
```
kkysen, line 2640: We don't need to backport this because we already changed `tile_start_off` to a `usize`, right?

kkysen (approved):

src/decode.c
fbossen: You are correct.

PR #758, opened 2024-02-23
`enum Rav1dDecodeFrameType`: Make a real `enum`

folkertdev:


randomPoison (approved):
Looks good to me! As we go forward and make more of these types into enums we may want to use [strum](https://crates.io/crates/strum) or a similar crate to automatically derive the conversions, but for now the manual impls you wrote are good.

kkysen:
> Looks good to me! As we go forward and make more of these types into enums we may want to use [strum](https://crates.io/crates/strum) or a similar crate to automatically derive the conversions, but for now the manual impls you wrote are good.

We do already have `strum` as a dependency btw.

kkysen (approved):

PR #759, opened 2024-02-23
`struct EdgeFlags`: Make `bitflags!`

folkertdev:


src/intra_edge.rs
```diff
@@ -144,38 +157,27 @@ unsafe fn init_mode_node(
     top_has_right: bool,
     left_has_bottom: bool,
 ) {
-    init_edges(
-        &mut nwc.node,
+    // this is an invariant: this function is only called with BL_128X128 and BL_64X64.
+    // then this function recurses, decreasing the block level at each step, until bottoming out
+    // at BL_16x16.
+    assert!(bl != BL_8X8);
```
folkertdev, line 163: I think this logical step is required to make the logic safe here. It is a change to the logic (versus the C code) but hopefully it's close enough.

randomPoison (approved):
LGTM!

kkysen (commented):
Sorry again for the overlap @folkertdev.  In making `intra_edge.rs` safe and `const`, I did some of this already.  I didn't make `EdgeFlags` a `bitflags!` yet, though, so we can use those changes once rebased and everything made `const fn`-compatible, since all of the uses in `intra_edge.rs` are now in `const fn`s.  I'll open that PR shortly.

folkertdev:
that's fine, #762 looks really nice!

I'll build this on top of that again when it gets merged (one extra change is that `&` and `|` are not const for the bitflags, so we'll need to use the `union` and `intersection` associated functions)

src/intra_edge.rs
```diff
@@ -4,17 +4,52 @@ use crate::src::levels::BL_16X16;
...
+    pub(crate) const fn select(&self, select: bool) -> Self {
+        if select {
+            *self
+        } else {
+            Self::empty()
+        }
+    }
+}
+
+impl core::ops::Shr<u32> for EdgeFlags {
```
kkysen, line 46:
```suggestion
impl Shr<u32> for EdgeFlags {
```

kkysen (changes requested):

kkysen:
@folkertdev, wanted to let you know #762 is merged now.

src/intra_edge.rs
```diff
@@ -47,94 +82,82 @@ struct ModeSelMem {
...
+    node.o = edge_flags;
+
+    node.h[0] = edge_flags | EdgeFlags::ALL_LEFT_HAS_BOTTOM;
+    node.h[1] = edge_flags & (EdgeFlags::ALL_LEFT_HAS_BOTTOM | EdgeFlags::EDGE_I420_TOP_HAS_RIGHT);
+
+    node.v[0] = edge_flags | EdgeFlags::ALL_TOP_HAS_RIGHT;
+    node.v[1] = edge_flags
+        & (EdgeFlags::ALL_TOP_HAS_RIGHT
+            | EdgeFlags::EDGE_I420_LEFT_HAS_BOTTOM
+            | EdgeFlags::EDGE_I422_LEFT_HAS_BOTTOM);
```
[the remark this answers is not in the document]
folkertdev: yes (I also like the formatting much better)

src/intra_edge.rs
```diff
@@ -103,58 +160,59 @@ impl EdgeBranch {
...
-            edge_flags & EDGE_TOP_HAS_RIGHT,
+            edge_flags.union(EdgeFlags::EDGE_TOP_HAS_RIGHT),
+            EdgeFlags::EDGE_TOP_HAS_RIGHT.union(if bl == BL_16X16 {
+                edge_flags.intersection(EdgeFlags::union_all([
+                    EdgeFlags::EDGE_I420_LEFT_HAS_BOTTOM,
+                    EdgeFlags::EDGE_I422_LEFT_HAS_BOTTOM,
+                ]))
+            } else {
+                EdgeFlags::empty()
+            }),
```
folkertdev: this is the most messy expression. I've tried playing around with it a bit using the laws on union and intersection, but nothing is clearly better to me than this which is at least close to the original.

src/intra_edge.rs
```diff
@@ -4,17 +4,52 @@ use crate::src::levels::BL_16X16;
...
-pub const EDGE_I444_TOP_HAS_RIGHT: EdgeFlags = 1;
+bitflags! {
+    #[derive(Clone, Copy, PartialEq, Eq, Hash, Default)]
+    pub(crate) struct EdgeFlags: u8 {
+        const EDGE_I444_TOP_HAS_RIGHT = 1 << 0;
+        const EDGE_I422_TOP_HAS_RIGHT = 1 << 1;
+        const EDGE_I420_TOP_HAS_RIGHT = 1 << 2;
+        const EDGE_I444_LEFT_HAS_BOTTOM = 1 << 3;
+        const EDGE_I422_LEFT_HAS_BOTTOM = 1 << 4;
+        const EDGE_I420_LEFT_HAS_BOTTOM = 1 << 5;
```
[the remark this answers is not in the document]
folkertdev: sure, renames are easy. 

an aside: what do those numbers 4, 2 and 0 mean?

src/intra_edge.rs
```diff
@@ -103,58 +158,59 @@ impl EdgeBranch {
...
-            edge_flags & EDGE_TOP_HAS_RIGHT,
+            edge_flags.union(EdgeFlags::TOP_HAS_RIGHT),
+            EdgeFlags::TOP_HAS_RIGHT.union(if bl == BL_16X16 {
+                edge_flags.intersection(EdgeFlags::union_all([
+                    EdgeFlags::I420_LEFT_HAS_BOTTOM,
+                    EdgeFlags::I422_LEFT_HAS_BOTTOM,
+                ]))
+            } else {
+                EdgeFlags::empty()
+            }),
```
folkertdev, line 190: this is the most messy expression. I've tried playing around with it a bit using the laws on union and intersection, but nothing is clearly better to me than this which is at least close to the original.

randomPoison: Could you use `select` here to simplify slightly? Seems like the `if else` here is doing the same thing as `select`.

```rust
EdgeFlags::TOP_HAS_RIGHT.union(
    edge_flags
        .intersection(EdgeFlags::union_all([
            EdgeFlags::I420_LEFT_HAS_BOTTOM,
            EdgeFlags::I422_LEFT_HAS_BOTTOM,
        ]))
        .select(bl == BL_16X16),
)
```

src/intra_edge.rs
```diff
@@ -214,34 +270,18 @@ impl<const SB128: bool, const N_BRANCH: usize, const N_TIP: usize>
...
-                } else {
-                    EDGE_TOP_HAS_RIGHT
-                }) | (if !(n == 0 || (n == 2 && left_has_bottom)) {
-                    0 as EdgeFlags
-                } else {
-                    EDGE_LEFT_HAS_BOTTOM
-                });
-                self.tip[tip.index as usize] = EdgeTip::new(edge_flags);
+                let thr = EdgeFlags::TOP_HAS_RIGHT.select(!(n == 3 || (n == 1 && !top_has_right)));
+                let lhb = EdgeFlags::LEFT_HAS_BOTTOM.select(n == 0 || n == 2 && left_has_bottom);
```
randomPoison, line 283: Could you put parens around `n == 2 && left_has_bottom` to make order of operations more clear?

```suggestion
                let lhb = EdgeFlags::LEFT_HAS_BOTTOM.select(n == 0 || n == (2 && left_has_bottom));
```

randomPoison (approved):
Looks good! Just had a couple of minor questions/comments.

src/intra_edge.rs
```diff
@@ -103,58 +158,57 @@ impl EdgeBranch {
...
-                | (if bl == BL_16X16 {
-                    edge_flags & EDGE_I420_TOP_HAS_RIGHT
-                } else {
-                    0 as EdgeFlags
-                }),
-            EDGE_LEFT_HAS_BOTTOM,
-            edge_flags & EDGE_LEFT_HAS_BOTTOM,
+            edge_flags.union(EdgeFlags::LEFT_HAS_BOTTOM),
+            EdgeFlags::LEFT_HAS_BOTTOM.union(
+                (edge_flags.intersection(EdgeFlags::I420_TOP_HAS_RIGHT)).select(bl == BL_16X16),
```
kkysen, line 173:
```suggestion
                edge_flags.intersection(EdgeFlags::I420_TOP_HAS_RIGHT).select(bl == BL_16X16),
```

src/intra_edge.rs
```diff
@@ -103,58 +158,57 @@ impl EdgeBranch {
...
+            ),
+            EdgeFlags::TOP_HAS_RIGHT,
+            edge_flags.intersection(EdgeFlags::TOP_HAS_RIGHT),
         ];
 
         let tls = [
-            EDGE_TOP_HAS_RIGHT | EDGE_LEFT_HAS_BOTTOM,
-            edge_flags & EDGE_LEFT_HAS_BOTTOM,
-            edge_flags & EDGE_TOP_HAS_RIGHT,
+            EdgeFlags::TOP_HAS_RIGHT.union(EdgeFlags::LEFT_HAS_BOTTOM),
```
kkysen, line 194: `union_all`? I think it looks best with `.union` when the LHS is `edge_flags`, but when the LHS and RHS are equivalent, a symmetric way seems nicer.

src/intra_edge.rs
```diff
@@ -214,34 +268,18 @@ impl<const SB128: bool, const N_BRANCH: usize, const N_TIP: usize>
...
-            }) | (if left_has_bottom {
-                EDGE_LEFT_HAS_BOTTOM
-            } else {
-                0 as EdgeFlags
-            }),
-            bl,
-        );
+        let thr = EdgeFlags::TOP_HAS_RIGHT.select(top_has_right);
+        let lhb = EdgeFlags::LEFT_HAS_BOTTOM.select(left_has_bottom);
+        let mut branch = EdgeBranch::new(thr.union(lhb), bl);
```
kkysen, line 273: I think I prefer an inline `union_all` since `thr` and `lhb` are kinda confusing.

src/intra_edge.rs
```diff
@@ -103,58 +158,57 @@ impl EdgeBranch {
...
-            0 as EdgeFlags,
+            edge_flags.union(EdgeFlags::TOP_HAS_RIGHT),
+            edge_flags.union(EdgeFlags::LEFT_HAS_BOTTOM),
+            EdgeFlags::empty(),
         ];
         let tts = [
-            EDGE_TOP_HAS_RIGHT | EDGE_LEFT_HAS_BOTTOM,
-            edge_flags & EDGE_TOP_HAS_RIGHT,
-            edge_flags & EDGE_LEFT_HAS_BOTTOM,
+            EdgeFlags::TOP_HAS_RIGHT.union(EdgeFlags::LEFT_HAS_BOTTOM),
```
kkysen, line 204: Same here

src/intra_edge.rs
```diff
@@ -214,34 +268,18 @@ impl<const SB128: bool, const N_BRANCH: usize, const N_TIP: usize>
...
-                    EDGE_TOP_HAS_RIGHT
-                }) | (if !(n == 0 || (n == 2 && left_has_bottom)) {
-                    0 as EdgeFlags
-                } else {
-                    EDGE_LEFT_HAS_BOTTOM
-                });
-                self.tip[tip.index as usize] = EdgeTip::new(edge_flags);
+                let thr = EdgeFlags::TOP_HAS_RIGHT.select(!(n == 3 || (n == 1 && !top_has_right)));
+                let lhb = EdgeFlags::LEFT_HAS_BOTTOM.select(n == 0 || (n == 2 && left_has_bottom));
+                self.tip[tip.index as usize] = EdgeTip::new(thr.union(lhb));
```
kkysen, line 282: Same here.

src/intra_edge.rs
```diff
@@ -4,19 +4,63 @@ use crate::src::levels::BL_128X128;
...
+    pub(crate) const fn select(&self, select: bool) -> Self {
+        if select {
+            *self
+        } else {
+            Self::empty()
+        }
+    }
+}
+
+impl Shr<u32> for EdgeFlags {
```
kkysen, line 57: Could you `impl` this for `Rav1dPixelLayout` directly?  Instead of `u32` I mean.  I think the shift makes more sense operating on both types rather than a more arbitrary `u32`.

kkysen (changes requested):
I still had some minor comments.  It mostly looks very good.

src/intra_edge.rs
folkertdev: that makes this format to 3 lines instead of 1, I'm happy to go with that but the parens did have a reason

src/intra_edge.rs
folkertdev: yes totally agree!

folkertdev: though actually this could be `EdgeFlags::all()`

kkysen: > though actually this could be `EdgeFlags::all()`

Oh true, even better.

src/intra_edge.rs
```diff
@@ -77,22 +122,33 @@ impl EdgeTip {
...
+                EdgeFlags::I422_LEFT_HAS_BOTTOM,
+            ])),
         ];
         let node = EdgeNode { o, h, v };
 
         let split = [
-            EDGE_TOP_HAS_RIGHT | EDGE_LEFT_HAS_BOTTOM,
-            (edge_flags & EDGE_TOP_HAS_RIGHT) | EDGE_I422_LEFT_HAS_BOTTOM,
-            edge_flags | EDGE_I444_TOP_HAS_RIGHT,
+            EdgeFlags::union_all([EdgeFlags::TOP_HAS_RIGHT, EdgeFlags::LEFT_HAS_BOTTOM]),
```
kkysen, line 142:
```suggestion
            EdgeFlags::all(),
```

src/intra_edge.rs
```diff
@@ -103,58 +159,59 @@ impl EdgeBranch {
...
-                    0 as EdgeFlags
-                }),
-            EDGE_TOP_HAS_RIGHT,
-            edge_flags & EDGE_TOP_HAS_RIGHT,
+            edge_flags.union(EdgeFlags::TOP_HAS_RIGHT),
+            EdgeFlags::TOP_HAS_RIGHT.union(
+                edge_flags
+                    .intersection(EdgeFlags::union_all([
+                        edge_flags.intersection(EdgeFlags::I420_LEFT_HAS_BOTTOM),
+                        edge_flags.intersection(EdgeFlags::I422_LEFT_HAS_BOTTOM),
```
kkysen, line 188:
```suggestion
                        EdgeFlags::I420_LEFT_HAS_BOTTOM,
                        EdgeFlags::I422_LEFT_HAS_BOTTOM,
```

kkysen (changes requested):
Almost there; found two more little simplifications.

kkysen (approved):

PR #760, opened 2024-02-23
`enum BlockLevel`: Make a real `enum`

folkertdev:
a bit more experimental, and some naming choices to be made here. I'll leave some inline comments

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -93,9 +93,6 @@ use crate::src::levels::BlockSize;
 use crate::src::levels::MotionMode;
 use crate::src::levels::RectTxfmSize;
 use crate::src::levels::TxfmSize;
-use crate::src::levels::BL_128X128;
-use crate::src::levels::BL_64X64;
-use crate::src::levels::BL_8X8;
 use crate::src::levels::CFL_PRED;
 use crate::src::levels::COMP_INTER_AVG;
 use crate::src::levels::COMP_INTER_NONE;
@@ -1613,7 +1610,7 @@ unsafe fn decode_b_inner(
     let cw4 = w4 + ss_hor >> ss_hor;
     let ch4 = h4 + ss_ver >> ss_ver;
 
-    b.bl = bl as u8;
+    b.bl = bl;
     b.bp = bp as u8;
     b.bs = bs as u8;
 
@@ -3492,7 +3489,7 @@ unsafe fn decode_sb(
     edge_index: EdgeIndex,
 ) -> Result<(), ()> {
     let ts = &mut *t.ts;
-    let hsz = 16 >> bl;
+    let hsz = 16 >> bl as u8;
     let have_h_split = f.bw > t.bx + hsz;
     let have_v_split = f.bh > t.by + hsz;
 
@@ -3500,12 +3497,15 @@ unsafe fn decode_sb(
     let intra_edge = &IntraEdges::DEFAULT;
 
     if !have_h_split && !have_v_split {
-        assert!(bl < BL_8X8);
+        let Some(next_bl) = bl.decrease() else {
+            unreachable!("BlockLevel::BL_8X8 should never make it here");
+        };
+
         return decode_sb(
             c,
             t,
             f,
-            bl + 1,
+            next_bl,
             intra_edge.branch(sb128, edge_index).split[0],
         );
     }
@@ -3519,10 +3519,10 @@ unsafe fn decode_sb(
     let pc = if t.frame_thread.pass == 2 {
         None
     } else {
-        if false && bl == BL_64X64 {
+        if false && bl == BlockLevel::BL_64X64 {
             println!(
                 "poc={},y={},x={},bl={},r={}",
-                frame_hdr.frame_offset, t.by, t.bx, bl, ts.msac.rng,
+                frame_hdr.frame_offset, t.by, t.bx, bl as u8, ts.msac.rng,
             );
         }
         bx8 = (t.bx & 31) >> 1;
@@ -3549,7 +3549,7 @@ unsafe fn decode_sb(
             if debug_block_info!(f, t) {
                 println!(
                     "poc={},y={},x={},bl={},ctx={},bp={}: r={}",
-                    frame_hdr.frame_offset, t.by, t.bx, bl, ctx, bp, ts.msac.rng,
+                    frame_hdr.frame_offset, t.by, t.bx, bl as u8, ctx, bp, ts.msac.rng,
                 );
             }
         } else {
@@ -3578,41 +3578,44 @@ unsafe fn decode_sb(
                 t.bx -= hsz;
             }
             PARTITION_SPLIT => {
-                if bl == BL_8X8 {
-                    let tip = intra_edge.tip(sb128, edge_index);
-                    assert!(hsz == 1);
-                    decode_b(c, t, f, bl, BS_4x4, bp, tip.split[0])?;
-                    let tl_filter = t.tl_4x4_filter;
-                    t.bx += 1;
-                    decode_b(c, t, f, bl, BS_4x4, bp, tip.split[1])?;
-                    t.bx -= 1;
-                    t.by += 1;
-                    decode_b(c, t, f, bl, BS_4x4, bp, tip.split[2])?;
-                    t.bx += 1;
-                    t.tl_4x4_filter = tl_filter;
-                    decode_b(c, t, f, bl, BS_4x4, bp, tip.split[3])?;
-                    t.bx -= 1;
-                    t.by -= 1;
-                    if cfg!(target_arch = "x86_64") && t.frame_thread.pass != 0 {
-                        // In 8-bit mode with 2-pass decoding the coefficient buffer
-                        // can end up misaligned due to skips here.
-                        // Work around the issue by explicitly realigning the buffer.
-                        let p = (t.frame_thread.pass & 1) as usize;
-                        ts.frame_thread[p].cf =
-                            (((ts.frame_thread[p].cf as uintptr_t) + 63) & !63) as *mut DynCoef;
+                match bl.decrease() {
+                    None => {
+                        let tip = intra_edge.tip(sb128, edge_index);
+                        assert!(hsz == 1);
+                        decode_b(c, t, f, bl, BS_4x4, bp, tip.split[0])?;
+                        let tl_filter = t.tl_4x4_filter;
+                        t.bx += 1;
+                        decode_b(c, t, f, bl, BS_4x4, bp, tip.split[1])?;
+                        t.bx -= 1;
+                        t.by += 1;
+                        decode_b(c, t, f, bl, BS_4x4, bp, tip.split[2])?;
+                        t.bx += 1;
+                        t.tl_4x4_filter = tl_filter;
+                        decode_b(c, t, f, bl, BS_4x4, bp, tip.split[3])?;
+                        t.bx -= 1;
+                        t.by -= 1;
+                        if cfg!(target_arch = "x86_64") && t.frame_thread.pass != 0 {
+                            // In 8-bit mode with 2-pass decoding the coefficient buffer
+                            // can end up misaligned due to skips here.
+                            // Work around the issue by explicitly realigning the buffer.
+                            let p = (t.frame_thread.pass & 1) as usize;
+                            ts.frame_thread[p].cf =
+                                (((ts.frame_thread[p].cf as uintptr_t) + 63) & !63) as *mut DynCoef;
+                        }
                     }
-                } else {
-                    let branch = intra_edge.branch(sb128, edge_index);
-                    decode_sb(c, t, f, bl + 1, branch.split[0])?;
-                    t.bx += hsz;
-                    decode_sb(c, t, f, bl + 1, branch.split[1])?;
-                    t.bx -= hsz;
-                    t.by += hsz;
-                    decode_sb(c, t, f, bl + 1, branch.split[2])?;
-                    t.bx += hsz;
-                    decode_sb(c, t, f, bl + 1, branch.split[3])?;
-                    t.bx -= hsz;
-                    t.by -= hsz;
+                    Some(next_bl) => {
+                        let branch = intra_edge.branch(sb128, edge_index);
+                        decode_sb(c, t, f, next_bl, branch.split[0])?;
+                        t.bx += hsz;
+                        decode_sb(c, t, f, next_bl, branch.split[1])?;
+                        t.bx -= hsz;
+                        t.by += hsz;
+                        decode_sb(c, t, f, next_bl, branch.split[2])?;
+                        t.bx += hsz;
+                        decode_sb(c, t, f, next_bl, branch.split[3])?;
+                        t.bx -= hsz;
+                        t.by -= hsz;
+                    }
                 }
             }
             PARTITION_T_TOP_SPLIT => {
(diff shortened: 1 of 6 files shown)


src/env.rs
```diff
@@ -101,7 +100,8 @@ pub fn get_partition_ctx(
     yb8: c_int,
     xb8: c_int,
 ) -> u8 {
-    (a.partition[xb8 as usize] >> (4 - bl) & 1) + ((l.partition[yb8 as usize] >> (4 - bl) & 1) << 1)
+    (a.partition[xb8 as usize] >> (bl.invert() as u8) & 1)
+        + ((l.partition[yb8 as usize] >> (bl.invert() as u8) & 1) << 1)
```
folkertdev, line 104: this logic is kind of nasty: what does this mean? I'm guessing the partition just stores the value in the logical order 8 .. 128 while now the order becomes 128 .. 8?

src/decode.rs
```diff
@@ -3597,14 +3600,14 @@ unsafe fn decode_sb(
                     }
                 } else {
                     let branch = &*(node as *const EdgeBranch);
-                    decode_sb(c, t, f, bl + 1, branch.split[0])?;
+                    decode_sb(c, t, f, bl.saturating_decrease(), branch.split[0])?;
```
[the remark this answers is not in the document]
randomPoison: I've noticed this pattern as well, so the naming makes sense to me. Maybe name it just `decrease` though? There's no non-saturating decrease so the name currently is maybe overly verbose.

src/levels.rs
```diff
@@ -16,13 +16,42 @@ pub const TX_16X16: TxfmSize = 2;
...
-pub const N_BL_LEVELS: usize = 5;
-pub const BL_8X8: BlockLevel = 4;
-pub const BL_16X16: BlockLevel = 3;
-pub const BL_32X32: BlockLevel = 2;
-pub const BL_64X64: BlockLevel = 1;
-pub const BL_128X128: BlockLevel = 0;
+#[repr(u8)]
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BlockLevel {
+    BL_8X8 = 4,
```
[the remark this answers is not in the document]
randomPoison: Yeah, I think it'd be better to use the Rust naming convention here. I kinda think in this case the `SCREAMING_SNAKE_CASE` version is *slightly* more readable, but probably not a good enough reason to deviate from conventions.

kkysen:
Hey @folkertdev, I was working on `intra_edges.rs` last night, too, and make it all safe and statically initialized.  Sorry for the overlap; we'll have to merge a bunch of stuff.  I didn't make things `bitflags!` or `enum`s yet though.

src/levels.rs
```diff
@@ -16,13 +16,42 @@ pub const TX_16X16: TxfmSize = 2;
...
+impl BlockLevel {
+    pub(crate) fn saturating_decrease(self) -> Self {
+        match self {
+            BlockLevel::BL_8X8 => BlockLevel::BL_8X8,
+            BlockLevel::BL_16X16 => BlockLevel::BL_8X8,
+            BlockLevel::BL_32X32 => BlockLevel::BL_16X16,
+            BlockLevel::BL_64X64 => BlockLevel::BL_32X32,
+            BlockLevel::BL_128X128 => BlockLevel::BL_64X64,
+        }
+    }
```
kkysen, line 43: I think we can just have this return an `Option` for `BL_8X8`, since it seems like all of the places check `== BL_8X8` before, and we can fold that into an `Option` check.

Also, the methods don't need to be `pub(crate)`, right?  They can just be `pub`.  Since we're in a non externally `pub` module.

kkysen (changes requested):
Also, I opened #762 now.

src/decode.rs
```diff
@@ -3495,13 +3492,19 @@ unsafe fn decode_sb(
...
     let ts = &mut *t.ts;
-    let hsz = 16 >> bl;
+    let hsz = 16 >> bl as u8;
     let have_h_split = f.bw > t.bx + hsz;
     let have_v_split = f.bh > t.by + hsz;
 
     if !have_h_split && !have_v_split {
-        assert!(bl < BL_8X8);
-        return decode_sb(c, t, f, bl + 1, (*(node as *const EdgeBranch)).split[0]);
+        assert!(bl != BlockLevel::BL_8X8);
```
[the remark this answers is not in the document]
kkysen: Since we are already doing things like `bl + 1`/`bl.decrease()`, I think deriving `Ord` is fine.  That said, if these are resolved by having `.decrease()` return an `Option` that we check (and thus do the `BL_8X8` check), that's fine, too.

src/levels.rs
```diff
@@ -16,13 +16,42 @@ pub const TX_16X16: TxfmSize = 2;
...
-pub const BL_128X128: BlockLevel = 0;
+#[repr(u8)]
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BlockLevel {
+    BL_8X8 = 4,
+    BL_16X16 = 3,
+    BL_32X32 = 2,
+    BL_64X64 = 1,
+    #[default]
+    BL_128X128 = 0,
```
kkysen, line 27: Could you flip the order of these?  `c2rust transpile` transpiles them backwards for some reason, and especially when the numeric value matters it looks very weird backwards.

src/env.rs
folkertdev: actually, now that I look at this again, the associativity is different, the expression is really 

```
((a.partition[xb8 as usize] >> (bl.invert() as u8)) & 1)
```

the shift happens before the bitwise and, which is consistent with https://doc.rust-lang.org/reference/expressions.html#expression-precedence, `>>` has higher precedence than `&`.

src/decode.rs
```diff
@@ -3519,10 +3519,10 @@ unsafe fn decode_sb(
     let pc = if t.frame_thread.pass == 2 {
         None
     } else {
-        if false && bl == BL_64X64 {
+        if false && bl == BlockLevel::Bl64x64 {
             println!(
                 "poc={},y={},x={},bl={},r={}",
-                frame_hdr.frame_offset, t.by, t.bx, bl, ts.msac.rng,
+                frame_hdr.frame_offset, t.by, t.bx, bl as u8, ts.msac.rng,
```
folkertdev, line 3525: is the output format important here? otherwise using the `Debug` instance is probably more informative (so you get `BlockLevel::Bl8x8` instead of `4`).

src/intra_edge.rs
```diff
@@ -324,17 +323,21 @@ impl<const SB128: bool, const N_BRANCH: usize, const N_TIP: usize>
...
+            BlockLevel::Bl128x128
+        } else {
+            BlockLevel::Bl64x64
+        };
         (self, indices) = self.init_mode_node(EdgeIndex::root(), bl, indices, true, false);
 
-        let mut bl = BL_128X128;
-        while bl <= BL_32X32 {
+        let mut bl = BlockLevel::Bl128x128 as u8;
+        while bl <= BlockLevel::Bl32x32 as u8 {
```
folkertdev, line 340: so sad this can't be a `for bl in [BlockLevel::Bl128x128, BlockLevel::Bl64x64, BlockLevel::Bl32x32]`. one day...

src/env.rs
```diff
@@ -101,7 +100,11 @@ pub fn get_partition_ctx(
     yb8: c_int,
     xb8: c_int,
 ) -> u8 {
-    (a.partition[xb8 as usize] >> (4 - bl) & 1) + ((l.partition[yb8 as usize] >> (4 - bl) & 1) << 1)
+    // the right-most ("index zero") bit of the partition represents the 8x8 block level,
+    // but the BlockLevel enum represents the variants numerically in the opposite order
+    // (128x128 = 0, 8x8 = 4). The shift corrects the ordering.
+    let shift = 4 - bl as u8;
+    ((a.partition[xb8 as usize] >> shift) & 1) + (((l.partition[yb8 as usize] >> shift) & 1) << 1)
```
folkertdev, line 107: would this be better?
```suggestion
    let get = |partition: &[_], index| partition[index as usize] >> (4 - bl as u8) & 1;
    get(&a.partition.0, xb8) + get(&l.partition.0, yb8) << 1
```

src/decode.rs
randomPoison: This is debug output so I don't think keeping the format is super important. The debug output should also be just `Bl8x8` which is nice and concise, so feel free to debug print the value now that it's a proper enum.

src/decode.rs
```diff
@@ -3492,20 +3489,23 @@ unsafe fn decode_sb(
...
     let have_v_split = f.bh > t.by + hsz;
 
     let sb128 = f.seq_hdr().sb128 != 0;
     let intra_edge = &IntraEdges::DEFAULT;
 
     if !have_h_split && !have_v_split {
-        assert!(bl < BL_8X8);
+        let Some(next_bl) = bl.decrease() else {
+            unreachable!("BlockLevel::BL_8X8 should never make it here");
+        };
```
randomPoison, line 3502: Why do a `let else` here instead of `expect`?

src/decode.rs
```diff
@@ -3708,13 +3711,16 @@ unsafe fn decode_sb(
             is_split = b.bl != bl;
         }
 
-        assert!(bl < BL_8X8);
+        let Some(next_bl) = bl.decrease() else {
+            unreachable!("BlockLevel::BL_8X8 should never make it here");
+        };
```
randomPoison, line 3716: This could also be an `expect` instead of a `let else`.

src/env.rs
randomPoison: I think either way is fine. The small bit of duplicated logic between the two isn't really an issue imo, but I think the closure approach is also nice because it makes it clear that the logic is the same for both of them.

src/env.rs
```diff
@@ -101,7 +100,11 @@ pub fn get_partition_ctx(
     yb8: c_int,
     xb8: c_int,
 ) -> u8 {
-    (a.partition[xb8 as usize] >> (4 - bl) & 1) + ((l.partition[yb8 as usize] >> (4 - bl) & 1) << 1)
+    // the right-most ("index zero") bit of the partition represents the 8x8 block level,
+    // but the BlockLevel enum represents the variants numerically in the opposite order
+    // (128x128 = 0, 8x8 = 4). The shift corrects the ordering.
```
randomPoison, line 105: Minor suggestion:

```suggestion
    // (128x128 = 0, 8x8 = 4). The shift reverses the ordering.
```

src/intra_edge.rs
```diff
@@ -173,7 +169,7 @@ impl EdgeBranch {
             EdgeFlags::LEFT_HAS_BOTTOM.union(
                 edge_flags
                     .intersection(EdgeFlags::I420_TOP_HAS_RIGHT)
-                    .select(bl == BL_16X16),
+                    .select(matches!(bl, BlockLevel::Bl16x16)),
```
randomPoison, line 172: This could still be a `==` comparison right? We're deriving `Eq` for `BlockLevel`. This also applies to the other places where you changed `==` to `matches!`.

src/intra_edge.rs
```diff
@@ -298,7 +294,10 @@ impl<const SB128: bool, const N_BRANCH: usize, const N_TIP: usize>
                 branch.split[n as usize] = child_branch;
                 (self, indices) = self.init_mode_node(
                     child_branch,
-                    bl + 1,
+                    match bl.decrease() {
+                        None => panic!("BlockLevel::BL_8X8 should never make it here"),
+                        Some(next_bl) => next_bl,
+                    },
```
randomPoison, line 300:
```suggestion
                    bl.decrease().expect("BlockLevel::BL_8X8 should never make it here"),
```

src/levels.rs
```diff
@@ -1,6 +1,10 @@
 use std::ffi::c_uint;
 use std::ops::Neg;
 
+use strum::EnumCount;
+
+use crate::src::enum_map::DefaultValue;
```
randomPoison, line 6: Could you remove the extra lines between `use` statements so that imports are grouped and can be automatically sorted?

```suggestion
use strum::EnumCount;
use crate::src::enum_map::DefaultValue;
```

src/tables.rs
```diff
@@ -1,3 +1,5 @@
+use strum::EnumCount;
+
```
randomPoison, line 2: Please remove the extra line here and sort the imports.

```suggestion
use strum::EnumCount;
```

randomPoison (approved):
Looks good! I left a handful of comments, but they're pretty much all stylistic so I'm going to approve now and we can merge once you've made any necessary updates.

src/env.rs
fbossen: It may be simpler to define `let has_bl = |x| (x >> (4 - bl as u8)) & 1;` and then use as `has_bl(a.partition[xb8 as usize]) + has_bl(b.partition[yb8 as usize]) * 2`. Note that I generally prefer using `* 2` instead of `<< 1` in such cases to avoid adding extra parentheses. Also the precedence of `*` is generally much better understood than the precedence of `<<`.

folkertdev: yes, good call!

src/intra_edge.rs
randomPoison: Okay I suspected that might've been the reason. Thanks for clarifying.

kkysen: One day... https://github.com/joshtriplett/rfcs/blob/is/text/3573-is.md.

src/intra_edge.rs
kkysen: I think expect isn't a `const fn`.

src/intra_edge.rs
```diff
@@ -324,17 +323,21 @@ impl<const SB128: bool, const N_BRANCH: usize, const N_TIP: usize>
 
         let sb128 = SB128 as u8;
 
-        let mut bl = BL_128X128;
-        while bl <= BL_32X32 {
-            indices.branch[bl as usize].index = level_index(bl + sb128);
+        let mut bl = BlockLevel::Bl128x128 as u8;
+        while bl <= BlockLevel::Bl32x32 as u8 {
```
kkysen, line 327: Can this be a `while let Some` with `.decrease`?

src/intra_edge.rs
```diff
@@ -234,7 +230,7 @@ impl DefaultValue for EdgeTip {
 }
 
 impl DefaultValue for EdgeBranch {
-    const DEFAULT: Self = Self::new(EdgeFlags::empty(), 0 as BlockLevel);
+    const DEFAULT: Self = Self::new(EdgeFlags::empty(), BlockLevel::DEFAULT);
```
kkysen, line 233:
```suggestion
    const DEFAULT: Self = Self::new(EdgeFlags::empty(), BlockLevel::BL_128X128);
```

128 isn't an obvious default so I don't think we want to `impl DefaultValue` for it and we can just use the variant directly here.

src/intra_edge.rs
folkertdev: yeah, in fact even `unreachable!` isn't allowed in const contexts which was surprising to me (intuitively it's just some pretty formatting around a `panic!`)

src/intra_edge.rs
folkertdev: I don't see how. upon entering the loop you immediately `.decrease` so it's impossible to cover `128x128` that way. We could add a `.increase`  that goes the opposite direction but I don't think that's worth it.

```
        let mut bl = BlockLevel::Bl16x16 ;
        while let Some(bl) = bl.increase() {
            indices.branch[bl as usize].index = level_index(bl as u8 + sb128);
        }
```

in particular it is kind of subtle that `16x16` is actually skipped here.

kkysen: Oh yeah, true, I didn't think about that.  Nevermind then.

kkysen (approved):

PR #762, opened 2024-02-24
`mod intra_edge`: Make fully safe and `const fn`-initialized

### LINKED ISSUE #756: Make mode tree node access safe
The node tree stored in `Rav1dContext_intra_edge` is currently largely initialized and accessed by passing around raw pointers to its data.

`ModeSelMem` holds onto raw node pointers and should be updated to use safe operations. `ModeSelMem` appears to only be used during initialization, so it shouldn't be an issue to use a less performant but safe approach.

`EdgeBranch::split` contains `*mut EdgeNode`s. We may need to turn these into indices to make this safe.

A recurring unsafe pattern is to pass around `*const EdgeNode`, which points to the first field of an `EdgeBranch` or an `EdgeTip`. The `*const EdgeNode` is then "upcast" to either `*const EdgeBranch` or `*const EdgeTip`. Some examples of this are in `init_edges` and `decode_sb`. We'll need to find an alternate way to express this in safe Rust.

kkysen:
* Fixes #756.

### DIFF
--- a/src/intra_edge.rs
+++ b/src/intra_edge.rs
@@ -176,62 +176,65 @@ impl EdgeBranch {
     }
 }
 
-fn init_mode_node<const SB128: bool, const N_BRANCH: usize, const N_TIP: usize>(
-    tree: &mut IntraEdge<SB128, N_BRANCH, N_TIP>,
-    branch_index: EdgeIndex,
-    bl: BlockLevel,
-    indices: &mut EdgeIndices,
-    top_has_right: bool,
-    left_has_bottom: bool,
-) {
-    let mut branch = EdgeBranch::new(
-        (if top_has_right {
-            EDGE_TOP_HAS_RIGHT
-        } else {
-            0 as EdgeFlags
-        }) | (if left_has_bottom {
-            EDGE_LEFT_HAS_BOTTOM
-        } else {
-            0 as EdgeFlags
-        }),
-        bl,
-    );
-    if bl == BL_16X16 {
-        let mut n = 0;
-        while n < B as u8 {
-            let (tip, next) = indices.tip.pop_front();
-            indices.tip = next;
-            branch.split[n as usize] = tip;
-            let edge_flags = (if n == 3 || (n == 1 && !top_has_right) {
-                0 as EdgeFlags
-            } else {
+impl<const SB128: bool, const N_BRANCH: usize, const N_TIP: usize>
+    IntraEdge<SB128, N_BRANCH, N_TIP>
+{
+    fn init_mode_node(
+        &mut self,
+        branch_index: EdgeIndex,
+        bl: BlockLevel,
+        indices: &mut EdgeIndices,
+        top_has_right: bool,
+        left_has_bottom: bool,
+    ) {
+        let mut branch = EdgeBranch::new(
+            (if top_has_right {
                 EDGE_TOP_HAS_RIGHT
-            }) | (if !(n == 0 || (n == 2 && left_has_bottom)) {
-                0 as EdgeFlags
             } else {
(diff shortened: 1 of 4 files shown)


src/intra_edge.rs
```diff
@@ -16,6 +13,41 @@ pub const EDGE_I420_TOP_HAS_RIGHT: EdgeFlags = 4;
 pub const EDGE_I422_TOP_HAS_RIGHT: EdgeFlags = 2;
 pub const EDGE_I444_TOP_HAS_RIGHT: EdgeFlags = 1;
 
+pub const EDGE_LEFT_HAS_BOTTOM: EdgeFlags =
+    EDGE_I444_LEFT_HAS_BOTTOM | EDGE_I422_LEFT_HAS_BOTTOM | EDGE_I420_LEFT_HAS_BOTTOM;
+pub const EDGE_TOP_HAS_RIGHT: EdgeFlags =
+    EDGE_I444_TOP_HAS_RIGHT | EDGE_I422_TOP_HAS_RIGHT | EDGE_I420_TOP_HAS_RIGHT;
+
+const B: usize = 4;
```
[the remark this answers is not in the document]
folkertdev: which maybe means it can be derived later as  `const B: usize = BlockLevel::COUNT - 1;` , subtracting one because 8x8 is special and not relevant here?

src/intra_edge.rs
```diff
@@ -38,182 +70,334 @@ pub struct EdgeBranch {
...
-                !(n == 3 || (n == 1 && !top_has_right)),
-                n == 0 || (n == 2 && left_has_bottom),
-            );
+        assert!(indices.tip.index == self.tip.len() as u8);
+
+        self
+    }
+
+    /// Check that all indices are in bound so that bounds checks are not needed at runtime.
+    const fn check_indices(self) -> Self {
```
folkertdev, line 304: huh, the optimizer takes that into account? that is extremely neat!

kkysen: Nope lol.  I did it manually with an `unsafe`.  The optimizer is generally pretty bad at this kind of non local optimization and understanding any invariants on fields.

src/intra_edge.rs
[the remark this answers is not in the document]
kkysen: I called it `B` like the B in B tree, maybe standing for branching factor or something.  I didn't look closely into why it's 4, though.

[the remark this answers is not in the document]
kkysen: > ah, I had this confused with the use of `BlockLevel` as an index. If I'm reading the code correctly the 4 is there because when you go from e.g. `16x16` to `32x32`, the number of tiles is multiplied by 4. So even if an extra block level gets added, this constant would just remain the same (it's just a consequence of doing operations in 2D)

Ah, that makes more sense.

src/intra_edge.rs
```diff
@@ -179,14 +188,15 @@ impl EdgeBranch {
 impl<const SB128: bool, const N_BRANCH: usize, const N_TIP: usize>
     IntraEdge<SB128, N_BRANCH, N_TIP>
 {
-    fn init_mode_node(
-        &mut self,
+    #[must_use]
```
randomPoison, line 208: Why the `#[must_use]` attribute here?

randomPoison (approved):
Looks good, just one minor question.

src/intra_edge.rs
kkysen: It's not really necessary, but since I switched from `&mut Self` to `Self -> Self`, I wanted to make sure the return value wasn't forgotten about, since `this.init_mode_node()` needed to be replaced with `this = this.init_mode_node()`.

PR #764, opened 2024-02-27
`struct AlignedVec`: Make generic over alignment

kkysen:
This makes `AlignedVec` generic over alignments by using a marker `trait AlignedByteChunk` `impl`ed for types with the same alignment and size (e.x. `Align64<[u8; 64]>`.  Since `#[repr(align(N))]` is not a const generic, const generics couldn't be used for this, but this works nicely.  We check that `mem::size_of::<C>() == mem::align_of::<C>()` in `AlignedVec`'s constructors to ensure `AlignedByteChunk` is `impl`ed on valid types.  Thus, I also ensured that the `AlignedVec`s were actually constructed, not transmuted through `unsafe { mem::zeroed() }`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -33,6 +33,7 @@ use crate::src::fg_apply;
 use crate::src::internal::Rav1dContext;
 use crate::src::internal::Rav1dContextTaskThread;
 use crate::src::internal::Rav1dContextTaskType;
+use crate::src::internal::Rav1dFrameContext_frame_thread;
 use crate::src::internal::Rav1dFrameData;
 use crate::src::internal::Rav1dTask;
 use crate::src::internal::Rav1dTaskContext;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -324,6 +326,7 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
     let mut n: c_uint = 0 as c_int as c_uint;
     while n < (*c).n_fc {
         let f: &mut Rav1dFrameData = &mut *((*c).fc).offset(n as isize);
+        addr_of_mut!(f.frame_thread).write(Default::default());
```
randomPoison, line 329: Why did you add this? This is to initialize the field the first time right (hence using `write` instead of writing to the field normally)? Adding this makes sense if we weren't doing it already, but was there a reason you needed to add this in this PR? I don't see how your changes would have made this necessary in a way that it wasn't before.

randomPoison: Ah nevermind, I reread the PR description and I see you mentioned this.

randomPoison (approved):

PR #765, opened 2024-02-27
`struct Rav1dFrameContext_lf::lr_mask`: Make into a `Vec`

randomPoison:
~~In order to make this work I had to inline `setup_tile` into its one call site in order to address borrow conflicts. Once making `lr_mask` into a `Vec`, it became necessary to pass `f` as a `&mut Rav1dFrameData` in order to get mutable access to `lr_mask`. This caused a large number of borrow conflicts in the calling function because a large amount of immutable borrows are held across the call site. Most notable is that we're calling `setup_tile` within a `for tile in &f.tiles`, which means we can't pass `&mut f` into a function within the loop body.~~

~~I don't like this solution, mostly because I don't see this being a solution we can apply in the general case, but I couldn't think of a cleaner way to address the borrow issue here. The main alternative here I considered was to not pass `f` as a whole into `setup_tile` and instead manually pass in all of the values it needed. That would allow us to only mutably borrow `f.lf`, which would prevent the borrow conflicts. However `setup_tile` accesses several fields of `f`, and it would have been extremely verbose to pass to pass them all manually. Since there was only one callsite and only one field needed to be mutably borrowed, inlining seemed simplest.~~

NOTE: The inlining mentioned above is no longer necessary after #766.

`read_restoration_info` had a different issue where both `&f` and `&mut f.lf` were being passed into the same function, which can't happen. In this case `f` was only being used in calls to `debug_block_info!`, so I opted to hoist the call to `debug_block_info!` to the call site so that the compiler could see that the field borrows were disjoint.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -152,7 +152,6 @@ use crate::src::lf_mask::rav1d_calc_eih;
 use crate::src::lf_mask::rav1d_calc_lf_values;
 use crate::src::lf_mask::rav1d_create_lf_mask_inter;
 use crate::src::lf_mask::rav1d_create_lf_mask_intra;
-use crate::src::lf_mask::Av1Restoration;
 use crate::src::lf_mask::Av1RestorationUnit;
 use crate::src::log::Rav1dLog as _;
 use crate::src::loopfilter::rav1d_loop_filter_dsp_init;
@@ -3924,9 +3923,9 @@ unsafe fn setup_tile(
             if sb128x >= f.sr_sb128w {
                 continue;
             }
-            &(*f.lf.lr_mask.offset((sb_idx + sb128x) as isize)).lr[p][u_idx as usize]
+            &f.lf.lr_mask[(sb_idx + sb128x) as usize].lr[p][u_idx as usize]
         } else {
-            &(*f.lf.lr_mask.offset(sb_idx as isize)).lr[p][unit_idx as usize]
+            &f.lf.lr_mask[sb_idx as usize].lr[p][unit_idx as usize]
         };
 
         let lr = lr_ref.get();
@@ -3946,13 +3945,12 @@ unsafe fn setup_tile(
 }
 
 unsafe fn read_restoration_info(
-    t: &mut Rav1dTaskContext,
-    f: &Rav1dFrameData,
+    ts: &mut Rav1dTileState,
     lr: &mut Av1RestorationUnit,
     p: usize,
     frame_type: Rav1dRestorationType,
+    debug_block_info: bool,
 ) {
-    let ts = &mut *t.ts;
     let lr_ref = ts.lr_ref[p];
 
     if frame_type == RAV1D_RESTORATION_SWITCHABLE {
@@ -4006,7 +4004,7 @@ unsafe fn read_restoration_info(
         lr.filter_h[2] = msac_decode_lr_subexp(ts, lr_ref.filter_h[2], 3, 17);
         lr.sgr_weights = lr_ref.sgr_weights;
         ts.lr_ref[p] = *lr;
-        if debug_block_info!(f, t) {
+        if debug_block_info {
             println!(
                 "Post-lr_wiener[pl={},v[{},{},{}],h[{},{},{}]]: r={}",
                 p,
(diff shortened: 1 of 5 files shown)


src/decode.rs
```diff
@@ -4144,7 +4038,7 @@ pub(crate) unsafe fn rav1d_decode_tile_sbrow(
             cdef_idx[0] = -1;
             t.cur_sb_cdef_idx_ptr = cdef_idx.as_mut_ptr();
         }
-        let frame_hdr = f.frame_hdr();
+        let frame_hdr = f.frame_hdr.as_ref().unwrap();
```
kkysen, line 4041:
```suggestion
        let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
```

kkysen (commented):
Let me look into this a bit more, but have you tried `.take()`ing any of the fields, like `.lr_mask`, passing it separately, and then putting it back?

kkysen:
> Let me look into this a bit more, but have you tried `.take()`ing any of the fields, like `.lr_mask`, passing it separately, and then putting it back?

Update: It looks like that works when I tried it out.

kkysen:
Another idea I remember trying out before: Make `Av1RestorationUnit` a `Cell<Av1RestorationUnit>` as `Av1RestorationUnit: Copy`.  Then use `Cell::get` and `Cell:set` in `fn setup_tile`.

kkysen:
> Another idea I remember trying out before: Make `Av1RestorationUnit` a `Cell<Av1RestorationUnit>` as `Av1RestorationUnit: Copy`. Then use `Cell::get` and `Cell:set` in `fn setup_tile`.

I think the `Cell` option is the nicest here.  What do you think?

kkysen:
> > Another idea I remember trying out before: Make `Av1RestorationUnit` a `Cell<Av1RestorationUnit>` as `Av1RestorationUnit: Copy`. Then use `Cell::get` and `Cell:set` in `fn setup_tile`.
> 
> I think the `Cell` option is the nicest here. What do you think?

See #766 for that; it was pretty simple.

randomPoison:
I think I like the `Cell` approach best since there are only a couple of places that actually use this data and we don't generally need to take references to it. I'll review #766 then rebase this on top after that's merged.

src/decode.rs
```diff
@@ -3947,10 +3946,10 @@ unsafe fn setup_tile(
 
 unsafe fn read_restoration_info(
     t: &mut Rav1dTaskContext,
-    f: &Rav1dFrameData,
     lr: &mut Av1RestorationUnit,
     p: usize,
     frame_type: Rav1dRestorationType,
+    debug_block_info: bool,
```
kkysen, line 3952: Since we precompute `debug_block_info`, we need to make sure `fn read_restoration_info` doesn't modify `debug_block_info!(f, t)`.  This is true here, but I think we should make it obvious by not passing `t: &mut Rav1dTaskContext`, but just `ts: &mut Rav1dTileState`.  No other fields of `t` are used.  That'll also probably help a little bit with borrowck issues later.

kkysen (changes requested):

src/decode.rs
```diff
@@ -4143,7 +4142,7 @@ pub(crate) unsafe fn rav1d_decode_tile_sbrow(
             cdef_idx[0] = -1;
             t.cur_sb_cdef_idx_ptr = cdef_idx.as_mut_ptr();
         }
-        let frame_hdr = f.frame_hdr();
+        let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
```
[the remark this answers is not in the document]
randomPoison: We still need it in order for the compiler to see that the field borrows are disjoint. Using `frame_hdr()` borrows all of `f`, so it was causing a conflict of the later attempt to mutably borrow `f.lr`.

src/decode.rs
randomPoison: Good idea, done!

src/decode.rs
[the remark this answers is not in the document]
kkysen: Ah, okay.  This way's good then.

kkysen (approved):

PR #768, opened 2024-02-27
`struct Rav1dFrameContext_lf::tx_lpf_right_edge`: Make into `Vec`

randomPoison:
`tx_lpf_right_edge` was being used to store two arrays in a single allocation, with the second pointer pointing halfway into the same allocation as the first pointer. I opted to split the field into two `Vec`s because it seemed like the two halves of the allocation were being treated as disjoint in practice. The new field names use the `_y` and `_uv` suffixes based on how each of the arrays is used.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4237,23 +4237,15 @@ pub(crate) unsafe fn rav1d_decode_tile_sbrow(
     // backup t->a/l.tx_lpf_y/uv at tile boundaries to use them to "fix"
     // up the initial value in neighbour tiles when running the loopfilter
     let mut align_h = f.bh + 31 & !31;
-    slice::from_raw_parts_mut(
-        f.lf.tx_lpf_right_edge[0],
-        (align_h * tile_col + t.by + sb_step).try_into().unwrap(),
-    )[(align_h * tile_col + t.by).try_into().unwrap()..][..sb_step.try_into().unwrap()]
-        .copy_from_slice(&t.l.tx_lpf_y.0[(t.by & 16) as usize..][..sb_step.try_into().unwrap()]);
+    let (tx_lpf_right_edge_y, tx_lpf_right_edge_uv) = f.lf.tx_lpf_right_edge_mut();
+    tx_lpf_right_edge_y[(align_h * tile_col + t.by) as usize..][..sb_step as usize]
+        .copy_from_slice(&t.l.tx_lpf_y.0[(t.by & 16) as usize..][..sb_step as usize]);
     let ss_ver = (f.cur.p.layout == Rav1dPixelLayout::I420) as c_int;
     align_h >>= ss_ver;
-    slice::from_raw_parts_mut(
-        f.lf.tx_lpf_right_edge[1],
-        (align_h * tile_col + (t.by >> ss_ver) + (sb_step >> ss_ver))
-            .try_into()
-            .unwrap(),
-    )[(align_h * tile_col + (t.by >> ss_ver)).try_into().unwrap()..]
-        [..(sb_step >> ss_ver).try_into().unwrap()]
+    tx_lpf_right_edge_uv[(align_h * tile_col + (t.by >> ss_ver)) as usize..]
+        [..(sb_step >> ss_ver) as usize]
         .copy_from_slice(
-            &t.l.tx_lpf_uv.0[((t.by & 16) >> ss_ver) as usize..]
-                [..(sb_step >> ss_ver).try_into().unwrap()],
+            &t.l.tx_lpf_uv.0[((t.by & 16) >> ss_ver) as usize..][..(sb_step >> ss_ver) as usize],
         );
 
     Ok(())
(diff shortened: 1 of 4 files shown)


src/lf_apply.rs
```diff
@@ -560,10 +560,8 @@ pub(crate) unsafe fn rav1d_loopfilter_sbrow_cols<BD: BitDepth>(
     let hmax = (1 as c_uint) << hmask;
     let endy4 = (starty4 + cmp::min(f.h4 - sby * sbsz, sbsz)) as c_uint;
     let uv_endy4: c_uint = endy4.wrapping_add(ss_ver as c_uint) >> ss_ver;
-    let mut lpf_y: *const u8 = &mut *(*(f.lf.tx_lpf_right_edge).as_ptr().offset(0))
-        .offset((sby << sbl2) as isize) as *mut u8;
-    let mut lpf_uv: *const u8 = &mut *(*(f.lf.tx_lpf_right_edge).as_ptr().offset(1))
-        .offset((sby << sbl2 - ss_ver) as isize) as *mut u8;
+    let mut lpf_y = &mut f.lf.tx_lpf_right_edge_y[(sby << sbl2) as usize..];
+    let mut lpf_uv = &mut f.lf.tx_lpf_right_edge_uv[(sby << sbl2 - ss_ver) as usize..];
```
kkysen, line 564:
```suggestion
    let mut lpf_y = &f.lf.tx_lpf_right_edge_y[(sby << sbl2) as usize..];
    let mut lpf_uv = &f.lf.tx_lpf_right_edge_uv[(sby << sbl2 - ss_ver) as usize..];
```

src/lf_apply.rs
```diff
@@ -614,16 +612,15 @@ pub(crate) unsafe fn rav1d_loopfilter_sbrow_cols<BD: BitDepth>(
...
                 ) as isize))[sidx_0 as usize];
                 *fresh6 = (*fresh6 as c_uint | smask_0) as u16;
                 y_0 = y_0.wrapping_add(1);
                 uv_mask <<= 1;
             }
         }
-        lpf_y = lpf_y.offset(halign as isize);
-        lpf_uv = lpf_uv.offset((halign >> ss_ver) as isize);
+        lpf_y = &mut lpf_y[halign as usize..];
+        lpf_uv = &mut lpf_uv[(halign >> ss_ver) as usize..];
```
kkysen, line 623:
```suggestion
        lpf_y = &lpf_y[halign as usize..];
        lpf_uv = &lpf_uv[(halign >> ss_ver) as usize..];
```

src/decode.rs
```diff
@@ -4237,19 +4237,12 @@ pub(crate) unsafe fn rav1d_decode_tile_sbrow(
     // backup t->a/l.tx_lpf_y/uv at tile boundaries to use them to "fix"
     // up the initial value in neighbour tiles when running the loopfilter
     let mut align_h = f.bh + 31 & !31;
-    slice::from_raw_parts_mut(
-        f.lf.tx_lpf_right_edge[0],
-        (align_h * tile_col + t.by + sb_step).try_into().unwrap(),
-    )[(align_h * tile_col + t.by).try_into().unwrap()..][..sb_step.try_into().unwrap()]
+    f.lf.tx_lpf_right_edge_y[(align_h * tile_col + t.by).try_into().unwrap()..]
+        [..sb_step.try_into().unwrap()]
```
kkysen, line 4241:
```suggestion
    f.lf.tx_lpf_right_edge_y[(align_h * tile_col + t.by) as usize..]
        [..sb_step as usize]
```

You can change these to `as usize`s instead of `.try_into().unwrap()`s by the way.  I had done that out of precaution since it was being used as a length in `slice::from_raw_parts_mut` and if it overflowed, it'd be UB, as opposed to an out-of-bounds slicing panic.

Same for the others like this.

src/internal.rs
```diff
@@ -459,11 +459,11 @@ pub struct Rav1dFrameContext_lf {
     pub cdef_buf_plane_sz: [c_int; 2], /* stride*sbh*4 */
     pub cdef_buf_sbh: c_int,
     pub lr_buf_plane_sz: [c_int; 2], /* (stride*sbh*4) << sb128 if n_tc > 1, else stride*4 */
-    pub re_sz: c_int,                /* h */
     pub lim_lut: Align16<Av1FilterLUT>,
     pub last_sharpness: c_int,
     pub lvl: [[[[u8; 2]; 8]; 4]; 8], /* [8 seg_id][4 dir][8 ref][2 is_gmv] */
-    pub tx_lpf_right_edge: [*mut u8; 2],
+    pub tx_lpf_right_edge_y: Vec<u8>, /* len = h */
+    pub tx_lpf_right_edge_uv: Vec<u8>, /* len = h */
```
kkysen, line 466: I'm not sure it helps so much to split them into separate `Vec`s.  It is a simple implementation, but I don't think keeping them contiguous, which clearly had some benefit, is too hard.  Can we do this?

```rust
impl Rav1dFrameContext_lf {
    pub fn tx_lpf_right_edge(&self) -> [&[u8]; 2] {
        let mid = self.tx_lpf_right_edge.len() / 2;
        let (y, uv) = self.tx_lpf_right_edge.split_at(mid);
        [y, uv]
    }

    pub fn tx_lpf_right_edge_mut(&mut self) -> [&mut [u8]; 2] {
        let mid = self.tx_lpf_right_edge.len() / 2;
        let (y, uv) = self.tx_lpf_right_edge.split_at_mut(mid);
        [y, uv]
    }
}
```

Then usages are the same as before (though we could use a tuple instead of a `[_; 2]`; either is fine).

kkysen (changes requested):

kkysen:
Oh also there's an unused `use crate::src::internal::Rav1dFrameContext_frame_thread;` import in `src/lib.rs`.

src/internal.rs
randomPoison: I gave it a shot and didn't run into any borrow issues, so seems like this works!

src/internal.rs
```diff
@@ -459,11 +459,10 @@ pub struct Rav1dFrameContext_lf {
     pub cdef_buf_plane_sz: [c_int; 2], /* stride*sbh*4 */
     pub cdef_buf_sbh: c_int,
     pub lr_buf_plane_sz: [c_int; 2], /* (stride*sbh*4) << sb128 if n_tc > 1, else stride*4 */
-    pub re_sz: c_int,                /* h */
     pub lim_lut: Align16<Av1FilterLUT>,
     pub last_sharpness: c_int,
     pub lvl: [[[[u8; 2]; 8]; 4]; 8], /* [8 seg_id][4 dir][8 ref][2 is_gmv] */
-    pub tx_lpf_right_edge: [*mut u8; 2],
+    pub tx_lpf_right_edge: Vec<u8>,  /* len = h */
```
kkysen, line 465:
```suggestion
    pub tx_lpf_right_edge: Vec<u8>,  // len = h * 2
```

kkysen (approved):
LGTM, just one little comment about the commented length should be double now.

PR #769, opened 2024-02-28
`fn BitDepth::pxstride`: Make generic over `stride` type and `BitDepth`

### LINKED ISSUE #636: Account for negative strides
See https://github.com/memorysafety/rav1d/pull/588#discussion_r1425759037.

kkysen:
This way we don't have to do `BD::pxstride(stride as usize) as isize`, which is verbose and wrong for negative strides.  There may still be some places where strides are always positive, though, so it's kept generic.  This also lets us fix #636 gradually; this is the first major step in fixing #636.  We'll fully fix that later.

@fbossen, hopefully this'll help you in your PRs not have to write the hundreds of extra `as usize) as isize` casts.

### DIFF
--- a/include/common/bitdepth.rs
+++ b/include/common/bitdepth.rs
@@ -174,7 +174,11 @@ pub trait BitDepth: Clone + Copy {
         clip(pixel, 0.into(), self.bitdepth_max())
     }
 
-    fn pxstride(n: usize) -> usize;
+    fn pxstride(n: usize) -> usize {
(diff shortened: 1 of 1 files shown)


fbossen:
> @fbossen, hopefully this'll help you in your PRs not have to write the hundreds of extra `as usize) as isize` casts.

Indeed, this should help. I will clean up the casts after this is merged.

randomPoison (approved):

PR #771, opened 2024-02-28
Backport from dav1d 1.3.0: add `dav1d_version_api` function

fbossen:


randomPoison (approved):

src/lib.rs
```diff
@@ -119,6 +119,27 @@ pub unsafe extern "C" fn dav1d_version() -> *const c_char {
...
 
+/**
+ * Get library API version.
+ *
+ * @return A value in the format 0x00XXYYZZ, where XX is the major version,
+ *         YY the minor version, and ZZ the patch version.
+ */
+#[no_mangle]
+#[cold]
+pub unsafe extern "C" fn dav1d_version_api() -> c_uint {
```
kkysen, line 130: This can be safe right?  Also, do we need to parse it from a string?

kkysen: The others are all `unsafe` either because we haven't made them safe yet, or because they take raw pointers and need to remain `unsafe`.  This one is just `() -> u32`, so it can be safe.

I think we can for now just define them as `const` `u8`s.  In the future, if we match `rav1d`'s `cargo` version to this, we can use `env!("CARGO_PKG_VERSION_MAJOR")`, `env!("CARGO_PKG_VERSION_MINOR")`, and `env!("CARGO_PKG_VERSION_PATCH")` that `cargo` will set for us.

src/lib.rs
```diff
@@ -119,6 +119,27 @@ pub unsafe extern "C" fn dav1d_version() -> *const c_char {
...
+    let DAV1D_API_VERSION = "6.8.0";
+    let dav1d_api_version_vec = DAV1D_API_VERSION
+        .split('.')
+        .map(|x| x.parse::<c_uint>().unwrap())
+        .collect::<Vec<c_uint>>();
+    let DAV1D_API_VERSION_MAJOR = dav1d_api_version_vec[0];
+    let DAV1D_API_VERSION_MINOR = dav1d_api_version_vec[1];
+    let DAV1D_API_VERSION_PATCH = dav1d_api_version_vec[2];
+
+    DAV1D_API_VERSION_MAJOR << 16 | DAV1D_API_VERSION_MINOR << 8 | DAV1D_API_VERSION_PATCH << 0
```
kkysen, line 140:
```suggestion
    const MAJOR: u8 = 6;
    const MINOR: u8 = 8;
    const PATCH: u8 = 0;
    
    u32::from_be_bytes([0, MAJOR, MINOR, PATCH])
```

tools/dav1d.rs
```diff
@@ -344,6 +345,17 @@ unsafe fn main_0(argc: c_int, argv: *const *mut c_char) -> c_int {
...
+    if version_api != 0x00060800 {
+        fprintf(
+            stderr,
+            b"API Version mismatch (library: %08x, executable: %08x)\n\0" as *const u8
+                as *const c_char,
+            version_api,
+            0x00060800,
+        );
+        return 1 as c_int;
+    }
```
kkysen, line 358: Did `dav1d.c` get updated, too?

fbossen: Good point. I added the corresponding patch and updated the rust code to match it.

src/lib.rs
fbossen: I made the function safe.

I had looked at the `cargo` package version, but wasn't convinced it was the right thing to do.

src/lib.rs
```diff
@@ -119,6 +119,27 @@ pub unsafe extern "C" fn dav1d_version() -> *const c_char {
...
+///
+/// Get the `dav1d` library C API version.
+///
+/// @return A value in the format 0x00XXYYZZ, where XX is the major version,
+///         YY the minor version, and ZZ the patch version.
+///
+#[no_mangle]
+#[cold]
+pub extern "C" fn dav1d_version_api() -> c_uint {
+    <u32>::from_be_bytes([
```
kkysen, line 135: Do you need the `<>` around `u32` here?

kkysen (approved):

src/lib.rs
fbossen: No...

PR #777, opened 2024-02-29
`struct Rav1dFrameContext_lf::cdef_line`: Convert pointers to offsets

randomPoison:
The pointers in `cdef_line` were pointing into the buffer owned by `cdef_line_buf`. In converting them to offsets, I had to make sure that that offsets were consistently in terms of pixels (rather than bytes). The original code was inconsistent here, with the logic working with `u8` pointers (and so was using byte offsets), but the code that uses `cdef_line` treats the pointers as `BD::Pixel` pointers, so the offset logic is calculated in terms of pixels. In converting to offsets I normalized on pixel offsets. To do so I had to add a couple of helpers:

* `BPC::pxstride` - Does the same thing as `BitDepth::pxstride`, but works in non generic code. This was needed to calculate pixel offsets during the initialization logic.
* `BitDepth::cast_pixel_slice{_mut}` - Converts a `&[u8]` to a `&[Pixel]`. This was needed so that we could index into the buffer using pixel indices within our bidepth-dependent code.

In practice most of the usages of `cdef_line` is to calculate pointers to be passed into asm functions, so the cleaned up code is still doing some pointer arithmetic, though we never dereference those pointers from Rust code.

### DIFF
--- a/include/common/bitdepth.rs
+++ b/include/common/bitdepth.rs
@@ -12,6 +12,7 @@ use std::ops::Div;
 use std::ops::Mul;
 use std::ops::Rem;
 use std::ops::Shr;
+use std::slice;
 
 pub trait FromPrimitive<T> {
     fn from_prim(t: T) -> Self;
@@ -206,6 +207,34 @@ pub trait BitDepth: Clone + Copy {
     fn bitdepth_max(&self) -> Self::Pixel;
 
     fn get_intermediate_bits(&self) -> u8;
+
+    fn cast_pixel_slice(bytes: &[u8]) -> &[Self::Pixel] {
+        let size = mem::size_of::<Self::Pixel>();
+
+        // Check that the number of elements is a multiple of the new element
+        // size and that the alignment is correct for the new element type.
+        debug_assert!(bytes.len() % size == 0);
+        assert!(bytes.as_ptr() as usize % mem::align_of::<Self::Pixel>() == 0);
+
+        let len = bytes.len() / size;
+
+        // SAFETY: We've checked that alignment and the number of elements is correct.
+        unsafe { slice::from_raw_parts(bytes.as_ptr().cast(), len) }
+    }
+
+    fn cast_pixel_slice_mut(bytes: &mut [u8]) -> &mut [Self::Pixel] {
+        let size = mem::size_of::<Self::Pixel>();
+
+        // Check that the number of elements is a multiple of the new element
+        // size and that the alignment is correct for the new element type.
+        debug_assert!(bytes.len() % size == 0);
+        assert!(bytes.as_ptr() as usize % mem::align_of::<Self::Pixel>() == 0);
+
+        let len = bytes.len() / size;
+
+        // SAFETY: We've checked that alignment and the number of elements is correct.
+        unsafe { slice::from_raw_parts_mut(bytes.as_mut_ptr().cast(), len) }
+    }
 
     const PREP_BIAS: i16;
 
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -3,6 +3,7 @@ use crate::include::common::bitdepth::BitDepth16;
 use crate::include::common::bitdepth::BitDepth8;
 use crate::include::common::bitdepth::DynCoef;
 use crate::include::common::bitdepth::DynPixel;
+use crate::include::common::bitdepth::BPC;
 use crate::include::common::intops::apply_sign64;
 use crate::include::common::intops::iclip;
 use crate::include::common::intops::iclip_u8;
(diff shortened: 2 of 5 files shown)


include/common/bitdepth.rs
```diff
@@ -193,6 +204,26 @@ pub trait BitDepth: Clone + Copy {
 
     fn get_intermediate_bits(&self) -> u8;
 
+    fn cast_pixel_slice(bytes: &[u8]) -> &[Self::Pixel] {
+        let len = Self::pxstride(bytes.len());
+
+        assert!(bytes.len() % len == 0);
+        assert!(bytes.as_ptr() as usize % mem::align_of::<Self::Pixel>() == 0);
```
kkysen, line 211: We don't want to have to check this everytime we access the pixels, so it'd be better to check it once during creation.  I'd been planning to create a type for this.  This is probably fine in the meantime, though, just worst perf.

include/common/bitdepth.rs
```diff
@@ -193,6 +204,26 @@ pub trait BitDepth: Clone + Copy {
 
     fn get_intermediate_bits(&self) -> u8;
 
+    fn cast_pixel_slice(bytes: &[u8]) -> &[Self::Pixel] {
+        let len = Self::pxstride(bytes.len());
+
+        assert!(bytes.len() % len == 0);
```
kkysen, line 210:
```suggestion
        assert!(len * mem::size_of::<Self::Pixel>() <= bytes.len());
```

Shoudn't we check the length is in bounds?

include/common/bitdepth.rs
```diff
@@ -96,6 +97,16 @@ impl BPC {
...
+
+    pub fn pxstride<T>(self, n: T) -> T
+    where
+        T: Copy + From<u8> + Div<Output = T>,
+    {
+        match self {
+            BPC::BPC8 => n,
+            BPC::BPC16 => n / 2.into(),
+        }
+    }
```
kkysen, line 109: This is good but I want to slightly rework things so we aren't redundant.  I'll open a PR.

src/decode.rs
```diff
@@ -4399,35 +4400,39 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
...
         // f.lf.cdef_buf_plane_sz = [0, 0];
         f.lf.cdef_line_buf.resize(alloc_sz, 0);
-        let mut ptr = f.lf.cdef_line_buf.as_mut_ptr();
 
-        ptr = ptr.offset(32);
+        let bpc = BPC::from_bitdepth_max(f.bitdepth_max);
+        let y_stride_px = bpc.pxstride(f.cur.stride[0]);
+        let uv_stride_px = bpc.pxstride(f.cur.stride[1]);
+
+        let mut offset: usize = bpc.pxstride(32) as usize;
```
kkysen, line 4408:
```suggestion
        let mut offset = bpc.pxstride(32usize);
```

kkysen (changes requested):
I had some comments about the pixel casting checks, but the rest of the offset stuff all looks good.

include/common/bitdepth.rs
randomPoison: Wait now that I think about it that explanation isn't correct, because we're also changing the type of the pointer. I'm pretty sure we don't need to check the length explicitly, but maybe I should change the logic to use `size_of::<Self::Pixel>` instead of pxstride 🤔

randomPoison: Okay I've rewritten it to use `size_of`, which I think makes it clearer that we're checking the length correctly without relying on the behavior of `pxstride`.

include/common/bitdepth.rs
kkysen: See #780.

include/common/bitdepth.rs
```diff
@@ -193,6 +204,34 @@ pub trait BitDepth: Clone + Copy {
 
     fn get_intermediate_bits(&self) -> u8;
 
+    fn cast_pixel_slice(bytes: &[u8]) -> &[Self::Pixel] {
+        let size = mem::size_of::<Self::Pixel>();
+
+        // Check that the number of elements is a multiple of the new element
+        // size and that the alignment is correct for the new element type.
+        assert!(bytes.len() % size == 0);
```
kkysen, line 212:
```suggestion
        debug_assert!(bytes.len() % size == 0);
```

Not being a multiple shouldn't happen, but it's not UB if it does, so this can just be a `debug_assert!`.  If we use `BPC::pxstride` from #780`, it does that same `debug_assert!` with `%`.  Up to you if you think we should still just inline it here, which does make the safety much clearer.

Same for the `mut` version, too, of course.

kkysen (changes requested):

include/common/bitdepth.rs
kkysen: > What were you thinking wrt making a custom type to handle it? Something like a `PixelBuf` type that wraps a `Vec<u8>` and has helpers for accessing the buffer as either a `&[u8]` or a `&[Pixel]`?

Yeah, basically.  It's a bit trickier for `Rav1dPicture::data` specifically since it supports custom allocators, but that'd be the gist.  The alignment/length stuff would all be checked during construction, and so these safe transmute casts would be no-ops other than dividing the length.

include/common/bitdepth.rs
randomPoison: I'm going to leave the assert inlined, but I'll change it to `debug_assert!`. I had initially been thinking of the case where the source buffer has an odd number of elements, so converting to a `&[u16]` buffer would have a half element at the end. But since we're doing integer division in that case the new length will always be truncated, so we would actually omit that last byte entirely. Though in practice none of this should matter because the buffers are created with the correct length 🤷

kkysen (changes requested):
I think you merged the changes from #780 wrong; they're all overwritten here with the old versions.

randomPoison:
Oh weird, idk how I did that. Should be fixed now.

kkysen (approved):

PR #778, opened 2024-03-01
`enum ObuMetaType`: make a real enum

folkertdev:


src/levels.rs
```diff
@@ -1,13 +1,16 @@
...
 use std::ops::Neg;
-use strum::EnumCount;
-
-pub type ObuMetaType = c_uint;
-pub const OBU_META_TIMECODE: ObuMetaType = 5;
-pub const OBU_META_ITUT_T35: ObuMetaType = 4;
-pub const OBU_META_SCALABILITY: ObuMetaType = 3;
-pub const OBU_META_HDR_MDCV: ObuMetaType = 2;
-pub const OBU_META_HDR_CLL: ObuMetaType = 1;
+use strum::{EnumCount, FromRepr};
```
kkysen, line 3: Could you put one import per line?  I've been doing that with these rust-analyzer settings in my `.vscode/settings.json`:

```json
    "rust-analyzer.imports.granularity.group": "item",
    "rust-analyzer.imports.prefix": "crate",
```

src/levels.rs
```diff
@@ -1,13 +1,16 @@
...
+
+#[repr(u32)]
```
kkysen, line 5: Why do we need this?  Shouldn't the below code still work fine if it defaults to `u8`?

kkysen (changes requested):
Looks good!  Just a couple little things.

src/levels.rs
folkertdev: generally yes, but it's slightly less convenient. To convert from `u32` to `ObuMetaType` you'd now need to go via `u8`, something like

```rust
impl TryFrom<c_uint> for ObuMetaType {
    type Error = Rav1dError;

    fn try_from(value: c_uint) -> Result<Self, Self::Error> {
        let value = u8::try_from(value).map_err(|_| Rav1dError::EINVAL)?;
        Self::from_repr(value).ok_or(Rav1dError::EINVAL)
    }
}
```

that's just more code for no real gain in my view, but happy to go with that solution.

src/levels.rs
```diff
@@ -1,13 +1,27 @@
...
 use strum::EnumCount;
+use strum::FromRepr;
 
-pub type ObuMetaType = c_uint;
-pub const OBU_META_TIMECODE: ObuMetaType = 5;
-pub const OBU_META_ITUT_T35: ObuMetaType = 4;
-pub const OBU_META_SCALABILITY: ObuMetaType = 3;
-pub const OBU_META_HDR_MDCV: ObuMetaType = 2;
-pub const OBU_META_HDR_CLL: ObuMetaType = 1;
+#[repr(u8)]
```
kkysen, line 7:
```suggestion
```

src/levels.rs
```diff
@@ -1,13 +1,27 @@
...
+}
+
+impl TryFrom<u32> for ObuMetaType {
+    type Error = Rav1dError;
+
+    fn try_from(value: u32) -> Result<Self, Self::Error> {
+        let value = u8::try_from(value).map_err(|_| Rav1dError::EINVAL)?;
+        Self::from_repr(value).ok_or(Rav1dError::EINVAL)
+    }
+}
```
kkysen, line 24:
```suggestion
```

src/obu.rs
```diff
@@ -2349,14 +2344,14 @@ unsafe fn parse_obus(
...
-            let meta_type = gb.get_uleb128() as ObuMetaType;
+            let meta_type = gb.get_uleb128();
             let meta_type_len = ((gb.pos() - init_bit_pos) >> 3) as c_int;
             if gb.has_error() != 0 {
                 return Err(EINVAL);
             }
 
-            match meta_type {
-                OBU_META_HDR_CLL => {
+            match ObuMetaType::try_from(meta_type).ok() {
```
kkysen, line 2353:
```suggestion
            match ObuMetaType::from_repr(meta_type as usize) {
```

kkysen (changes requested):
Could you just do this instead?  Without a `#[repr]`, it seems `from_repr` defaults to `usize`, which we can just `as` cast to.

kkysen (approved):

PR #781, opened 2024-03-03
`enum Rav1dObuType`: make a real enum

folkertdev:


include/dav1d/headers.rs
```diff
@@ -59,24 +59,27 @@ pub(crate) const RAV1D_REFS_PER_FRAME: usize = DAV1D_REFS_PER_FRAME;
...
+
+#[derive(Clone, Copy, PartialEq, Eq, FromRepr)]
+pub enum Rav1dObuType {
+    SeqHdr = 1,
+    Td = 2,
+    FrameHdr = 3,
+    TileGrp = 4,
+    Metadata = 5,
+    Frame = 6,
+    RedundantFramehdr = 7,
```
kkysen, line 80:
```suggestion
    RedundantFrameHdr = 7,
```

kkysen (approved):
Just one little naming thing for one of the variants but otherwise it all LGTM!

src/obu.rs
```diff
@@ -2132,21 +2124,22 @@ unsafe fn parse_obus(
 
     // obu header
     gb.get_bit(); // obu_forbidden_bit
-    let r#type = gb.get_bits(4) as Rav1dObuType;
-    let has_extension = gb.get_bit() as c_int;
-    let has_length_field = gb.get_bit() as c_int;
+    let raw_type = gb.get_bits(4);
+    let r#type = Rav1dObuType::from_repr(raw_type as usize);
+    let has_extension = gb.get_bit();
+    let has_length_field = gb.get_bit();
```
folkertdev, line 2130: a general question: how important is the performance of `GetBits`? its `refill` method only reads data one byte at a time, which in microbenchmarks is quite inefficient. But maybe in practice this is just not a performance-sensitive part of the code?

This is the logic we've taken from `zlib-ng`: https://github.com/memorysafety/zlib-rs/blob/main/zlib-rs/src/inflate/bitreader.rs#L103

src/obu.rs
```diff
@@ -2132,21 +2124,22 @@ unsafe fn parse_obus(
 
     // obu header
     gb.get_bit(); // obu_forbidden_bit
-    let r#type = gb.get_bits(4) as Rav1dObuType;
-    let has_extension = gb.get_bit() as c_int;
-    let has_length_field = gb.get_bit() as c_int;
+    let raw_type = gb.get_bits(4);
+    let r#type = Rav1dObuType::from_repr(raw_type as usize);
```
folkertdev, line 2128: we need to keep the raw integer around for logging later. 

It would be nice if rust could represent this in some other way. `non_exhaustive` seems like it would work here, but that is only an API thing: it does not change the repr. I haven't been able to find an RFC in this direction, though surely one must exist somewhere.

kkysen: What do you mean exactly?

src/obu.rs
kkysen: I don't believe it's very performance critical.  I remember @randomPoison did some benchmarks and I don't think it mattered much, since it's just used to parse the headers.  It would be not too hard to swap out the implementation, though, like the approach you took, but we did already make it safe.  I also looked into `bitvec` a bit but the generated code didn't look as optimized.  I was also hoping to make APIs like `.get_bits` const generic and return `uN` like types, as that also avoids bounds checking for a bunch of arrays.  It sucks that Rust doesn't have `uN` native types, though.  I explored that a bit but it wasn't a priority since we already made it safe, but once we finish safety, I'd like to revisit it to squeeze some more performance out.

kkysen (approved):

src/obu.rs
folkertdev: the problem is this restriction on enums https://doc.rust-lang.org/nomicon/other-reprs.html

> Fieldless enums with repr(C) or repr(u*) still may not be set to an integer value without a corresponding variant, even though this is permitted behavior in C or C++. It is undefined behavior to (unsafely) construct an instance of an enum that does not match one of its variants. (This allows exhaustive matches to continue to be written and compiled as normal.)

It would be nice for this sort of low-level code (I've also seen this come up in zlib and ntpd-rs) to opt-in to not having that restriction. Then any match on the type would have to be non-exhaustive, but you could do something like

```rust
#[repr(u8, no_niche)]
enum MyEnum {
    Foo,
    Bar,
}

match my_enum {
    Foo => x,
    Bar => y,
    other @ _ => report_error(other as u8), 
}
```

which could make this work because this mapping is now total

```rust
3u8 as MyEnum
```

there is some talk about range types which maybe would even allow you to write

```rust
#[repr(u8)]
enum MyEnum {
    Foo = 1,
    Bar = 2,
    Rest(u8) = 0 | 3..
}
```

which then can be matched on exhaustively

folkertdev: I've asked about it on the rust zullip https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/enum.20range.20variants/near/424528113 

Looks like this idea does have a lot of overlap with the range types idea, but there are some problems to fix around the representation and guarantees in FFI. Hopefully I've nerdsniped the right people :)

PR #783, opened 2024-03-04
`enum Rav1dTxfmMode`: make a real enum

folkertdev:


include/dav1d/headers.rs
```diff
@@ -79,16 +79,31 @@ pub(crate) const RAV1D_OBU_TD: Rav1dObuType = DAV1D_OBU_TD;
...
+pub const DAV1D_TX_SWITCHABLE: Dav1dTxfmMode = Rav1dTxfmMode::Switchable as Dav1dTxfmMode;
 
-pub(crate) type Rav1dTxfmMode = c_uint;
-pub(crate) const _RAV1D_N_TX_MODES: usize = DAV1D_N_TX_MODES;
-pub(crate) const RAV1D_TX_SWITCHABLE: Rav1dTxfmMode = DAV1D_TX_SWITCHABLE;
-pub(crate) const RAV1D_TX_LARGEST: Rav1dTxfmMode = DAV1D_TX_LARGEST;
-pub(crate) const RAV1D_TX_4X4_ONLY: Rav1dTxfmMode = DAV1D_TX_4X4_ONLY;
+#[derive(Clone, Copy, PartialEq, Eq, FromRepr, EnumCount)]
+pub enum Rav1dTxfmMode {
+    Only4x4 = 0,
```
folkertdev, line 89: had to get slightly creative with the name here. `r#` is kind of ugly, but an option I guess if preserving the name exactly is important.

kkysen (approved):

include/dav1d/headers.rs
kkysen: This should be totally fine I think.

PR #787, opened 2024-03-05
`enum Rav1dAdaptiveBool`: make a real enum

folkertdev:


src/obu.rs
```diff
@@ -2015,8 +2012,8 @@ unsafe fn parse_frame_hdr(
         showable_frame,
         error_resilient_mode,
         disable_cdf_update,
-        allow_screen_content_tools,
-        force_integer_mv,
+        allow_screen_content_tools: allow_screen_content_tools as c_int,
+        force_integer_mv: force_integer_mv as c_int,
```
folkertdev, line 2016: can the field types be changed for `Rav1dFrameHeader`? it is `repr(C)` so the layout would change when turning these fields into booleans

kkysen (approved):

PR #788, opened 2024-03-06
`enum Rav1dRestorationType`: make a real enum

folkertdev:
and turn some fields into `bool`.


include/dav1d/headers.rs
```diff
@@ -2164,14 +2171,20 @@ pub struct Rav1dFrameHeader_restoration {
 impl From<Dav1dFrameHeader_restoration> for Rav1dFrameHeader_restoration {
     fn from(value: Dav1dFrameHeader_restoration) -> Self {
         let Dav1dFrameHeader_restoration { r#type, unit_size } = value;
-        Self { r#type, unit_size }
+        Self {
+            r#type: r#type.map(|e| Rav1dRestorationType::from_repr(e as usize).unwrap()),
+            unit_size,
+        }
     }
 }
```
folkertdev, line 2179: maybe this should be a `TryFrom` now?

kkysen: Yeah, that'd be good.

kkysen: Actually, leave it as an `.unwrap()`.  This is called when round-tripping across the C `DAV1D_API`, so there's not really an opportunity to bubble up an error.

src/obu.rs
```diff
@@ -1302,48 +1301,59 @@ fn parse_cdef(
...
+                let mut unit_size_0 = 6 + seqhdr.sb128;
+                if gb.get_bit() {
+                    unit_size_0 += 1;
+                    if seqhdr.sb128 == 0 {
+                        unit_size_0 += gb.get_bit() as c_int;
+                    }
+                }
+
+                let unit_size_1 = if (r#type[1] != Rav1dRestorationType::None
+                    || r#type[2] != Rav1dRestorationType::None)
```
kkysen, line 1343:
```suggestion
                let unit_size_1 = if !(matches!(r#type, [_, Rav1dRestorationType::None, Rav1dRestorationType::None]))
```

kkysen (approved):

src/obu.rs
folkertdev: sadly your suggestion formats as

```rust
                let unit_size_1 = if !(matches!(
                    r#type,
                    [_, Rav1dRestorationType::None, Rav1dRestorationType::None]
                )) && seqhdr.ss_hor == 1
                    && seqhdr.ss_ver == 1
                {
                    unit_size_0 - gb.get_bit() as c_int
                } else {
                    unit_size_0
                };
```
which I think is less clear

kkysen: Oh yeah, it's fine as is then.  Maybe if `Rav1dRestorationType::None` were replaced with `Option`'s `None` it'd be better, but that seems like more work than it's worth.

PR #789, opened 2024-03-06
`enum Rav1dWarpedMotionType`: make a real enum

folkertdev:


src/refmvs.rs
```diff
@@ -710,12 +710,14 @@ pub(crate) unsafe fn rav1d_refmvs_find(
...
-                mv::INVALID
-            };
+
+        gmv[0] = if (*rf.frm_hdr).gmv[r#ref.r#ref[0] as usize - 1].r#type
+            > Rav1dWarpedMotionType::Translation
+        {
+            tgmv[0]
+        } else {
+            mv::INVALID
+        };
```
folkertdev, line 720: this could be
```suggestion
                gmv[0] = match (*rf.frm_hdr).gmv[r#ref.r#ref[0] as usize - 1].r#type {
                    Rav1dWarpedMotionType::RotZoom | Rav1dWarpedMotionType::Affine => tgmv[0],
                    Rav1dWarpedMotionType::Identity | Rav1dWarpedMotionType::Translation => mv::INVALID,
            };

```

the code relies on the ordering, which is always confusing (no way you don't need to look at the definition of this type when trying to understand the code) but also moving away from that is fragile if the ordering ever changes. Although because the match is exhaustive the compiler will force you to look here anyway if a new variant gets added and reordering existing variants is unlikely.

kkysen: I think using `>` is fine since `>` is used similarly in a bunch of other places, and because it's done that way in C.  So if we need to backport something, it'll be easy with `>`.  Personally, I'd probably do something more similar to this, in a method, but I think it's a good idea to stick to how it's done in C, and it's not terrible.

kkysen (approved):

kkysen:
@folkertdev, could you fix the conflicts, then I'll merge? Thanks!

kkysen (approved):

PR #791, opened 2024-03-10
`enum Rav1dProfile`: make a real enum

folkertdev:
This one does not have a `Dav1d` counterpart. The variant names are a based on the spec (https://aomediacodec.github.io/av1-spec/av1-spec.pdf, A.2, page 638)


kkysen (approved):
Thanks for finding the named values for this!

PR #792, opened 2024-03-11
`Rav1dFrameContext_lf::lr_line_buf`: Make into `AlignedVec64`

randomPoison:


kkysen (approved):

src/decode.rs
```diff
@@ -4449,18 +4449,14 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
     if y_stride * num_lines as isize != f.lf.lr_buf_plane_sz[0] as isize
         || uv_stride * num_lines as isize * 2 != f.lf.lr_buf_plane_sz[1] as isize
     {
```
kkysen, line 4451: Based @fbossen's comment here, https://github.com/memorysafety/rav1d/pull/793#issuecomment-1994241745, it seems this `if` should be removed, right?

I think we can also get rid of `lr_buf_plane_sz` and just calculate it from `f.sr_cur.p.stride`, `f.sbh`, `seq_hdr.sb128`, and `c.tc.len()` the one time we need it later on.

PR #793, opened 2024-03-11
`struct Rav1dFrameContext_lf::lr_lpf_line`: Convert to offsets

randomPoison:


src/lf_apply.rs
```diff
@@ -161,29 +161,21 @@ pub(crate) unsafe fn rav1d_copy_lpf<BD: BitDepth>(
...
-    let dst_offset: [usize; 2] = [
+    let yuv_offset: [usize; 2] = [
         (tt_off as isize * y_stride - cmp::min(y_span, 0)) as usize,
         (tt_off as isize * uv_stride - cmp::min(uv_span, 0)) as usize,
     ];
+    let dst_offset = [
+        f.lf.lr_lpf_line[0].wrapping_add_signed(cmp::min(y_span, 0)) + yuv_offset[0],
+        f.lf.lr_lpf_line[1].wrapping_add_signed(cmp::min(uv_span, 0)) + yuv_offset[1],
+        f.lf.lr_lpf_line[2].wrapping_add_signed(cmp::min(uv_span, 0)) + yuv_offset[1],
+    ];
```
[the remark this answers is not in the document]
randomPoison: Splitting up `lr_line_buf` into separate slices isn't necessary here. We were passing `dst[i]` plus an offset into `backup_lpf`, so the most straightforward thing to do is combine the offsets and pass all of `lr_line_buf` in along with the calculated offset. I don't see any reason to go through the extra steps of `split_at_mut`ing the slice only to still pass in a buffer+offset pair.

randomPoison:
This is blocked on #801. The test failures we're seeing originate from the out-of-bounds pointer access that is already present on `main`, and moving to slice operations triggered an array OOB panic. As far as I can tell this change didn't introduce the issue, so we'll need to fix #801 first and rebase this on top.

src/lr_apply.rs
```diff
@@ -47,12 +46,8 @@ unsafe fn lr_stripe<BD: BitDepth>(
     let sby = y + (if y != 0 { 8 << ss_ver } else { 0 }) >> 6 - ss_ver + seq_hdr.sb128;
     let have_tt = (c.tc.len() > 1) as c_int;
     let lpf_stride = BD::pxstride(stride);
-    let lpf_plane_sz = BD::pxstride(f.lf.lr_buf_plane_sz[(plane != 0) as usize] as isize);
-    let mut lpf_offset = cmp::max(lpf_stride - lpf_plane_sz, 0);
```
randomPoison, line 51: @fbossen I ended up removing `lpf_plane_sz` and `lpf_offset` in order to make the calculation here the same as the original C. You had added those in #746 to handle negative strides I think. Do we still need those with the fixed `pxstride`? Was there a reason why we needed those values in Rust but not in the original C?

fbossen: The reason I added `lpf_plane_sz` and `lpf_offset` is that I defined `lpf` as a slice that begins at the memory section being effectively used and needed the compute the memory location of the beginning of the slice (and cover the case of a negative stride). If, instead of defining a new slice `lpf`, you use `lr_line_buf`, then removing `lpf_plane_sz` and `lpf_offset` looks appropriate.

fbossen (approved):

fbossen (commented):
Upon further testing, it looks like I am hitting an error on one of the test bitstreams. Let me look into it and report more details later.

src/lf_apply.rs
```diff
@@ -161,29 +161,21 @@ pub(crate) unsafe fn rav1d_copy_lpf<BD: BitDepth>(
     let y_span = lr_plane_sz[0] as isize - y_stride;
```
fbossen, line 161: I think you should be able to remove `y_span` and `uv_span` here as well. They were used to compute the base address of the slice. It seems superfluous to subtract `y_span` in `yuv_offset[0]` and add it back in `dst_offset[0]`

fbossen:
> Upon further testing, it looks like I am hitting an error on one of the test bitstreams. Let me look into it and report more details later.

The issues boils down to the following: in `fn rav1d_decode_frame_init` `f.lf.lr_line_buf` is reallocated if needed. The offsets `lr_lpf_line` are updated upon reallocation. However, there are cases where `f.lf.lr_line_buf` doesn't need to be reallocated but the offsets need updating: for example, if the picture width doubles and the pixel size is halved. (e.g., `y_stride` doesn't change but `y_stride_px` does).

src/lr_apply.rs
```diff
@@ -102,7 +97,9 @@ unsafe fn lr_stripe<BD: BitDepth>(
             p.as_mut_ptr().add(p_offset).cast(),
             stride,
             left.as_ptr().cast(),
-            lpf.as_ptr().offset(lpf_offset).cast(),
+            // NOTE: The calculated pointer may point to before the beginning of
+            // `lr_line_buf`, so we must use `wrapping_offset` here.
```
kkysen, line 101:
```suggestion
            // NOTE: The calculated pointer may point to before the beginning of
            // `lr_line_buf`, so we must use `.wrapping_offset` here.
            // `.wrapping_offset` is needed since `.offset` requires the pointer is in bounds,
            // which `.wrapping_offset` does not, and delays that requirement to when the pointer is dereferenced.
```

randomPoison:
> The issues boils down to the following: in `fn rav1d_decode_frame_init` `f.lf.lr_line_buf` is reallocated if needed. The offsets `lr_lpf_line` are updated upon reallocation. However, there are cases where `f.lf.lr_line_buf` doesn't need to be reallocated but the offsets need updating

Huh, is that being caused by the changes in this PR? I haven't changed anything related to when we do the re-allocation and update `lr_lpf_line`, so I would think the behavior there is the same as before. I can remove the `if` condition around the initialization logic so that we always resize and recompute `lr_lpf_line`, that would fix the issue I think.

fbossen:
> Huh, is that being caused by the changes in this PR? I haven't changed anything related to when we do the re-allocation and update `lr_lpf_line`, so I would think the behavior there is the same as before.

Yes. Changes in this PR cause the issue. If `lr_lpf_line` was always in units of `u8`, then there would be no issue. But `lr_lpf_line` is now in units of `u8` or `u16` depending on the bit depth. So if the bit depth changes, but the amount (in bytes) of allocated memory doesn't, the values of `lr_lpf_line` require updating.

randomPoison:
Okay, thanks for confirming! I've removed the conditional logic for initializing `lr_lpf_line` so it should be updated properly now if the bitdepth changes.

fbossen (approved):
LGTM

PR #797, opened 2024-03-11
`enum Rav1dMatrixCoefficients`: make a real enum

folkertdev:
https://aomediacodec.github.io/av1-spec/av1-spec.pdf#page=130.

### DIFF
--- a/include/dav1d/headers.rs
+++ b/include/dav1d/headers.rs
@@ -543,38 +543,67 @@ pub(crate) const RAV1D_TRC_UNKNOWN: Rav1dTransferCharacteristics = DAV1D_TRC_UNK
 pub(crate) const _RAV1D_TRC_BT709: Rav1dTransferCharacteristics = DAV1D_TRC_BT709;
 
 pub type Dav1dMatrixCoefficients = c_uint;
+pub const DAV1D_MC_IDENTITY: Dav1dMatrixCoefficients = Rav1dMatrixCoefficients::IDENTITY.to_dav1d();
+pub const DAV1D_MC_BT709: Dav1dMatrixCoefficients = Rav1dMatrixCoefficients::BT709.to_dav1d();
+pub const DAV1D_MC_UNKNOWN: Dav1dMatrixCoefficients = Rav1dMatrixCoefficients::UNKNOWN.to_dav1d();
+pub const DAV1D_MC_FCC: Dav1dMatrixCoefficients = Rav1dMatrixCoefficients::FCC.to_dav1d();
+pub const DAV1D_MC_BT470BG: Dav1dMatrixCoefficients = Rav1dMatrixCoefficients::BT470BG.to_dav1d();
+pub const DAV1D_MC_BT601: Dav1dMatrixCoefficients = Rav1dMatrixCoefficients::BT601.to_dav1d();
+pub const DAV1D_MC_SMPTE240: Dav1dMatrixCoefficients = Rav1dMatrixCoefficients::SMPTE240.to_dav1d();
+pub const DAV1D_MC_SMPTE_YCGCO: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::SMPTE_YCGCO.to_dav1d();
+pub const DAV1D_MC_BT2020_NCL: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::BT2020_CL.to_dav1d();
+pub const DAV1D_MC_BT2020_CL: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::BT2020_CL.to_dav1d();
+pub const DAV1D_MC_SMPTE2085: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::SMPTE2085.to_dav1d();
+pub const DAV1D_MC_CHROMAT_NCL: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::CHROMAT_NCL.to_dav1d();
+pub const DAV1D_MC_CHROMAT_CL: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::CHROMAT_CL.to_dav1d();
+pub const DAV1D_MC_ICTCP: Dav1dMatrixCoefficients = Rav1dMatrixCoefficients::ICTCP.to_dav1d();
+// dav1d defines this value as 255, but the specification defines it as 3
 pub const DAV1D_MC_RESERVED: Dav1dMatrixCoefficients = 255;
-pub const DAV1D_MC_ICTCP: Dav1dMatrixCoefficients = 14;
-pub const DAV1D_MC_CHROMAT_CL: Dav1dMatrixCoefficients = 13;
-pub const DAV1D_MC_CHROMAT_NCL: Dav1dMatrixCoefficients = 12;
-pub const DAV1D_MC_SMPTE2085: Dav1dMatrixCoefficients = 11;
-pub const DAV1D_MC_BT2020_CL: Dav1dMatrixCoefficients = 10;
-pub const DAV1D_MC_BT2020_NCL: Dav1dMatrixCoefficients = 9;
-pub const DAV1D_MC_SMPTE_YCGCO: Dav1dMatrixCoefficients = 8;
-pub const DAV1D_MC_SMPTE240: Dav1dMatrixCoefficients = 7;
-pub const DAV1D_MC_BT601: Dav1dMatrixCoefficients = 6;
-pub const DAV1D_MC_BT470BG: Dav1dMatrixCoefficients = 5;
-pub const DAV1D_MC_FCC: Dav1dMatrixCoefficients = 4;
-pub const DAV1D_MC_UNKNOWN: Dav1dMatrixCoefficients = 2;
-pub const DAV1D_MC_BT709: Dav1dMatrixCoefficients = 1;
-pub const DAV1D_MC_IDENTITY: Dav1dMatrixCoefficients = 0;
 
-pub(crate) type Rav1dMatrixCoefficients = c_uint;
-pub(crate) const _RAV1D_MC_RESERVED: Rav1dMatrixCoefficients = DAV1D_MC_RESERVED;
-pub(crate) const _RAV1D_MC_ICTCP: Rav1dMatrixCoefficients = DAV1D_MC_ICTCP;
-pub(crate) const _RAV1D_MC_CHROMAT_CL: Rav1dMatrixCoefficients = DAV1D_MC_CHROMAT_CL;
-pub(crate) const _RAV1D_MC_CHROMAT_NCL: Rav1dMatrixCoefficients = DAV1D_MC_CHROMAT_NCL;
-pub(crate) const _RAV1D_MC_SMPTE2085: Rav1dMatrixCoefficients = DAV1D_MC_SMPTE2085;
-pub(crate) const _RAV1D_MC_BT2020_CL: Rav1dMatrixCoefficients = DAV1D_MC_BT2020_CL;
-pub(crate) const _RAV1D_MC_BT2020_NCL: Rav1dMatrixCoefficients = DAV1D_MC_BT2020_NCL;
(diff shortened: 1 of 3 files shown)


include/dav1d/headers.rs
```diff
@@ -543,38 +543,69 @@ pub(crate) const RAV1D_TRC_UNKNOWN: Rav1dTransferCharacteristics = DAV1D_TRC_UNK
...
+    Bt601 = 6,
+    Smpte240 = 7,
+    Smpte_ycgco = 8,
+    Bt2020_ncl = 9,
+    Bt2020_cl = 10,
+    Smpte2085 = 11,
+    Chromat_ncl = 12,
+    Chromat_cl = 13,
+    Ictcp = 14,
+    Reserved = 255,
```
folkertdev, line 593: this `Reserved` variant is not in the spec (https://aomediacodec.github.io/av1-spec/av1-spec.pdf, page 118/119), is this a C implementation detail maybe?

include/dav1d/headers.rs
```diff
@@ -543,38 +543,69 @@ pub(crate) const RAV1D_TRC_UNKNOWN: Rav1dTransferCharacteristics = DAV1D_TRC_UNK
...
+pub const DAV1D_MC_ICTCP: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::Ictcp as Dav1dMatrixCoefficients;
+pub const DAV1D_MC_RESERVED: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::Reserved as Dav1dMatrixCoefficients;
+
+#[derive(Clone, Copy, PartialEq, Eq, FromRepr)]
+pub enum Rav1dMatrixCoefficients {
+    Identity = 0,
+    Bt709 = 1,
+    Unknown = 2,
```
kkysen, line 581:
```suggestion
    Unknown = 2,
    Reserved = 3,
```

include/dav1d/headers.rs
kkysen: Yeah, that does seem weird, especially since 3 is supposed to be reserved.

include/dav1d/headers.rs
```diff
@@ -543,38 +543,69 @@ pub(crate) const RAV1D_TRC_UNKNOWN: Rav1dTransferCharacteristics = DAV1D_TRC_UNK
...
+    Rav1dMatrixCoefficients::Chromat_ncl as Dav1dMatrixCoefficients;
+pub const DAV1D_MC_CHROMAT_CL: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::Chromat_cl as Dav1dMatrixCoefficients;
+pub const DAV1D_MC_ICTCP: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::Ictcp as Dav1dMatrixCoefficients;
+pub const DAV1D_MC_RESERVED: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::Reserved as Dav1dMatrixCoefficients;
+
+#[derive(Clone, Copy, PartialEq, Eq, FromRepr)]
+pub enum Rav1dMatrixCoefficients {
```
kkysen, line 578: Could you keep the case-insensitive names the same?  So for things like `IDENTITY` and `CHROMAT`, `Identity` and `Chromat` are okay, but keep all caps for acronyms, and leave the `_`s.  You can add a `#[allow(non_camel_case_types)]
` to the `enum`.

src/obu.rs
```diff
@@ -393,11 +391,11 @@ fn parse_seq_hdr(c: &mut Rav1dContext, gb: &mut GetBits) -> Rav1dResult<Rav1dSeq
     if color_description_present != 0 {
         pri = gb.get_bits(8) as Rav1dColorPrimaries;
         trc = gb.get_bits(8) as Rav1dTransferCharacteristics;
-        mtrx = gb.get_bits(8) as Rav1dMatrixCoefficients;
+        mtrx = Rav1dMatrixCoefficients::from_repr(gb.get_bits(8) as usize).unwrap();
```
kkysen, line 394: What values is this panicking on?  It's unfortunate that `from_repr` just returns an `Option` instead of `Result` so we don't see what the value was in the error message.  Is it `Reserved = 3`?  Or something `> 14`?

kkysen (changes requested):

src/obu.rs
folkertdev: it seems to be just arbitrary values

```
554/792 dav1d:testdata-8 / Syntax_AV1_mainb8ss420_432x240_001_vq_aom_ctest_4.2  FAIL     0.01 s (exit status 101)

--- command ---
22:37:06 /home/folkertdev/rust/rav1d/tests/../target/release/dav1d -q --threads 1 -i /home/folkertdev/rust/rav1d/build/../tests/dav1d-test-data/8-bit/vq_suite/Syntax_AV1_mainb8ss420_432x240_001_vq_aom_ctest_4.2.stripped.ivf --verify 74cd1c55dc974ee2d13bf6b2a38aeecb
--- stderr ---
thread 'main' panicked at src/obu.rs:397:21:
Rav1dMatrixCoefficients::from_repr(74) will return None
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
-------

555/792 dav1d:testdata-8 / Syntax_AV1_mainb8ss420_432x240_002_vq_aom_ctest_4.2  FAIL     0.00 s (exit status 101)

--- command ---
22:37:06 /home/folkertdev/rust/rav1d/tests/../target/release/dav1d -q --threads 1 -i /home/folkertdev/rust/rav1d/build/../tests/dav1d-test-data/8-bit/vq_suite/Syntax_AV1_mainb8ss420_432x240_002_vq_aom_ctest_4.2.stripped.ivf --verify a9249d720bf9a7dcddce9e358e9aa967
--- stderr ---
thread 'main' panicked at src/obu.rs:397:21:
Rav1dMatrixCoefficients::from_repr(71) will return None
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
-------

556/792 dav1d:testdata-8 / Syntax_AV1_mainb8ss420_432x240_003_vq_aom_ctest_4.2  FAIL     0.01 s (exit status 101)

--- command ---
22:37:06 /home/folkertdev/rust/rav1d/tests/../target/release/dav1d -q --threads 1 -i /home/folkertdev/rust/rav1d/build/../tests/dav1d-test-data/8-bit/vq_suite/Syntax_AV1_mainb8ss420_432x240_003_vq_aom_ctest_4.2.stripped.ivf --verify 713a7b52a34dea39a6ff3f1064fe5a2c
--- stderr ---
thread 'main' panicked at src/obu.rs:397:21:
Rav1dMatrixCoefficients::from_repr(244) will return None
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
-------
```

So that suggests that, to mirror the `dav1d` behavior it should be 


```rust
pub struct Rav1dMatrixCoefficients(u8);

impl Rav1dMatrixCoefficients {
    pub const IDENTITY : Self = Self(0);
    // ...
}
```

that should be fine, because no (exhaustive) matches  are ever performed on this "enum" anyway

src/obu.rs
```diff
@@ -392,11 +390,11 @@ fn parse_seq_hdr(c: &mut Rav1dContext, gb: &mut GetBits) -> Rav1dResult<Rav1dSeq
     if color_description_present != 0 {
         pri = gb.get_bits(8) as Rav1dColorPrimaries;
         trc = gb.get_bits(8) as Rav1dTransferCharacteristics;
-        mtrx = gb.get_bits(8) as Rav1dMatrixCoefficients;
+        mtrx = Rav1dMatrixCoefficients(gb.get_bits(8) as u8)
```
folkertdev, line 393: this could be e.g. `Rav1dMatrixCoefficients::from_byte` and then the field does not need to be `pub`

include/dav1d/headers.rs
```diff
@@ -543,38 +543,67 @@ pub(crate) const RAV1D_TRC_UNKNOWN: Rav1dTransferCharacteristics = DAV1D_TRC_UNK
...
+    Rav1dMatrixCoefficients::BT2020_CL.to_dav1d();
+pub const DAV1D_MC_SMPTE2085: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::SMPTE2085.to_dav1d();
+pub const DAV1D_MC_CHROMAT_NCL: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::CHROMAT_NCL.to_dav1d();
+pub const DAV1D_MC_CHROMAT_CL: Dav1dMatrixCoefficients =
+    Rav1dMatrixCoefficients::CHROMAT_CL.to_dav1d();
+pub const DAV1D_MC_ICTCP: Dav1dMatrixCoefficients = Rav1dMatrixCoefficients::ICTCP.to_dav1d();
+// dav1d defines this value as 255, but the specficiation defines it as 3
 pub const DAV1D_MC_RESERVED: Dav1dMatrixCoefficients = 255;
```
folkertdev, line 567: is there someone we can ask about this? the value here needs to be 255 I think to maintain dav1d compatability (although this value is never used in the codebase)

folkertdev: in practice, at the moment, neither `3` or `255` are used by the tests

src/obu.rs
folkertdev: that can work, but you would loose information on a roundtrip `Ravid::from(David::from(x)) != x` any more (this is already partial of course, but I think otherwise the equality holds today).

Also, at this point, what is the advantage of an enum (versus a struct with constants), if in practice its variants do not map cleanly to the values it is supposed to represent?

kkysen: I *think* we've removed all of the `Dav1d` to `Rav1d` conversions, but I've left the `impl From`s and `impl TryFrom`s in case we need to add them back.  So I don't think having `Dav1d::from(Rav1d::from(x)) != x` is that bad, though you're right that it could be confusing.

The advantage, admittedly very small, of an `enum` is that it could be smaller.

folkertdev: I'm happy to go with the enum, but one final thing I thought of is that if the `mtrx` ever ends up in a (debug) log, that might be confusing. The log will report that the `Reserved` matrix coefficients are used even though the source file might not actually be doing that (it's actually using a value that is not defined by the spec).

include/dav1d/headers.rs
```diff
@@ -543,38 +543,71 @@ pub(crate) const RAV1D_TRC_UNKNOWN: Rav1dTransferCharacteristics = DAV1D_TRC_UNK
...
-pub(crate) const RAV1D_MC_UNKNOWN: Rav1dMatrixCoefficients = DAV1D_MC_UNKNOWN;
-pub(crate) const _RAV1D_MC_BT709: Rav1dMatrixCoefficients = DAV1D_MC_BT709;
-pub(crate) const RAV1D_MC_IDENTITY: Rav1dMatrixCoefficients = DAV1D_MC_IDENTITY;
+
+#[derive(Clone, Copy, PartialEq, Eq, FromRepr)]
+#[allow(non_camel_case_types)]
+pub enum Rav1dMatrixCoefficients {
+    Identity = 0,
+    BT709 = 1,
+    Unspecified = 2,
```
folkertdev, line 582: this is the name used in the specification (dav1d uses `Unknown`).

include/dav1d/headers.rs
```diff
@@ -543,38 +543,71 @@ pub(crate) const RAV1D_TRC_UNKNOWN: Rav1dTransferCharacteristics = DAV1D_TRC_UNK
...
+    FCC = 4,
+    BT470BG = 5,
+    BT601 = 6,
+    SMPTE240 = 7,
+    SMPTE_YCGCO = 8,
+    BT2020_NCL = 9,
+    BT2020_CL = 10,
+    SMTPE2085 = 11,
+    CHROMAT_NCL = 12,
+    CHROMAT_CL = 13,
```
folkertdev, line 593: `CromatNcl` looked weird to me. Now this is consistent with `BT2020_NCL`

src/obu.rs
kkysen: After thinking about it some more, I think your `u8` newtype wrapper is a less risky approach.  If in the future we need to squeeze any more bytes or something, we could consider a different approach, and the newtype approach is still quite ergonomic.

kkysen (approved):

PR #800, opened 2024-03-12
`enum TaskType`: make a real enum

folkertdev:


src/thread_task.rs
```diff
@@ -604,7 +589,7 @@ unsafe fn check_tile(t: *mut Rav1dTask, f: &mut Rav1dFrameData, frame_mt: c_int)
             }
             match current_block_14 {
                 2370887241019905314 => {
-                    let p3 = f.refp[n as usize].progress.as_ref().unwrap()[(tp == 0) as usize]
+                    let p3 = f.refp[n as usize].progress.as_ref().unwrap()[(!tp) as usize]
```
kkysen, line 592:
```suggestion
                    let p3 = f.refp[n as usize].progress.as_ref().unwrap()[!tp as usize]
```

src/thread_task.rs
```diff
@@ -1295,35 +1271,35 @@ pub unsafe fn rav1d_worker_task(c: &Rav1dContext, task_thread: Arc<Rav1dTaskCont
...
-                    RAV1D_TASK_TYPE_RECONSTRUCTION_PROGRESS => {
+                    TaskType::ReconstructionProgress => {
                         // dummy to cover for no post-filters
                     }
-                    RAV1D_TASK_TYPE_ENTROPY_PROGRESS => {
+                    TaskType::EntropyProgress => {
                         // dummy to convert tile progress to frame
                     }
-                    _ => {
+                    TaskType::FgPrep | TaskType::FgApply => {
```
kkysen, line 1302: I think this should remain `_`, as it was a `default:` in C.

kkysen (changes requested):
Mostly very good, a couple of minor things.

src/thread_task.rs
folkertdev: I assumed that was because there are no exhaustive matches in C. I don't see any downsides to the exhaustive match here 

- when you read the code, it's instantly clear what cases are left over and handled by this fallback
- when a new case gets added the exhaustiveness checker will ensure the match is updated
- when a case is removed it's clear what to do

there are cases where making a match exhaustive is very verbose, but that's not the case here either

kkysen: Hmm, if you've thought about it carefully, I think it's okay to change then.  There's not really any risk of something going wrong here.

kkysen (approved):

PR #806, opened 2024-03-15
Fix aarch64 build and test aarch64 on macOS

### LINKED ISSUE #773: rav1d fails to build on aarch64
I tried to build rav1d on aarch64 and get the following error:

```
# cargo build --release
error: cannot produce proc-macro for `atomig-macro v0.3.0` as the target `aarch64-unknown-linux-gnu` does not support these crate types
```

This is using the latest rust on Ubuntu `rustc 1.76.0-nightly (6b771f6b5 2023-11-15)`

thedataking:
- delete `.cargo/config.toml` since it can cause `cargo build` without `--target` to fail. Closes #773.
- test `aarch64` on macOS/arm64 which runs much faster than under QEMU. 
  + Old QEMU/arm64 workflow restricted to specific branches so it is effectively disabled.
- test `armv7` without docker which avoids the time spent provisioning and starting docker on each run.


randomPoison (approved):

kkysen (commented):
We need to keep a good way to run tests with qemu locally.  Is it possible to do this in `build.rs` somehow, maybe using [`cargo:rustc-flags=FLAGS`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-flags)?  Maybe we can have a local-only solution for testing, but if `build.rs` can accommodate this, I think that'd be ideal.

.github/workflows/build-and-test-aarch64-darwin.yml
```diff
@@ -0,0 +1,41 @@
...
+  test-on-macos-aarch64:
+    runs-on: macos-14
+    name: test on macos-14-aarch64
+    steps:
+      - name: install prerequisites
+        run: |
+          rustup target add armv7-unknown-linux-gnueabihf
+          brew tap messense/macos-cross-toolchains
+          brew install \
+            armv7-unknown-linux-gnueabihf \
```
kkysen, line 13: Why do we need `armv7-unknown-linux-gnueabihf` on aarch64 macOS?  Is this a typo?

.github/workflows/build-and-test-aarch64-darwin.yml
```diff
@@ -0,0 +1,41 @@
...
+            meson \
+            docker
```
kkysen, line 15: Do we still need docker?

.github/workflows/build-and-test-aarch64-darwin.yml
thedataking: It is a leftover from me trying to run the armv7 in docker. I ended up keeping the QEMU armv7 workflow instead. Removed.

Issue #811, opened 2024-03-15
Backport updates from dav1d 1.3.0


randomPoison:

fbossen:
Remaining:
- https://code.videolan.org/videolan/dav1d/-/commit/90a45d89de3a42058943efbab5fd9d35ae933dc7
- https://code.videolan.org/videolan/dav1d/-/commit/d437510e9d4061a32845ece547b28cadd716900d
- https://code.videolan.org/videolan/dav1d/-/commit/97becd73726c3d4c4bb8793d2215f846da8795af

Issue #805, opened 2024-03-15
Performance regressions against dav1d master on x86_64


negge:
Similar to https://github.com/memorysafety/rav1d/issues/804, the performance of `target/release/dav1d` on `x86_64` is worse than `dav1d 1.4.0-83-g872e470` on both 8-bit and 10-bit videos in both single-threaded and multi-threaded configurations. There is something especially odd going on with 10-bit multi-threaded where the CPU utilization is almost a 1/3 lower than dav1d, and thus takes 154% longer to decode.

**Single-Threaded**

| | dav1d 1.4.0-83-g872e470 | rav1d 966d63c1 | % delta |
| ------------- | ------------- | ------------- | ------------- |
| 8-bit Wall time (s) | 1:41.30 | 1:49.85 | 8.44% |
| 10-bit Wall time (s) | 2:15.80  | 2:25.11 | 6.86% |
| 8-bit RSS (kbytes) | 35024 | 36352 | 3.79% |
| 10-bit RSS (kbytes) | 62228 | 63628 | 2.24% |

**Multi-Threaded**

| | dav1d 1.4.0-83-g872e470 | rav1d 966d63c1 | % delta |
| ------------- | ------------- | ------------- | ------------- |
| 8-bit Wall time (s) | 0:18.98 |  0:20.77 | 9.43% |
| 10-bit Wall time (s) |  0:21.56 | 0:54.86 | 154.45% |
| 8-bit RSS (kbytes) | 170752 | 214948 | 25.88% |
| 10-bit RSS (kbytes) | 276168  | 291084 | 5.40%  |

 **Single-Threaded Data**

```
nathan@vulpes ~/git/dav1d/build $ /usr/bin/time -v tools/dav1d --threads 1 -i ~/Videos/Chimera/Chimera-AV1-8bit-1920x1080-6736kbps.ivf -o /dev/null
dav1d 1.4.0-83-g872e470 - by VideoLAN
Decoded 8929/8929 frames (100.0%) - 88.15/23.98 fps (3.68x)
	Command being timed: "tools/dav1d --threads 1 -i /home/nathan/Videos/Chimera/Chimera-AV1-8bit-1920x1080-6736kbps.ivf -o /dev/null"
	User time (seconds): 101.09
	System time (seconds): 0.11
	Percent of CPU this job got: 99%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 1:41.30
	Average shared text size (kbytes): 0
[...]

PR #821, opened 2024-03-15
`fn dav1d_load_tmvs_sse4`: backport `x86_64` asm function from dav1d 1.3.0

fbossen:
Relates to #811 and #805.

### DIFF
--- a/src/refmvs.rs
+++ b/src/refmvs.rs
@@ -150,9 +150,9 @@ pub struct refmvs_block(pub refmvs_block_unaligned);
 
 #[repr(C)]
 pub(crate) struct refmvs_frame {
-    /// A pointer to a `struct refmvs_frame` may be passed to a function of type `load_tmvs_fn`.
-    /// However, the `frm_hdr` pointer is not accessed in such a function. Thus, it is
-    /// safe to have a pointer to `Rav1dFrameHeader` instead of `Dav1dFrameHeader` here
+    /// A pointer to a [`refmvs_frame`] may be passed to a [`load_tmvs_fn`] function.
+    /// However, the [`Self::frm_hdr`] pointer is not accessed in such a function (see [`load_tmvs_c`]).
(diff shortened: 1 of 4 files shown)


fbossen:
Relates to #811 and #805

src/refmvs.rs
```diff
@@ -142,6 +150,9 @@ pub struct refmvs_block(pub refmvs_block_unaligned);
 
 #[repr(C)]
 pub(crate) struct refmvs_frame {
+    /// A pointer to a `struct refmvs_frame` may be passed to a function of type `load_tmvs_fn`.
+    /// However, the `frm_hdr` pointer is not accessed in such a function. Thus, it is
+    /// safe to have a pointer to `Rav1dFrameHeader` instead of `Dav1dFrameHeader` here
```
kkysen, line 155:
```suggestion
    /// A pointer to a [`refmvs_frame`] may be passed to a [`load_tmvs_fn`] function.
    /// However, the [`Self::frm_hdr`] pointer is not accessed in such a function (see [`load_tmvs_c`]).
    /// Thus, it is safe to have a pointer to [`Rav1dFrameHeader`] instead of [`Dav1dFrameHeader`] here.
```

This is unfortunate, as previously `refmvs_frame` was not passed to asm and didn't need to remain `#[repr(C)]` and compatible.  Luckily this field isn't used in asm `fn`s, though now I'm worried that it may in the future.  Hopefully not, as that would mess up a lot of things.

Eventually, I think we should make this field `pub _frm_hdr: *const ()` and pass `&Rav1dFrameHeader` down the callstack where it's needed (if this possible, which I hope it is, since no `Rav1dRef` is stored here).  That fixes the lifetime issue of the pointer and makes it clear that this field should not be accessed.

kkysen (changes requested):
I just had a suggestion for the doc comment.  The rest of my comment there is more about the future of the field, but this PR is good to go besides the comment suggestion.

kkysen (approved):

Issue #803, opened 2024-03-14
aarch64-unknown-linux-gnu/release/dav1d incorrectly lists and parses cpu architectures for x86_64


negge:
Building for `aarch64` with the command `cargo build --release --target aarch64-unknown-linux-gnu` as described in https://github.com/memorysafety/rav1d/issues/773#issuecomment-1971788036 correctly produces a 64-bit `dav1d` binary.
```
negge@arm1:~/git/rav1d# file target/aarch64-unknown-linux-gnu/release/dav1d
target/aarch64-unknown-linux-gnu/release/dav1d: ELF 64-bit LSB executable, ARM aarch64, version 1 (GNU/Linux), statically linked, BuildID[sha1]=bc36ccb2760f4280851164dc8a5feddad1553535, for GNU/Linux 3.7.0, with debug_info, not stripped
```

However, looking at the help, it incorrectly shows cpumask specializations for `x86_64` architectures.
```
negge@arm1:~/git/rav1d# target/aarch64-unknown-linux-gnu/release/dav1d --help
target/aarch64-unknown-linux-gnu/release/dav1d: unrecognized option '--help'
Usage: target/aarch64-unknown-linux-gnu/release/dav1d [options]

Supported options:
 --input/-i $file:     input file
 --output/-o $file:    output file (%n, %w or %h will be filled in for per-frame files)
 --demuxer $name:      force demuxer type ('ivf', 'section5' or 'annexb'; default: detect from content)
 --muxer $name:        force muxer type ('md5', 'yuv', 'yuv4mpeg2' or 'null'; default: detect from extension)
                       use 'frame' as prefix to write per-frame files; if filename contains %n, will default to writing per-frame files
 --quiet/-q:           disable status messages
 --frametimes $file:   dump frame times to file
 --limit/-l $num:      stop decoding after $num frames
 --skip/-s $num:       skip decoding of the first $num frames
 --realtime [$fract]:  limit framerate, optional argument to override input framerate
 --realtimecache $num: set the size of the cache in realtime mode (default: 0)
 --version/-v:         print version and exit
 --threads $num:       number of threads (default: 0)
 --framedelay $num:    maximum frame delay, capped at $threads (default: 0);
                       set to 1 for low-latency decoding
 --filmgrain $num:     enable film grain application (default: 1, except if muxer is md5 or xxh3)
 --oppoint $num:       select an operating point of a scalable AV1 bitstream (0 - 31)
[...]

PR #825, opened 2024-03-16
Link `librav1d` against C `dav1d` and `seek_stress` instead of Rust versions

### LINKED ISSUE #807: Use original C code for dav1d CLI
Currently we're using a c2rust translated version of the dav1d CLI. Instead we would like to use the original C for dav1d and have it link against rav1d. This would help test that the rav1d C API matches dav1d, and would fix issues like #803.

thedataking:
Closes #807 and #803.

When the `test_rust` meson option (on by default) is set, use a custom meson command to build `librav1d.a` and link `dav1d` and `seek_stress` with it instead of `libdav1d`. 

Notes:
- Using the subprojects feature of meson doesn't seem to be an option since it requires a specific directory layout.
- Cross-compiling isn't supported.
- The macOS build behaves differently than local macOS, possibly due to meson version differences. Didn't root cause.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -472,7 +472,27 @@ endif
 # Generate config.h
 config_h_target = configure_file(output: 'config.h', configuration: cdata)
 
+# Use --out-dir instead of --target-dir to fully control where librav1d.a goes.
+# Using --target-dir would put it in a subdirectory. The --out-dir option is
+# currently unstable so we need to tell cargo to enable those.
+# FIXME: this will only work if the host matches the build target.
+cargo_command = [
+    'cargo', 'build', '--lib',
+    '-Z', 'unstable-options',
+    '--out-dir', meson.current_build_dir()
+]
+if get_option('buildtype') == 'release'
+    cargo_command += ['--release']
+endif
 
+librav1d = custom_target(
+    'librav1d',
+    output: 'librav1d.a',
+    input: ['lib.rs'],
+    command: cargo_command,
+    install: true,
+    install_dir: get_option('libdir')
+)
 
 #
 # Include subdir meson.build files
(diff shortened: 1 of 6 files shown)


meson.build
```diff
@@ -472,7 +472,27 @@ endif
...
 config_h_target = configure_file(output: 'config.h', configuration: cdata)
 
+# Use --out-dir instead of --target-dir to fully control where librav1d.a goes.
+# Using --target-dir would put it in a subdirectory. The --out-dir option is
+# currently unstable so we need to tell cargo to enable those.
+# FIXME: this will only work if the host matches the build target.
+cargo_command = [
+    'cargo', 'build', '--lib',
+    '-Z', 'unstable-options',
+    '--out-dir', meson.current_build_dir()
```
kkysen, line 482: Do we need to use unstable options?  Can we just copy the build artifacts to a different directory?

kkysen (commented):
Could you clarify, does this preserve the ability to build `librav1d` linked against Rust `dav1d` and `seek_stress` binaries?  If not, I'd like to keep that, as that's type checked, unlike linking against C.  It might suffice just to compile this to get the type checking, as opposed to running tests with both the C and Rust CLI binaries.

Also, with this change, how do I build and test things and also run `cargo` independently so I get normal errors that aren't swallowed?  Is it intended that I run `cargo` independently first to see errors, then run `meson` which runs `cargo`, whose compilation is cached?

Also, could we use `cargo`'s `--profile` option instead of `--release`?

thedataking:
> Could you clarify, does this preserve the ability to build librav1d linked against Rust dav1d and seek_stress binaries?

Yes, `cargo build` works as usual this only affects how things are built when using meson. When you cargo build, it links against the Rust version of the dav1d CLI tools which are also fixed in this PR.

> Also, could we use cargo's --profile option instead of --release?

I think that would be confusing. The idea is to match build types such that if we configure meson for a debug build, it links against libravid.a with debug symbols.

tools/dav1d_cli_parse.rs
randomPoison, line 1: Are these CPU flag changes necessary as part of this, or was this you fixing another issue in the same PR? It's fine either way, I just don't see the obvious connection with the build system changes.

randomPoison (approved):
Looks good to me.

tools/dav1d_cli_parse.rs
kkysen: I think this is to fix #803.

thedataking: > I think this is to fix https://github.com/memorysafety/rav1d/issues/803.

Correct. 

> Are these CPU flag changes necessary as part of this, or was this you fixing another issue in the same PR?

No, this could have been split into two easier-to-review PRs. Sorry for the confusion.

PR #826, opened 2024-03-16
CI: Fix and clean up aarch64 test workflows

thedataking:
Fix breakage after removing `.cargo/config.toml` in #806  plus a little cleanup.

The QEMU workflows for `aarch64` and `armv7` are merged into a single new file to reduce duplicate logic. This also re-enables the QEMU tests for `aarch64` by default since they don't bottleneck testing runs.

Test runs for aarch64 on `macos-14` now include the variants with frame delay since they run fast.


.github/workflows/build-and-test-arm.yml
```diff
@@ -0,0 +1,64 @@
...
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ matrix.target }}-cargo-and-target-${{ hashFiles('**/Cargo.lock') }}
+      - name: cargo build for ${{ matrix.target }}
+        run: |
+          rustup target add ${{ matrix.target }}
+          RUSTFLAGS="-C target-feature=+crt-static -C linker=${{ matrix.linker }}" \
+            cargo build --release --target ${{ matrix.target }}
```
kkysen, line 52: I'm not sure if we were doing this before, but we should be testing with checked release builds.  Something like this:

```toml
# Fast but checked.
# For CI and sometimes local development with qemu.
#
# That is, it compiles with all runtime checks and limited debug info (as opposed to none).
[profile.checked-release]
inherits = "release"
opt-level = 3
debug = "limited"
debug-assertions = true
overflow-checks = true
lto = "thin"
codegen-units = 16
```

kkysen (approved):
Nice!  It's much nicer with the matrix.

PR #830, opened 2024-03-17
`mod intra_edge`: Backport memory reduction from dav1d 1.2.1

fbossen:
Eliminate duplicate and constant `intra_edge` table entries

### DIFF
--- a/src/intra_edge.rs
+++ b/src/intra_edge.rs
@@ -18,17 +18,22 @@ bitflags! {
 }
 
 impl EdgeFlags {
-    const LEFT_HAS_BOTTOM: Self = Self::union_all([
+    pub const LEFT_HAS_BOTTOM: Self = Self::union_all([
         Self::I444_LEFT_HAS_BOTTOM,
         Self::I422_LEFT_HAS_BOTTOM,
         Self::I420_LEFT_HAS_BOTTOM,
     ]);
 
-    const TOP_HAS_RIGHT: Self = Self::union_all([
+    pub const TOP_HAS_RIGHT: Self = Self::union_all([
         Self::I444_TOP_HAS_RIGHT,
         Self::I422_TOP_HAS_RIGHT,
         Self::I420_TOP_HAS_RIGHT,
     ]);
+
+    pub const EDGE_ALL_TR_AND_BL: Self =
+        Self::union_all([Self::LEFT_HAS_BOTTOM, Self::TOP_HAS_RIGHT]);
+
+    pub const EDGE_NONE: Self = Self::empty();
 
     pub const fn union_all<const N: usize>(flags: [Self; N]) -> Self {
         let mut i = 0;
@@ -99,19 +104,15 @@ pub struct EdgeNode {
 #[repr(C)]
 pub struct EdgeTip {
     pub node: EdgeNode,
-    pub split: [EdgeFlags; B],
+    pub split: [EdgeFlags; 3],
 }
 
 #[repr(C)]
 pub struct EdgeBranch {
     pub node: EdgeNode,
-    pub tts: [EdgeFlags; 3],
-    pub tbs: [EdgeFlags; 3],
-    pub tls: [EdgeFlags; 3],
-    pub trs: [EdgeFlags; 3],
-    pub h4: [EdgeFlags; 4],
-    pub v4: [EdgeFlags; 4],
-    pub split: [EdgeIndex; B],
+    pub h4: EdgeFlags,
+    pub v4: EdgeFlags,
+    pub split: [EdgeIndex; 4],
 }
 
 impl EdgeTip {
@@ -135,7 +136,6 @@ impl EdgeTip {
         let node = EdgeNode { o, h, v };
 
         let split = [
-            EdgeFlags::all(),
             edge_flags
                 .intersection(EdgeFlags::TOP_HAS_RIGHT)
                 .union(EdgeFlags::I422_LEFT_HAS_BOTTOM),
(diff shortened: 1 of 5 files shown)


fbossen:
Relates to #286

src/intra_edge.rs
```diff
@@ -18,18 +18,23 @@ bitflags! {
...
+    pub const TOP_HAS_RIGHT: Self = Self::union_all([
         Self::I444_TOP_HAS_RIGHT,
         Self::I422_TOP_HAS_RIGHT,
         Self::I420_TOP_HAS_RIGHT,
     ]);
 
+    pub const EDGE_ALL_TR_AND_BL: Self =
+        Self::union_all([Self::LEFT_HAS_BOTTOM, Self::TOP_HAS_RIGHT]);
+
+    pub const EDGE_NONE: Self = Self::empty();
```
folkertdev, line 36: I think we can drop the `EDGE_` prefix from these constants. They'll always occur qualified as `EdgeFlags::*` so the edge bit is just repeated.

src/intra_edge.rs
```diff
@@ -164,62 +164,27 @@ impl EdgeBranch {
...
-                    .select(matches!(bl, BlockLevel::Bl16x16)),
-            ),
-            EdgeFlags::TOP_HAS_RIGHT,
-            edge_flags.intersection(EdgeFlags::TOP_HAS_RIGHT),
-        ];
+        let h4 = EdgeFlags::LEFT_HAS_BOTTOM.union(
+            edge_flags
+                .intersection(EdgeFlags::I420_TOP_HAS_RIGHT)
+                .select(matches!(bl, BlockLevel::Bl16x16)),
+        );
```
folkertdev, line 171: the convention we went with here is that variables can use `union`, so `foo.union(...)`, but when the left-hand side is an enum variant we use `union_all`:

```suggestion
        let h4 = EdgeFlags::union_all([
            EdgeFlags::LEFT_HAS_BOTTOM,
            edge_flags
                .intersection(EdgeFlags::I420_TOP_HAS_RIGHT)
                .select(matches!(bl, BlockLevel::Bl16x16)),
        ]);
```

src/intra_edge.rs
```diff
@@ -164,62 +164,27 @@ impl EdgeBranch {
...
-            EdgeFlags::empty(),
-        ];
+        let v4 = EdgeFlags::TOP_HAS_RIGHT.union(
+            edge_flags
+                .intersection(EdgeFlags::union_all([
+                    EdgeFlags::I420_LEFT_HAS_BOTTOM,
+                    EdgeFlags::I422_LEFT_HAS_BOTTOM,
+                ]))
+                .select(matches!(bl, BlockLevel::Bl16x16)),
+        );
```
folkertdev, line 180: like above

```suggestion
        let v4 = EdgeFlags::union_all([
            EdgeFlags::TOP_HAS_RIGHT,
            edge_flags
                .intersection(EdgeFlags::union_all([
                    EdgeFlags::I420_LEFT_HAS_BOTTOM,
                    EdgeFlags::I422_LEFT_HAS_BOTTOM,
                ]))
                .select(matches!(bl, BlockLevel::Bl16x16)),
        ]);
```

folkertdev (commented):
I think this looks good except for some minor style/consistency nitpicks

fbossen:
> I think this looks good except for some minor style/consistency nitpicks

Thanks for the feedback. I made the suggested changes.

folkertdev (approved):

src/intra_edge.rs
```diff
@@ -164,62 +163,29 @@ impl EdgeBranch {
...
-            edge_flags.union(EdgeFlags::TOP_HAS_RIGHT),
-            EdgeFlags::empty(),
-        ];
+            edge_flags
+                .intersection(EdgeFlags::union_all([
+                    EdgeFlags::I420_LEFT_HAS_BOTTOM,
+                    EdgeFlags::I422_LEFT_HAS_BOTTOM,
+                ]))
+                .select(matches!(bl, BlockLevel::Bl16x16)),
+        ]);
```
kkysen, line 181:
```suggestion
        let v4 = edge_flags
            .intersection(EdgeFlags::union_all([
                EdgeFlags::I420_LEFT_HAS_BOTTOM,
                EdgeFlags::I422_LEFT_HAS_BOTTOM,
            ]))
            .select(matches!(bl, BlockLevel::Bl16x16))
            .union(EdgeFlags::TOP_HAS_RIGHT),
```

src/intra_edge.rs
```diff
@@ -18,18 +18,22 @@ bitflags! {
 }
 
 impl EdgeFlags {
-    const LEFT_HAS_BOTTOM: Self = Self::union_all([
+    pub const LEFT_HAS_BOTTOM: Self = Self::union_all([
```
kkysen, line 21:
```suggestion
    pub const ALL_LEFT_HAS_BOTTOM: Self = Self::union_all([
```

I should've done this when I added this, but could add this in a separate commit on this PR?

src/intra_edge.rs
```diff
@@ -18,18 +18,22 @@ bitflags! {
...
         Self::I444_LEFT_HAS_BOTTOM,
         Self::I422_LEFT_HAS_BOTTOM,
         Self::I420_LEFT_HAS_BOTTOM,
     ]);
 
-    const TOP_HAS_RIGHT: Self = Self::union_all([
+    pub const TOP_HAS_RIGHT: Self = Self::union_all([
```
kkysen, line 27:
```suggestion
    pub const ALL_TOP_HAS_RIGHT: Self = Self::union_all([
```

Same reasoning and commit as above.

kkysen (changes requested):
Mostly looks very good, but I had a few more suggestions.  Also, by using 2-byte `EdgeIndex`s instead of 8-byte pointers, our table should be even smaller than C's 🎉 .

kkysen (approved):

PR #832, opened 2024-03-18
`enum InterIntraType`: make a real enum

folkertdev:


src/decode.rs
```diff
@@ -2970,27 +2971,30 @@ unsafe fn decode_b_inner(
...
-                    ) as u8;
-                if b.interintra_type() == INTER_INTRA_WEDGE {
+                *b.interintra_type_mut() = if rav1d_msac_decode_bool_adapt(
+                    &mut ts.msac,
+                    &mut ts.cdf.m.interintra_wedge[wedge_ctx as usize],
+                ) {
+                    InterIntraType::Wedge
+                } else {
+                    InterIntraType::Blend
+                };
```
kkysen, line 2981: Did you check if the `if` generates equally efficient code as the `+`?  `fn decode_b` is in the inner decoding loop, so it's likely quite performance sensitive.

kkysen: It does: https://godbolt.org/z/EoWGo61nv.

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum InterIntraType {
    None = 0,
    Blend = 1,
    Wedge = 2,
}

extern "C" {
    fn rav1d_msac_decode_bool_adapt() -> bool;
}

pub fn using_if() -> InterIntraType {
    if unsafe { rav1d_msac_decode_bool_adapt() } {
        InterIntraType::Wedge
    } else {
        InterIntraType::Blend
    }
}

pub fn using_add() -> u8 {
    InterIntraType::Blend as u8 + unsafe { rav1d_msac_decode_bool_adapt() } as u8
}
```

They both use `inc` (x86) and `cinc` (arm).

src/levels.rs
```diff
@@ -209,10 +209,12 @@ pub const COMP_INTER_AVG: CompInterType = 2;
...
-pub type InterIntraType = u8;
-pub const INTER_INTRA_WEDGE: InterIntraType = 2;
-pub const INTER_INTRA_BLEND: InterIntraType = 1;
-pub const INTER_INTRA_NONE: InterIntraType = 0;
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum InterIntraType {
+    None = 0,
+    Blend = 1,
+    Wedge = 2,
+}
```
kkysen, line 217: What if we removed `InterIntraType::None` and used `Option::None` instead?  It'd make a bunch of the `match`es exhaustive since the initial `None` check can be done using `if let Some`.

kkysen (approved):
Approving since the changes all LGTM, but you could try implementing the `Option` version I suggested, too.

src/levels.rs
folkertdev: yes that's nicer

src/decode.rs
folkertdev: the version that uses an `Option` is even better: the bool is just transmuted into a `Option<InterIntraType>` behind the scenes.

src/decode.rs
```diff
@@ -2970,27 +2968,30 @@ unsafe fn decode_b_inner(
...
-                    ) as u8;
-                if b.interintra_type() == INTER_INTRA_WEDGE {
+                *b.interintra_type_mut() = if rav1d_msac_decode_bool_adapt(
+                    &mut ts.msac,
+                    &mut ts.cdf.m.interintra_wedge[wedge_ctx as usize],
+                ) {
+                    Some(InterIntraType::Wedge)
+                } else {
+                    Some(InterIntraType::Blend)
+                };
```
kkysen, line 2978: nit:

```suggestion
                *b.interintra_type_mut() = Some(if rav1d_msac_decode_bool_adapt(
                    &mut ts.msac,
                    &mut ts.cdf.m.interintra_wedge[wedge_ctx as usize],
                ) {
                    InterIntraType::Wedge
                } else {
                    InterIntraType::Blend
                });
```

kkysen (approved):

src/decode.rs
folkertdev: yeah? I generally don't like many lines of logic in data constructors, but feel free to commit that if you prefer

kkysen: How about this? 9626bb7

folkertdev: Looks good!

kkysen (approved):

PR #838, opened 2024-03-18
Backport memory optimization for palette mode from dav1d 1.3.0

fbossen:
Allocate memory for palette mode according to bit depth of pixel instead of fixed `u16` size.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -79,6 +79,8 @@ use crate::src::refmvs::refmvs_temporal_block;
 use crate::src::refmvs::refmvs_tile;
 use crate::src::refmvs::Rav1dRefmvsDSPContext;
 use crate::src::refmvs::RefMvsFrame;
+use crate::src::unstable_extensions::as_chunks;
+use crate::src::unstable_extensions::as_chunks_mut;
 use atomig::Atomic;
 use libc::ptrdiff_t;
 use std::cell::UnsafeCell;
@@ -442,6 +444,36 @@ impl CodedBlockInfo {
 
 #[derive(Default)]
 #[repr(C)]
+pub struct AlignedVec64Pal {
+    data: AlignedVec64<u8>,
+}
+
+impl AlignedVec64Pal {
+    pub fn resize(&mut self, n: usize) {
+        self.data.resize(n * 8 * 3, Default::default());
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.data.is_empty()
+    }
+
+    pub fn as_slice<BD: BitDepth>(&self) -> &[[[BD::Pixel; 8]; 3]] {
+        as_chunks::<3, [BD::Pixel; 8]>(
+            as_chunks::<8, BD::Pixel>(BD::cast_pixel_slice(&self.data)).0,
+        )
+        .0
+    }
+
+    pub fn as_slice_mut<BD: BitDepth>(&mut self) -> &mut [[[BD::Pixel; 8]; 3]] {
+        as_chunks_mut::<3, [BD::Pixel; 8]>(
+            as_chunks_mut::<8, BD::Pixel>(BD::cast_pixel_slice_mut(&mut self.data)).0,
+        )
+        .0
+    }
+}
+
+#[derive(Default)]
+#[repr(C)]
 pub struct Rav1dFrameContext_frame_thread {
     /// Indices: 0: reconstruction, 1: entropy.
     pub next_tile_row: [c_int; 2],
@@ -453,9 +485,9 @@ pub struct Rav1dFrameContext_frame_thread {
 
     /// Indexed using `(t.b.y >> 1) * (f.b4_stride >> 1) + (t.b.x >> 1)`.
     /// Inner indices are `[3 plane][8 idx]`.
-    /// Allocated as a flat array. `pal_as_slice` and `pal_as_slice_mut` should be
+    /// Allocated as a flat array. `pal.as_slice` and `pal.as_slice_mut` should be
     /// used to access elements of type `[[BitDepth::Pixel; 8]; 3]`
-    pub pal: AlignedVec64<u8>,
+    pub pal: AlignedVec64Pal,
 
     /// Iterated over inside tile state.
     pub pal_idx: AlignedVec64<u8>,
@@ -465,23 +497,6 @@ pub struct Rav1dFrameContext_frame_thread {
 
     /// Start offsets per tile
     pub tile_start_off: Vec<u32>,
-}
-
-use crate::src::unstable_extensions::as_chunks;
-use crate::src::unstable_extensions::as_chunks_mut;
-
-impl Rav1dFrameContext_frame_thread {
-    pub fn pal_as_slice<BD: BitDepth>(&self) -> &[[[BD::Pixel; 8]; 3]] {
-        as_chunks::<3, [BD::Pixel; 8]>(as_chunks::<8, BD::Pixel>(BD::cast_pixel_slice(&self.pal)).0)
-            .0
-    }
-
-    pub fn pal_as_slice_mut<BD: BitDepth>(&mut self) -> &mut [[[BD::Pixel; 8]; 3]] {
-        as_chunks_mut::<3, [BD::Pixel; 8]>(
-            as_chunks_mut::<8, BD::Pixel>(BD::cast_pixel_slice_mut(&mut self.pal)).0,
-        )
-        .0
-    }
 }
 
 #[derive(Default)]
@@ -814,6 +829,7 @@ pub struct Cf;
 impl BitDepthDependentType for Cf {
     type T<BD: BitDepth> = Align64<[BD::Coef; 32 * 32]>;
 }
+
 pub struct AlPal;
 
 impl BitDepthDependentType for AlPal {
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -444,11 +444,11 @@ impl CodedBlockInfo {
 
 #[derive(Default)]
 #[repr(C)]
-pub struct AlignedVec64Pal {
+pub struct Pal {
     data: AlignedVec64<u8>,
 }
 
-impl AlignedVec64Pal {
+impl Pal {
     pub fn resize(&mut self, n: usize) {
         self.data.resize(n * 8 * 3, Default::default());
     }
@@ -487,7 +487,7 @@ pub struct Rav1dFrameContext_frame_thread {
     /// Inner indices are `[3 plane][8 idx]`.
     /// Allocated as a flat array. `pal.as_slice` and `pal.as_slice_mut` should be
     /// used to access elements of type `[[BitDepth::Pixel; 8]; 3]`
-    pub pal: AlignedVec64Pal,
+    pub pal: Pal,
 
     /// Iterated over inside tile state.
     pub pal_idx: AlignedVec64<u8>,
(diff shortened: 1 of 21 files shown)


src/internal.rs
```diff
@@ -424,6 +441,23 @@ pub struct Rav1dFrameContext_frame_thread {
     pub tile_start_off: Vec<u32>,
 }
 
+use crate::src::unstable_extensions::as_chunks;
+use crate::src::unstable_extensions::as_chunks_mut;
```
kkysen, line 445: Can you put these at the top of the file?

src/internal.rs
```diff
@@ -656,6 +690,11 @@ pub struct Cf;
 impl BitDepthDependentType for Cf {
     type T<BD: BitDepth> = Align64<[BD::Coef; 32 * 32]>;
 }
+pub struct AlPal;
```
kkysen, line 693:
```suggestion

pub struct AlPal;
```

src/recon.rs
```diff
@@ -145,6 +146,28 @@ pub(crate) type backup_ipred_edge_fn = unsafe fn(&Rav1dFrameData, &mut Rav1dTask
...
+
+pub(crate) type read_pal_plane_fn = unsafe fn(
+    &mut Rav1dTaskContext,
+    &mut Rav1dFrameData,
+    &mut Av1Block,
+    bool,
+    u8,
+    usize,
+    usize,
+) -> ();
```
kkysen, line 160: Could you add the arg names for these, too?

src/recon.rs
```diff
@@ -2564,13 +2587,15 @@ pub(crate) unsafe fn rav1d_recon_b_intra<BD: BitDepth>(
...
+                let pal: *const BD::Pixel = if t.frame_thread.pass != 0 {
                     let index = (((t.by as isize >> 1) + (t.bx as isize & 1)) * (f.b4_stride >> 1)
                         + ((t.bx >> 1) + (t.by & 1)) as isize)
                         as isize;
-                    f.frame_thread.pal[index as usize][0].as_ptr()
+                    f.frame_thread.pal_as_slice::<BD>()[index as usize][0].as_ptr()
                 } else {
-                    (t.scratch.c2rust_unnamed_0.pal[0]).as_ptr()
+                    let interintra_edge_pal =
+                        BD::select(&mut t.scratch.c2rust_unnamed_0.interintra_edge_pal);
```
kkysen, line 2597:
```suggestion
                        BD::select(&t.scratch.c2rust_unnamed_0.interintra_edge_pal);
```

src/internal.rs
```diff
@@ -412,7 +427,9 @@ pub struct Rav1dFrameContext_frame_thread {
 
     /// Indexed using `(t.by >> 1) * (f.b4_stride >> 1) + (t.bx >> 1)`.
     /// Inner indices are `[3 plane][8 idx]`.
-    pub pal: AlignedVec64<[[u16; 8]; 3]>,
+    /// Allocated as a flat array. `pal_as_slice` and `pal_as_slice_mut` should be
+    /// used to access elements of type `[[BitDepth::Pixel; 8]; 3]`
+    pub pal: AlignedVec64<u8>,
```
kkysen, line 432: This might run into a borrowck issue later, as `.pal_as_slice()` borrows the whole `&Rav1dFrameContext_frame_thread` instead `.pal` just borrowing the field.  We might want to newtype `pal` and have a `.as_slice()` method on it instead.  That way we can do `.pal.as_slice()`, which will allow borrowck to see we're only borrowing the one field.

src/recon.rs
```diff
@@ -4828,3 +4858,301 @@ pub(crate) unsafe fn rav1d_backup_ipred_edge<BD: BitDepth>(
...
+    bw4: c_int,
+    bh4: c_int,
+) {
+    let pal = if t.frame_thread.pass != 0 {
+        let index = ((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
+            + ((t.bx >> 1) + (t.by & 1)) as isize;
+        &f.frame_thread.pal_as_slice::<BD>()[index as usize][0]
+    } else {
+        let interintra_edge_pal =
+            BD::select_mut(&mut t.scratch.c2rust_unnamed_0.interintra_edge_pal);
```
kkysen, line 4876:
```suggestion
            BD::select(&t.scratch.c2rust_unnamed_0.interintra_edge_pal);
```

src/recon.rs
```diff
@@ -4828,3 +4858,301 @@ pub(crate) unsafe fn rav1d_backup_ipred_edge<BD: BitDepth>(
...
+        &interintra_edge_pal.0.pal[0]
+    };
+    for al_pal in &mut BD::select_mut(&mut t.al_pal)[0][bx4 as usize..][..bw4 as usize] {
```
kkysen, line 4879:
```suggestion
    for al_pal in BD::select_mut(&mut t.al_pal)[0][bx4 as usize..][..bw4 as usize] {
```

I think the `&mut` is redundant here, since `BD::select_mut` already returns a `&mut`.

src/recon.rs
```diff
@@ -4828,3 +4858,301 @@ pub(crate) unsafe fn rav1d_backup_ipred_edge<BD: BitDepth>(
...
+        al_pal[0] = *pal;
+    }
+    for al_pal in &mut BD::select_mut(&mut t.al_pal)[1][by4 as usize..][..bh4 as usize] {
```
kkysen, line 4882:
```suggestion
    for al_pal in BD::select_mut(&mut t.al_pal)[1][by4 as usize..][..bh4 as usize] {
```

src/recon.rs
```diff
@@ -4828,3 +4858,301 @@ pub(crate) unsafe fn rav1d_backup_ipred_edge<BD: BitDepth>(
...
+    bw4: c_int,
+    bh4: c_int,
+) {
+    let pal = if t.frame_thread.pass != 0 {
+        let index = ((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
+            + ((t.bx >> 1) + (t.by & 1)) as isize;
+        &f.frame_thread.pal_as_slice_mut::<BD>()[index as usize]
+    } else {
+        let interintra_edge_pal =
+            BD::select_mut(&mut t.scratch.c2rust_unnamed_0.interintra_edge_pal);
```
kkysen, line 4901:
```suggestion
            BD::select(&t.scratch.c2rust_unnamed_0.interintra_edge_pal);
```

src/recon.rs
```diff
@@ -4828,3 +4858,301 @@ pub(crate) unsafe fn rav1d_backup_ipred_edge<BD: BitDepth>(
...
+    f: &mut Rav1dFrameData,
+    bx4: c_int,
+    by4: c_int,
+    bw4: c_int,
+    bh4: c_int,
+) {
+    let pal = if t.frame_thread.pass != 0 {
+        let index = ((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
+            + ((t.bx >> 1) + (t.by & 1)) as isize;
+        &f.frame_thread.pal_as_slice_mut::<BD>()[index as usize]
```
kkysen, line 4898:
```suggestion
        &f.frame_thread.pal_as_slice::<BD>()[index as usize]
```

kkysen (changes requested):
Left some comments, still reviewing though.

src/recon.rs
fbossen: Removing `&mut` doesn't seem to work here.

src/recon_tmpl.c
```diff
@@ -2200,3 +2201,178 @@ void bytefn(dav1d_backup_ipred_edge)(Dav1dTaskContext *const t) {
...
+    bytefn(dav1d_read_pal_plane)(t, b, 1, sz_ctx, bx4, by4);
+
+    // V pal coding
+    Dav1dTileState *const ts = t->ts;
+    const Dav1dFrameContext *const f = t->f;
+    pixel *const pal = t->frame_thread.pass ?
+        f->frame_thread.pal[((t->by >> 1) + (t->bx & 1)) * (f->b4_stride >> 1) +
+                            ((t->bx >> 1) + (t->by & 1))][2] :
+        bytefn(t->scratch.pal)[2];
+    const int bpc = BITDEPTH == 8 ? 8 : f->cur.p.bpc;
```
randomPoison, line 2358: I notice that in a couple of places the C code is doing this bit of logic to get a `bpc` value, but in the Rust translation you just use `f.cur.p.bpc` directly. Is there a reason why this extra logic was needed in C but not in Rust?

randomPoison (commented):
The translations of the C changes to Rust look correct to me. I didn't see an obvious source for the CI test failures though. I'll give this a final approval once CI has been fixed.

src/recon_tmpl.c
kkysen: I think it's a very minor optimization that allows bpc8 to use a constant.

fbossen: Yes, it's an optimization for when `BITDEPTH == 8`. `bpc` is then a constant, meaning you don't need a register to hold its value and don't need to load `f->cur.p.bpc`.

randomPoison (approved):

fbossen:
> Left some comments, still reviewing though.

@kkysen Have you had a chance to look at it some more?

kkysen (changes requested):
I'm still looking into the generic alignment issue.  Also, I'm assuming the large functions were just moved.  The rest looks good except for a couple comments I made.  Sorry I forgot to review this again; I know there've been a bunch of merge conflicts.

src/internal.rs
```diff
@@ -836,15 +891,17 @@ pub union Rav1dTaskContext_scratch_levels_pal {
...
+    pub pal: [[BD::Pixel; 8]; 3], /* [3 plane][8 palette_idx] */
 }
 
-pub struct InterIntraEdge;
+pub struct InterIntraEdgePal;
 
-impl BitDepthDependentType for InterIntraEdge {
-    type T<BD: BitDepth> = Align64<InterIntraEdgeBD<BD>>;
+impl BitDepthDependentType for InterIntraEdgePal {
+    type T<BD: BitDepth> = Align64<InterIntraEdgePalBD<BD>>;
```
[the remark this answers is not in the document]
kkysen: If we can't use something like `elain` to do the `BD::Pixel`-dependent alignment right, this should be `Align16`.

src/internal.rs
```diff
@@ -836,15 +891,17 @@ pub union Rav1dTaskContext_scratch_levels_pal {
 
 #[derive(Clone, Copy)]
 #[repr(C)]
-pub struct InterIntraEdgeBD<BD: BitDepth> {
+pub struct InterIntraEdgePalBD<BD: BitDepth> {
     pub interintra: [BD::Pixel; 64 * 64],
     pub edge: [BD::Pixel; 257],
+    _align: [BD::Pixel; 7],       // Align pal to 8 times BD::Pixel
+    pub pal: [[BD::Pixel; 8]; 3], /* [3 plane][8 palette_idx] */
```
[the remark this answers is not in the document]
kkysen: Using `elain`:

```suggestion
    pub pal: [Align<{ mem::align_of::<BD::Pixel>() * 8 }, [BD::Pixel; 8]>; 3], /* [3 plane][8 palette_idx] */
```

where we redefine our `Align*` types as `Align<const N: usize, T>`, which store an `elain::Align<N>`.

kkysen (approved):
LGTM.  I'll follow-up on how we can improve the alignment stuff, but this way is definitely good enough for now.

PR #867, opened 2024-03-20
`struct Rav1dFrameData`: Add `index` field

rinon:
Store the index of a frame context in the array of frame contexts, rather than doing pointer arithmetic to determine the index.


rinon:
This is a pretty trivial change to remove some pointer arithmetic

kkysen (approved):

PR #868, opened 2024-03-20
`struct CdfThreadContext`: convert to `Arc`

### LINKED ISSUE #824: `CdfThreadContext{_data}`: Make fields safe
`CdfThreadContext`:

- [ ] `ref` - A `Rav1dRef`. Should be replaced with an `Arc`.
  - Also covered by https://github.com/memorysafety/rav1d/issues/641
  - The `data.cdf` and `progress` fields point into this data.
- [ ] `data` - A union containing a `*mut CdfContext`. The union only has two fields, so it may make sense to unpack it and inline the fields directly in `CdfThreadContext`.
- [ ] `progress` - Points to just past the end of the `CdfContext` object stored in `ref` (i.e. the object pointed to by `data.cdf`).

It may make the most sense to convert `ref` into something like `Arc<(CdfContext, AtomicU32)>` so that both the `CdfContext` and progress counter can be accessed directly.

`CdfThreadContext_data`:

* `cdf` - Points to the data owned by the `ref` field in the parent `CdfThreadContext` struct.

Either inline contents directly into `CdfThreadContext` or make into an enum.

rinon:
Convert `CdfThreadContext` from a `Rav1dRef` of `CdfContext` + `progress` atomic to an `Arc`. This is complicated by the fact that this structure has several states, which are now represented by an `enum` and optional `progress` field.

* Closes #824

### DIFF
--- a/src/cdf.rs
+++ b/src/cdf.rs
@@ -5,7 +5,6 @@ use crate::src::align::Align16;
 use crate::src::align::Align32;
 use crate::src::align::Align4;
 use crate::src::align::Align8;
-use crate::src::error::Rav1dError::ENOMEM;
 use crate::src::error::Rav1dResult;
 use crate::src::internal::Rav1dContext;
 use crate::src::levels::BlockLevel;
@@ -16,20 +15,18 @@ use crate::src::levels::N_COMP_INTER_PRED_MODES;
 use crate::src::levels::N_INTRA_PRED_MODES;
 use crate::src::levels::N_TX_SIZES;
 use crate::src::levels::N_UV_INTRA_PRED_MODES;
-use crate::src::r#ref::rav1d_ref_create_using_pool;
-use crate::src::r#ref::rav1d_ref_dec;
-use crate::src::r#ref::rav1d_ref_inc;
-use crate::src::r#ref::Rav1dRef;
 use crate::src::tables::dav1d_partition_type_count;
-use libc::memcpy;
 use std::cmp;
 use std::ffi::c_int;
 use std::ffi::c_uint;
-use std::ffi::c_void;
-use std::ptr;
+use std::mem;
 use std::sync::atomic::AtomicU32;
+use std::sync::Arc;
+use std::sync::RwLock;
+use std::sync::RwLockWriteGuard;
 use strum::EnumCount;
 
+#[derive(Clone)]
 #[repr(C)]
 pub struct CdfContext {
     pub m: CdfModeContext,
@@ -39,6 +36,7 @@ pub struct CdfContext {
     pub dmv: CdfMvContext,
 }
 
+#[derive(Clone)]
 #[repr(C)]
 pub struct CdfMvContext {
     pub comp: [CdfMvComponent; 2],
@@ -133,18 +131,39 @@ pub struct CdfModeContext {
 }
 
 #[derive(Clone)]
-#[repr(C)]
-pub struct CdfThreadContext {
-    pub r#ref: *mut Rav1dRef,
-    pub data: CdfThreadContext_data,
-    pub progress: *mut AtomicU32,
+pub enum CdfThreadContext {
+    QCat(c_uint),
+    Cdf(Arc<CdfThreadContext_data>),
 }
 
-#[derive(Clone, Copy)]
+impl CdfThreadContext {
+    pub fn progress(&self) -> Option<&AtomicU32> {
+        match self {
+            Self::QCat(_) => None,
+            Self::Cdf(cdf) => cdf.progress.as_ref(),
+        }
+    }
+
+    pub fn cdf_write(&self) -> RwLockWriteGuard<CdfContext> {
+        match self {
+            Self::QCat(_) => panic!("Expected a Cdf"),
+            Self::Cdf(cdf) => cdf.cdf.try_write().ok().unwrap(),
+        }
+    }
+}
+
+impl Default for CdfThreadContext {
+    fn default() -> Self {
+        Self::QCat(0)
+    }
+}
+
 #[repr(C)]
-pub union CdfThreadContext_data {
-    pub cdf: *mut CdfContext,
-    pub qcat: c_uint,
+pub struct CdfThreadContext_data {
+    /// This context should not be contended but needs interior mutability. It
+    /// should always be accessible with `.try_*` methods.
+    cdf: RwLock<CdfContext>,
+    pub progress: Option<AtomicU32>,
 }
 
 const fn cdf0d<const P: usize, const N: usize>(probs: [u16; P]) -> [u16; N] {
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -947,7 +946,6 @@ impl Drop for Rav1dContext {
             let _ = mem::take(&mut self.tiles);
             let mut n_4 = 0;
             while n_4 < 8 {
-                rav1d_cdf_thread_unref(&mut *(self.cdf).as_mut_ptr().offset(n_4 as isize));
```
randomPoison, line 950: Do we not need to set the `Option`s in `self.cdf` to `None` here the way we do for the other calls to `rav1d_cdf_thread_unref`?

src/cdf.rs
```diff
@@ -133,18 +130,31 @@ pub struct CdfModeContext {
...
+    }
 }
 
-#[derive(Clone, Copy)]
 #[repr(C)]
-pub union CdfThreadContext_data {
-    pub cdf: *mut CdfContext,
-    pub qcat: c_uint,
+pub struct CdfThreadContext_data {
+    pub cdf: RwLock<CdfContext>,
```
randomPoison, line 156: Is the `RwLock` here just for interior mutability? Do we expect any contention on it? From the usage of `unwrap` when locking it I assume not?

randomPoison (approved):
Looks good. I had a couple of questions but no specific changes to request.

src/cdf.rs
```diff
@@ -5075,66 +5085,40 @@ unsafe fn get_qcat_idx(q: c_int) -> c_int {
...
-        progress: ptr::null_mut(),
+) -> Rav1dResult<CdfThreadContext> {
+    let progress = if have_frame_mt != 0 {
+        Some(AtomicU32::new(0))
+    } else {
+        None
     };
-    rav1d_ref_dec(&mut cdf.r#ref);
+    Ok(CdfThreadContext::Cdf(Arc::new(CdfThreadContext_data {
+        cdf: mem::zeroed(),
```
kkysen, line 5121: Why do we need to introduce a new `mem::zeroed()`, or is this split off an existing, larger `mem::zeroed()`?

rinon: Arc needs to not be zero-initialized, so this is split from the parent's zeroed init.

src/cdf.rs
rinon: Yes, it is just for interior mutability, we only use `try_` variants of locking.

src/lib.rs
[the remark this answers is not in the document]
rinon: The array is owned, drop will be called on its children which will unref the Arc, no?

rinon: Addressing in other thread

src/cdf.rs
```diff
@@ -5075,66 +5085,40 @@ unsafe fn get_qcat_idx(q: c_int) -> c_int {
...
-        (*dst).kfym = default_kf_y_mode_cdf;
-        (*dst).coef = av1_default_coef_cdf[(*src).data.qcat as usize].clone();
-        (*dst).mv.joint = default_mv_joint_cdf;
-        (*dst).dmv.joint = default_mv_joint_cdf;
-        (*dst).dmv.comp[1] = default_mv_component_cdf.clone();
-        (*dst).dmv.comp[0] = (*dst).dmv.comp[1].clone();
-        (*dst).mv.comp[1] = (*dst).dmv.comp[0].clone();
-        (*dst).mv.comp[0] = (*dst).mv.comp[1].clone();
-    };
+pub unsafe fn rav1d_cdf_thread_copy(dst: &mut CdfContext, src: &CdfThreadContext) {
```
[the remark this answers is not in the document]
rinon: Could it, probably, but does it need to be? It could also be a method, etc. There's plenty of things we could do here but I'd like to just move on right now.

randomPoison (approved):

src/thread_task.rs
```diff
@@ -1130,12 +1130,12 @@ pub unsafe fn rav1d_worker_task(c: &Rav1dContext, task_thread: Arc<Rav1dTaskCont
...
                                     rav1d_cdf_thread_update(
                                         frame_hdr,
-                                        &mut *f.out_cdf.data.cdf,
+                                        &mut f.out_cdf.cdf().unwrap().try_write().unwrap(),
                                         &(*(f.ts).offset(frame_hdr.tiling.update as isize)).cdf,
                                     );
                                 }
                                 if c.n_fc > 1 as c_uint {
-                                    (*f.out_cdf.progress).store(
+                                    f.out_cdf.progress().unwrap().store(
```
kkysen, line 1138: This could be an `if let Some` for the progress.

src/decode.rs
```diff
@@ -4619,7 +4618,10 @@ pub(crate) unsafe fn rav1d_decode_frame_init_cdf(
     let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
 
     if frame_hdr.refresh_context != 0 {
-        rav1d_cdf_thread_copy(f.out_cdf.data.cdf, &f.in_cdf);
+        rav1d_cdf_thread_copy(
+            &mut f.out_cdf.cdf().unwrap().try_write().unwrap(),
```
kkysen, line 4622: `cdf().unwrap()` and `refresh_context != 0` are checking the same thing, so this can be an `if let Some`.

src/decode.rs
```diff
@@ -4845,9 +4847,10 @@ pub(crate) unsafe fn rav1d_decode_frame(c: &Rav1dContext, f: &mut Rav1dFrameData
             res = rav1d_decode_frame_main(c, f);
             let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
             if res.is_ok() && frame_hdr.refresh_context != 0 && f.task_thread.update_set {
+                let out_cdf = f.out_cdf.cdf().unwrap();
```
kkysen, line 4850: The `refresh_context != 0` and `.unwrap()` can be combined with an `if let Some`.

src/thread_task.rs
```diff
@@ -1130,12 +1130,12 @@ pub unsafe fn rav1d_worker_task(c: &Rav1dContext, task_thread: Arc<Rav1dTaskCont
                                 if error_0 == 0 {
                                     rav1d_cdf_thread_update(
                                         frame_hdr,
-                                        &mut *f.out_cdf.data.cdf,
+                                        &mut f.out_cdf.cdf().unwrap().try_write().unwrap(),
```
kkysen, line 1133: The `cdf().unwrap()` is redundant with the `refresh_context != 0` above.

src/cdf.rs
```diff
@@ -133,18 +130,37 @@ pub struct CdfModeContext {
...
+        match self {
+            Self::QCat(_) => None,
+            Self::Cdf(cdf) => Some(&cdf.cdf),
+        }
+    }
+}
+
+impl Default for CdfThreadContext {
+    fn default() -> Self {
+        Self::QCat(0)
```
kkysen, line 156: nit: There's nothing special about 0 here, we just want some default value, so I think `Default::default()` expresses that intent slightly better.  Not a big deal, though.

```suggestion
        Self::QCat(Default::default())
```

kkysen (changes requested):
I'd very much like a doc comment or methods (preferred) on the `cdf: RwLock<CdfContext>` field indicating that only `.try_*()` methods should be used on it.

Also, most of the `.unwrap()`s added here can be folded into their checks a few lines above.  That might be less important since `thread_task.rs` should hopefully not be too performance sensitive, but I think it might be clearer as well to do the checks only once .

src/cdf.rs
rinon: There is something special about zero, the original C code zero-initialized this field. I'd prefer to make that explicit here.

src/thread_task.rs
```diff
@@ -1011,7 +1011,7 @@ pub unsafe fn rav1d_worker_task(c: &Rav1dContext, task_thread: Arc<Rav1dTaskCont
                         }
                         let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
                         if frame_hdr.refresh_context != 0 && !f.task_thread.update_set {
-                            (*f.out_cdf.progress).store(
+                            f.out_cdf.progress().unwrap().store(
```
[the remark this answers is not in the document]
rinon: Why? It isn't clear those are the same thing.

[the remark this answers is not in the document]
rinon: I'd prefer not to remove the assert, as I think it is clearer.

src/decode.rs
rinon: Yes, but I have changed `cdf()` to `cdf_write()` which tries to lock the field and requires an unwrap anyway.

src/decode.rs
rinon: `cdf_write()` now tries to lock, so we need the unwrap

src/thread_task.rs
rinon: `cdf_write()` now tries to lock, so we need the unwrap.

src/cdf.rs
kkysen: Oh, there is?  Then your way is better.

src/cdf.rs
```diff
@@ -133,18 +131,39 @@ pub struct CdfModeContext {
...
+    }
 }
 
-#[derive(Clone, Copy)]
 #[repr(C)]
-pub union CdfThreadContext_data {
-    pub cdf: *mut CdfContext,
-    pub qcat: c_uint,
+pub struct CdfThreadContext_data {
+    /// This context should not be contentended but needs interior mutability.
```
kkysen, line 163:
```suggestion
    /// This context should not be contended but needs interior mutability.
```

src/cdf.rs
```diff
@@ -133,18 +131,39 @@ pub struct CdfModeContext {
...
+            Self::Cdf(cdf) => cdf.progress.as_ref(),
+        }
+    }
+
+    pub fn cdf_write(&self) -> Option<RwLockWriteGuard<CdfContext>> {
+        match self {
+            Self::QCat(_) => None,
+            Self::Cdf(cdf) => cdf.cdf.try_write().ok(),
+        }
+    }
```
kkysen, line 152: If the result is always `.unwrap()`ed, why not put the `.unwrap()` here?  And what about `cdf_read`?

kkysen (changes requested):
For the `.unwrap()`s that I thought could be folded into existing checks, I still think many of them can, but I'm fine leaving them as is unless performance/code size matters here and we can be more careful about combining the checks.

src/cdf.rs
rinon: cdf read wasn't needed because we only access the field for reading in the same module and we already needed to destructure the enum to do it (`rav1d_cdf_thread_copy`)

kkysen (approved):

PR #869, opened 2024-03-20
`enum CompInterType`: make a real enum

folkertdev:
this PR looks big but most lines are because I needed to reverse two branches so that 

```rust
if comp_type.is_none() { ... } else { ... }
```

can turn into

```rust
if let Some(comp_type) = comp_type { ... }  else { ... }
```

the first commit is most changes, the second one just the switching of branches and using the variable bound by the if-let

### DIFF
--- a/src/levels.rs
+++ b/src/levels.rs
@@ -204,10 +204,10 @@ pub const NEARESTMV_NEARESTMV: CompInterPredMode = 0;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub enum CompInterType {
-    COMP_INTER_WEIGHTED_AVG = 1,
-    COMP_INTER_AVG = 2,
-    COMP_INTER_SEG = 3,
-    COMP_INTER_WEDGE = 4,
+    WeightedAvg = 1,
+    Avg = 2,
+    Seg = 3,
+    Wedge = 4,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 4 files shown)


src/levels.rs
```diff
@@ -202,12 +202,13 @@ pub const NEARESTMV_NEWMV: CompInterPredMode = 2;
...
-pub const COMP_INTER_AVG: CompInterType = 2;
-pub const COMP_INTER_WEIGHTED_AVG: CompInterType = 1;
-pub const COMP_INTER_NONE: CompInterType = 0;
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum CompInterType {
+    COMP_INTER_WEIGHTED_AVG = 1,
+    COMP_INTER_AVG = 2,
+    COMP_INTER_SEG = 3,
+    COMP_INTER_WEDGE = 4,
+}
```
kkysen, line 211:
```suggestion
pub enum CompInterType {
    WeightedAvg = 1,
    Avg = 2,
    Seg = 3,
    Wedge = 4,
}
```

kkysen (changes requested):
Still reviewing the second commit.

kkysen (approved):
The 2nd commit is too annoying to review.  It's just the `if let Some` and swapping the branches' order, right?  If that's it, it LGTM.

folkertdev:
yes, that and it removes these lines

```rust
// TODO use if let in the surrounding if structure
let comp_inter_type = b.c2rust_unnamed.c2rust_unnamed_0.comp_type.unwrap();
```

because that variable is now bound by the `if let`

PR #870, opened 2024-03-20
Build and test for RISC-V

### LINKED ISSUE #775: Fix features and `build.rs` to compile with default features even on platforms with no asm support
See https://github.com/memorysafety/rav1d/issues/774#issuecomment-1970246931.

### LINKED ISSUE #861: Build and test for RISC-V
- [ ] Add RISC-V asm (#812)
- [x] Fix build on RISC-V (#775)

thedataking:
* Closes #775.
* Closes #861.

Build for QEMU user with:
```
RUSTFLAGS="-C target-feature=+crt-static,+v -C linker=riscv64-linux-gnu-gcc"  \
    cargo build --target riscv64gc-unknown-linux-gnu
```

(Note to self: remove `,+v` in above command to test RISC-V sans assembly.)

Note: the above command seems to work on Ubuntu 22.04 which comes with ld version 2.38 but fail on Ubuntu 20.04 where ld is version 2.34.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -13,8 +13,6 @@ mod asm {
     enum Arch {
         X86(ArchX86),
         Arm(ArchArm),
-        #[allow(dead_code)]
-        Unknown,
     }
 
     #[derive(Clone, Copy, PartialEq, Eq)]
@@ -128,7 +126,6 @@ mod asm {
         let use_nasm = match arch {
             Arch::X86(..) => true,
             Arch::Arm(..) => false,
-            _ => unimplemented!("should not get here"),
         };
 
         let define_prefix = if use_nasm { "%" } else { " #" };
@@ -251,13 +248,11 @@ mod asm {
             Arch::X86(ArchX86::X86_32) => x86_all,
             Arch::X86(ArchX86::X86_64) => x86_64_all,
             Arch::Arm(..) => arm_all,
-            _ => unimplemented!("should not get here"),
         };
 
         let asm_file_dir = match arch {
             Arch::X86(..) => ["x86", "."],
             Arch::Arm(..) => ["arm", pointer_width],
-            _ => unimplemented!("should not get here"),
         };
         let asm_extension = if use_nasm { "asm" } else { "S" };
 
--- a/build.rs
+++ b/build.rs
@@ -13,6 +13,8 @@ mod asm {
     enum Arch {
         X86(ArchX86),
         Arm(ArchArm),
+        #[allow(dead_code)]
+        Unknown,
     }
 
     #[derive(Clone, Copy, PartialEq, Eq)]
@@ -68,18 +70,14 @@ mod asm {
         let pointer_width = env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap();
         let features = env::var("CARGO_CFG_TARGET_FEATURE").unwrap();
 
-        let arch = arch.parse::<Arch>().unwrap();
+        // Nothing to do on unknown architectures
+        let Ok(arch) = arch.parse::<Arch>() else {
+            return;
+        };
         let os = os.as_str();
         let vendor = vendor.as_str();
         let pointer_width = pointer_width.as_str();
         let features = features.split(',').collect::<HashSet<_>>();
-
-        let rustc_cfg = match arch {
-            Arch::X86(ArchX86::X86_32) => "nasm_x86",
-            Arch::X86(ArchX86::X86_64) => "nasm_x86_64",
-            Arch::Arm(..) => "asm_neon",
-        };
-        println!("cargo:rustc-cfg={rustc_cfg}");
 
         let mut defines = Vec::new();
         let mut define = |define: Define| {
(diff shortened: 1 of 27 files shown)


src/decode.rs
```diff
@@ -1,5 +1,7 @@
 use crate::include::common::attributes::ctz;
+#[cfg(feature = "bitdepth_16")]
 use crate::include::common::bitdepth::BitDepth16;
+#[cfg(feature = "bitdepth_8")]
 use crate::include::common::bitdepth::BitDepth8;
```
kkysen, line 5: Are these just to get rid of warnings?  Same for the other ones.

build.rs
```diff
@@ -13,6 +13,8 @@ mod asm {
     enum Arch {
         X86(ArchX86),
         Arm(ArchArm),
+        #[allow(dead_code)]
+        Unknown,
```
kkysen, line 17: Why do we need this?  Your `else { return; }` below should take care of it.

build.rs
```diff
@@ -68,19 +70,15 @@ mod asm {
         let pointer_width = env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap();
         let features = env::var("CARGO_CFG_TARGET_FEATURE").unwrap();
 
-        let arch = arch.parse::<Arch>().unwrap();
+        // Nothing to do on unknown architectures
+        let Ok(arch) = arch.parse::<Arch>() else {
+            return;
+        };
```
kkysen, line 76: This fixes #775, right?

src/cpu.rs
```diff
@@ -42,6 +44,14 @@ bitflags! {
     }
 }
 
+#[cfg(any(target_arch = "riscv64", target_arch = "riscv32"))]
+bitflags! {
+    #[derive(Clone, Copy)]
+    pub struct CpuFlags: c_uint {
+        const V = 1 << 0;
+    }
+}
```
kkysen, line 53: We should `git cherry-pick` the corresponding `dav1d` commits for `riscv` asm support so that our C stays in sync.

kkysen (changes requested):

src/decode.rs
thedataking: yes

build.rs
thedataking: Otherwise I get this: "warning: variant `Unknown` is never constructed".

src/cpu.rs
thedataking: Done

build.rs
kkysen: Sorry that might've been confusing.  I didn't mean the `#[allow(dead_code)]`.  Can we remove `Unknown`?  Then we can remove the `_ => unimplemented!("should not get here")`s that are never reached because of the `else { return; }`.

src/cpu.rs
```diff
@@ -72,6 +82,8 @@ impl CpuFlags {
             CpuFlags::AVX512ICL,
             #[cfg(target_feature = "neon")]
             CpuFlags::NEON,
+            #[cfg(any(target_arch = "riscv64", target_arch = "riscv32"))]
```
kkysen, line 85:
```suggestion
            #[cfg(target_feature = "v")]
```

kkysen: We also need the runtime detection equivalent that can go below the arm ones:

```rust
        #[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
        if std::arch::is_riscv_feature_detected!("v") {
            flags |= Self::V;
        }
```

This also requires us changing in `lib.rs`

```rust
#![cfg_attr(target_arch = "arm", feature(stdsimd))]
```

to this

```rust
#![cfg_attr(any(target_arch = "arm", target_arch = "riscv32", target_arch = "riscv64"), feature(stdsimd))]
```

kkysen (changes requested):

src/cpu.rs
thedataking: It looks like the code changed after you made this suggestion. I used the following:

```
#![cfg_attr(target_arch = "arm", feature(stdarch_arm_feature_detection))]
#![cfg_attr(any(target_arch = "riscv32", target_arch = "riscv64"), feature(stdsimd))]
```

LMK if this seems right, I'm not terribly familiar with Rust feature detection.

kkysen: Yeah that looks right.  They split those features into more fine grained ones.

thedataking: I think we want to keep using this `cfg` (instead of `#[cfg(target_feature = "v")]`) because we want a CPU flags type even if we're on a RISC-V that doesn't have the vector instructions. Only declaring the `V` constant when the feature is detected is also cumbersome since we reference it other places (and need to reference it in more places as the RISC-V assembly gets patched in).

```Rust
#[cfg(any(target_arch = "riscv64", target_arch = "riscv32"))]
bitflags! {
    #[derive(Clone, Copy)]
    pub struct CpuFlags: c_uint {
        const V = 1 << 0;
    }
}
```

kkysen: Yeah, for the type itself, it should be cfg'd on riscv.  But the compile time method is the one where we cfg based on the target features, like we do for the other arches.

kkysen (changes requested):
Everything LGTM now except for in `fn compile_time_detect`, the `cfg` needs to be `#[cfg(target_feature = "v")]` because that `fn` is specifically for compile-time detection of features.  As you can see, we similarly do `#[cfg(target_feature = "neon")]` for `CpuFlags::NEON`.

kkysen (approved):

PR #871, opened 2024-03-20
Cleanup state machine control flow

### LINKED ISSUE #840: Cleanup state machine control flow in `thread_task.rs`
The following functions contain `current_block` state machine code that needs to be cleaned up:

* [x] `reset_task_cur`
* [x] `insert_tasks`
* [ ] `check_tile`

thedataking:
* Closes #840.

### DIFF
--- a/src/thread_task.rs
+++ b/src/thread_task.rs
@@ -215,10 +215,10 @@ unsafe fn insert_tasks(
                     return;
                 }
                 // same sby
-                if tasks[first].type_0 as c_uint > tasks[t].type_0 as c_uint {
+                if tasks[first].type_0 > tasks[t].type_0 {
                     break 'next;
                 }
-                if (tasks[first].type_0 as c_uint) < tasks[t].type_0 as c_uint {
+                if (tasks[first].type_0) < tasks[t].type_0 {
                     insert_tasks_between(c, f, first, last, prev_t, Some(t), cond_signal);
                     return;
                 }
@@ -226,24 +226,16 @@ unsafe fn insert_tasks(
             }
 
             // sort by tile-id
-            if !matches!(
-                tasks[first].type_0,
-                TaskType::TileReconstruction | TaskType::TileEntropy
-            ) {
-                unreachable!();
-            }
-            if !(tasks[first].type_0 == tasks[t].type_0) {
-                unreachable!();
-            }
-            if !(tasks[t].sby == tasks[first].sby) {
-                unreachable!();
-            }
+            assert!(
+                tasks[first].type_0 == TaskType::TileReconstruction
+                    || tasks[first].type_0 == TaskType::TileEntropy
+            );
+            assert!(tasks[first].type_0 == tasks[t].type_0);
+            assert!(tasks[t].sby == tasks[first].sby);
             let p = tasks[first].type_0 == TaskType::TileEntropy;
             let t_tile_idx = first - tasks.tile_tasks[p as usize].unwrap();
             let p_tile_idx = t - tasks.tile_tasks[p as usize].unwrap();
-            if !(t_tile_idx != p_tile_idx) {
-                unreachable!();
-            }
+            assert!(t_tile_idx != p_tile_idx);
             if !(t_tile_idx > p_tile_idx) {
                 insert_tasks_between(c, f, first, last, prev_t, Some(t), cond_signal);
                 return;
--- a/src/thread_task.rs
+++ b/src/thread_task.rs
@@ -500,46 +500,37 @@ unsafe fn check_tile(t_idx: Rav1dTaskIndex, f: &Rav1dFrameData, frame_mt: c_int)
         let p_b: c_uint = (((*t).sby + 1) << f.sb_shift + 2) as c_uint;
         let tile_sby = (*t).sby - ((*ts).tiling.row_start >> f.sb_shift);
         let lowest_px = &f.lowest_pixel_mem[(*ts).lowest_pixel + tile_sby as usize];
-        let mut current_block_14: u64;
-        let mut n = (*t).deps_skip;
-        while n < 7 {
-            let mut lowest: c_uint = 0;
-            if tp {
-                lowest = p_b;
-                current_block_14 = 2370887241019905314;
-            } else {
-                let y = if lowest_px[n as usize][0] == i32::MIN {
-                    i32::MIN
+        for n in t.deps_skip..7 {
+            'next: {
+                let lowest = if tp {
+                    p_b
                 } else {
-                    lowest_px[n as usize][0] + 8
-                };
-                let uv = if lowest_px[n as usize][1] == i32::MIN {
-                    i32::MIN
-                } else {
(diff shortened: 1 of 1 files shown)


randomPoison:
It looks like you might need to change your base for the PR or rebase again, one of stephens commits is included in the PR.

randomPoison:
For `insert_tasks`, I'm confused about how we got state machine control flow in the first place since there are no `goto`s in the function. So I'm also not clear on why the `match`-based rewrite you did is necessary. Am I missing something here?

randomPoison:
Ah, is it because the original was doing `continue` inside of a for loop? In the original C you still run the loop update logic after a `continue`, but in Rust where we're using a `while` loop the update logic goes at the end of the loop, so `continue` would skip it.

src/thread_task.rs
```diff
@@ -207,43 +207,36 @@ unsafe fn insert_tasks(
...
-                if tasks[first].type_0 as c_uint > tasks[t].type_0 as c_uint {
-                    current_block_34 = 11174649648027449784;
-                } else {
-                    if (tasks[first].type_0 as c_uint) < tasks[t].type_0 as c_uint {
-                        insert_tasks_between(c, f, first, last, prev_t, Some(t), cond_signal);
-                        return;
-                    }
-                    current_block_34 = 15904375183555213903;
-                }
+            (_, _, TaskType::TileEntropy, _) => {
```
randomPoison, line 224: I'm having a hard time convincing myself that the new control flow here is correct. As an example, in the cases where both task types are `TileEntropy` and both `sby` values are equal, the original C would hit the fallthrough case at the bottom of the loop, whereas I believe it would hit this match arm. Am I misreading this?

randomPoison (commented):
`reset_task_cur` and `check_tile` look good to me, but `insert_tasks` may have an issue. Not 100% sure, though. I found one case that seems like it would execute differently with your `match`-based version.

kkysen:
> Ah, is it because the original was doing `continue` inside of a for loop?

I think so.  `c2rust transpile` does that sometimes.

src/thread_task.rs
```diff
@@ -207,43 +207,36 @@ unsafe fn insert_tasks(
...
+            tasks[first].sby,
+        ) {
+            // entropy coding precedes other steps
+            (TaskType::TileEntropy, _, f_type, _) if f_type > TaskType::TileEntropy => { /* continue */
             }
-        } else {
-            if tasks[first].type_0 == TaskType::TileEntropy {
+            // both are entropy
+            (_, t_sby, _, f_sby) if f_sby > t_sby => { /* continue */ }
+            (_, t_sby, _, f_sby) if f_sby < t_sby => {
```
kkysen, line 220: Like @randomPoison, I'm not sure this is correct.  Shouldn't the second case be if `t.type_0 != TaskType::TileEntropy && first.type_0 == TaskType::TileEntropy`?

Note that if we get the logic wrong here, I don't think we get a test failure, just slower multithreading, so we need to be careful.

I think if you do my suggestion above of defining `t` and `first` variables and then `match (t.type_0, first.type_0)`ing, it'll be easier to understand the logic.

src/thread_task.rs
```diff
@@ -508,13 +508,10 @@ unsafe fn check_tile(t_idx: Rav1dTaskIndex, f: &Rav1dFrameData, frame_mt: c_int)
         let tile_sby = (*t).sby - ((*ts).tiling.row_start >> f.sb_shift);
         let lowest_px: *const [c_int; 2] =
             (*((*ts).lowest_pixel).offset(tile_sby as isize)).as_mut_ptr() as *const [c_int; 2];
-        let mut current_block_14: u64;
-        let mut n = (*t).deps_skip;
-        while n < 7 {
-            let mut lowest: c_uint = 0;
+        for n in tasks[t_idx].deps_skip..7 {
```
kkysen, line 512:
```suggestion
        for n in t.deps_skip..7 {
```

src/thread_task.rs
```diff
@@ -528,26 +525,20 @@ unsafe fn check_tile(t_idx: Rav1dTaskIndex, f: &Rav1dFrameData, frame_mt: c_int)
                 };
                 let max = cmp::max(y, uv);
                 if max == i32::MIN {
-                    current_block_14 = 7651349459974463963;
-                } else {
-                    lowest = iclip(max, 1 as c_int, f.refp[n as usize].p.p.h) as c_uint;
-                    current_block_14 = 2370887241019905314;
+                    tasks[t_idx].deps_skip += 1;
```
kkysen, line 532: Can you just set `tasks[t_idx].deps_skip = 7` at the end of the loop?

src/thread_task.rs
```diff
@@ -528,26 +525,20 @@ unsafe fn check_tile(t_idx: Rav1dTaskIndex, f: &Rav1dFrameData, frame_mt: c_int)
...
+                f.refp[n as usize].progress.as_ref().unwrap()[!tp as usize].load(Ordering::SeqCst);
+            if p3 < lowest {
+                return 1;
             }
-            n += 1;
+            f.task_thread
+                .error
+                .fetch_or((p3 == FRAME_ERROR) as c_int, Ordering::SeqCst);
+
             tasks[t_idx].deps_skip += 1;
```
kkysen, line 546: Same here for moving outside of the loop.

kkysen (changes requested):

thedataking:
> It looks like you might need to change your base for the PR or rebase again, one of stephens commits is included in the PR.

Yup, I had the wrong base 😓

src/thread_task.rs
thedataking: The loop can exit early so I don't think that would be right. Even if it was, it wouldn't make a difference.

src/thread_task.rs
```diff
@@ -207,43 +207,36 @@ unsafe fn insert_tasks(
...
-                if tasks[first].sby < tasks[t].sby {
-                    insert_tasks_between(c, f, first, last, prev_t, Some(t), cond_signal);
-                    return;
-                }
-                current_block_34 = 15904375183555213903;
+        match (
+            tasks[t].type_0,
+            tasks[t].sby,
+            tasks[first].type_0,
+            tasks[first].sby,
```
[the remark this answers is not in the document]
thedataking: we need `first` when we call `insert_tasks_between(.., first, ...)`. I'm switching to an approach without a `match`.

src/thread_task.rs
thedataking: Yeah, I think you are right. Might have done this too late in the evening :| 

I talked to Stephen and he showed me his break-out-of block approach which is closer to the C and easier to review.

randomPoison (approved):
Looks much better! I was going to suggest the labeled break approach if you were having trouble getting the control flow to match, glad that worked!

src/thread_task.rs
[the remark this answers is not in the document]
kkysen: Ahh.  Without a `match` like C does it is good, too.

PR #875, opened 2024-03-20
Backport memory optimization from dav1d 1.3.0

fbossen:
Define union to share memory space for `ac` and `txtp_map`

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -719,9 +719,16 @@ impl BitDepthDependentType for InterIntraEdge {
 
 #[derive(Clone, Copy)]
 #[repr(C)]
+pub union Rav1dTaskContext_scratch_ac_txtp_map {
+    pub ac: [i16; 1024],      // intra only
+    pub txtp_map: [u8; 1024], // inter only
+}
+
+#[derive(Clone, Copy)]
+#[repr(C)]
 pub struct Rav1dTaskContext_scratch_levels_pal_ac_interintra_edge {
     pub c2rust_unnamed: Rav1dTaskContext_scratch_levels_pal,
-    pub ac: [i16; 1024],
+    pub ac_txtp_map: Rav1dTaskContext_scratch_ac_txtp_map,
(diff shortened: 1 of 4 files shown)


src/internal.rs
```diff
@@ -717,11 +717,18 @@ impl BitDepthDependentType for InterIntraEdge {
     type T<BD: BitDepth> = Align64<InterIntraEdgeBD<BD>>;
 }
 
+#[derive(Clone, Copy)]
+#[repr(C)]
+pub union Rav1dTaskContext_scratch_ac_txtp_map {
+    pub ac: [i16; 1024],      // intra only
+    pub txtp_map: [u8; 1024], // inter only
+}
```
randomPoison, line 725: Rather than defining this as a union the way it is in C, could we define it as an enum? We're trying to cleanup unsafe code, and introducing another union means more places are going to require unsafe operations to access the union fields. The overhead of the extra discriminant field shouldn't outweigh the memory savings from moving `txtp_map` into the union/enum.

If there's some reason why that'd be a bad idea let me know.

randomPoison (changes requested):

src/internal.rs
fbossen: There are plenty of such unions in `Rav1dTaskContext_scratch`. I was just following the current pattern. Maybe all the unions could be fixed together in another PR? (I guess we'll have to be careful with potential alignment issues).

kkysen: I think the union is fine for now, we can merge it with a union and make it an enums later.  We might want to do all the inter intra enums together somehow.

kkysen (approved):

src/internal.rs
randomPoison: Makes sense 👍

randomPoison (approved):

src/internal.rs
kkysen: Oh you could replace the `1024` with `32 * 32`.

PR #876, opened 2024-03-20
`enum Rav1dRestorationType`: backport optimization from dav1d 1.3.0

fbossen:
Embed `sgr_idx` into `enum Rav1dRestorationType` to avoid storing it separately. Define `enum Rav1dRestorationType` as `repr(u8)` to match C code and save some memory.

### DIFF
--- a/include/dav1d/headers.rs
+++ b/include/dav1d/headers.rs
@@ -1,6 +1,10 @@
 use crate::src::enum_map::EnumKey;
 use std::ffi::c_int;
 use std::ffi::c_uint;
+use std::fmt;
+use std::fmt::Debug;
+use std::fmt::Display;
+use std::fmt::Formatter;
 use std::ops::BitAnd;
 use std::ops::Deref;
 use std::sync::atomic::AtomicU64;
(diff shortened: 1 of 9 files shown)


include/dav1d/headers.rs
```diff
@@ -193,13 +193,38 @@ pub const DAV1D_RESTORATION_WIENER: Dav1dRestorationType =
 pub const DAV1D_RESTORATION_SGRPROJ: Dav1dRestorationType =
     Rav1dRestorationType::SgrProj as Dav1dRestorationType;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, FromRepr, Default)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, FromRepr, Default)]
+#[repr(u8)]
```
kkysen, line 197: We shouldn't need this.  It'll choose the smallest size anyways and specifying a repr disables niche optimizations I believe.

include/dav1d/headers.rs
```diff
@@ -193,13 +193,38 @@ pub const DAV1D_RESTORATION_WIENER: Dav1dRestorationType =
...
+    SgrProj14,
+    SgrProj15,
+}
+
+impl Rav1dRestorationType {
+    pub fn discriminant(&self) -> u8 {
+        // SAFETY: Because `Self` is marked `repr(u8)`, its layout is a `repr(C)` `union`
+        // between `repr(C)` structs, each of which has the `u8` discriminant as its first
+        // field, so we can read the discriminant without offsetting the pointer.
+        unsafe { *<*const _>::from(self).cast::<u8>() }
```
kkysen, line 226: We shouldn't use unsafe here.  You can just do `as u8` since the enum is data-less (only the variants).

kkysen (changes requested):

kkysen (commented):
@fbossen, could I push a change?  I think it's easier to show that way than explaining it in scattered comments.  I want to add an `enum SgrIdx` to simplify this.

fbossen:
> @fbossen, could I push a change? I think it's easier to show that way than explaining it in scattered comments. I want to add an `enum SgrIdx` to simplify this.

No problem. Please do.

fbossen:
> @fbossen, how does [f72d0b9](https://github.com/memorysafety/rav1d/commit/f72d0b902c4189202901f85f2fe47a4be1a04c2b) look?

LGTM

kkysen (approved):

PR #877, opened 2024-03-20
`struct Rav1dFrameContext_frame_thread`: backport memory optimization from dav1d 1.3.0

fbossen:
Combine `eob` and `txtp` into single value.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -394,13 +394,6 @@ impl Rav1dFrameContext_bd_fn {
 
 #[derive(Default)]
 #[repr(C)]
-pub struct CodedBlockInfo {
-    pub eob: [i16; 3], /* plane */
-    pub txtp: [u8; 3], /* plane */
-}
-
-#[derive(Default)]
-#[repr(C)]
 pub struct Rav1dFrameContext_frame_thread {
     /// Indices: 0: reconstruction, 1: entropy.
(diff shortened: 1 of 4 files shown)


src/internal.rs
```diff
@@ -408,7 +401,7 @@ pub struct Rav1dFrameContext_frame_thread {
     /// Indexed using `t.by * f.b4_stride + t.bx`.
     pub b: Vec<Av1Block>,
 
-    pub cbi: Vec<CodedBlockInfo>,
+    pub cbi: Vec<[i16; 3]>, /* bits 0-4: txtp, bits 5-15: eob */
```
kkysen, line 404: We should probably newtype `i16` as `CodedBlockInfo`.  The bit-twiddling everywhere is pretty error-prone and brittle.  I can push a commit to do that if that's okay; I tested it out.

kkysen (changes requested):

src/internal.rs
kkysen: How does 7375ba3 look?

fbossen: > How does [7375ba3](https://github.com/memorysafety/rav1d/commit/7375ba3f3b4dc5160996606f7c3ba07ba113b0fa) look?

LGTM. I guess that, unlike in C, `<<` is not UB for signed integers.

kkysen (approved):

PR #878, opened 2024-03-20
`struct FrameTileThreadData::lowest_pixel_mem`: Make into `Vec`

### LINKED ISSUE #712: `FrameTileThreadData`: Make fields safe
Convert all raw pointer fields to an appropriate safe smart pointer or collection type.

- [x] `lowest_pixel_mem` - `malloc` allocation, can be a `Vec`.

randomPoison:
Also remove `FrameTileThreadData` because it only had a single field and was only used in one place. `lowest_pixel_mem` is now a direct field of `Rav1dFrameData`.

@rinon let me know if this looks like it'll conflict with your changes. From what I saw it doesn't look like you changed `FrameTileThreadData` so I'm not expecting it to conflict, but we can wait to merge this until after your changes are merged if that'll help.

* Closes #712

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -201,7 +201,6 @@ use crate::src::thread_task::TILE_ERROR;
 use crate::src::warpmv::rav1d_find_affine_int;
 use crate::src::warpmv::rav1d_get_shear_params;
 use crate::src::warpmv::rav1d_set_affine_mv2d;
-use libc::free;
 use libc::malloc;
 use libc::ptrdiff_t;
 use libc::uintptr_t;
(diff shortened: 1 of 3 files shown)


src/decode.rs
```diff
@@ -4325,18 +4324,11 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
...
-            if f.tile_thread.lowest_pixel_mem.is_null() {
-                f.tile_thread.lowest_pixel_mem_sz = 0;
-                return Err(ENOMEM);
-            }
-            f.tile_thread.lowest_pixel_mem_sz = lowest_pixel_mem_sz;
-        }
-        let mut lowest_pixel_ptr = f.tile_thread.lowest_pixel_mem;
+        // TODO: Fallible allocation
+        f.lowest_pixel_mem
+            .resize(lowest_pixel_mem_sz as usize, Default::default());
```
kkysen, line 4329:
```suggestion
            .resize_with(lowest_pixel_mem_sz as usize, Default::default);
```

kkysen (approved):
LGTM besides using `.resize_with` instead.

src/decode.rs
randomPoison: Why though? What's the difference between these two? Data is `Copy` so the end result is the same.

kkysen: The docs recommend it for `Default::default` and it optimizes a little bit better, since `.resize` does a write for `n - 1`, cloning them, and then writes the last one without the clone.  `.resize_with` just write all `n` at once, so the codegen is better, and the source is a bit cleaner.

randomPoison: Noted. I'll use `resize_with` going forward.

PR #879, opened 2024-03-21
`enum TxClass`: make a real enum

folkertdev:


src/recon.rs
```diff
@@ -913,7 +919,7 @@ unsafe fn decode_coefs<BD: BitDepth>(
                     }
                 }
             }
-            1 => {
+            TxClass::H => {
                 let lo_ctx_offsets = None;
                 let stride: ptrdiff_t = 16 as c_int as ptrdiff_t;
                 let shift: c_uint = ((*t_dim).lh as c_int + 2) as c_uint;
```
folkertdev, line 925: here the changes start repeating, because the C source uses a large macro for the switch cases here. That macro captures a lot of values. Or, well, it's C, so it does not capture anything, but in rust we'd need to pass all the values in somehow.

That'll be messy, so I think it's best to leave that for the future

src/recon.rs
```diff
@@ -708,18 +708,22 @@ unsafe fn decode_coefs<BD: BitDepth>(
...
-                    y = rc & mask;
-                } else if TX_CLASS_2D as c_int == TX_CLASS_H as c_int {
-                    x = eob as c_uint & mask;
-                    y = (eob >> shift) as c_uint;
-                    rc = eob as c_uint;
-                } else {
-                    x = eob as c_uint & mask;
-                    y = (eob >> shift) as c_uint;
-                    rc = x << shift2 | y;
+                match tx_class {
```
folkertdev, line 711: this relies on constant propagation to perform this match at compile time, but based on some tests even rust's MIR passes find this now (otherwise LLVM would)

randomPoison (approved):
The changes look correct, and I think replacing the various constants with `tx_class` should help us to deduplicate that code down the line 👍

src/levels.rs
```diff
@@ -81,10 +81,12 @@ pub const DCT_ADST: TxfmType = 2;
...
 
-pub type TxClass = c_uint;
-pub const TX_CLASS_V: TxClass = 2;
-pub const TX_CLASS_H: TxClass = 1;
-pub const TX_CLASS_2D: TxClass = 0;
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum TxClass {
+    TwoD = 0,
+    H = 1,
+    V = 2,
```
kkysen, line 88: Do we need the specified values since we're not parsing this from numbers (e.x. `from_repr`)?

kkysen (approved):
I've never seen an `if if` in the wild before 🤣.

src/levels.rs
folkertdev: we don't, fixed

kkysen (approved):

PR #887, opened 2024-03-21
`fn pthread_attr_setstacksize`: Remove stack size setting

### LINKED ISSUE #864: Set thread stack size using Rust APIs instead of pthread APIs

kkysen:
* Fixes #864.

See #889 for why stack size setting was fully removed instead of ported.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,7 +37,6 @@ use crate::src::internal::Rav1dTaskContext;
 use crate::src::internal::Rav1dTaskContext_task_thread;
 use crate::src::internal::TaskThreadData;
 use crate::src::log::Rav1dLog as _;
-use crate::src::mem::freep;
 use crate::src::mem::rav1d_alloc_aligned;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -247,7 +244,7 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
         return Err(ENOMEM);
     }
     let stack_size: usize = 1024 * 1024 * get_stack_size_internal(&mut thread_attr);
-    pthread_attr_setstacksize(&mut thread_attr, stack_size);
+    pthread_attr_destroy(&mut thread_attr);
```
randomPoison, line 247: Why destroy `thread_attr` here? Don't we already have logic for destroying it at the appropriate time?

kkysen: Nope, we deleted it accidentally I think.

randomPoison (approved):

PR #888, opened 2024-03-21
`struct Rav1dTileState_frame_thread::pal_idx`: Make into offset

randomPoison:
The first commit converts the field to an `Option<usize>` because the pointer was being set to null in one case. However all usages of `pal_idx` were asserting that it's not null before using, so I don't think making this an `Option` is actually necessary. The second commit removes the `Option`, but I left it as a separate commit in case it's problematic for some reason I'm missing.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -2063,9 +2063,8 @@ unsafe fn decode_b_inner(
                 let p = t.frame_thread.pass & 1;
                 let frame_thread = &mut ts.frame_thread[p as usize];
                 let len = usize::try_from(bw4 * bh4 * 16).unwrap();
-                let pal_idx_offset = frame_thread.pal_idx.as_mut().unwrap();
-                let pal_idx = &mut f.frame_thread.pal_idx[*pal_idx_offset..];
-                *pal_idx_offset += len;
+                let pal_idx = &mut f.frame_thread.pal_idx[frame_thread.pal_idx..][..len];
+                frame_thread.pal_idx += len;
                 pal_idx
             } else {
                 &mut t.scratch.c2rust_unnamed_0.pal_idx
@@ -2091,9 +2090,8 @@ unsafe fn decode_b_inner(
                 let p = t.frame_thread.pass & 1;
                 let frame_thread = &mut ts.frame_thread[p as usize];
                 let len = usize::try_from(cbw4 * cbh4 * 16).unwrap();
-                let pal_idx_offset = frame_thread.pal_idx.as_mut().unwrap();
-                let pal_idx = &mut f.frame_thread.pal_idx[*pal_idx_offset..];
-                *pal_idx_offset += len;
+                let pal_idx = &mut f.frame_thread.pal_idx[frame_thread.pal_idx..];
+                frame_thread.pal_idx += len;
                 pal_idx
             } else {
                 &mut t.scratch.c2rust_unnamed_0.pal_idx[(bw4 * bh4 * 16) as usize..]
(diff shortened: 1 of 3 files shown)


randomPoison:
@kkysen I made the one change you requested. I'm going to merge this under the assumption there are no other issues. If I somehow missed something I can fix it in another PR.

kkysen:
> @kkysen I made the one change you requested. I'm going to merge this under the assumption there are no other issues. If I somehow missed something I can fix it in another PR.

Oh I hadn't fully reviewed it.  I just make a couple suggestions for the part that I had looked at so far.

randomPoison:
Ah, woops! Please review and let me know if there are serious issues. I can always revert the merge if necessary.

src/decode.rs
```diff
@@ -2087,17 +2086,16 @@ unsafe fn decode_b_inner(
...
-            let pal_idx;
-            if t.frame_thread.pass != 0 {
+            let pal_idx = if t.frame_thread.pass != 0 {
                 let p = t.frame_thread.pass & 1;
                 let frame_thread = &mut ts.frame_thread[p as usize];
-                assert!(!(frame_thread.pal_idx).is_null());
                 let len = usize::try_from(cbw4 * cbh4 * 16).unwrap();
-                pal_idx = std::slice::from_raw_parts_mut(frame_thread.pal_idx, len);
-                frame_thread.pal_idx = frame_thread.pal_idx.offset(len as isize);
+                let pal_idx = &mut f.frame_thread.pal_idx[frame_thread.pal_idx..];
```
kkysen, line 2093:
```suggestion
                let pal_idx = &mut f.frame_thread.pal_idx[frame_thread.pal_idx..][..len];
```

src/recon.rs
```diff
@@ -2521,18 +2521,15 @@ pub(crate) unsafe fn rav1d_recon_b_intra<BD: BitDepth>(
...
                     let p = t.frame_thread.pass & 1;
-                    if ((*ts).frame_thread[p as usize].pal_idx).is_null() {
-                        unreachable!();
-                    }
-                    pal_idx = (*ts).frame_thread[p as usize].pal_idx;
-                    (*ts).frame_thread[p as usize].pal_idx =
-                        ((*ts).frame_thread[p as usize].pal_idx).offset((bw4 * bh4 * 16) as isize);
+                    let frame_thread = &mut (*ts).frame_thread[p as usize];
+                    let pal_idx = &f.frame_thread.pal_idx[frame_thread.pal_idx..];
+                    frame_thread.pal_idx += (bw4 * bh4 * 16) as usize;
```
kkysen, line 2528:
```suggestion
                    let len = (bw4 * bh4 * 16) as usize;
                    let pal_idx = &f.frame_thread.pal_idx[frame_thread.pal_idx..][..len];
                    frame_thread.pal_idx += len;
```

If this becomes a `DisjointMut` (it looks like it probably would?), we want these borrows to be more fine-grained, so we want to cut off the excess length to only what we use.  So for these `.split_at_mut()` like places, we should add `[..len]`s.

src/recon.rs
```diff
@@ -2904,41 +2901,34 @@ pub(crate) unsafe fn rav1d_recon_b_intra<BD: BitDepth>(
...
                             as isize;
-                        pal = &f.frame_thread.pal[index as usize][0] as *const [u16; 8];
-                        pal_idx = (*ts).frame_thread[p as usize].pal_idx;
-                        (*ts).frame_thread[p as usize].pal_idx = ((*ts).frame_thread[p as usize]
-                            .pal_idx)
-                            .offset((cbw4 * cbh4 * 16) as isize);
+                        let pal_idx_offset = &mut (*ts).frame_thread[p as usize].pal_idx;
+                        let pal_idx = &f.frame_thread.pal_idx[*pal_idx_offset..];
+                        *pal_idx_offset += (cbw4 * cbh4 * 16) as usize;
+                        (&f.frame_thread.pal[index as usize], pal_idx)
```
kkysen, line 2912:
```suggestion
                        let len = (cbw4 * cbh4 * 16) as usize;
                        let pal_idx = &f.frame_thread.pal_idx[*pal_idx_offset..][..len];
                        *pal_idx_offset += len;
                        (&f.frame_thread.pal[index as usize], pal_idx)
```

kkysen (commented):
It's mostly all fine, I just had more of the same comment from before in that I think we should cut off the excess length of these slices that are `.split_at_mut()` like.

randomPoison:
I've put up https://github.com/memorysafety/rav1d/pull/928 to address this feedback.

PR #890, opened 2024-03-22
Add `DisjointMut` wrapper to allow concurrent mutable slicing

rinon:
Applies the wrapper to `struct Rav1dFrameContext_frame_thread::pal`.  When debug assertions are enabled, this wrapper tracks borrows and ensures that mutable borrowed regions are actually exclusive for their lifetime. In release mode, borrowing is unchecked and relies on correct disjoint access.

### DIFF
--- a/src/align.rs
+++ b/src/align.rs
@@ -13,7 +13,6 @@ use std::ops::Deref;
 use std::ops::DerefMut;
 use std::ops::Index;
 use std::ops::IndexMut;
-use std::ptr;
 use std::slice;
 
 use crate::src::disjoint_mut::AsMutPtr;
@@ -232,13 +231,11 @@ pub type AlignedVec64<T> = AlignedVec<T, Align64<[u8; 64]>>;
 unsafe impl<T: Copy, C: AlignedByteChunk> AsMutPtr for AlignedVec<T, C> {
     type Target = T;
 
-    unsafe fn as_mut_slice(ptr: *mut Self) -> *mut [Self::Target] {
-        let len = unsafe { (*ptr).len() };
+    unsafe fn as_mut_ptr(ptr: *mut Self) -> *mut Self::Target {
         // SAFETY: .as_mut_ptr() does not materialize a mutable reference to the
         // underlying slice so we can still allow immutable references into this
         // slice.
-        let data = unsafe { (*ptr).as_mut_ptr() };
-        ptr::slice_from_raw_parts_mut(data, len)
+        unsafe { (*ptr).as_mut_ptr() }
     }
 
     fn len(&self) -> usize {
--- a/src/align.rs
+++ b/src/align.rs
@@ -13,7 +13,10 @@ use std::ops::Deref;
 use std::ops::DerefMut;
 use std::ops::Index;
 use std::ops::IndexMut;
+use std::ptr;
 use std::slice;
+
+use crate::src::disjoint_mut::AsMutPtr;
 
 /// [`Default`] isn't `impl`emented for all arrays `[T; N]`
 /// because they were implemented before `const` generics
@@ -190,6 +193,18 @@ impl<T: Copy, C: AlignedByteChunk> AlignedVec<T, C> {
     }
 }
 
+impl<T: Copy, C: AlignedByteChunk> AsRef<[T]> for AlignedVec<T, C> {
+    fn as_ref(&self) -> &[T] {
+        self.as_slice()
+    }
+}
+
+impl<T: Copy, C: AlignedByteChunk> AsMut<[T]> for AlignedVec<T, C> {
+    fn as_mut(&mut self) -> &mut [T] {
+        self.as_mut_slice()
+    }
+}
+
 impl<T: Copy, C: AlignedByteChunk> Deref for AlignedVec<T, C> {
     type Target = [T];
 
@@ -213,3 +228,20 @@ impl<T: Copy, C: AlignedByteChunk> Default for AlignedVec<T, C> {
 
 pub type AlignedVec32<T> = AlignedVec<T, Align32<[u8; 32]>>;
 pub type AlignedVec64<T> = AlignedVec<T, Align64<[u8; 64]>>;
+
+unsafe impl<T: Copy, C: AlignedByteChunk> AsMutPtr for AlignedVec<T, C> {
+    type Target = T;
+
+    unsafe fn as_mut_slice(ptr: *mut Self) -> *mut [Self::Target] {
+        let len = unsafe { (*ptr).len() };
+        // SAFETY: .as_mut_ptr() does not materialize a mutable reference to the
+        // underlying slice so we can still allow immutable references into this
+        // slice.
+        let data = unsafe { (*ptr).as_mut_ptr() };
+        ptr::slice_from_raw_parts_mut(data, len)
+    }
+
+    fn len(&self) -> usize {
+        self.len()
+    }
+}
--- a/src/disjoint_mut.rs
+++ b/src/disjoint_mut.rs
@@ -473,10 +473,15 @@ mod debug {
         }
 
         fn remove_range(&self, range: (Bound<usize>, Bound<usize>), mutable: bool) {
-            let mut ranges = if mutable {
-                self.ranges.mutable.lock().unwrap()
+            let ranges = if mutable {
+                self.ranges.mutable.lock()
             } else {
-                self.ranges.immutable.lock().unwrap()
+                self.ranges.immutable.lock()
+            };
+            let Ok(mut ranges) = ranges else {
+                // Another thread has panicked holding a range lock. We can't
+                // remove anything.
+                return;
             };
             let idx = ranges
                 .iter()
@@ -584,6 +589,30 @@ impl<V: Copy, C: AlignedByteChunk> DisjointMut<AlignedVec<V, C>> {
     }
 }
 
+#[test]
+fn test_overlapping_immut() {
+    let mut v: DisjointMut<Vec<u8>> = Default::default();
+    v.resize(10, 0u8);
+
+    let guard1 = unsafe { v.index(0..5) };
+    let guard2 = unsafe { v.index(2..) };
+
+    assert_eq!(guard1[2], guard2[0]);
+}
+
+#[test]
+#[cfg_attr(debug_assertions, should_panic)]
+fn test_overlapping_mut() {
+    let mut v: DisjointMut<Vec<u8>> = Default::default();
+    v.resize(10, 0u8);
+
+    let guard1 = unsafe { v.index(0..5) };
+    let mut guard2 = unsafe { v.index_mut(2..) };
+
+    guard2[0] = 42;
+    assert_eq!(guard1[2], 42);
+}
+
 #[cfg(debug_assertions)]
 #[test]
 fn test_pointer_write_debug() {
@@ -616,7 +645,7 @@ fn test_pointer_write_release() {
     let mut v: DisjointMut<Vec<[u8; 4]>> = Default::default();
     v.resize(10, [0u8; 4]);
 
-    let borrow = v.index(0..);
+    let borrow = unsafe { v.index(0..) };
     let ptr = v.as_mut_ptr().wrapping_offset(3) as *mut u8;
     unsafe {
         ptr.wrapping_offset(2).write(42);
@@ -629,8 +658,8 @@ fn test_pointer_write_release() {
     // assert_eq!(borrow[3][2], 0);
 
     // We are fine to re-borrow at this point now that the write is done.
-    assert_eq!(v.index(4)[0], 0);
-    assert_eq!(v.index(3)[2], 42);
+    assert_eq!(unsafe { v.index(4)[0] }, 0);
+    assert_eq!(unsafe { v.index(3)[2] }, 42);
 }
 
 #[test]
--- a/src/disjoint_mut.rs
+++ b/src/disjoint_mut.rs
@@ -1,0 +1,675 @@
+//! Wrapper that allows concurrent, disjoint mutation of a slice-like owned
+//! structure.
+
+#![deny(unsafe_op_in_unsafe_fn)]
+// TODO(SJC): Remove when we use the whole module.
+#![allow(unused)]
+
+use std::cell::UnsafeCell;
+use std::fmt::Debug;
+use std::marker::PhantomData;
+use std::ops::Bound;
+use std::ops::Deref;
+use std::ops::DerefMut;
+use std::ops::Index;
+use std::ops::IndexMut;
+use std::ops::Range;
+use std::ops::RangeBounds;
+use std::ops::RangeFrom;
+use std::ops::RangeInclusive;
+use std::ops::RangeTo;
+use std::ops::RangeToInclusive;
+use std::ptr;
+
+use crate::src::align::AlignedByteChunk;
+use crate::src::align::AlignedVec;
+
+/// Wraps an indexable collection to allow unchecked concurrent mutable borrows.
+///
+/// This wrapper allows users to concurrently mutably borrow disjoint regions or
+/// elements from a collection. This is necessary to allow multiple threads to
+/// concurrently read and write to disjoint pixel data from the same arrays and
+/// vectors.
+///
+/// In debug mode (debug assertions enabled), indexing returns a guard which
+/// acts as a lock for the borrowed region and borrows are checked to ensure
+/// that mutably borrowed regions are actually disjoint with all other borrows
+/// for the lifetime of the returned borrow guard.
+#[derive(Default)]
+#[cfg_attr(not(debug_assertions), repr(transparent))]
+pub struct DisjointMut<T: AsMutPtr> {
+    inner: UnsafeCell<T>,
+
+    #[cfg(debug_assertions)]
+    ranges: debug::DisjointMutDebugRanges,
+}
+
+#[cfg_attr(not(debug_assertions), repr(transparent))]
+pub struct DisjointMutGuard<'a, T: AsMutPtr, V: ?Sized> {
+    slice: &'a mut V,
+
+    phantom: PhantomData<&'a DisjointMut<T>>,
+
+    #[cfg(debug_assertions)]
+    parent: &'a DisjointMut<T>,
+    #[cfg(debug_assertions)]
+    range: (Bound<usize>, Bound<usize>),
+}
+
+impl<'a, T: AsMutPtr, V: ?Sized> Deref for DisjointMutGuard<'a, T, V> {
+    type Target = V;
+
+    fn deref(&self) -> &Self::Target {
+        self.slice
+    }
+}
+
+impl<'a, T: AsMutPtr, V: ?Sized> DerefMut for DisjointMutGuard<'a, T, V> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.slice
+    }
+}
+
+#[cfg_attr(not(debug_assertions), repr(transparent))]
+pub struct DisjointImmutGuard<'a, T: AsMutPtr, V: ?Sized> {
+    slice: &'a V,
+
+    phantom: PhantomData<&'a DisjointMut<T>>,
+
+    #[cfg(debug_assertions)]
+    parent: &'a DisjointMut<T>,
+    #[cfg(debug_assertions)]
+    range: (Bound<usize>, Bound<usize>),
+}
+
+impl<'a, T: AsMutPtr, V: ?Sized> Deref for DisjointImmutGuard<'a, T, V> {
+    type Target = V;
+
+    fn deref(&self) -> &Self::Target {
+        self.slice
+    }
+}
+
+/// Convert from a mutable pointer to a collection to a mutable pointer to the
+/// underlying slice without ever creating a mutable reference to the slice.
+///
+/// This trait exists for the same reason as [`Vec::as_mut_ptr`] - we want to
+/// create a mutable pointer to the underlying slice without ever creating a
+/// mutable reference to the slice.
+///
+/// # Safety
+///
+/// This trait must not ever create a mutable reference to the underlying slice,
+/// as it may be (partially) immutably borrowed concurrently.
+pub unsafe trait AsMutPtr: Sized {
+    type Target;
+
+    /// Convert a mutable pointer to a collection to a mutable pointer to the
+    /// underlying slice.
+    ///
+    /// # Safety
+    ///
+    /// This method may dereference `ptr` as an immutable reference, so this
+    /// pointer must be safely dereferenceable.
+    unsafe fn as_mut_slice(ptr: *mut Self) -> *mut [Self::Target];
+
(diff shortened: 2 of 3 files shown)


src/disjoint_mut.rs
```diff
@@ -0,0 +1,417 @@
...
+                        (Bound::Unbounded, _) | (_, Bound::Unbounded) => true,
+                    }
+                }
+            }
+        };
+    }
+
+    impl_disjoint_mut_index!(Range<usize>);
+    impl_disjoint_mut_index!(RangeFrom<usize>);
+    impl_disjoint_mut_index!(RangeTo<usize>);
```
randomPoison, line 293: Do we also want to include `RangeInclusive` and `RangeToInclusive` here? We're also missing `RangeFull`, but I don't think that matters here because borrowing the full range would defeat the purpose of tracking disjoint borrows.

On a related note, would it make sense (or even be possible) to do this with a blanket impl rather than a macro? I would think we could do something like:

```rust
impl<T: RangeBounds> DisjointMutIndex for T {
    ...
}
```

And that would cover all the range types automatically. I think this would work because the macro logic is already assuming `Self: RangeBounds`, but I might be missing a reason why a blanket impl wouldn't work.

src/disjoint_mut.rs
```diff
@@ -0,0 +1,417 @@
...
+                        (_, Bound::Unbounded) => {}
+                    }
+                    match (other.end_bound(), self.start_bound()) {
+                        (Bound::Included(i), Bound::Included(j)) => i >= j,
+                        (Bound::Included(i), Bound::Excluded(j))
+                        | (Bound::Excluded(i), Bound::Included(j)) => i > j,
+                        (Bound::Excluded(i), Bound::Excluded(j)) => i - 1 > *j,
+                        (Bound::Unbounded, _) | (_, Bound::Unbounded) => true,
+                    }
+                }
```
randomPoison, line 286: I was having a hard time understanding this logic enough to confirm that it's accurate, so I ended up writing a unit test to check various edge cases:

```rust
#[test]
fn test_range_overlap() {
    // Range overlap.
    assert!((5..7).overlaps(&(4..10)));
    assert!((4..10).overlaps(&(5..7)));
    
    // RangeFrom overlap.
    assert!((5..).overlaps(&(4..10)));
    assert!((4..10).overlaps(&(5..)));

    // RangeTo overlap.
    assert!((..7).overlaps(&(4..10)));
    assert!((4..10).overlaps(&(..7)));

    // RangeInclusive overlap.
    assert!((5..=7).overlaps(&(7..10)));
    assert!((7..10).overlaps(&(5..=7)));

    // RangeToInclusive overlap.
    assert!((..=7).overlaps(&(7..10)));
    assert!((7..10).overlaps(&(..=7)));

    // Range no overlap.
    assert!(!(5..7).overlaps(&(10..20)));
    assert!(!(10..20).overlaps(&(5..7)));

    // RangeFrom no overlap.
    assert!(!(15..).overlaps(&(4..10)));
    assert!(!(4..10).overlaps(&(15..)));

    // RangeTo no overlap.
    assert!(!(..7).overlaps(&(10..20)));
    assert!(!(10..20).overlaps(&(..7)));

    // RangeInclusive no overlap.
    assert!(!(5..=7).overlaps(&(8..10)));
    assert!(!(8..10).overlaps(&(5..=7)));

    // RangeToInclusive no overlap.
    assert!(!(..=7).overlaps(&(8..10)));
    assert!(!(8..10).overlaps(&(..=7)));
}
```

I think it'd be nice to include a unit test like this for potentially tricky logic, though we're not running Rust tests in CI so it'd only be helpful for manually verifying changes. I also included tests for `RangeInclusive` and `RangeToInclusive` in case you decide to add support for those. Up to you if you want to add it or not.

randomPoison (approved):
The implementation here looks good, just had a couple of comments/questions.

I do, however, wonder if this checking is fine-grained enough to work in all the cases where we need it. In the case where we have multi-dimensional arrays, this tracking will only be able to track if borrows of the top-level array are disjoint. If multiple threads try to borrow the same element of the outermost array, but only access disjoint parts of the inner array, then the current logic would count that as a conflict.

I'm not sure if that's a concern in practice, I don't know the data access patterns well enough. I'm also not sure how we'd make the tracking more fine-grained, since tracking borrows through multiple layers of nested array in a generic way seems hard. Not sure what do do about that other than hope that this coarse tracking is enough 🤷

src/decode.rs
```diff
@@ -770,9 +769,14 @@ unsafe fn read_pal_plane(
     let used_cache = &used_cache[..i];
 
     // parse new entries
+    #[allow(unused_mut)]
```
kkysen, line 772:
```suggestion
```

Why's this needed?  You're doing `&mut pal_guard[pli]` so it seems like it would need to be `mut`.

src/decode.rs
```diff
@@ -852,9 +856,14 @@ unsafe fn read_pal_uv(
     // V pal coding
     let ts = &mut *t.ts;
 
+    #[allow(unused_mut)]
```
kkysen, line 859:
```suggestion
```

src/decode.rs
```diff
@@ -2241,10 +2253,12 @@ unsafe fn decode_b_inner(
...
+                let pal_guard;
                 let pal = if t.frame_thread.pass != 0 {
                     let index = ((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
                         + ((t.bx >> 1) + (t.by & 1)) as isize;
-                    &f.frame_thread.pal[index as usize]
+                    pal_guard = f.frame_thread.pal.index(index as usize);
+                    &pal_guard
                 } else {
                     &t.scratch.c2rust_unnamed_0.pal
                 };
```
kkysen, line 2264: Since `pal` is just a 128-bit `&[u16; 8]`, we can just copy it and have `pal` be a `[u16; 8]`.  That way we don't need the guard.

```rust
            let pal = if t.frame_thread.pass != 0 {
                let index = ((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
                    + ((t.bx >> 1) + (t.by & 1)) as isize;
                f.frame_thread.pal.index(index as usize)[0]
            } else {
                t.scratch.c2rust_unnamed_0.pal[0]
            };
            for al_pal in &mut t.al_pal[0][bx4 as usize..][..bw4 as usize] {
                al_pal[0] = pal;
            }
            for al_pal in &mut t.al_pal[1][by4 as usize..][..bh4 as usize] {
                al_pal[0] = pal;
            }
```

src/decode.rs
```diff
@@ -769,9 +769,14 @@ unsafe fn read_pal_plane(
...
-        &mut f.frame_thread.pal[(((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
-            + ((t.bx >> 1) + (t.by & 1)) as isize) as usize][pli]
+        pal_guard = f.frame_thread.pal.index_mut(
+            (((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
+                + ((t.bx >> 1) + (t.by & 1)) as isize) as usize,
+        );
+        &mut pal_guard[pli]
     } else {
         &mut t.scratch.c2rust_unnamed_0.pal[pli]
     };
```
kkysen, line 782: We don't need a non-transient guard here.  We can do the 128-bit write at the end instead of holding the `&mut` the whole time:

```rust
    let mut pal = [0; 8];
    if i < pal_sz {
        let mut prev = rav1d_msac_decode_bools(&mut ts.msac, f.cur.p.bpc as u32) as u16;
        pal[i] = prev;
        i += 1;

        if i < pal_sz {
            let mut bits = f.cur.p.bpc as u32 + rav1d_msac_decode_bools(&mut ts.msac, 2) - 3;
            let max = (1 << f.cur.p.bpc) - 1;

            loop {
                let delta = rav1d_msac_decode_bools(&mut ts.msac, bits) as u16;
                prev = cmp::min(prev + delta + not_pl, max);
                pal[i] = prev;
                i += 1;
                if prev + not_pl >= max {
                    pal[i..].fill(max);
                    break;
                } else {
                    bits = cmp::min(bits, 1 + ulog2((max - prev - not_pl) as u32) as u32);
                    if !(i < pal_sz) {
                        break;
                    }
                }
            }
        }

        // merge cache+new entries
        let mut n = 0;
        let mut m = used_cache.len();
        for i in 0..pal_sz {
            if n < used_cache.len() && (m >= pal_sz || used_cache[n] <= pal[m]) {
                pal[i] = used_cache[n];
                n += 1;
            } else {
                pal[i] = pal[m];
                m += 1;
            }
        }
    } else {
        pal[..used_cache.len()].copy_from_slice(&used_cache);
    }

    if t.frame_thread.pass != 0 {
        f.frame_thread.pal.index_mut(
            (((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
                + ((t.bx >> 1) + (t.by & 1)) as isize) as usize,
        )[pli] = pal;
    } else {
        t.scratch.c2rust_unnamed_0.pal[pli] = pal;
    };
```

src/decode.rs
```diff
@@ -851,9 +856,14 @@ unsafe fn read_pal_uv(
...
-        &mut f.frame_thread.pal[(((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
-            + ((t.bx >> 1) + (t.by & 1)) as isize) as usize][2]
+        pal_guard = f.frame_thread.pal.index_mut(
+            (((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
+                + ((t.bx >> 1) + (t.by & 1)) as isize) as usize,
+        );
+        &mut pal_guard[2]
     } else {
         &mut t.scratch.c2rust_unnamed_0.pal[2]
     };
```
kkysen, line 869: Same for this one, we can make the guard transient:

```rust
    let pal_sz = b.pal_sz()[1] as usize;
    let mut pal = [0; 8];
    if rav1d_msac_decode_bool_equi(&mut ts.msac) {
        let bits = f.cur.p.bpc as u32 + rav1d_msac_decode_bools(&mut ts.msac, 2) - 4;
        let mut prev = rav1d_msac_decode_bools(&mut ts.msac, f.cur.p.bpc as c_uint) as u16;
        pal[0] = prev;
        let max = (1 << f.cur.p.bpc) - 1;
        for pal in &mut pal[1..pal_sz] {
            let mut delta = rav1d_msac_decode_bools(&mut ts.msac, bits) as i16;
            if delta != 0 && rav1d_msac_decode_bool_equi(&mut ts.msac) {
                delta = -delta;
            }
            prev = ((prev as i16 + delta) as u16) & max;
            *pal = prev;
        }
    } else {
        pal[..pal_sz]
            .fill_with(|| rav1d_msac_decode_bools(&mut ts.msac, f.cur.p.bpc as c_uint) as u16);
    }
    if t.frame_thread.pass != 0 {
        f.frame_thread.pal.index_mut(
            (((t.by >> 1) + (t.bx & 1)) as isize * (f.b4_stride >> 1)
                + ((t.bx >> 1) + (t.by & 1)) as isize) as usize,
        )[2] = pal;
    } else {
        t.scratch.c2rust_unnamed_0.pal[2] = pal;
    };
```

kkysen (changes requested):

src/disjoint_mut.rs
rinon: I tried to do that but we run into the problem of making a blanket impl for a trait that we don't control while specializing for usize. Someone else might come along and impl RangeBounds for usize.

```
error[E0119]: conflicting implementations of trait `disjoint_mut::BoundsExt` for type `usize`
   --> src/disjoint_mut.rs:118:9
    |
96  | impl BoundsExt for usize {
    | ------------------------ first implementation here
...
118 |         impl<T: RangeBounds<V>, V> BoundsExt for T {
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `usize`
    |
    = note: downstream crates may implement trait `std::ops::RangeBounds<_>` for type `usize`
```

src/decode.rs
rinon: It's not obviously clear to me that this is the same without copying the data out of the array first. It might be, but I'd have to think on this. What's wrong with the mutable reference here for now?

kkysen: To me, it's simpler code.  We create `pal` and then write it, so it's easier to reason about things.  We don't have to reason about lifetimes.  `pal` is only written to, so this should be fine to do.

rinon: It's not quite write-only, `pal[m]` is read from and I'm having trouble convincing myself that every entry is written to (i.e. the original contents are never read). I'd prefer a simpler diff here, considering that the "guard" is zero-cost in release mode.

kkysen: > It's not quite write-only, `pal[m]` is read from and I'm having trouble convincing myself that every entry is written to (i.e. the original contents are never read). I'd prefer a simpler diff here, considering that the "guard" is zero-cost in release mode.

Hmm, yeah, that's not as trivial I thought.  Nevermind then.

src/disjoint_mut.rs
```diff
@@ -0,0 +1,615 @@
...
+
+    fn len(&self) -> usize {
+        N
+    }
+}
+
+impl<V, const A: usize, const B: usize> DisjointMut<[[V; A]; B]> {
+    // TODO(SJC): Add debug ranges for this method or make this return a raw pointer?
+    #[allow(unused)]
+    pub unsafe fn nd_index_mut(&self, index: [usize; 2]) -> &mut V {
```
randomPoison, line 485: Are we expecting to add versions of this for more deeply nested arrays? If so maybe name this `index_mut_2d` so that we can more easily add 3d or 4d versions as needed.

kkysen (commented):
For the previous suggestions I made about minimizing the `pal` borrows, they're not too important for this PR.  Also, I realized I think I need `DisjointMut` to `Arc`ify the segmaps, too (if I don't want to have blocking `RwLock`s).

src/disjoint_mut.rs
randomPoison: So if we use `nd_index_mut` we're not getting any disjointness checking, right? What's the reason for this function then? Based on the TODO is the plan to add debug checks here?

src/recon.rs
```diff
@@ -2532,7 +2532,7 @@ pub(crate) unsafe fn rav1d_recon_b_intra<BD: BitDepth>(
                     let index = (((t.by as isize >> 1) + (t.bx as isize & 1)) * (f.b4_stride >> 1)
                         + ((t.bx >> 1) + (t.by & 1)) as isize)
                         as isize;
-                    f.frame_thread.pal[index as usize][0].as_ptr()
+                    f.frame_thread.pal.index(index as usize)[0].as_ptr()
```
randomPoison, line 2535: Is this a valid way to use `DisjointMut`? i.e. without sticking the guard on the stack? It looks like this logic is getting a pointer to one of the nested arrays within `pal`, but you're not keeping the guard active while that pointer is being used. Is this actually going to catch incorrectly overlapping borrows?

randomPoison:
Another thought: Would it make sense to always return a guard, even in release builds? I'm playing around with `DisjointMut` on a branch and I'm noticing that we can get warnings in release mode that don't show up in debug builds because of the change from `DisjointMutGuard<T>` and `&mut T`, e.g. if you do something like:

```rust
let mut guard = foo.index_mut(index);
let item = &mut *guard;
```

In debug this compiles without issues, but in release you get warnings about unused `mut` on `guard` because that variable is now a `&mut T` and so the variable itself doesn't need to be marked as `mut`. Similarly if you manually `drop` the guard you'll get warnings in release builds for trying to drop a reference.

src/disjoint_mut.rs
```diff
@@ -0,0 +1,417 @@
...
+        ) -> &mut <<T as Deref>::Target as Index<I>>::Output
+        where
+            T: DerefMut,
+            <T as Deref>::Target: IndexMut<I>,
+        {
+            // SAFETY: The preconditions of our `index_mut` safety imply that
+            // the indexed region we are mutably borrowing is not concurrently
+            // borrowed and will not be borrowed during the lifetime of the
+            // returned reference.
+            unsafe { &mut (*self.inner.get()).deref_mut()[index] }
```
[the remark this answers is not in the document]
rinon: We can't move the reference out unless we create a temporary. I'd prefer to just leave it, clearer.

src/disjoint_mut.rs
rinon: Correct, I was planning to add a 2d range debug thing, just haven't yet.

src/disjoint_mut.rs
```diff
@@ -0,0 +1,615 @@
...
+                    Bound::Excluded(i) => i + 1,
+                    Bound::Unbounded => 0,
+                };
+                let end = match self.end_bound() {
+                    Bound::Included(i) => i + 1,
+                    Bound::Excluded(i) => *i,
+                    Bound::Unbounded => len,
+                };
+                if start <= end && start < len && end <= len {
+                    unsafe { slice::from_raw_parts_mut((slice as *mut T).add(start), end - start) }
```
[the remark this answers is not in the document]
rinon: Good catch, thanks.

src/recon.rs
rinon: Fixed

src/disjoint_mut.rs
randomPoison: Were you planning on adding the 2d debug range as part of this PR, or was that meant to be a separate PR? I worry that if we merge this without that we're going to end up with a hole in the `DisjointMut` coverage any place that we try to use `nd_index_mut`, but I'm not sure if you have a reason to think that won't be an issue in practice.

randomPoison (approved):
Looks good, I just had the one remaining question about adding the debug checking for `nd_index_mut`.

src/disjoint_mut.rs
```diff
@@ -0,0 +1,704 @@
...
+        I: DisjointMutIndex<
+            [<T as AsMutPtr>::Target],
+            Output = <[<T as AsMutPtr>::Target] as Index<I>>::Output,
+        >,
+    {
+        let range = index.to_bounds();
+        // SAFETY: The preconditions of our `index_mut` safety imply that the
+        // indexed region we are mutably borrowing is not concurrently borrowed
+        // and will not be borrowed during the lifetime of the returned
+        // reference.
```
kkysen, line 208:
```suggestion
        // SAFETY: The safety preconditions of `index` implies that the
        // indexed region we are immutably borrowing is not concurrently mutably borrowed
        // and will not be mutably borrowed during the lifetime of the returned
        // reference.
```

src/disjoint_mut.rs
```diff
@@ -0,0 +1,704 @@
...
+        assert!(index[0] < B);
+        assert!(index[1] < A);
+        let array = self.inner.get() as *mut [V; A];
+        let array = array.wrapping_add(index[0]) as *mut V;
+        // SAFETY: The preconditions of our `index_mut`  and `nd_index_mut`
+        // safety imply that the indexed region we are mutably borrowing is not
+        // concurrently borrowed and will not be borrowed during the lifetime of
+        // the returned reference. We ensure that the indices are in-bounds with
+        // the two asserts above.
+        unsafe { &mut *array.wrapping_add(index[1]) }
```
kkysen, line 584: What are the `.wrapping_add`s for vs. just `.add`?

src/disjoint_mut.rs
kkysen: > Are we expecting to add versions of this for more deeply nested arrays? If so maybe name this `index_mut_2d` so that we can more easily add 3d or 4d versions as needed.

I agree, this one's only for 2d arrays, right?

src/disjoint_mut.rs
```diff
@@ -0,0 +1,704 @@
...
+        ptr.wrapping_offset(2).write(42);
+    }
+
+    // Miri complains if we dereference the guard after the pointer write here
+    // instead of dereferencing and borrowing the slice before the write. I
+    // think this is a Miri limitation although to be safe we might want to
+    // change how we store the borrow in the guard. This is not an issue in
+    // release mode without the guards (see test_pointer_write_release below).
+    //
+    // assert_eq!(guard[4][0], 0);
```
kkysen, line 613: Should we have `#[should_panic]` tests?

kkysen (changes requested):
Most of this looks really good I think!  I just had a few potentially important points:
* Do we need a `ptr.cast_const()` for one of the `.len()`s?  One suggestion to fix this and avoid similar things is flipping `fn as_mut_ptr` and `fn as_mut_slice` in terms of which is implemented in terms of the other.
* I think `fn index` needs to be `unsafe` just like `fn index_mut`.
* I agree with @randomPoison that `index_mut_2d` makes more sense than `nd_index_mut`.
* Should we have `#[should_panic]` tests?  IMO it's more important for this to know that in correct code will panic with `#[cfg(debug_assertions)]`.

src/disjoint_mut.rs
rinon: I thought for a bit on how to do this properly, and then realized the only place we use it could be atomics instead... So I just did that and removed this function entirely. It was always a bit out of place.

src/align.rs
```diff
@@ -213,3 +228,20 @@ impl<T: Copy, C: AlignedByteChunk> Default for AlignedVec<T, C> {
 
 pub type AlignedVec32<T> = AlignedVec<T, Align32<[u8; 32]>>;
 pub type AlignedVec64<T> = AlignedVec<T, Align64<[u8; 64]>>;
+
+unsafe impl<T: Copy, C: AlignedByteChunk> AsMutPtr for AlignedVec<T, C> {
+    type Target = T;
+
+    unsafe fn as_mut_slice(ptr: *mut Self) -> *mut [Self::Target] {
+        let len = unsafe { (*ptr).len() };
```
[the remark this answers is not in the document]
rinon: I don't think it was needed for Vec, I'm going to remove it from there. We're not creating a mutable reference, we're calling an immutable method.

src/disjoint_mut.rs
```diff
@@ -0,0 +1,704 @@
...
+    }
+
+    #[allow(unused)]
+    pub fn get_mut(&mut self) -> &mut T {
+        &mut *self.inner.get_mut()
+    }
+
+    /// Mutably borrow a slice or element.
+    ///
+    /// This mutable borrow is unchecked and callers must ensure that no other
```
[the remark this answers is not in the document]
rinon: This used to be documenting only the release version of the method, corrected.

src/disjoint_mut.rs
rinon: Good idea, I'll write some tomorrow.

rinon: Nope, I take it back. Wrote one and realized that we panic in a destructor which is unrecoverable. It's fine, I don't think we should spend more time here.

kkysen: Ahh. Yeah if tests are tricky it's not worth it.

PR #891, opened 2024-03-22
Reset debug to defaults

thedataking:
Fixes situations where debug builds wouldn't link without dead code elimination and resets debug builds to the defaults.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,13 +45,6 @@ asm = []
 bitdepth_8 = []
 bitdepth_16 = []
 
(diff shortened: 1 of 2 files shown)


randomPoison (approved):
Ah cool, I was hoping your `bpc_fn!` approach could be used to fix the dead code elimination requirement. Looks good to me!

kkysen:
Oh, I already had a mostly complete fix for this from a while ago that I was waiting on #683 to finish.  This change itself looks good, though, so it's good enough probably.

PR #892, opened 2024-03-22
`struct Rav1dTransferCharacteristics`: make a newtype wrapper

folkertdev:
https://aomediacodec.github.io/av1-spec/av1-spec.pdf#page=129


folkertdev:
it does yes

randomPoison (approved):

PR #897, opened 2024-03-23
`fn dav1d_parse_sequence_header`: backport simplification from dav1d 1.2.1

fbossen:
Avoid creating an entire decoder instance to just parse a sequence header.

### DIFF
--- a/src/getbits.rs
+++ b/src/getbits.rs
@@ -190,12 +190,10 @@ impl<'a> GetBits<'a> {
         self.index * u8::BITS as usize - self.bits_left as usize
     }
 
-    #[inline]
     pub const fn byte_pos(&self) -> usize {
         self.index
     }
 
-    #[inline]
     pub const fn has_pending_bits(&self) -> bool {
         self.state != 0 || self.bits_left != 0
     }
--- a/src/getbits.rs
+++ b/src/getbits.rs
@@ -189,4 +189,14 @@ impl<'a> GetBits<'a> {
     pub const fn pos(&self) -> usize {
         self.index * u8::BITS as usize - self.bits_left as usize
     }
+
+    #[inline]
+    pub const fn byte_pos(&self) -> usize {
+        self.index
+    }
+
+    #[inline]
+    pub const fn has_pending_bits(&self) -> bool {
+        self.state != 0 || self.bits_left != 0
+    }
 }
--- a/src/obu.c
+++ b/src/obu.c
@@ -35,6 +35,7 @@
 
 #include "common/frame.h"
 #include "common/intops.h"
+#include "common/validate.h"
 
 #include "src/decode.h"
 #include "src/getbits.h"
@@ -44,13 +45,13 @@
 #include "src/ref.h"
 #include "src/thread_task.h"
 
-static int parse_seq_hdr_error(Dav1dContext *const c) {
-    dav1d_log(c, "Error parsing sequence header\n");
+static int parse_seq_hdr_error(void) {
     return DAV1D_ERR(EINVAL);
 }
 
-static int parse_seq_hdr(Dav1dContext *const c, GetBits *const gb,
-                         Dav1dSequenceHeader *const hdr)
+static NOINLINE int parse_seq_hdr(Dav1dSequenceHeader *const hdr,
+                                  GetBits *const gb,
+                                  const int strict_std_compliance)
 {
 #define DEBUG_SEQ_HDR 0
 
@@ -60,7 +61,7 @@ static int parse_seq_hdr(Dav1dContext *const c, GetBits *const gb,
 
     memset(hdr, 0, sizeof(*hdr));
     hdr->profile = dav1d_get_bits(gb, 3);
-    if (hdr->profile > 2) return parse_seq_hdr_error(c);
+    if (hdr->profile > 2) return parse_seq_hdr_error();
 #if DEBUG_SEQ_HDR
     printf("SEQHDR: post-profile: off=%u\n",
            dav1d_get_bits_pos(gb) - init_bit_pos);
@@ -68,7 +69,7 @@ static int parse_seq_hdr(Dav1dContext *const c, GetBits *const gb,
 
     hdr->still_picture = dav1d_get_bit(gb);
     hdr->reduced_still_picture_header = dav1d_get_bit(gb);
-    if (hdr->reduced_still_picture_header && !hdr->still_picture) return parse_seq_hdr_error(c);
+    if (hdr->reduced_still_picture_header && !hdr->still_picture) return parse_seq_hdr_error();
 #if DEBUG_SEQ_HDR
     printf("SEQHDR: post-stillpicture_flags: off=%u\n",
            dav1d_get_bits_pos(gb) - init_bit_pos);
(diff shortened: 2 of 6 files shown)


src/obu.rs
```diff
@@ -549,6 +541,47 @@ fn parse_seq_hdr(c: &mut Rav1dContext, gb: &mut GetBits) -> Rav1dResult<Rav1dSeq
...
 
+pub(crate) unsafe fn rav1d_parse_sequence_header(
+    mut data: &[u8],
+) -> Rav1dResult<DRav1d<Rav1dSequenceHeader, Dav1dSequenceHeader>> {
+    while !data.is_empty() {
+        let mut gb = GetBits::new(data);
+
+        gb.get_bit(); // obu_forbidden_bit
+        let raw_type = gb.get_bits(4);
+        let r#type = Rav1dObuType::from_repr(raw_type as usize);
```
kkysen, line 552:
```suggestion
        let r#type = Rav1dObuType::from_repr(gb.get_bits(4) as usize).unwrap();
```

src/obu.rs
```diff
@@ -549,6 +541,47 @@ fn parse_seq_hdr(c: &mut Rav1dContext, gb: &mut GetBits) -> Rav1dResult<Rav1dSeq
...
+            let len = gb.get_uleb128() as usize;
+            if len > data.len() {
+                return Err(EINVAL);
+            }
+            len
+        } else {
+            data.len()
+        };
+
+        if r#type == Some(Rav1dObuType::SeqHdr) {
```
kkysen, line 568:
```suggestion
        if r#type == Rav1dObuType::SeqHdr {
```

src/obu.rs
```diff
@@ -549,6 +541,47 @@ fn parse_seq_hdr(c: &mut Rav1dContext, gb: &mut GetBits) -> Rav1dResult<Rav1dSeq
     })
 }
 
+pub(crate) unsafe fn rav1d_parse_sequence_header(
+    mut data: &[u8],
+) -> Rav1dResult<DRav1d<Rav1dSequenceHeader, Dav1dSequenceHeader>> {
+    while !data.is_empty() {
```
kkysen, line 547:
```suggestion
    let res = Err(ENOENT);

    while !data.is_empty() {
```

src/obu.rs
```diff
@@ -549,6 +541,47 @@ fn parse_seq_hdr(c: &mut Rav1dContext, gb: &mut GetBits) -> Rav1dResult<Rav1dSeq
...
+        }
+
+        if gb.has_error() != 0 {
+            return Err(EINVAL);
+        }
+
+        data = &data[obu_end..]
+    }
+
+    Err(ENOENT)
```
kkysen, line 582:
```suggestion
    res.map(DRav1d::from_rav1d)
```

src/obu.rs
```diff
@@ -549,6 +541,47 @@ fn parse_seq_hdr(c: &mut Rav1dContext, gb: &mut GetBits) -> Rav1dResult<Rav1dSeq
...
+            }
+            len
+        } else {
+            data.len()
+        };
+
+        if r#type == Some(Rav1dObuType::SeqHdr) {
+            if let Ok(seq_hdr) = parse_seq_hdr(&mut gb, false) {
+                return Ok(DRav1d::from_rav1d(seq_hdr));
+            }
```
kkysen, line 571:
```suggestion
            let res = parse_seq_hdr(&mut gb, false);
            res?;
```

src/obu.c
```diff
@@ -283,6 +278,45 @@ static int parse_seq_hdr(Dav1dContext *const c, GetBits *const gb,
...
+        }
+
+        if (type == DAV1D_OBU_SEQ_HDR) {
+            if ((res = parse_seq_hdr(out, &gb, 0)) < 0) return res;
+            if (gb.ptr > obu_end) return DAV1D_ERR(EINVAL);
+            dav1d_bytealign_get_bits(&gb);
+        }
+
+        if (gb.error) return DAV1D_ERR(EINVAL);
+        assert(gb.state == 0 && gb.bits_left == 0);
```
kkysen, line 313: I think you missed this line.

src/obu.c
```diff
@@ -283,6 +278,45 @@ static int parse_seq_hdr(Dav1dContext *const c, GetBits *const gb,
...
+        const uint8_t *obu_end = gb.ptr_end;
+        if (has_length_field) {
+            const size_t len = dav1d_get_uleb128(&gb);
+            if (len > (size_t)(obu_end - gb.ptr)) return DAV1D_ERR(EINVAL);
+            obu_end = gb.ptr + len;
+        }
+
+        if (type == DAV1D_OBU_SEQ_HDR) {
+            if ((res = parse_seq_hdr(out, &gb, 0)) < 0) return res;
+            if (gb.ptr > obu_end) return DAV1D_ERR(EINVAL);
```
kkysen, line 308: This line needs to be ported, too.

kkysen (changes requested):

src/obu.rs
fbossen: Not sure that's the right thing to do. Not all 16 possible values of `Rav1dObuType` are defined. As per the AV1 specification, values that aren't recognized should simply be ignored. We shouldn't panic when encountering one of the undefined values.

kkysen: Oh okay, I might've misread that.

src/obu.rs
fbossen: I made `res` mutable so that it can be modified in the loop

src/obu.rs
fbossen: I removed the `let` to avoid creating a shadow variable

src/obu.c
fbossen: Those are private fields, so I added a `has_pending_bits` method

src/obu.c
fbossen: Done

fbossen:
Note sure whether the API function `dav1d_parse_sequence_header` is tested in CI. Ended up using `-s 5` when running the decoder to test.

src/obu.rs
```diff
@@ -549,6 +541,51 @@ fn parse_seq_hdr(c: &mut Rav1dContext, gb: &mut GetBits) -> Rav1dResult<Rav1dSeq
...
+            if gb.byte_pos() > obu_end {
+                return Err(EINVAL);
+            }
+            gb.bytealign();
+        }
+
+        if gb.has_error() != 0 {
+            return Err(EINVAL);
+        }
+        assert_eq!(gb.has_pending_bits(), false);
```
kkysen, line 581:
```suggestion
        assert!(!gb.has_pending_bits());
```

kkysen (approved):
LGTM!  Could you just use `assert!` instead of `assert_eq!` on a `bool`?

src/obu.rs
fbossen: Done

src/getbits.rs
```diff
@@ -189,4 +189,14 @@ impl<'a> GetBits<'a> {
     pub const fn pos(&self) -> usize {
         self.index * u8::BITS as usize - self.bits_left as usize
     }
+
+    #[inline]
```
kkysen, line 193:
```suggestion
```

We shouldn't need manual `#[inline]`s where C doesn't have them and looking at the assembly doesn't show a difference.  If it's a short `const fn`, it's basically `#[inline]` anyways, and in practice, `rustc` now makes all short functions `#[inline]` (that's why we need `#[inline(never)]` on godbolt for short functions).

src/getbits.rs
```diff
@@ -189,4 +189,14 @@ impl<'a> GetBits<'a> {
...
+    pub const fn byte_pos(&self) -> usize {
+        self.index
+    }
+
+    #[inline]
```
kkysen, line 198:
```suggestion
```

src/obu.rs
```diff
@@ -124,7 +126,10 @@ impl Debug {
     }
 }
 
-fn parse_seq_hdr(c: &mut Rav1dContext, gb: &mut GetBits) -> Rav1dResult<Rav1dSequenceHeader> {
+fn parse_seq_hdr(
```
kkysen, line 129:
```suggestion
#[inline(never)]
fn parse_seq_hdr(
```

Porting C's `NOINLINE` addition.

kkysen (changes requested):
Sorry I missed these couple things before.  We should port the `NOINLINE` C added.

kkysen (approved):

PR #898, opened 2024-03-23
`enum DrlProximity`: make a real enum

folkertdev:


src/decode.rs
```diff
@@ -2555,58 +2552,68 @@ unsafe fn decode_b_inner(
...
-            *b.drl_idx_mut() = NEAREST_DRL;
+            *b.drl_idx_mut() = DrlProximity::Nearest;
             if b.inter_mode() == NEWMV_NEWMV {
                 if n_mvs > 1 {
                     // NEARER, NEAR or NEARISH
                     let drl_ctx_v1 = get_drl_context(&mvstack, 0);
-                    *b.drl_idx_mut() += rav1d_msac_decode_bool_adapt(
+                    if rav1d_msac_decode_bool_adapt(
                         &mut ts.msac,
                         &mut ts.cdf.m.drl_bit[drl_ctx_v1 as usize],
```
folkertdev, line 2562: this addition with a bool is hard to translate. Having an `increment` function on `DrlProximity` leads to worse codegen. Instead I've basically written out what the incrementing would do. I think that's clearer, but it is kind of verbose.

src/decode.rs
```diff
@@ -2895,36 +2904,38 @@ unsafe fn decode_b_inner(
             } else {
                 has_subpel_filter = true;
                 *b.inter_mode_mut() = NEWMV;
-                *b.drl_idx_mut() = NEAREST_DRL;
+                *b.drl_idx_mut() = DrlProximity::Nearest;
                 if n_mvs > 1 {
                     // NEARER, NEAR or NEARISH
```
folkertdev, line 2909: this comment is misleading, because `NEARISH` can never actually be assigned here

src/decode.rs
randomPoison: The translation here looks correct, but I think an `increment` method would have also been fine. Even if codegen isn't optimal, I think we'd rather wait until profiling before worrying too much about the performance cost. But the version you have here is easier to understand in my opinion, so what you have now is fine.

randomPoison (approved):

src/decode.rs
kkysen: True, you can fix the comment.

kkysen (approved):
Whoops, forgot to submit these comments I made over the weekend.  I like the explicitness of the setting of the `DrlProximity`s now vs. the previous `+=`s, which were more confusing.  We can merge it once its rebased, and you can fix that wrong comment, too.

src/decode.rs
```diff
@@ -2562,58 +2559,68 @@ unsafe fn decode_b_inner(
             }
 
             let im = &dav1d_comp_inter_pred_modes[b.inter_mode() as usize];
-            *b.drl_idx_mut() = NEAREST_DRL;
+            *b.drl_idx_mut() = DrlProximity::Nearest;
             if b.inter_mode() == NEWMV_NEWMV {
                 if n_mvs > 1 {
-                    // NEARER, NEAR or NEARISH
+                    // NEARER or NEAR
```
kkysen, line 2565:
```suggestion
                    // `Nearer` or `Near`
```

src/levels.rs
```diff
@@ -196,11 +196,13 @@ pub const GLOBALMV: InterPredMode = 2;
...
-pub const NEAR_DRL: DRL_PROXIMITY = 2;
-pub const NEARER_DRL: DRL_PROXIMITY = 1;
-pub const NEAREST_DRL: DRL_PROXIMITY = 0;
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, FromRepr)]
+pub enum DrlProximity {
+    Nearest = 0,
+    Nearer = 1,
+    Near = 2,
+    Nearish = 3,
+}
```
kkysen, line 205:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
pub enum DrlProximity {
    Nearest,
    Nearer,
    Near,
    Nearish,
}
```

We don't actually use `from_repr`.

kkysen (changes requested):

kkysen (approved):
LGTM, just needs a rebase.

PR #899, opened 2024-03-23
`enum BlockSize`: make a real enum

folkertdev:


src/tables.rs
```diff
@@ -672,47 +655,47 @@ pub static dav1d_filter_mode_to_y_mode: [u8; 5] = [
...
-    | 1 << BS_16x8
-    | 1 << BS_16x4
-    | 1 << BS_8x32
-    | 1 << BS_8x16
-    | 1 << BS_8x8
-    | 1 << BS_8x4
-    | 1 << BS_4x16
-    | 1 << BS_4x8
-    | 1 << BS_4x4
+    | 1 << BlockSize::Bs32x32 as u8
```
randomPoison, line 663: I think for readability it'd be better to glob import the `BlockSize` variants at the top of the and then use the variants directly, like what you did for `dav1d_block_sizes`. Same applies to `wedge_allowed_mask` and `interintra_allowed_mask`.

randomPoison (approved):
Just one minor style request, but changes look good 👍

src/tables.rs
folkertdev: did that now, but it does format to something slightly uglier

randomPoison: Could you stick a `#[rustfmt::skip]` on there? If not I leave it up to you to decide which approach you like better since it's just a style issue.

folkertdev: nah it's allright, it's fine this way

kkysen: It doesn't really matter whichever way it is, but that's why I didn't suggest using `use BlockSize::*` for those.  We could also put all of the `Bs`s in an array (without the `as u8`s) and then do a `const_for!` loop to `|` them together.  It's fine as is, though, so I'll merge it.

PR #902, opened 2024-03-25
`struct Rav1dTileState_frame_thread::cf`: Make into offset

randomPoison:
To avoid borrow checker issues I modified `decode_coefs` to use a new `CfSelect` enum to determine which of the two possible coef bufs to use. I also added coef versions of `pxstride` and `cast_pixel_slice` so that I could correctly convert between byte offsets and coef offsets.

I also had to make a couple of functions take a `&mut Rav1dFrameData` (previously they were taking a shared reference) in order to get mutable access to the coef buffers. @rinon does this conflict with your current work?

### DIFF
--- a/include/common/bitdepth.rs
+++ b/include/common/bitdepth.rs
@@ -12,7 +12,8 @@ use std::ops::Div;
 use std::ops::Mul;
 use std::ops::Rem;
 use std::ops::Shr;
-use std::slice;
+use zerocopy::AsBytes;
+use zerocopy::FromBytes;
 
 pub trait FromPrimitive<T> {
     fn from_prim(t: T) -> Self;
@@ -104,12 +105,26 @@ impl BPC {
         }
     }
 
+    /// Converts a value in bytes to a value in pixel units.
+    ///
     /// `T` is generally meant to be `usize` or `isize`.
     pub fn pxstride<T>(&self, n: T) -> T
     where
         T: Copy + Eq + From<u8> + Div<Output = T> + Rem<Output = T>,
     {
         let scale = (self.bitdepth() / 8).into();
+        debug_assert!(n % scale == 0.into());
+        n / scale
+    }
+
+    /// Converts a value in bytes to a value in coef units.
+    ///
+    /// `T` is generally meant to be `usize` or `isize`.
+    pub fn coef_stride<T>(&self, n: T) -> T
+    where
+        T: Copy + Eq + From<u8> + Div<Output = T> + Rem<Output = T>,
+    {
+        let scale = (self.bitdepth() / 4).into();
         debug_assert!(n % scale == 0.into());
         n / scale
     }
@@ -134,7 +149,9 @@ pub trait BitDepth: Clone + Copy {
         + FromPrimitive<c_uint>
         + ToPrimitive<i16>
         + ToPrimitive<c_int>
-        + ToPrimitive<c_uint>;
+        + ToPrimitive<c_uint>
+        + FromBytes
+        + AsBytes;
 
     type Coef: Copy
         + FromPrimitive<c_int>
@@ -142,6 +159,8 @@ pub trait BitDepth: Clone + Copy {
         + ToPrimitive<c_int>
         + ToPrimitive<c_uint>
         + Add<Output = Self::Coef>
+        + FromBytes
+        + AsBytes
         + Display;
 
     type Entry: Copy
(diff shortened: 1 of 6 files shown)


randomPoison:
Oh, also the first commit doesn't compile here. I split this up into two commits to more easily roll back my changes if I hit issues. We should squash this instead of merging it to avoid having a non-compiling commit in our history.

include/common/bitdepth.rs
```diff
@@ -113,6 +115,56 @@ impl BPC {
...
+    // size and that the alignment is correct for the new element type.
+    debug_assert!(bytes.len() % size == 0);
+    assert!(bytes.as_ptr() as usize % mem::align_of::<T>() == 0);
+
+    let len = bytes.len() / size;
+
+    // SAFETY: We've checked that alignment and the number of elements is
+    // correct, caller is responsible for ensuring target type is valid.
+    unsafe { slice::from_raw_parts_mut(bytes.as_mut_ptr().cast(), len) }
 }
```
kkysen, line 168: Can we replace these with [`zerocopy`](https://docs.rs/zerocopy)'s [`FromBytes::slice_from`](https://docs.rs/zerocopy/latest/zerocopy/trait.FromBytes.html#method.slice_from) and [`FromBytes::slice_from_mut`](https://docs.rs/zerocopy/latest/zerocopy/trait.FromBytes.html#method.slice_from_mut)?  For the 4 functions ([pixel, coef] x [&, &mut]), we can just use `FromBytes::slice_from(_mut)(bytes).unwrap()`.

This way `zerocopy` can do all of the safety checking for us here, and it does it with traits, which is a bit safer.  It's also written by google, so it should be fine as a dependency.  And I checked that these functions are zero cost abstractions.  We should also store the coef buffers as `i16`s instead of `u8`s, since they're either `i16` or `i32`, and this way the cast to `i16` is a no-op.  With our approach, we'd have to write `cast_i16_slice(_mut)` functions for that, but with `zerocopy`, we get that generically for free.

If you want, I can put this up as a separate PR.

kkysen (changes requested):

include/common/bitdepth.rs
```diff
@@ -113,6 +115,56 @@ impl BPC {
         debug_assert!(n % scale == 0.into());
         n / scale
     }
+
+    /// Converts a value in bytes to a value in coef units.
+    ///
+    /// `T` is generally meant to be `usize` or `isize`.
+    pub fn coefstride<T>(&self, n: T) -> T
```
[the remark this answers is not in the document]
randomPoison: Yeah that's fine, I almost did that initially and I don't mind the naming being slightly different than `pxstride`.

include/common/bitdepth.rs
randomPoison: I think using zerocopy for the slice transmutes makes sense, I've added that. I don't think making the buffer `i16`s makes sense though. The cast should already be functionally a no-op, or at least very cheap, and I'd like to stick to the pattern of having byte buffers that either get cast to `Pixel` or `Coef` buffers.

include/common/bitdepth.rs
```diff
@@ -113,6 +115,56 @@ impl BPC {
         debug_assert!(n % scale == 0.into());
         n / scale
     }
+
+    /// Converts a value in bytes to a value in coef units.
```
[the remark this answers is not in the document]
randomPoison: I don't think this wording is any more clear than what I already have.

include/common/bitdepth.rs
kkysen: The cast to a larger type has a branch and panicking code, though it's still quite cheap.  The bpc8 casts are no-ops, though, when using the generic code if the target type is the same.  In general, though, I think it's clearer if we had a `type MinPixel = BitDepth8::Pixel` and `type MinCoef = BitDepth8::Coef` that we used instead of `u8` for these buffers.  Then it's clear that it's pixels or coefs vs. just some other `u8`s.

kkysen: Maybe we can make the definition of `Dyn{Pixel,Coef}` that.

src/recon.rs
```diff
@@ -1584,6 +1589,15 @@ unsafe fn decode_coefs<BD: BitDepth>(
...
 }
 
+#[derive(Clone, Copy)]
+enum CfSelect {
+    // Use `f.frame_thread.cf` at the specified offset.
+    Frame(usize),
+
+    // Use `t.cf`.
+    Task,
+}
```
kkysen, line 1599: I don't want to nitpick much, so no change necessary now, but in the future, I think it'd be better to make these things doc comments.

src/recon.rs
```diff
@@ -1688,23 +1702,16 @@ unsafe fn read_coef_tree<BD: BitDepth>(
...
         let mut cf_ctx: u8 = 0;
         let eob;
-        let cf: *mut BD::Coef;
+        let cf;
         let mut cbi_idx = 0;
         if (*t).frame_thread.pass != 0 {
             let p = (*t).frame_thread.pass & 1;
-            if ((*ts).frame_thread[p as usize].cf).is_null() {
-                unreachable!();
-            }
```
kkysen, line 1697: Is this check important?  We store `0` instead now, and so this null check is gone now.

kkysen (approved):

include/common/bitdepth.rs
kkysen: Could we just open an issue to do this later at some point?

kkysen (approved):
LGTM, just make sure to squash merge as you said as there was a non-compiling commit.

include/common/bitdepth.rs
randomPoison: Sure, open an issue to look into it. It's probably only worth doing after profiling if we find that the cast is causing performance issues, but can't hurt to track the potential improvement.

src/recon.rs
randomPoison: I couldn't leave the check in because sometimes the index is 0. Given that the code was running successfully without hitting the check before, I think we can be confident that null ptr case wasn't happening in practice. If we're really worried about that happening and feel that we need the assert then I can wrap the field in `Option` that way we have to check if the index is valid, but personally I don't think that check makes as much sense once we're using an offset.

include/common/bitdepth.rs
kkysen: * Tracked in #932.

PR #909, opened 2024-03-26
`struct Av1Restoration`: Replace `Cell` with `RwLock`

rinon:
`Rav1dFrameContext_lf::lr_mask` is shared across threads, and therefore must be a thread-safe `Sync` type. This changes replaces an existing usage of `Cell` to provide interior mutability with non-blocking usage of `RwLock`.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -3748,15 +3748,14 @@ unsafe fn setup_tile(
             &f.lf.lr_mask[sb_idx as usize].lr[p][unit_idx as usize]
         };
 
-        let lr = lr_ref.get();
-        let lr = Av1RestorationUnit {
+        let mut lr = lr_ref.try_write().unwrap();
+        *lr = Av1RestorationUnit {
             filter_v: [3, -7, 15],
             filter_h: [3, -7, 15],
             sgr_weights: [-32, 31],
-            ..lr
+            ..*lr
         };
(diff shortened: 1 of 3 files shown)


src/decode.rs
```diff
@@ -3989,14 +3989,13 @@ unsafe fn setup_tile(
...
-        let lr = lr_ref.get();
+        let lr = lr_ref.try_write().unwrap();
         let lr = Av1RestorationUnit {
             filter_v: [3, -7, 15],
             filter_h: [3, -7, 15],
             sgr_weights: [-32, 31],
-            ..lr
+            ..*lr
         };
-        lr_ref.set(lr);
```
randomPoison, line 3999: Why are we no longer writing the value to `lr_ref`? It looks like the original was actually updating the data in `lr_ref`, but the new version doesn't.

kkysen (commented):
Can this be an atomic?  I see a way that's not too hard that I could write up.

kkysen:
> Can this be an atomic? I see a way that's not too hard that I could write up.

Moved this to #913.

src/decode.rs
rinon: Good catch. It doesn't seem to affect our tests, but this is definitely wrong. Should be fixed now.

randomPoison (approved):

PR #911, opened 2024-03-26
`fn get_cur_frame_segid`: Remove negative offsets

kkysen:
`fn get_cur_frame_segid` is always called with a whole slice (or will be once segmaps are `Arc`anized), so we only have to worry about negative offsets within this `fn`.  I think this should work for negative strides?

Actually `stride` can't be negative here.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -607,29 +607,23 @@ pub unsafe fn get_cur_frame_segid(
     let negative_adjustment = have_left as usize + have_top as usize * stride.unsigned_abs();
     cur_seg_map = cur_seg_map
         .add((bx as isize + by as isize * stride - negative_adjustment as isize) as usize);
-    if have_left && have_top {
-        let l = *cur_seg_map.add(stride.unsigned_abs());
-        let a = *cur_seg_map.add(1);
-        let al = *cur_seg_map.add(0);
-        let seg_ctx = if l == a && al == l {
-            2
-        } else if l == a || al == l || a == al {
-            1
-        } else {
-            0
-        };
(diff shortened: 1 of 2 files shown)


src/env.rs
```diff
@@ -604,30 +604,26 @@ pub unsafe fn get_cur_frame_segid(
...
-            *cur_seg_map.offset(-1)
-        } else if have_top {
-            *cur_seg_map.offset(-stride as isize)
-        } else {
-            0
-        };
-        (seg_id, seg_ctx)
+    let negative_adjustment = have_left as usize + have_top as usize * stride.unsigned_abs();
+    cur_seg_map = cur_seg_map
+        .add((bx as isize + by as isize * stride - negative_adjustment as isize) as usize);
```
kkysen, line 609: @fbossen, when `stride` is negative, is `by` also going to be negative?  Because if `by` is positive and `stride` is negative, this `add` is probably going to be negative (ignoring the `as usize` at the end).  But `fn get_cur_frame_segid` is always called with `f.cur_segmap`, which is always the start of an allocation, so it can't be negatively offset.  So is `stride` always positive here, or is `by` negative when `stride` is negative?

fbossen: Looks like `stride` can't be negative here. The function is called with `stride =  f.b4_stride = (f.bw + 31 & !31) as ptrdiff_t;` (`f.bw` is positive).

kkysen: Awesome!  Good to know `f.bw` is positive.  That helps a lot.  I'll simply things then.

randomPoison (approved):

PR #912, opened 2024-03-26
Abort on panic

randomPoison:
Currently tests hang if a thread panics, meaning that if a lot of tests panic CI can hang for a long time. For testing purposes we just want to abort on panic to ensure we get quick test output in those cases.


rinon (approved):

kkysen (commented):
Do we want to do this for debug, too?

randomPoison:
I added it to the debug profile as well.

Issue #822, opened 2024-03-15
`struct refmvs_frame`: Make fields safer without changing layout


randomPoison:
Also cleanup `rav1d_refmvs_init` and `rav1d_refmvs_clear`.

- [x] `frm_hdr` - Pointer to `f.frame_hdr`. We may be able to pass `frame_hdr` into the places where it's used. `frm_hdr` is only used from `rav1d_refmvs_find` and `add_temporal_candidate`, which is only called from `rav1d_refmvs_find`. So if we can pass `f.frame_hdr` into `rav1d_refmvs_find` and then down into `add_temporal_candidate` we should be able to remove this field.  Since `frm_hdr` is not used by `load_tmvs_c` (and thus not accessed by asm), we should change it to a `*const ()`, signifying that it should never be used while keeping the layout the same.  It is dangerous to keep it as a `*const Rav1dFrameHeader`, since it's a `*const Dav1dFrameHeader` in C.  And passing the `&Rav1dFrameHeader` through the callstack removes the lifetime issue as well.
- [x] `rp` - Points to `f.mvs`.  `rp` is also not accessed in `load_tmvs_c`, so hopefully we can change it (in a layout preserving manner).

fbossen:
There may be a conflict with #821 where a pointer to a `struct refmvs_frame` is passed to an asm function.

randomPoison:
Ah, that's a good point. Even before #821 `refmvs_frame` is being passed into asm functions, so I'm not sure we can actually make changes to it safely. The sister struct `refmvs_tile` seems to only be used from Rust so I think we can still clean that one up.

kkysen:
I think we should do part of this.  At least, `frame_hdr` should be made a `*const ()` and passed separately as a reference through the callstack as @randomPoison proposed above.  It seems dangerous to leave a `*const Rav1dFrameHeader` when C has a `*const Dav1dFrameHeader`.  We can make `refmvs_frame` safer; we just can't change its layout at all, and we can't fields that are accessed in `load_tmvs_c`.  See also https://github.com/memorysafety/rav1d/pull/821#discussion_r1529542020.

kkysen:
* Fixed by #983.

kkysen:
* Fixed by #983.

PR #916, opened 2024-03-26
`struct refmvs_frame::frm_hdr`: Replace raw ptr with `*const ()`

kkysen:
* Fixes `frm_hdr` field of #822.

This removes the `*const Rav1dFrameHeader` in the field, passing it through the call stack.  But we can't change the field's layout since `*const refmvs_frame` is passed to asm, so we make the field `*const ()` and `ptr::null()`. This also fixes the problem of accidentally conflating `*const Rav1dFrameHeader` and `*const Dav1dFrameHeader` in that field.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -1449,6 +1449,8 @@ unsafe fn decode_b_inner(
         && (bw4 > ss_hor || t.bx & 1 != 0)
         && (bh4 > ss_ver || t.by & 1 != 0);
 
+    let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
+
     if t.frame_thread.pass == 2 {
         if b.intra != 0 {
             bd_fn.recon_b_intra(f, t, bs, intra_edge_flags, b);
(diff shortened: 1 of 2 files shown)


src/decode.rs
```diff
@@ -2251,10 +2254,11 @@ unsafe fn decode_b_inner(
                 }
             }
         }
-        if f.frame_hdr().frame_type.is_inter_or_switch() || f.frame_hdr().allow_intrabc {
+        let frame_hdr = f.frame_hdr();
```
randomPoison, line 2257: It looks like you're redeclaring `frame_hdr` a few times in this function. Was that for borrowing reasons, i.e. to avoid holding a reference when we need to pass `&mut f` to a function?

randomPoison (approved):
Changes look good, I just had a minor question about redeclaring a variable.

src/decode.rs
kkysen: Yeah, I think I've minimized the number of redeclarations of `frame_hdr` that still works with the borrow checker now (in this `fn`).  In some places we pass a `&mut f`, so we need this, and then in other contexts, the `&mut f` is over, so a previous `frame_hdr` works.

Issue #845, opened 2024-03-18
`refmvs.rs`: Unsafe cleanup


randomPoison:
There are ~350 unsafe operations happening in `refmvs.rs`.

- [ ] Cleanup unsafe operations happening within functions.
- [ ] Mark any functions that can be made fully safe as such.
- [ ] Add `unsafe` blocks around any remaining unsafe operations along with standard safety documentation.
- [ ] Add safety documentation for functions that need to remain unsafe.
- [ ] Add `#![deny(unsafe_op_in_unsafe_fn)]` to file.

randomPoison:
The functionality in `refmvs.rs` has already been made as safe as possible. The last couple of operations are in Rust fallback functions, so I'm going to close this.

Issue #820, opened 2024-03-15
Cleanup c2rust output in `refmvs.rs`


randomPoison:
A lot of the code in `refmvs.rs` has already been cleaned up, but there are still places where we're using raw pointers that could be converted to safe references. A good chunk of the raw pointer operations in here come from the pointers in `refmvs_tile` and `refmvs_frame`, see #706 for issues tracking cleanup of those structs.

* [ ] Cleanup helper functions to take safe arguments where possible (may take raw pointers as arguments currently).
* [ ] General cleanup of c2rust generated code in functions.
* [ ] Restore comments from original C.

The following functions still need to be cleaned:

- [x] `scan_row`
- [x] `scan_col`
- [x] `rav1d_refmvs_save_tmvs`
- [x] `rav1d_refmvs_tile_sbrow_init`
- [x] `load_tmvs_c`
- [x] `save_tmvs_c`
- [x] `rav1d_refmvs_init_frame`

PR #920, opened 2024-03-27
`fn rav1d_refmvs_save_tmvs`: Cleanup

kkysen:
* Fixes part of #820.
* Fixes part of #845.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4281,7 +4281,7 @@ pub(crate) unsafe fn rav1d_decode_tile_sbrow(
     {
         rav1d_refmvs_save_tmvs(
             &c.refmvs_dsp,
-            &mut t.rt,
+            &t.rt,
             ts.tiling.col_start >> 1,
             ts.tiling.col_end >> 1,
             t.by >> 1,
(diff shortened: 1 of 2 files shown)


src/decode.rs
```diff
@@ -4802,7 +4802,8 @@ unsafe fn rav1d_decode_frame_main(c: &Rav1dContext, f: &mut Rav1dFrameData) -> R
                 rav1d_decode_tile_sbrow(c, &mut t, f).map_err(|()| EINVAL)?;
             }
             if f.frame_hdr().frame_type.is_inter_or_switch() {
-                rav1d_refmvs_save_tmvs(&c.refmvs_dsp, &mut t.rt, 0, f.bw >> 1, t.by >> 1, by_end);
+                let by = t.by; // For borrowck.
+                rav1d_refmvs_save_tmvs(&c.refmvs_dsp, &mut t.rt, 0, f.bw >> 1, by >> 1, by_end);
```
kkysen, line 4806: I have no idea why I have to do this, since it works fine in the other `rav1d_refmvs_save_tmvs` call when we pass `&mut t.rt` and `t.by > 1`, too.

kkysen: Still have no idea why this happened, but I removed it another way (`&mut t.rt` isn't necessary, only a `&` is needed).

src/refmvs.rs
```diff
@@ -1136,31 +1136,27 @@ pub(crate) unsafe fn rav1d_refmvs_find(
...
-    row_end8 = cmp::min(row_end8, (*rf).ih8);
-    col_end8 = cmp::min(col_end8, (*rf).iw8);
-    let stride: ptrdiff_t = (*rf).rp_stride;
-    let ref_sign: *const u8 = ((*rf).mfmv_sign).as_ptr();
-    let rp: *mut refmvs_temporal_block = (*rf).rp.offset(row_start8 as isize * stride);
-
-    (*dsp).save_tmvs.expect("non-null function pointer")(
+    let rf = &*rt.rf;
+    assert!(row_start8 >= 0);
+    assert!((row_end8 - row_start8) as c_uint <= 16);
```
randomPoison, line 1148: I think the cast here is redundant:

```suggestion
    assert!((row_end8 - row_start8) <= 16);
```

randomPoison (approved):

src/refmvs.rs
kkysen: C deliberately does the unsigned cast here.  I think they're folding in an `assert!(_ >= 0)` here.

PR #924, opened 2024-03-27
`type save_tmvs_fn`: Make `rr`, `ref_sign` args array ptrs

kkysen:
Since we pass them as pointers, we can change `*const T` args to `*const [T; N]` if we're always passing an array, as their representation is the same and so asm will treat them the same.  But for the fallbacks, we have less unsafety since we get an array instead of just a pointer with unknown length.

### DIFF
--- a/src/refmvs.rs
+++ b/src/refmvs.rs
@@ -20,6 +20,7 @@ use std::cmp;
 use std::ffi::c_int;
 use std::ffi::c_uint;
 use std::ffi::c_void;
+use std::mem;
 use std::ptr;
 
 #[cfg(feature = "asm")]
@@ -38,7 +39,7 @@ extern "C" {
     fn dav1d_save_tmvs_ssse3(
         rp: *mut refmvs_temporal_block,
         stride: ptrdiff_t,
-        rr: *const *const refmvs_block,
+        rr: *const [*const refmvs_block; 31],
         ref_sign: *const [u8; 7],
         col_end8: c_int,
         row_end8: c_int,
@@ -80,7 +81,7 @@ extern "C" {
     fn dav1d_save_tmvs_avx2(
         rp: *mut refmvs_temporal_block,
         stride: ptrdiff_t,
-        rr: *const *const refmvs_block,
+        rr: *const [*const refmvs_block; 31],
         ref_sign: *const [u8; 7],
         col_end8: c_int,
         row_end8: c_int,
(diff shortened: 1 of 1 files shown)


src/refmvs.rs
```diff
@@ -1154,16 +1154,12 @@ pub(crate) unsafe fn rav1d_refmvs_save_tmvs(
     let stride = rf.rp_stride;
     let ref_sign = &rf.mfmv_sign;
     let rp = rf.rp.offset(row_start8 as isize * stride);
+    let rr = (&rt.r[37 - 31..][..31] as *const [_]).cast::<[_; 31]>();
+    // SAFETY: We're just slicing an array to one with length 31 and casting `*mut` to `*const`.
+    let rr = unsafe { &*rr };
```
randomPoison, line 1159: The logic for doing the slicing here is harder to follow than I would like, specifically the `&rt.r[37 - 31..][..31]` part. Doing `&rt.r[6..]` would be simpler and be closer to the original logic of `offset(6)`.

Also do we need the unsafe `&*rr`? We're passing `rr` as a pointer into `save_tmvs`, so I don't see why we need to dereference the pointer first just to have it be implicitly cast back to a pointer.

kkysen: I had `[6..][..31]` originally, but I thought it'd be slightly more brittle if they array length changed in the future and more clear this way that we're going from an array with length 37 to one with length 31.  I can change it back to `[6..]`, though, if you think that's clearer.

Also, you're right about the `unsafe`.  I'll remove it here.

kkysen: Actually, for the `unsafe`, I thought it makes the `unsafe { &*rr }` in the fallback `fn` more trivial, since you can justify it by just saying it came from a reference that decayed into a pointer.  Putting this justification on the `unsafe` in the fallback `fn` doesn't seem ideal since they are far apart.

randomPoison: Yeah, switch the logic to `[6..]`, we don't really expect these fields to change so we don't have to be so defensive against things like the array length changing.

wrt the unsafe logic, I think it makes more sense to put the justification in the fallback function. That logic has to be unsafe regardless since it's going from pointer to reference, and the place where the pointer is created is always going to be far from the place it's used. I think putting the dereference operation here just adds clutter without buying safety or clarity.

kkysen: I switched it to `[6..]`, but I think it's better to keep the `unsafe` safety justification here.  If we do things this way, all the pointers passed to fn ptrs will be valid.  Otherwise, we can't make that simple assumption for the `unsafe`s, and I don't think it's good to have to reason about each one individually.

kkysen: I figured out how to do the array slice without `unsafe`, so that part's not an issue anymore.

randomPoison (approved):

PR #927, opened 2024-03-27
`enum InterPredMode` and `enum CompInterPredMode`: make real enums

folkertdev:


src/levels.rs
```diff
@@ -204,16 +197,36 @@ pub enum DrlProximity {
...
+impl From<InterPredMode> for CompInterPredMode {
+    fn from(value: InterPredMode) -> Self {
+        match value {
+            InterPredMode::Nearest => CompInterPredMode::NearestNearest,
+            InterPredMode::Near => CompInterPredMode::NearNear,
+            InterPredMode::Global => CompInterPredMode::NearestNew,
+            InterPredMode::New => CompInterPredMode::NewNearest,
+        }
+    }
+}
```
folkertdev, line 229: this conversion happens in practice, though I don't really understand why it makes sense. Maybe the `CompInterPredMode` field is just implicitly a `InterPredMode` in certain circumstances? or there is something more subtle going on.

src/levels.rs
```diff
@@ -204,16 +197,36 @@ pub enum DrlProximity {
...
-pub const NEWMV_NEARESTMV: CompInterPredMode = 3;
-pub const NEARESTMV_NEWMV: CompInterPredMode = 2;
-pub const NEARMV_NEARMV: CompInterPredMode = 1;
-pub const NEARESTMV_NEARESTMV: CompInterPredMode = 0;
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum InterPredMode {
+    Nearest = 0,
+    Near = 1,
+    Global = 2,
+    New = 3,
```
folkertdev, line 205: I've kept the number here for now because the ordering is important for the mapping to `CompInterPredMode`

src/levels.rs
folkertdev: but e.g. `Global` mapping to `NearestNew` is weird when `GlobalGlobal` exists.

kkysen: That is very weird.  I have no idea why that's done.  It looks like a correct translation of the C, though.  @fbossen, have any idea?

Maybe the same fields are being used for the two `enum`s, but never at the same time?  I'm curious if the tests would still pass if we changed the conversion to what would be expected.

kkysen (changes requested):

src/levels.rs
folkertdev: there is this one expression that mixes the two types

```rust
(b.inter_mode() == if is_comp { GLOBALMV_GLOBALMV } else { GLOBALMV }) as c_int;
```

so, I guess, if `is_comp` is true then `CompInterPredMode` is used, and otherwise just `InterPredMode`. That at least intuitively makes sense here, and also seems to be true in this file.

Not sure what to do with that though, `is_comp` is a runtime value so it's hard encode that into the type. Some `enum { Comp(_), NotComp(_) }` enum also doesn't seem great (but would provide some extra type safety).

PR #931, opened 2024-03-28
`fn decode_b_inner`: Fix borrow checker error

randomPoison:
Fixes breakage introduced by https://github.com/memorysafety/rav1d/pull/902

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -1446,8 +1446,7 @@ unsafe fn decode_b_inner(
     let has_chroma = f.cur.p.layout != Rav1dPixelLayout::I400
         && (bw4 > ss_hor || t.bx & 1 != 0)
         && (bh4 > ss_ver || t.by & 1 != 0);
-
-    let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
+    let frame_type = f.frame_hdr.as_ref().unwrap().frame_type;
 
     if t.frame_thread.pass == 2 {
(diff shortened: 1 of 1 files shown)


src/decode.rs
```diff
@@ -1446,8 +1446,12 @@ unsafe fn decode_b_inner(
...
         && (bw4 > ss_hor || t.bx & 1 != 0)
         && (bh4 > ss_ver || t.by & 1 != 0);
-
-    let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
+    let is_inter_or_switch = f
+        .frame_hdr
+        .as_ref()
+        .unwrap()
+        .frame_type
+        .is_inter_or_switch();
```
kkysen, line 1454: Could you just copy out `frame_type`?  Since `frame_type` is used elsewhere, too, and is `Copy`.  It'd probably help more to avoid other borrow checker errors in the future vs. just doing `is_inter_or_switch`.

randomPoison: Sure that also works. I also updated one other place that was using `is_inter_or_switch`.

kkysen: It looks there's still borrow checker errors, from another use of `frame_hdr.frame_type`, so `Copy`ing out the whole `frame_type` would fix that I think.  Also, there seem to be unused import warnings so if you want to fix these here, too, feel free to.

kkysen (approved):

PR #940, opened 2024-03-28
`fn rav1d_recon_b_inter`: Cleanup

kkysen:


src/recon.rs
```diff
@@ -3535,27 +3533,25 @@ pub(crate) unsafe fn rav1d_recon_b_inter<BD: BitDepth>(
...
-                        .r#ref
-                        .r#ref[0]
-                        > 0;
+                    is_sub8x8 &= (*r[0].offset((t.b.x - 1) as isize)).0.r#ref.r#ref[0] > 0;
                 }
-            }
+                r
+            } else {
+                // Never used, because `&=` can only turn `false` into `true`, not vice versa.
+                &[]
```
randomPoison, line 3543: I'm not sure I understand this comment. From the code I can see that if `is_sub8x8 == false` then the branch that uses `r` will never be taken, so the fact that the empty slice will never be used makes sense. But the comment here seems backwards to me, i.e. from the perspective of the left-hand value `&=` can turn true into false, but not the other way around. I think it'd be clearer to word the comment something like "Never used because `r` is only used if `is_sub8x8 == true`", though honestly I don't think a comment is even necessary here. I don't want to spend time hashing out the wording of the comment though, so I'll leave it up to you to decide if it's worth changing.

randomPoison (approved):

src/recon.rs
kkysen: Oh I might've messed up the true/false.  If you thought it was clear enough I'll just shorten it to "never used" then.

PR #943, opened 2024-04-01
`rust-toolchain.toml`: Upgrade to `nightly-2024-04-01`

### LINKED ISSUE #623: Replace `cstr!` with `c""` literals when 1.77 is released
rust-lang/rust#117472 stabilized `c""` string literals and is supposed to be released with 1.77, so when that comes out, we can update our toolchain, use `c""` string literals instead of the `cstr!` macro and the `cstr` crate.

See https://github.com/memorysafety/rav1d/pull/619#issuecomment-1849024403.

kkysen:
`rust-analyzer` has been showing me a ton of warnings lately, things like `non_camel_case_types`, which we allow for `rustc`, and which I realized is due to the `rust-analyzer` version being new, and our `rustc` toolchain being old.  Upgrading it to the latest nightly fixes this.  These warnings are especially annoying as they slow down my edit and compile cycle, and we are due for an upgrade anyways, though hopefully we can go to `stable` next (once #620 is merged, though it hasn't been enough of a priority).

We can also replace the `target_arch = "arm"`-dependent `#![feature(stdsimd)]` with the much more specific and more likely to stabilize soon `#![feature(stdarch_arm_feature_detection)]`.

* Fixes #623.

### DIFF
--- a/include/common/bitdepth.rs
+++ b/include/common/bitdepth.rs
@@ -239,7 +239,7 @@ pub trait BitDepth: Clone + Copy {
         Self::Pixel::mut_slice_from(bytes).unwrap()
     }
 
-    fn cast_coef_slice(bytes: &[u8]) -> &[Self::Coef] {
+    fn _cast_coef_slice(bytes: &[u8]) -> &[Self::Coef] {
         Self::Coef::slice_from(bytes).unwrap()
     }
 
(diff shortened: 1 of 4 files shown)


randomPoison (approved):
Thanks for fixing this. I've also noticed the warnings from rust-analyzer, will be nice to have those cleaned up.

PR #947, opened 2024-04-02
`struct TxLpfRightEdge`: Add inner mutability

rinon:
Adds inner mutability to the right edge split vector by using DisjointMut and narrower ranges for reading and writing data.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -92,8 +92,10 @@ use std::ffi::c_uint;
 use std::mem;
 use std::ops::Add;
 use std::ops::AddAssign;
+use std::ops::Deref;
 use std::ops::Index;
 use std::ops::IndexMut;
+use std::ops::Range;
 use std::ops::Sub;
 use std::ptr;
 use std::sync::atomic::AtomicBool;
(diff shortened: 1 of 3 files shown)


randomPoison (approved):

src/internal.rs
```diff
@@ -505,27 +507,49 @@ pub struct Rav1dFrameContext_frame_thread {
...
-        self.inner.split_at(mid)
+        assert!(index_y.end <= mid);
+        let (uv_start, uv_end) = (index_uv.start + mid, index_uv.end + mid);
+        (
+            self.inner.index(index_y),
+            self.inner.index(uv_start..uv_end),
+        )
+    }
+
+    pub fn copy_from_slice_y(&self, index: Range<usize>, src: &[u8]) {
```
kkysen, line 542: Don't all of these methods, `fn copy_from_slice_{y,uv}` and `fn get`, need to be `unsafe`?  Their safeness relies on the specific `index`es passed.  It's clear that `y` and `uv` are non-overlapping (statically known), but within `y` and `uv`, the `index`es could overlap, so these all need to be `unsafe`, right?

src/internal.rs
```diff
@@ -505,27 +507,49 @@ pub struct Rav1dFrameContext_frame_thread {
...
 }
 
 impl TxLpfRightEdge {
     #[allow(dead_code)]
-    pub const fn new() -> Self {
-        Self { inner: Vec::new() }
+    pub fn new() -> Self {
+        Self {
+            inner: DisjointMut::default(),
+        }
```
kkysen, line 518: The `fn new` here was specifically a `const fn new`, so if we're not gonna make it `const`, there's no point in having it, as we already `#[derive(Default)]`.

kkysen (changes requested):

PR #949, opened 2024-04-02
`struct Rav1dFrameData`: Add inner mutability to `lowest_pixel_mem`

rinon:


src/decode.rs
```diff
@@ -3040,7 +3040,7 @@ unsafe fn decode_b_inner(
 
     if t.frame_thread.pass == 1 && b.intra == 0 && frame_hdr.frame_type.is_inter_or_switch() {
         let sby = t.b.y - ts.tiling.row_start >> f.sb_shift;
-        let lowest_px = &mut f.lowest_pixel_mem[ts.lowest_pixel + sby as usize];
+        let lowest_px = &mut f.lowest_pixel_mem.index_mut(ts.lowest_pixel + sby as usize);
```
randomPoison, line 3043: Should this be something like

```suggestion
        let mut lowest_px = f.lowest_pixel_mem.index_mut(ts.lowest_pixel + sby as usize);
        let lowest_px = &mut *lowest_px;
```

It seems like you're taking a mutable reference to the guard returned by `index_mut`, rather than a reference to the data protected by the guard, which doesn't seem like the right thing to do.

randomPoison (approved):
Looks good, just one minor question

src/decode.rs
rinon: In this particular case it doesn't matter, but generally yes. Fixed.

PR #950, opened 2024-04-02
`struct Pal`: Add inner mutability to pal buffer

rinon:
The `pal` buffer is a type-erased `Pixel` buffer that needs to be accessed concurrently. This change adds support for casting `DisjointMut` guards from `[u8]` buffers to appropriate bitdepth-dependent types that implement `FromBytes`.

### DIFF
--- a/src/disjoint_mut.rs
+++ b/src/disjoint_mut.rs
@@ -13,6 +13,8 @@ use std::fmt::Debug;
 use std::fmt::Display;
 use std::fmt::Formatter;
 use std::marker::PhantomData;
+use std::mem;
+use std::mem::ManuallyDrop;
 use std::ops::Bound;
 use std::ops::Deref;
 use std::ops::DerefMut;
@@ -27,6 +29,8 @@ use std::ops::RangeToInclusive;
 use std::ptr;
 use std::ptr::addr_of_mut;
 use std::sync::Arc;
+use zerocopy::AsBytes;
+use zerocopy::FromBytes;
 
 /// Wraps an indexable collection to allow unchecked concurrent mutable borrows.
 ///
@@ -70,6 +74,23 @@ pub struct DisjointMutGuard<'a, T: ?Sized + AsMutPtr, V: ?Sized> {
     bounds: Bounds,
 }
 
+impl<'a, T: AsMutPtr> DisjointMutGuard<'a, T, [u8]> {
+    fn cast<V: AsBytes + FromBytes>(self) -> DisjointMutGuard<'a, T, V> {
+        // We don't want to drop the old guard, because we aren't changing or
+        // removing the bounds from parent here.
+        let mut old_guard = ManuallyDrop::new(self);
+        let bytes = mem::take(&mut old_guard.slice);
+        DisjointMutGuard {
+            slice: V::mut_from(bytes).unwrap(),
+            phantom: old_guard.phantom,
+            #[cfg(debug_assertions)]
+            parent: old_guard.parent,
+            #[cfg(debug_assertions)]
+            bounds: old_guard.bounds.clone(),
+        }
+    }
+}
+
 impl<'a, T: ?Sized + AsMutPtr, V: ?Sized> Deref for DisjointMutGuard<'a, T, V> {
     type Target = V;
 
@@ -94,6 +115,23 @@ pub struct DisjointImmutGuard<'a, T: ?Sized + AsMutPtr, V: ?Sized> {
     parent: &'a DisjointMut<T>,
     #[cfg(debug_assertions)]
     bounds: Bounds,
+}
+
+impl<'a, T: AsMutPtr> DisjointImmutGuard<'a, T, [u8]> {
+    fn cast<V: FromBytes>(self) -> DisjointImmutGuard<'a, T, V> {
+        // We don't want to drop the old guard, because we aren't changing or
+        // removing the bounds from parent here.
+        let mut old_guard = ManuallyDrop::new(self);
+        let bytes = mem::take(&mut old_guard.slice);
+        DisjointImmutGuard {
+            slice: V::ref_from(bytes).unwrap(),
+            phantom: old_guard.phantom,
+            #[cfg(debug_assertions)]
+            parent: old_guard.parent,
+            #[cfg(debug_assertions)]
+            bounds: old_guard.bounds.clone(),
+        }
+    }
 }
 
 impl<'a, T: ?Sized + AsMutPtr, V: ?Sized> Deref for DisjointImmutGuard<'a, T, V> {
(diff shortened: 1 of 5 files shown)


src/internal.rs
```diff
@@ -462,18 +464,28 @@ impl Pal {
         self.data.is_empty()
     }
 
-    pub fn as_slice<BD: BitDepth>(&self) -> &[[[BD::Pixel; 8]; 3]] {
-        as_chunks::<3, [BD::Pixel; 8]>(
-            as_chunks::<8, BD::Pixel>(BD::cast_pixel_slice(&self.data)).0,
-        )
-        .0
+    pub fn index<'a: 'b, 'b, BD: BitDepth>(
```
randomPoison, line 467: Don't `index` and `index_mut` still need to be `unsafe`? Safety when indexing into a `DisjointMut` relies on the caller ensuring that the index is actually disjoint from other accesses, so I think helper methods like this that are doing `DisjointMut` access under the hood still need to be `unsafe`

src/recon.rs
```diff
@@ -2458,7 +2458,7 @@ pub(crate) unsafe fn rav1d_recon_b_intra<BD: BitDepth>(
                         * (f.b4_stride >> 1)
                         + ((t.b.x >> 1) + (t.b.y & 1)) as isize)
                         as isize;
-                    f.frame_thread.pal.as_slice::<BD>()[index as usize][0].as_ptr()
+                    f.frame_thread.pal.index::<BD>(index as usize)[0].as_ptr()
```
randomPoison, line 2461: Do we need to store the guard on the stack here? This looks like we're taking a pointer to data in a `DisjointMut`, so I think we need to have a guard on the stack while that pointer is being used.

src/internal.rs
rinon: If `index_mut` is used properly (i.e. never overlapping any other concurrent slice reference), then we don't. My argument to myself there was that we only need to put the safety comment and unsafe requirements on that method because of that. If one only used `index` there's no way to make that unsafe, you can only make something break here by using `index_mut` incorrectly.

I'm open to making it unsafe, I just don't think it is required.

src/recon.rs
rinon: Good point, done.

src/internal.rs
randomPoison: Okay, that makes sense to me wrt `index` not needing to be `unsafe`. But I think we need to keep `index_mut` unsafe. Any case where the function depends on "if it's used properly" is by definition unsafe, I think.

rinon: Absolutely, I missed that I didn't here, fixing that.

randomPoison (approved):

src/disjoint_mut.rs
```diff
@@ -19,6 +21,9 @@ use std::ops::RangeTo;
 use std::ops::RangeToInclusive;
 use std::ptr;
 
+use zerocopy::AsBytes;
+use zerocopy::FromBytes;
+
```
kkysen, line 26: nit:

```suggestion
use zerocopy::AsBytes;
use zerocopy::FromBytes;
```

src/disjoint_mut.rs
```diff
@@ -54,6 +59,21 @@ pub struct DisjointMutGuard<'a, T: AsMutPtr, V: ?Sized> {
     range: (Bound<usize>, Bound<usize>),
 }
 
+impl<'a, T: AsMutPtr> DisjointMutGuard<'a, T, [u8]> {
+    pub fn cast<V: AsBytes + FromBytes>(self) -> DisjointMutGuard<'a, T, V> {
+        let mut old_guard = ManuallyDrop::new(self);
```
kkysen, line 64: Is this because you can't move out of a type that `impl Drop`s?

src/disjoint_mut.rs
```diff
@@ -54,6 +59,21 @@ pub struct DisjointMutGuard<'a, T: AsMutPtr, V: ?Sized> {
...
+        let bytes = mem::take(&mut old_guard.slice);
+        DisjointMutGuard {
+            slice: V::mut_from(bytes).unwrap(),
+            phantom: old_guard.phantom,
+            #[cfg(debug_assertions)]
+            parent: old_guard.parent,
+            #[cfg(debug_assertions)]
+            range: old_guard.range,
```
kkysen, line 72: Not sure if this works:

```suggestion
            phantom: old_guard.phantom,
            ..old_guard
```

src/disjoint_mut.rs
```diff
@@ -80,6 +100,21 @@ pub struct DisjointImmutGuard<'a, T: AsMutPtr, V: ?Sized> {
     range: (Bound<usize>, Bound<usize>),
 }
 
+impl<'a, T: AsMutPtr> DisjointImmutGuard<'a, T, [u8]> {
+    pub fn cast<V: AsBytes + FromBytes>(self) -> DisjointImmutGuard<'a, T, V> {
```
kkysen, line 104: Why do we need the `AsBytes` constraint here?  `fn ref_from` is just on `FromBytes`.

src/internal.rs
```diff
@@ -462,18 +464,43 @@ impl Pal {
...
+        &'a self,
+        index: usize,
+    ) -> DisjointImmutGuard<'b, AlignedVec64<u8>, [[BD::Pixel; 8]; 3]> {
+        let guard: DisjointImmutGuard<'a, AlignedVec64<u8>, [u8]> = self.data.index(
+            index * 8 * 3 * mem::size_of::<BD::Pixel>()
+                ..(index + 1) * 8 * 3 * mem::size_of::<BD::Pixel>(),
+        );
+        guard.cast()
```
kkysen, line 475:
```suggestion
        type E<BD: BitDepth> = [[BD::Pixel; 8]; 3];
        
        let guard: DisjointImmutGuard<'a, AlignedVec64<u8>, [u8]> = self.data.index(
            index * mem::size_of::<E<BD>>()
                ..(index + 1) * mem::size_of::<E<BD>>(),
        );
        guard.cast::<E<BD>>()
```

I'd like to avoid the possibility of arithmetic errors turning into panics or UB.  This way they'd be a type error.  Given there's `unsafe`ty involved, I think the extra precaution is worth it.  The same can be done for `index_mut`.  We also shouldn't be manually computing array sizes with arithmetic.  Size calculation of arrays is a bit more complex (e.x. if alignment every becomes an issue, it changes); we should leave it up to the compiler.

src/unstable_extensions.rs
```diff
@@ -71,7 +71,7 @@ pub unsafe fn as_chunks_unchecked_mut<const N: usize, T>(this: &mut [T]) -> &mut
 #[inline]
 #[track_caller]
 #[must_use]
-pub fn as_chunks_mut<const N: usize, T>(this: &mut [T]) -> (&mut [[T; N]], &mut [T]) {
+pub fn as_chunks_mut<'a, const N: usize, T>(this: &'a mut [T]) -> (&'a mut [[T; N]], &'a mut [T]) {
```
kkysen, line 74: Why do we need this?  `std`'s version doesn't have it, and I'd like to keep the `fn`s in this module identical to `std's implementation (barring things we need to change slightly for privacy and unstability).

kkysen (changes requested):

src/disjoint_mut.rs
rinon: Yes, exactly. We need to not drop the guard, as we are replacing it.

src/disjoint_mut.rs
rinon: I tried this, but because the types of the two structs differ, it would require a nightly feature.

src/disjoint_mut.rs
kkysen: Could you add that as a comment?  Otherwise it's all good.

src/disjoint_mut.rs
kkysen: Ahh, one day I guess we'll have nice things.

src/disjoint_mut.rs
rinon: copied that from the mut version, thanks.

rinon:
I fixed this up so that we don't allow direct access to the guard cast and instead require users to index using the desired output type. Please take another look.

src/disjoint_mut.rs
```diff
@@ -85,6 +106,23 @@ pub struct DisjointImmutGuard<'a, T: AsMutPtr, V: ?Sized> {
     bounds: Bounds,
 }
 
+impl<'a, T: AsMutPtr> DisjointImmutGuard<'a, T, [u8]> {
```
kkysen, line 109: It's fine for now, but as a future potential optimization, we should see if we can add alignment info to the `[u8]` here, as without it, `fn cast` has to do an alignment check, even though the buffer we're indexing into is usually already aligned enough.

kkysen (changes requested):
The implementation LGTM, I just had a question about the naming, and also if we can `impl Mul<usize> for Bounds`.

src/disjoint_mut.rs
```diff
@@ -296,6 +400,12 @@ impl Bounds {
         let b = &other.range;
         a.start < b.end && b.start < a.end
     }
+
+    fn multiply(self, multiple: usize) -> Bounds {
+        let start = self.range.start * multiple;
+        let end = self.range.end * multiple;
+        Self { range: start..end }
+    }
```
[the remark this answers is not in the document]
rinon: It could be, but I don't particularly want to change this.

src/disjoint_mut.rs
rinon: Good point.

src/disjoint_mut.rs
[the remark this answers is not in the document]
kkysen: Sure, that's fine.

kkysen (approved):

PR #951, opened 2024-04-02
`struct Rav1dFrameContext_frame_thread`: Add inner mutability

rinon:


randomPoison (approved):
Changes mostly look good, I just had a couple of minor comments.

src/recon.rs
```diff
@@ -1591,8 +1620,46 @@ enum CfSelect {
     Task,
 }
 
+impl CfSelect {
+    fn set<BD: BitDepth>(
+        self,
+        f: &Rav1dFrameData,
+        t: &mut Rav1dTaskContext,
```
randomPoison, line 1627: Now that I think about it, it's probably not a good idea to pass a `&mut Rav1dTaskContext` here just to use one field, i.e. you might want to take `t.cf` as the argument, rather than all of `t`. The current usages of this are dereferencing a `*mut Rav1dTaskContext` so you're not hitting borrow errors yet, but once `decode_coefs` is taking a `t: &mut Rav1dTaskContext` you're likely going to hit borrow errors trying to take a mutable reference to all of `t`.

randomPoison: The same also probably applies to `get`, I think.

rinon: Probably, but it's easy enough to fix IF we hit that, rather than making this a bit less clear preemptively IMO.

src/recon.rs
```diff
@@ -1591,6 +1620,44 @@ enum CfSelect {
     Task,
 }
 
+impl CfSelect {
+    fn set<BD: BitDepth>(
```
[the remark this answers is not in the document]
rinon: Making `set` unsafe, I think get can be safe (see discussion on #890 )

src/decode.rs
```diff
@@ -1151,35 +1151,40 @@ unsafe fn obmc_lowest_px(
...
     t: &mut Rav1dTaskContext,
-    f: &mut Rav1dFrameData,
+    f: &Rav1dFrameData,
     bl: BlockLevel,
     bs: BlockSize,
     bp: BlockPartition,
     intra_edge_flags: EdgeFlags,
 ) -> Result<(), ()> {
     // Pull out the current block from Rav1dFrameData so that we can operate on
     // it without borrow check errors.
```
kkysen, line 1161:
```suggestion
```

src/recon.rs
kkysen: I agree it'd be better to take `cf: {&,&mut} BitDepthUnion<Cf>` as an individual field here instead of all of `t`.  And I think it's clearer this way as well.

src/internal.rs
```diff
@@ -511,7 +511,7 @@ pub struct Rav1dFrameContext_frame_thread {
     pub next_tile_row: [AtomicI32; 2],
 
     /// Indexed using `t.b.y * f.b4_stride + t.b.x`.
-    pub b: Vec<Av1Block>,
+    pub b: Vec<Mutex<Av1Block>>,
```
kkysen, line 514: This increases memory usage by 25% (32 vs. 40 bytes).  I think it's better to make the fields atomic, as they're generally accessed individually from what I can remember.  And we also get rid of the `Mutex` overhead.  I can open an issue for this.

kkysen (changes requested):

src/internal.rs
rinon: I think it makes more sense to access the block as a whole rather than add atomicity inside. We don't access blocks concurrently afaict. I'll make this vec DisjointMut for now to avoid the memory bloat.

kkysen: I was about to work on making `Av1Block` use `enum`s, which is a noisy but not super complex change.  Should I hold off on that to avoid conflicts with this then?

src/recon.rs
```diff
@@ -1761,14 +1828,13 @@ unsafe fn read_coef_tree<BD: BitDepth>(
...
-                        &mut cf_buf[offset..]
+                        (f.frame_thread.cf.as_mut_ptr() as *mut BD::Coef).add(offset)
                     }
-                    CfSelect::Task => &mut BD::select_mut(&mut (*t).cf).0,
+                    CfSelect::Task => &mut BD::select_mut(&mut (*t).cf).0 as *mut BD::Coef,
                 };
                 if debug_block_info!(f, (*t).b) && 0 != 0 {
                     coef_dump(
-                        cf,
+                        |i| *cf.add(i),
```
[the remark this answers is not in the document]
rinon: Didn't see any easy way, but now that we support slices over buffers properly we can. Fixed.

src/internal.rs
rinon: Yes please, don't do that, that's not necessary.

kkysen: Okay I can wait, but what do you mean it's not necessary?  We need to remove the `union`s.

rinon: I'm sorry, I read that as atomics. I should read more carefully when I'm tired. Just make sure you're coordinating with @randomPoison to ensure you're not both planning to do that. I think it is best to wait until this lands though, since you haven't done it yet.

include/common/bitdepth.rs
```diff
@@ -239,12 +239,8 @@ pub trait BitDepth: Clone + Copy {
...
 
-    fn _cast_coef_slice(bytes: &[u8]) -> &[Self::Coef] {
-        Self::Coef::slice_from(bytes).unwrap()
-    }
-
-    fn cast_coef_slice_mut(bytes: &mut [u8]) -> &mut [Self::Coef] {
-        Self::Coef::mut_slice_from(bytes).unwrap()
+    fn cast_coef_mut(bytes: &mut [u8]) -> &mut Self::Coef {
+        Self::Coef::mut_from(bytes).unwrap()
     }
```
kkysen, line 244: Do we still need this?  It seems unused now.

kkysen (changes requested):
Overall LGTM, though a few minor things:
* I think `cast_coef_mut` isn't used now.  Do we intend on using it now that we have `.slice_as`, etc.?
* One place we added extra `as isize` casts that aren't needed.
* I'd prefer to take `cf: {&,&mut} BitDepthUnion<Cf>` directly instead all of `t` in `get` and `set.
* A bunch of places where we slice and then add the length to the offset, you added the new length as `len`, which duplicates the existing inline `+=` to the offset.  If we could use `len` in both places, that'd be perfect.

src/recon.rs
```diff
@@ -3634,8 +3723,12 @@ pub(crate) unsafe fn rav1d_recon_b_inter<BD: BitDepth>(
                             if t.frame_thread.pass != 2 {
                                 left_filter_2d
                             } else {
-                                f.frame_thread.b
-                                    [t.b.y as usize * f.b4_stride as usize + t.b.x as usize - 1]
+                                f.frame_thread
+                                    .b
+                                    .index(
+                                        (t.b.y as isize * f.b4_stride + t.b.x as isize - 1)
```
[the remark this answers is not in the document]
rinon: If that is the case then it probably shouldn't be `ptrdiff_t`. I'm leaving this to match other usage unless and until we change that.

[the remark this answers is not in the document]
kkysen: The other usages don't do the `isize` cast, though, only this one.

PR #952, opened 2024-04-02
`struct Rav1dFrameContext`: Separate into data and context

rinon:
Separates the frame data into intialized and then rarely updated data (`Rav1dFrameData`) and a few of the fields required to pick tasks before accessing the rest of the frame data to execute the task.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -25,6 +25,7 @@ use crate::src::cdf::rav1d_cdf_thread_init_static;
 use crate::src::cdf::rav1d_cdf_thread_update;
 use crate::src::cdf::CdfMvComponent;
 use crate::src::cdf::CdfMvContext;
+use crate::src::cdf::CdfThreadContext;
 use crate::src::ctx::CaseSet;
 use crate::src::dequant_tables::dav1d_dq_tbl;
 use crate::src::disjoint_mut::DisjointMut;
@@ -63,6 +64,7 @@ use crate::src::internal::Bxy;
 use crate::src::internal::Rav1dContext;
 use crate::src::internal::Rav1dContextTaskType;
 use crate::src::internal::Rav1dDSPContext;
+use crate::src::internal::Rav1dFrameContext;
 use crate::src::internal::Rav1dFrameData;
 use crate::src::internal::Rav1dTaskContext;
 use crate::src::internal::Rav1dTaskContext_scratch_pal;
@@ -168,7 +170,6 @@ use std::ffi::c_uint;
 use std::ffi::c_void;
 use std::iter;
 use std::mem;
-use std::ptr::addr_of_mut;
 use std::slice;
 use std::sync::atomic::AtomicI32;
 use std::sync::atomic::Ordering;
@@ -3650,6 +3651,7 @@ unsafe fn setup_tile(
     c: &Rav1dContext,
     ts: &mut Rav1dTileState,
     f: &Rav1dFrameData,
+    in_cdf: &CdfThreadContext,
     data: &[u8],
     tile_row: usize,
     tile_col: usize,
@@ -3680,7 +3682,7 @@ unsafe fn setup_tile(
         };
     }
 
-    rav1d_cdf_thread_copy(&mut ts.cdf, &f.in_cdf);
+    rav1d_cdf_thread_copy(&mut ts.cdf, in_cdf);
     ts.last_qidx = frame_hdr.quant.yac;
     ts.last_delta_lf.fill(0);
 
@@ -3857,7 +3859,7 @@ unsafe fn read_restoration_info(
 pub(crate) unsafe fn rav1d_decode_tile_sbrow(
     c: &Rav1dContext,
     t: &mut Rav1dTaskContext,
-    f: &mut Rav1dFrameData,
+    f: &Rav1dFrameData,
 ) -> Result<(), ()> {
     let seq_hdr = &***f.seq_hdr.as_ref().unwrap();
     let root_bl = if seq_hdr.sb128 != 0 {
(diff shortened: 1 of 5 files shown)


src/thread_task.rs
```diff
@@ -323,24 +331,29 @@ unsafe fn create_filter_sbrow(
...
     let num_tasks = (f.sbh * (1 + uses_2pass)) as usize;
     tasks.grow_tasks(num_tasks);
     let task_idx = Rav1dTaskIndex::Task((f.sbh * (pass & 1)) as usize);
     if pass & 1 != 0 {
-        f.frame_thread_progress.entropy = AtomicI32::new(0);
+        fc.frame_thread_progress
+            .try_read()
+            .unwrap()
+            .entropy
+            .store(0, Ordering::Relaxed);
```
randomPoison, line 344: It seems odd to have an atomic inside of an `RwLock`ed struct. Both give us interior mutability, and doing it this way means we're paying twice to mutate that field. Would it make sense to hoist the atomic field out of the `RwLock`?

randomPoison (commented):
I gave this a first look over and had a few comments. I can give this another pass when it's ready for final review.

src/thread_task.rs
rinon: I could move the RwLock down to just the Vecs, we only need it to be able to resize the Vec.

kkysen: The relaxed atomic load/store is pretty much free, though, so I don't know if it's the biggest deal.

src/thread_task.rs
```diff
@@ -923,6 +949,7 @@ pub unsafe fn rav1d_worker_task(c: &Rav1dContext, task_thread: Arc<Rav1dTaskCont
         tasks[t_idx].next = None;
         let tile_tasks = tasks.tile_tasks;
         let t = &mut tasks[t_idx];
+        // dbg!(&*t);
```
randomPoison, line 952: Unused debug comment that can be removed

src/thread_task.rs
```diff
@@ -1031,23 +1064,25 @@ pub unsafe fn rav1d_worker_task(c: &Rav1dContext, task_thread: Arc<Rav1dTaskCont
                                 p_0 += 1;
                             }
                             assert!(task_thread_lock.is_none(), "thread lock should not be held");
+                            // drop(fc_locked);
```
randomPoison, line 1067: Unused comment can be removed

src/thread_task.rs
```diff
@@ -1186,11 +1225,12 @@ pub unsafe fn rav1d_worker_task(c: &Rav1dContext, task_thread: Arc<Rav1dTaskCont
                                     t.type_0 = TaskType::Cdef;
                                     t.deblock_progress = 0 as c_int;
                                     t.recon_progress = t.deblock_progress;
-                                    add_pending(f, t_idx);
+                                    add_pending(fc, t_idx);
                                     assert!(
                                         task_thread_lock.is_none(),
                                         "thread lock should not be held"
                                     );
+                                    // drop(fc_locked);
```
randomPoison, line 1233: Unused comment

src/thread_task.rs
```diff
@@ -1319,23 +1364,25 @@ pub unsafe fn rav1d_worker_task(c: &Rav1dContext, task_thread: Arc<Rav1dTaskCont
                 }
             }
             if sby + 1 == sbh {
-                f.task_thread.done[0].store(1, Ordering::SeqCst);
+                fc.task_thread.done[0].store(1, Ordering::SeqCst);
             }
             assert!(task_thread_lock.is_none(), "thread lock should not be held");
+            // drop(fc_locked);
```
randomPoison, line 1370: unused comment

randomPoison (approved):
Looks good (modulo any changes necessary to fix the last bit of lock contention), my only comments were minor ones

src/decode.rs
```diff
@@ -4599,52 +4617,52 @@ pub(crate) unsafe fn rav1d_decode_frame_exit(
...
+        res = rav1d_decode_frame_init_cdf(c, fc, &mut f, &fc.in_cdf());
     }
     // wait until all threads have completed
     if res.is_ok() {
         if c.tc.len() > 1 {
-            res = rav1d_task_create_tile_sbrow(c, f, 0, 1);
-            let mut task_thread_lock = (*f.task_thread.ttd).delayed_fg.lock().unwrap();
-            (*f.task_thread.ttd).cond.notify_one();
+            res = rav1d_task_create_tile_sbrow(c, fc, &f, 0, 1);
+            drop(f);
```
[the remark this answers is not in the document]
rinon: Holding the lock for FrameData while trying to wait on the task thread lock caused contention. We need to drop the FrameData lock to allow other threads to make progress.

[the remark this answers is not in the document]
rinon: I added a comment to this effect.

rinon:
There is still lock contention at https://github.com/memorysafety/rav1d/blob/e2b43173887f12105be297417c66e33b83009606/src/decode.rs#L4584 that I can't replicate reliably enough to find. I'm going to do some experimentation to see if I can't figure out what's happening here. The alternative is to actually just block there, which might be the best alternative.

kkysen:
> There is still lock contention at
> 
> https://github.com/memorysafety/rav1d/blob/e2b43173887f12105be297417c66e33b83009606/src/decode.rs#L4584
> 
> that I can't replicate reliably enough to find. I'm going to do some experimentation to see if I can't figure out what's happening here. The alternative is to actually just block there, which might be the best alternative.

So this is just for destructors mostly?  I think taking the lock might be okay then.  We should check with some benchmarks and mark it with a comment, but if the perf looks okay I think it's fine.

rinon:
Local benchmark looks ok, I'm going to land this.

PR #958, opened 2024-04-08
`Rav1dTaskContext::a`: Make into offset and give `BlockContext` interior mutability

randomPoison:
Once `a` was made into an offset, accesses to `f.a[t.a]` ran into borrowing issues because we need to mutate that data through an immutable reference to `f`. To fix this I wrapped the fields of `BlockContext` in `DisjointMut`, as the fields are being accessed concurrently from multiple threads, with different threads accessing disjoint parts of the arrays. I also added support for using `CaseSet` with `DisjointMut` via a new `CaseSetter::set_disjoint` method, since all of the fields of `BlockContext` are used with `CaseSet` at some point.

I also broke up the `t` argument to `decode_coefs` in order to address borrow issues: The `l` argument to `decode_coefs` is always a reference to `t.l`, and `t` is also passed mutably into the function, which causes borrow errors. Only a few fields of `t` are used within `decode_coefs`, so passing the fields in individually addresses the borrow issues. This also makes the arguments safe, since previously `t` was being passed in as a raw pointer. This then in turn required that the `get` and `set` methods on `CfSelect` take a `&mut BitDepthUnion<Cf>` directly, rather than being passed the entire `Rav1dTaskContext`.

A few of these fields (e.g. `partition` and `uvmode`) could probably be made atomic instead of using `DisjointMut`, since only individual elements are read, rather than a full slice being taken, but I opted to use `DisjointMut` for all of them for simplicity. I think the only requirement to use atomics for some of these fields would be atomic support in `CaseSet`.


randomPoison:
Note to self: Squash this on merge because not all of the intermediate commits compile/pass tests.

.github/workflows/build-and-test-aarch64-darwin.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       matrix:
         build: [
           {name: "release", cargo_flags: "--release",  timeout_multiplier: 1},
-          {name: "debug", cargo_flags: "", timeout_multiplier: 3},
+          {name: "debug", cargo_flags: "", timeout_multiplier: 10},
```
rinon, line 13: Is this still needed? This branch does not (yet) include #967 .

src/decode.rs
```diff
@@ -1923,6 +1924,8 @@ unsafe fn decode_b_inner(
                 TileStateRef::Frame => &f.lf.lvl,
                 TileStateRef::Local => &ts.lflvlmem,
             };
+            let mut a_uv_guard;
+            let mut l_uv_guard;
```
rinon, line 1928: I'm not sure you need these guards, won't passing the mutable references as parameters keep the guards alive for the same lifetime?

src/decode.rs
```diff
@@ -2940,6 +2955,8 @@ unsafe fn decode_b_inner(
...
```
rinon, line 2959: Do you need these guards?

rinon (approved):

src/decode.rs
randomPoison: These guards are for the second pair of slices that get passed in as a tuple. Because of the way that the tuple is being created inline in the function call we need to explicitly create the guards.

src/decode.rs
randomPoison: Same as my other comment: We need the guards for the pair of slices that are being passed in as a tuple.

rinon (approved):

PR #960, opened 2024-04-10
`struct refmvs_block`: Remove `#[repr(packed)]` on `struct refmvs_block_unaligned`, as it's already 12 bytes

kkysen:
This is replaced with a static assert that `refmvs_block` is 12 bytes. `packed` works slightly differently in Rust than in C, which is why this still works, and `packed` is less well supported in Rust, so it's nice to remove it, as it frees up other possibilities, and just makes things simpler.


randomPoison (approved):
Looks like you need to rebase, but I reviewed the last couple of commits which I think are the ones that are meant to be in this PR.

PR #965, opened 2024-04-11
`struct Disjoint{Immut,Mut}Guard`: `impl Default` for slices

kkysen:
I'll need this to add `DisjointMut` to `RefMvsFrame::r` in #942.  You can wait until I open that PR to review that, as it'll provide more context for how this is meant to be used.


rinon:
This is ok, but I still don't quite understand why we need it. Asked questions on #966.

PR #966, opened 2024-04-11
`struct RefMvsFrame::r`: Add `DisjointMut`

kkysen:
This removes `&mut` accesses of `Rav1dFrameData` added in #942.

### DIFF
--- a/src/refmvs.rs
+++ b/src/refmvs.rs
@@ -5,10 +5,12 @@ use crate::include::dav1d::headers::Rav1dSequenceHeader;
 use crate::include::dav1d::headers::Rav1dWarpedMotionType;
 use crate::src::align::Align16;
 use crate::src::align::AlignedVec64;
+use crate::src::disjoint_mut::DisjointMut;
 use crate::src::env::fix_mv_precision;
 use crate::src::env::get_gmv_2d;
 use crate::src::env::get_poc_diff;
 use crate::src::error::Rav1dResult;
+use crate::src::ffi_safe::FFISafe;
 use crate::src::internal::Bxy;
 use crate::src::intra_edge::EdgeFlags;
 use crate::src::levels::mv;
@@ -16,6 +18,7 @@ use crate::src::levels::BlockSize;
 use crate::src::tables::dav1d_block_dimensions;
 use cfg_if::cfg_if;
 use libc::ptrdiff_t;
+use std::array;
 use std::cmp;
 use std::ffi::c_int;
 use std::ffi::c_uint;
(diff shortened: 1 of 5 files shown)


src/refmvs.rs
```diff
@@ -362,15 +363,35 @@ pub(crate) struct Rav1dRefmvsDSPContext {
...
+            }
+            let ri = rt.r[i];
+            if ri > rf.r.len() {
+                return Default::default();
+            }
+            // This is the range that will actually be accessed,
+            // but `splat_mv` expects a pointer offset `bx4` backwards.
+            rf.r.index_mut(ri + bx4..ri + bx4 + bw4)
+        });
+        let mut r: [_; 37] = array::from_fn(|i| {
```
rinon, line 387: This drops the guard by re-using the name, right? Shouldn't we hold the guard across the call to `splat_mv`?

src/refmvs.rs
```diff
@@ -362,15 +363,35 @@ pub(crate) struct Rav1dRefmvsDSPContext {
...
+            if r[i].is_empty() {
```
rinon, line 388: Couldn't we have used `Option<DisjointMutGuard<...>>` instead?

rinon (commented):
Was it just the one place that needed a default impl for the guards? It seems weird to impl default unless we absolutely need to. Option should be basically the same?

src/refmvs.rs
kkysen: No, `array::from_fn` creates a new array without consuming the old one, unlike `.map`.

kkysen:
> Was it just the one place that needed a default impl for the guards? It seems weird to impl default unless we absolutely need to. Option should be basically the same?

I think so.  I was previously using it somewhere else, too, where `impl Default`ing it was more useful so we could avoid `.unwrap()`s in release code, but that had to be removed/reworked for other reasons, so I think I could use an `Option` as you suggested, which is definitely a lot simpler.

src/refmvs.rs
kkysen: Yup, switched to `Option` now.

randomPoison (approved):

src/refmvs.rs
```diff
@@ -464,15 +483,17 @@ fn scan_row(
     cnt: &mut usize,
     r#ref: refmvs_refpair,
     gmv: &[mv; 2],
-    b: &[refmvs_block],
+    r: &DisjointMut<AlignedVec64<refmvs_block>>,
+    b_offset: usize,
+    // b: &[refmvs_block],
```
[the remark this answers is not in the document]
kkysen: Nope, just forgot to remove it.

PR #967, opened 2024-04-11
`test.sh`: Set `RUST_LIB_BACKTRACE=0` so `debug_assertions` tests don't take forever due to capturing a ton of backtraces

kkysen:
We still want `RUST_BACKTRACE=1` to print a backtrace on panicking, as that's very useful and doesn't slow tests down. But that also causes `Backtrace::capture()` to actually capture a backtrace, and since we do that in every `DisjointMut::{index,index_mut}` call, it makes the tests extremely slow. Setting `RUST_LIB_BACKTRACE=0` overrides this and makes the tests fast (enough) again. And it can always be overridden again on the command line.

### DIFF
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -43,6 +43,15 @@ test_args=(
     --suite testdata-multi
 )
 
+# We want `RUST_BACKTRACE=1` in order to print a backtrace on panicking,
+# as that's very useful and doesn't slow tests down.
+# But that also causes `Backtrace::capture()`` to actually capture a backtrace,
+# and since we do that in every `DisjointMut::{index,index_mut}` call,
+# it makes the tests extremely slow.
+# Setting `RUST_LIB_BACKTRACE=0` overrides this and makes the tests fast (enough) again.
+# And it can always be overridden again on the command line.
+#
+# See more in the docs for `Backtrace::capture`.
 export RUST_BACKTRACE=1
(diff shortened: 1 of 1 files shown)


rinon (approved):
This is ok with me, as we've accumulated more usage the debug builds have gotten pretty slow. My only concern is debugging weird heisenbugs that we can't reproduce, but at least we will have a panic, even if we don't have the other location that conflicted.

thedataking:
Also fine with this. 

@kkysen could we put a short comment on the added line so it is clear why we do that instead of just setting `RUST_BACKTRACE=0`? You have a very detailed commit message but others may not see it.

kkysen:
> Also fine with this.
> 
> @kkysen could we put a short comment on the added line so it is clear why we do that instead of just setting `RUST_BACKTRACE=0`? You have a very detailed commit message but others may not see it.

Sure, I'll add it there.

> After trying to educate myself on `RUST_LIB_BACKTRACE` I think it takes precedence over `RUST_BACKTRACE`. If so, do we still need to set `RUST_BACKTRACE=1`if `RUST_LIB_BACKTRACE` takes precedence? Is the type of backtrace generated between the two different?

`RUST_LIB_BACKTRACE` only takes precedence for when `Backtrace::capture` is called, but when there's a panic, only `RUST_BACKTRACE` is checked.  Since we want a backtrace on panicking and that doesn't slow things down at all, we set `RUST_BACKTRACE=1` and `RUST_LIB_BACKTRACE=0`.

PR #968, opened 2024-04-11
Make `DisjointMut::index()` safe

rinon:
We were not adding any useful information at the numerous call sites for `index()`, and the safety involved here is thread safety to avoid data races, which we don't require for this application when handling pixel data and other plain data collections. This change makes `index()` a safe function and only requires safety comments for mutable access.

### DIFF
--- a/src/disjoint_mut.rs
+++ b/src/disjoint_mut.rs
@@ -175,7 +175,10 @@ impl<T: AsMutPtr> DisjointMut<T> {
     ///
     /// Caller must ensure that no elements of the resulting borrowed slice or
     /// element are concurrently borrowed (immutably or mutably) at all during
-    /// the lifetime of the returned mutable borrow.
+    /// the lifetime of the returned mutable borrow. We require that the
+    /// referenced data must be plain data and not contain any pointers or
+    /// references to avoid other potential memory safety issues due to racy
+    /// access.
     pub unsafe fn index_mut<'a, I>(
         &'a self,
         index: I,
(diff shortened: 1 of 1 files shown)


randomPoison (approved):

src/disjoint_mut.rs
```diff
@@ -202,12 +205,12 @@ impl<T: AsMutPtr> DisjointMut<T> {
...
-    ///
-    /// Caller must ensure that no elements of the resulting borrowed slice or
-    /// element are concurrently mutably borrowed at all during the lifetime of
-    /// the returned borrow.
-    pub unsafe fn index<'a, I>(
+    /// Caller must take care that no elements of the resulting borrowed slice
+    /// or element are concurrently mutably borrowed at all by [`index_mut()`]
+    /// during the lifetime of the returned borrow. We do not consider this
+    /// method unsafe because it cannot result in a race condition without
+    /// calling [`index_mut()`] with an overlapping range.
```
kkysen, line 212:
```suggestion
    /// Caller must take care that no elements of the resulting borrowed slice
    /// or element are concurrently mutably borrowed at all by [`Self::index_mut`]
    /// during the lifetime of the returned borrow. We do not consider this
    /// method `unsafe` because it cannot result in a race condition without
    /// calling [`Self::index_mut`] with an overlapping range.
    ///
    /// Furthermore, because this method is not publically accessible outside of this crate,
    /// we can ensure all calls are safe as a whole rather than add a
    /// virtually identical `unsafe {}` block and `// SAFETY` comment for each one.
    /// Those `// SAFETY` comments wouldn't add any useful information,
    /// so we're instead moving the safety comment here and asserting all uses are safe altogether.
```

src/disjoint_mut.rs
```diff
@@ -175,7 +175,10 @@ impl<T: AsMutPtr> DisjointMut<T> {
     ///
     /// Caller must ensure that no elements of the resulting borrowed slice or
     /// element are concurrently borrowed (immutably or mutably) at all during
-    /// the lifetime of the returned mutable borrow.
+    /// the lifetime of the returned mutable borrow. We require that the
+    /// referenced data must be plain data and not contain any pointers or
+    /// references to avoid other potential memory safety issues due to racy
+    /// access.
```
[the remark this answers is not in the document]
rinon: I don't think we should, it's not quite strictly necessary, and we contain arrays as elements in some cases.

src/disjoint_mut.rs
rinon: I rewrote the comment, look ok now?

kkysen (approved):
The docs look good now!  Though one little nit: I don't think `` [`index_mut()`] `` links correctly; I think you need `` [`Self::index_mut`] ``.

PR #970, opened 2024-04-13
`struct TxLpfRightEdge`: Add inner mutability

rinon:
Adds inner mutability to the right edge split vector by using DisjointMut and narrower ranges for reading and writing data.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -92,8 +92,10 @@ use std::ffi::c_uint;
 use std::mem;
 use std::ops::Add;
 use std::ops::AddAssign;
+use std::ops::Deref;
 use std::ops::Index;
 use std::ops::IndexMut;
+use std::ops::Range;
 use std::ops::Sub;
 use std::ptr;
 use std::sync::atomic::AtomicBool;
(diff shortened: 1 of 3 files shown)


src/internal.rs
```diff
@@ -504,27 +506,49 @@ pub struct Rav1dFrameContext_frame_thread {
...
+        slice_mut.copy_from_slice(src);
     }
 
-    pub fn get_mut(&mut self) -> (&mut [u8], &mut [u8]) {
+    pub fn copy_from_slice_uv(&self, index: Range<usize>, src: &[u8]) {
         let mid = self.inner.len() / 2;
-        self.inner.split_at_mut(mid)
+        #[allow(unused_mut)]
+        let mut slice_mut = unsafe { self.inner.index_mut(index.start + mid..index.end + mid) };
+        slice_mut.copy_from_slice(src);
```
randomPoison, line 551: These methods don't seem to be used in this PR. What was the reason for adding them?

randomPoison (approved):

src/internal.rs
rinon: They're used at https://github.com/memorysafety/rav1d/pull/970/files#diff-aeece9508d349335fdebc30632ef26fa2851aa73336dea735c78d36cb7e77b2bR4071

PR #972, opened 2024-04-13
`struct DisjointMut`: Simplify bounds and improve error messages

kkysen:
When I was first looking into why `DisjointMut` slowed things down so drastically, I first tried simplifying the ranges/bounds.  I then discovered it was mostly the backtraces (fixed in #967), but I had already done this and started building stuff off of it, and I think it makes the logic a lot clearer (I had a hard time understanding the previous overlapping logic), specifically `fn overlap` and `fn get_mut`.  I also improved the error messages, making it more concise, and printing the `Display` version of the `Backtrace` that's normally printed on panics rather than the `Debug` version we'd been printing.

### DIFF
--- a/src/disjoint_mut.rs
+++ b/src/disjoint_mut.rs
@@ -6,7 +6,10 @@
 #![allow(unused)]
 
 use std::cell::UnsafeCell;
+use std::fmt;
 use std::fmt::Debug;
+use std::fmt::Display;
+use std::fmt::Formatter;
 use std::marker::PhantomData;
 use std::ops::Bound;
 use std::ops::Deref;
@@ -41,7 +44,7 @@ pub struct DisjointMut<T: AsMutPtr> {
     inner: UnsafeCell<T>,
 
     #[cfg(debug_assertions)]
-    ranges: debug::DisjointMutDebugRanges,
+    bounds: debug::DisjointMutAllBounds,
 }
 
 #[cfg_attr(not(debug_assertions), repr(transparent))]
@@ -53,7 +56,7 @@ pub struct DisjointMutGuard<'a, T: AsMutPtr, V: ?Sized> {
     #[cfg(debug_assertions)]
     parent: &'a DisjointMut<T>,
     #[cfg(debug_assertions)]
-    range: (Bound<usize>, Bound<usize>),
+    bounds: Bounds,
 }
 
 impl<'a, T: AsMutPtr, V: ?Sized> Deref for DisjointMutGuard<'a, T, V> {
@@ -79,7 +82,7 @@ pub struct DisjointImmutGuard<'a, T: AsMutPtr, V: ?Sized> {
     #[cfg(debug_assertions)]
     parent: &'a DisjointMut<T>,
     #[cfg(debug_assertions)]
-    range: (Bound<usize>, Bound<usize>),
+    bounds: Bounds,
 }
 
 impl<'a, T: AsMutPtr, V: ?Sized> Deref for DisjointImmutGuard<'a, T, V> {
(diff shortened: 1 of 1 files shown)


src/disjoint_mut.rs
```diff
@@ -407,143 +428,137 @@ mod debug {
...
+            for b in self.bounds.immutable.lock().unwrap().iter() {
+                check_overlaps(&bounds, true, b, false);
             }
-            let mut ranges = self.ranges.mutable.lock().unwrap();
-            for r in ranges.iter() {
-                if range.overlaps(r) {
-                    let thread = thread::current().id();
-                    panic!("{range:?} on thread {thread:?} overlaps with an existing mutable range: {r:#?}");
-                }
+            let mut all_bounds = self.bounds.mutable.lock().unwrap();
```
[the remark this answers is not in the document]
kkysen: Okay, I can change it to `mut_bounds` or something?

src/disjoint_mut.rs
```diff
@@ -407,143 +428,137 @@ mod debug {
...
-        mutable: Mutex<Vec<DisjointMutRange>>,
-
-        immutable: Mutex<Vec<DisjointMutRange>>,
+    pub struct DisjointMutAllBounds {
+        mutable: Mutex<Vec<DisjointMutBounds>>,
+
+        immutable: Mutex<Vec<DisjointMutBounds>>,
+    }
+
+    // Inline so that the thread doesn't switch during this call,
```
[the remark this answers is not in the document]
kkysen: I wasn't sure what was happening exactly, but before I inlined it, the kacktrace kept being only one stack frame.  I'm not really sure why exactly and inlining fixed it.  If there's a better way I'd definitely prefer that.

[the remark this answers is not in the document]
rinon: Was this in a release build or opt-dev? It's possible the unwinding info had issues and it couldn't backtrace correctly in the optimized builds. That's not an issue this is fixing so much as papering over, but it's also not an issue we are going to be able to fix.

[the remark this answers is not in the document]
kkysen: That might've been it, but it's hard to reproduce the conditions exactly.  I've changed things to instead use `#[track_caller]`, so at least the main error message (with no backtrace at all) points to where we've called `.index` or `.index_mut`.

rinon (approved):

src/disjoint_mut.rs
```diff
@@ -219,30 +224,32 @@ impl<T: AsMutPtr> DisjointMut<T> {
     /// during the lifetime of the returned borrow.
     ///
     /// [`index_mut`]: DisjointMut::index_mut
+    #[track_caller]
```
rinon, line 227: track_caller should be cfg_attr here, it's not zero-cost.

src/disjoint_mut.rs
```diff
@@ -407,143 +430,139 @@ mod debug {
...
     }
 
     impl<'a, T: AsMutPtr, V: ?Sized> DisjointMutGuard<'a, T, V> {
-        pub fn new(
-            parent: &'a DisjointMut<T>,
-            slice: &'a mut V,
-            range: (Bound<usize>, Bound<usize>),
-        ) -> Self {
-            parent.add_mut_range(&range);
+        #[track_caller]
```
rinon, line 532: cfg_attr

src/disjoint_mut.rs
```diff
@@ -179,24 +182,26 @@ impl<T: AsMutPtr> DisjointMut<T> {
     /// referenced data must be plain data and not contain any pointers or
     /// references to avoid other potential memory safety issues due to racy
     /// access.
+    #[track_caller]
```
rinon, line 185: cfg_attr

src/disjoint_mut.rs
kkysen: Oh, it's not? I thought it was just a debug info thing.  I can change that then.

src/disjoint_mut.rs
kkysen: I'll leave these ones since it's in `mod debug` already.

src/disjoint_mut.rs
kkysen: Would it still be useful here and in `DisjointMutIndex::get_mut` for the bounds check?

rinon: I don't think so, this is fine as is.

rinon (approved):

PR #975, opened 2024-04-15
`fn load_tmvs_c`: Add safe `rp_proj` arg

kkysen:


src/decode.rs
```diff
@@ -62,6 +62,7 @@ use crate::src::error::Rav1dError::EINVAL;
 use crate::src::error::Rav1dError::ENOMEM;
 use crate::src::error::Rav1dError::ENOPROTOOPT;
 use crate::src::error::Rav1dResult;
+use crate::src::ffi_safe::FFISafe;
```
randomPoison, line 65: What is this `FFISafe` type? I don't see it being added in this PR or the ones preceding it.

randomPoison (approved):

PR #981, opened 2024-04-18
Create safe slices from `cdef_line_buf`

rinon:


rinon:
Oops, this got merged as part of #950. Closing this, but if you could have a look at 217cd21b0e76596ae67b14c8029875104a460f22 still I'd appreciate it.

kkysen:
> Oops, this got merged as part of #950. Closing this, but if you could have a look at [217cd21](https://github.com/memorysafety/rav1d/commit/217cd21b0e76596ae67b14c8029875104a460f22) still I'd appreciate it.

What's the difference between `src_y_stride` and just `y_stride`?  Besides not knowing that, everything else in the commit LGTM.

rinon:
`y_stride` is the output stride (from `lr_stride[..]`), `src_y_stride` is the stride for `cdef_line_buf`. In the original C code we use `src_stride[..]` as the strides for `cdef_line_buf`, rather than `lr_stride[..]` The two strides can be different due to (I think) superscaling?

PR #982, opened 2024-04-18
Initialize `f.lowest_pixel_mem`

rinon:


rinon:
It does, I just didn't realize that. I'll rebase on your change.

kkysen:
Fixed in #978 already.

PR #983, opened 2024-04-18
`struct RefMvsFrame::{rp,rp_ref}`: Remove raw ptr fields and pass through the callstack

kkysen:


randomPoison (approved):

randomPoison:
Looks like this needs a rebase, but I reviewed the last commit which seems to be the only one for this PR. Let me know if I missed any commits that need review though.

kkysen:
> Looks like this needs a rebase, but I reviewed the last commit which seems to be the only one for this PR. Let me know if I missed any commits that need review though.

Oops.  It's the last two actually.

PR #987, opened 2024-04-19
Backport packing of palette indices from dav1d 1.3.0

fbossen:


src/decode.rs
```diff
@@ -708,7 +708,8 @@ fn order_palette(
 unsafe fn read_pal_indices(
     ts: &mut Rav1dTileState,
     scratch_pal: &mut Rav1dTaskContext_scratch_pal,
-    pal_idx: &mut [u8],
+    pal_tmp: &mut [u8],
```
rinon, line 711: Would it make sense to pass a reference to `t.scratch` instead? @randomPoison has been working on this structure and might have an opinion.

src/decode.rs
```diff
@@ -708,7 +708,8 @@ fn order_palette(
...
+    pal_idx: Option<&mut [u8]>, // if None, use pal_tmp instead of pad_idx
```
rinon, line 712: `pal_idx` instead of `pad_idx`?

src/decode.rs
```diff
@@ -731,32 +732,52 @@ unsafe fn read_pal_indices(
...
         }
     }
-    if h4 < bh4 {
-        let y_start = h4 * 4;
-        let len = bw4 * 4;
-        let (src, dests) = pal_idx.split_at_mut(stride * y_start);
-        let src = &src[stride * (y_start - 1)..][..len];
-        for y in 0..(bh4 - h4) * 4 {
-            dests[y * stride..][..len].copy_from_slice(src);
+    if let Some(pal_idx) = pal_idx {
```
rinon, line 754: Could we just do the following and not duplicate the code?

```rust
let pal_idx = pal_idx.unwrap_or(pal_tmp);
...
```

src/decode.rs
fbossen: Oops. Fixed it.

src/decode.rs
fbossen: I had tried that, but ran into borrowing issues in `pal_idx[i] = pal_tmp[2 * i + 0] | (pal_tmp[2 * i + 1] << 4)` and ended up duplicating the code.

fbossen: Actually, the duplicated code will be put in a separate function (with asm versions) in a future PR, so this is a temporary issue

rinon: Sounds good, I see.

PR #988, opened 2024-04-20
`trait DisjointMutIndex`: Add an `impl` for `(RangeFrom<usize>, RangeTo<usize>)` for `[start..][..len]` bounds

kkysen:
A majority of our slice ranges are of the form `[start..][..len]`.  This is easy to express with normal slices where we can do the slicing multiple times, but with `DisjointMut`, that's harder, so this adds support for `.index((start.., ..len))` to achieve the same.  It's not as clear what it means initially, but we use this idiom so much I think it might be worth it for clarity through brevity.

### DIFF
--- a/src/disjoint_mut.rs
+++ b/src/disjoint_mut.rs
@@ -232,17 +232,10 @@ impl<T: ?Sized + AsMutPtr> DisjointMut<T> {
     /// references to avoid other potential memory safety issues due to racy
     /// access.
     #[cfg_attr(debug_assertions, track_caller)]
-    pub unsafe fn index_mut<'a, I>(
-        &'a self,
-        index: I,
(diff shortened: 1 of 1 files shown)


src/disjoint_mut.rs
```diff
@@ -465,13 +451,22 @@ impl From<RangeToInclusive<usize>> for Bounds {
     }
 }
 
+impl From<(RangeFrom<usize>, RangeTo<usize>)> for Bounds {
```
rinon, line 454: I'd like to see documentation on this somewhere, it's not obvious from the implementation how it intended to be used.

rinon (approved):
Implementation looks fine, please add a short docstring with the info from the commit message.

PR #992, opened 2024-04-22
`struct Rav1dPictureData`: `Arc`ify

### LINKED ISSUE #866: `Rav1dPicture::ref` - Replace with `Arc`.
Split from https://github.com/memorysafety/rav1d/issues/641 because there is additional complexity around making the picture allocator safe.

kkysen:
* Fixes #866.


randomPoison (approved):
Looks good, just one comment but not necessarily a blocker.

PR #993, opened 2024-04-23
backport from `dav1d` 1.2.0: picture: allow storing an array of `Dav1dITUTT35` entries

kkysen:
Reopening #528 since it got closed because I deleted the target branch, which I guess happens when the PR branch is on a fork.

### DIFF
--- a/include/dav1d/headers.rs
+++ b/include/dav1d/headers.rs
@@ -57,16 +57,6 @@ where
     }
 }
 
-impl<R, D> DRav1d<Vec<R>, Vec<D>>
-where
-    R: Clone + Into<D>,
-{
-    pub fn push(&mut self, value: R) {
-        self.rav1d.push(value.clone());
-        self.dav1d.push(value.into());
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -34,6 +34,7 @@ paste = "1.0.14"
 raw-cpuid = "11.0.1"
 strum = { version = "0.25.0", features = ["derive"] }
 to_method = "1.1.0"
+try-lock = "0.2.5"
```
rinon, line 37: As discussed offline, I don't think we should add this dependency in this PR. I don't want to block this PR while we consider the new dependency, and it would be better to entirely switch over if we choose to use it.

rinon (commented):
Everything other than the addition of TryLock LGTM

Cargo.toml
kkysen: I'll replace it with a `Mutex` for now and then merge.

Issue #713, opened 2024-02-01
`Rav1dFrameContext`: Make fields safe


randomPoison:
Make all raw pointer fields into appropriate safe smart pointer or collection types. Note that the `Rav1dContext` pointer is handled separately in #695.

- [x] `mvs_ref`, `mvs` - Replace `Rav1dRef` with an `Arc`.
- [x] `ref_mvs`, `ref_mvs_ref` - Replace `Rav1dRef` with an `Arc`.
- [x] `cur_segmap`, `cur_segmap_ref` - Replace `Rav1dRef` with an `Arc`.
- [x] `prev_segmap`, `prev_segmap_ref` - Replace `Rav1dRef` with an `Arc`.
- [ ] `ts`, `n_ts` - Replace with `AlignedVec32`.
- [x] `dsp` - References `c.dsp` indexed by `seq_hdr.hbd`. Might replace with an offset, but might also make sense to remove entirely and directly do `c.dsp[seq_hdr.hbd]` where it's used.
- [x] `ipred_edge` - First element is the base pointer of the allocation, next two elements point into the allocation. Split up into an `AlignedVec64` and then an array of offsets.
- [x] `qm` - Points into `dav1d_qm_tbl`. Can probably be made into `&'static [u8]`s, but may need to be offsets depending on usage.
- [x] `a`, `a_sz` - Owned allocation, can be a `Vec`.

PR #995, opened 2024-04-23
`struct Rav1dFrameData::ipred_edge`: Make into `DisjointMut<AlignedVec<u8>>`

randomPoison:
Part of https://github.com/memorysafety/rav1d/issues/713

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -1,5 +1,4 @@
 use crate::include::common::attributes::ctz;
-use crate::include::common::bitdepth::DynPixel;
 use crate::include::common::bitdepth::BPC;
 use crate::include::common::intops::apply_sign64;
 use crate::include::common::intops::iclip;
@@ -115,7 +114,6 @@ use crate::src::lf_mask::Av1RestorationUnit;
 use crate::src::log::Rav1dLog as _;
 use crate::src::mem::rav1d_alloc_aligned;
 use crate::src::mem::rav1d_free_aligned;
-use crate::src::mem::rav1d_freep_aligned;
 use crate::src::msac::rav1d_msac_decode_bool;
 use crate::src::msac::rav1d_msac_decode_bool_adapt;
 use crate::src::msac::rav1d_msac_decode_bool_equi;
@@ -4318,21 +4316,9 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
     rav1d_calc_lf_values(&mut f.lf.lvl, &frame_hdr, &[0, 0, 0, 0]);
 
     let ipred_edge_sz = f.sbh * f.sb128w << hbd;
-    if ipred_edge_sz != f.ipred_edge_sz {
-        rav1d_freep_aligned(
-            &mut *f.ipred_edge.as_mut_ptr().offset(0) as *mut *mut DynPixel as *mut c_void,
-        );
-        f.ipred_edge[0] =
-            rav1d_alloc_aligned(ipred_edge_sz as usize * 128 * 3, 64) as *mut DynPixel;
-        let ptr = f.ipred_edge[0] as *mut u8;
-        if ptr.is_null() {
-            f.ipred_edge_sz = 0;
-            return Err(ENOMEM);
-        }
-        f.ipred_edge[1] = ptr.offset(ipred_edge_sz as isize * 128 * 1) as *mut DynPixel;
-        f.ipred_edge[2] = ptr.offset(ipred_edge_sz as isize * 128 * 2) as *mut DynPixel;
-        f.ipred_edge_sz = ipred_edge_sz;
-    }
+    // TODO: Fallible allocation
+    f.ipred_edge.resize(ipred_edge_sz as usize * 128 * 3, 0);
+    f.ipred_edge_off = bpc.pxstride(ipred_edge_sz as usize * 128);
 
     let re_sz = f.sb128h * frame_hdr.tiling.cols;
     // TODO: Fallible allocation
--- a/src/ipred_prepare.rs
+++ b/src/ipred_prepare.rs
@@ -127,6 +127,8 @@ pub fn rav1d_prepare_intra_edges<BD: BitDepth>(
     edge_flags: EdgeFlags,
     dst: &[BD::Pixel], // contains 4*h first rows of picture, last row in slice contains 4*w samples
     stride: ptrdiff_t,
+    // Buffer and offset pair. `isize` value is the base offset that should be used
+    // when indexing into the buffer.
     prefilter_toplevel_sb_edge: Option<(&DisjointMut<AlignedVec64<u8>>, isize)>,
     mut mode: IntraPredMode,
     angle: &mut c_int,
--- a/src/ipred_prepare.rs
+++ b/src/ipred_prepare.rs
@@ -1,6 +1,8 @@
 use crate::include::common::bitdepth::AsPrimitive;
 use crate::include::common::bitdepth::BitDepth;
+use crate::src::align::AlignedVec64;
 use crate::src::const_fn::const_for;
+use crate::src::disjoint_mut::DisjointMut;
 use crate::src::env::BlockContext;
 use crate::src::intra_edge::EdgeFlags;
 use crate::src::levels::IntraPredMode;
@@ -125,7 +127,7 @@ pub fn rav1d_prepare_intra_edges<BD: BitDepth>(
     edge_flags: EdgeFlags,
     dst: &[BD::Pixel], // contains 4*h first rows of picture, last row in slice contains 4*w samples
     stride: ptrdiff_t,
-    prefilter_toplevel_sb_edge: Option<&[BD::Pixel]>,
+    prefilter_toplevel_sb_edge: Option<(&DisjointMut<AlignedVec64<u8>>, isize)>,
     mut mode: IntraPredMode,
     angle: &mut c_int,
     tw: c_int,
@@ -169,6 +171,7 @@ pub fn rav1d_prepare_intra_edges<BD: BitDepth>(
     }
 
     // `dst_top` starts with either the top or top-left sample depending on whether have_left is true
+    let edge_buf_guard;
     let dst_top = if have_top
         && (av1_intra_prediction_edges[mode as usize]
             .needs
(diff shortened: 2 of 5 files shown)


randomPoison:
Wait until https://github.com/memorysafety/rav1d/pull/952 is merged and this is rebased on top of those changes before merging this.

src/ipred_prepare.rs
```diff
@@ -125,7 +127,7 @@ pub fn rav1d_prepare_intra_edges<BD: BitDepth>(
     edge_flags: EdgeFlags,
     dst: &[BD::Pixel], // contains 4*h first rows of picture, last row in slice contains 4*w samples
     stride: ptrdiff_t,
-    prefilter_toplevel_sb_edge: Option<&[BD::Pixel]>,
+    prefilter_toplevel_sb_edge: Option<(&DisjointMut<AlignedVec64<u8>>, isize)>,
```
kkysen, line 130: Could you add a doc comment for what the `isize` is?

src/ipred_prepare.rs
```diff
@@ -183,9 +186,10 @@ pub fn rav1d_prepare_intra_edges<BD: BitDepth>(
     {
         let px_have = cmp::min(8 * tw, 4 * (w - x)) as usize;
         let n = px_have + have_left as usize;
-        if let Some(prefilter_toplevel_sb_edge) = prefilter_toplevel_sb_edge {
-            let offset = (x * 4) as usize - have_left as usize;
-            &prefilter_toplevel_sb_edge[offset..][..n]
+        if let Some((edge_buf, base)) = prefilter_toplevel_sb_edge {
+            let offset = ((x * 4) as usize - have_left as usize).wrapping_add_signed(base);
+            edge_buf_guard = edge_buf.slice_as(offset..offset + n);
```
kkysen, line 191: After #988, I think we can do this:

```suggestion
            edge_buf_guard = edge_buf.slice_as((offset.., ..n));
```

src/recon.rs
```diff
@@ -2628,11 +2628,10 @@ pub(crate) unsafe fn rav1d_recon_b_intra<BD: BitDepth>(
                             ),
                         ]);
                         let top_sb_edge_slice = if t.b.y & f.sb_step - 1 == 0 {
-                            let mut top_sb_edge: *const BD::Pixel =
-                                f.ipred_edge[0] as *mut BD::Pixel;
                             let sby = t.b.y >> f.sb_shift;
-                            top_sb_edge = top_sb_edge.offset((f.sb128w * 128 * (sby - 1)) as isize);
-                            Some(slice::from_raw_parts(top_sb_edge, f.sb128w as usize * 128))
+                            let offset = f.ipred_edge_off as isize * 0
```
kkysen, line 2632: `f.ipred_edge_off as isize * 0 +` doesn't do anything.

src/recon.rs
```diff
@@ -2628,11 +2628,10 @@ pub(crate) unsafe fn rav1d_recon_b_intra<BD: BitDepth>(
...
+                                + (f.sb128w * 128 * (sby - 1)) as isize;
```
kkysen, line 2633: `f.sb128w` is always positive, so I think you can make this a `usize`.

src/recon.rs
kkysen: Oh, I guess this is just here for parity with below where it's not `0`?

src/recon.rs
```diff
@@ -4506,15 +4497,14 @@ pub(crate) unsafe fn rav1d_backup_ipred_edge<BD: BitDepth>(
...
     BD::pixel_copy(
-        &mut slice::from_raw_parts_mut(
-            f.ipred_edge[0].cast(),
-            (sby_off + x_off * 4 + (4 * (ts.tiling.col_end - x_off)))
-                .try_into()
-                .unwrap(),
-        )[(sby_off + x_off * 4).try_into().unwrap()..],
+        &mut f
+            .ipred_edge
+            .mut_slice_as(ipred_edge_off..ipred_edge_off + n),
```
kkysen, line 4505:
```suggestion
            .mut_slice_as((ipred_edge_off.., ..ipred_edge_off)),
```

src/recon.rs
```diff
@@ -4525,22 +4515,22 @@ pub(crate) unsafe fn rav1d_backup_ipred_edge<BD: BitDepth>(
             + (((t.b.y + f.sb_step) * 4 >> ss_ver) - 1) as isize * BD::pxstride(f.cur.stride[1]);
         let mut pl = 1;
         while pl <= 2 {
+            let ipred_edge_off =
+                (f.ipred_edge_off * pl) + (sby_off + (x_off * 4 >> ss_hor)) as usize;
+            let n = (4 * (ts.tiling.col_end - x_off) >> ss_hor)
+                .try_into()
+                .unwrap();
```
kkysen, line 4522:
```suggestion
            let n = 4 * (ts.tiling.col_end - x_off) as usize >> ss_hor;
```

src/recon.rs
```diff
@@ -4506,15 +4497,14 @@ pub(crate) unsafe fn rav1d_backup_ipred_edge<BD: BitDepth>(
     let y: *const BD::Pixel = (f.cur.data.data[0] as *const BD::Pixel)
         .offset((x_off * 4) as isize)
         .offset((((t.b.y + f.sb_step) * 4 - 1) as isize * BD::pxstride(f.cur.stride[0])) as isize);
+    let ipred_edge_off = (f.ipred_edge_off * 0) + (sby_off + x_off * 4) as usize;
+    let n = (4 * (ts.tiling.col_end - x_off)).try_into().unwrap();
```
kkysen, line 4501:
```suggestion
    let n = 4 * (ts.tiling.col_end - x_off) as usize;
```

I only used `.try_into().unwrap()` previously for a checked cast since we were blindly using it as a length without bounds checking.

src/recon.rs
```diff
@@ -4525,22 +4515,22 @@ pub(crate) unsafe fn rav1d_backup_ipred_edge<BD: BitDepth>(
...
             BD::pixel_copy(
-                &mut slice::from_raw_parts_mut(
-                    f.ipred_edge[pl as usize].cast(),
-                    (sby_off + (x_off * 4 >> ss_hor) + (4 * (ts.tiling.col_end - x_off) >> ss_hor))
-                        .try_into()
-                        .unwrap(),
-                )[(sby_off + (x_off * 4 >> ss_hor)).try_into().unwrap()..],
+                &mut f
+                    .ipred_edge
+                    .mut_slice_as(ipred_edge_off..ipred_edge_off + n),
```
kkysen, line 4526:
```suggestion
                    .mut_slice_as((ipred_edge_off.., ..ipred_edge_off)),
```

src/recon.rs
```diff
@@ -4525,22 +4515,22 @@ pub(crate) unsafe fn rav1d_backup_ipred_edge<BD: BitDepth>(
...
                 &slice::from_raw_parts(
-                    f.cur.data.data[pl as usize].cast(),
+                    f.cur.data.data[pl].cast(),
                     (uv_off + (4 * (ts.tiling.col_end - x_off) >> ss_hor) as isize)
                         .try_into()
                         .unwrap(),
                 )[uv_off.try_into().unwrap()..],
```
kkysen, line 4532:
```suggestion
                    f.cur.data.data[pl].cast().offset(uv_off),
                    n,
                ),
```

src/decode.rs
```diff
@@ -4309,21 +4307,9 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
...
-            f.ipred_edge_sz = 0;
-            return Err(ENOMEM);
-        }
-        f.ipred_edge[1] = ptr.offset(ipred_edge_sz as isize * 128 * 1) as *mut DynPixel;
-        f.ipred_edge[2] = ptr.offset(ipred_edge_sz as isize * 128 * 2) as *mut DynPixel;
-        f.ipred_edge_sz = ipred_edge_sz;
-    }
+    // TODO: Fallible allocation
+    f.ipred_edge.resize(ipred_edge_sz as usize * 128 * 3, 0);
+    f.ipred_edge_off = bpc.pxstride(ipred_edge_sz as usize * 128);
```
kkysen, line 4312: Can we just derive `f.ipred_edge_off` from `f.ipred_edge.len()` later?  Deriving it through division at use sites will allow for bounds check elision.

kkysen: Actually, we always add something to it first, so that doesn't matter.  Nevermind then.

kkysen, line 4312:
```suggestion
    f.ipred_edge.resize((ipred_edge_sz << hbd) * 128 * 3, 0);
    f.ipred_edge_off = ipred_edge_sz * 128);
```

src/decode.rs
```diff
@@ -4309,21 +4307,9 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
     rav1d_calc_lf_values(&mut f.lf.lvl, &frame_hdr, &[0, 0, 0, 0]);
 
     let ipred_edge_sz = f.sbh * f.sb128w << hbd;
```
kkysen, line 4309:
```suggestion
    let ipred_edge_sz = (f.sbh * f.sb128w) as usize;
```

kkysen (changes requested):

src/decode.rs
```diff
@@ -4309,21 +4307,9 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
...
-        if ptr.is_null() {
-            f.ipred_edge_sz = 0;
-            return Err(ENOMEM);
-        }
-        f.ipred_edge[1] = ptr.offset(ipred_edge_sz as isize * 128 * 1) as *mut DynPixel;
-        f.ipred_edge[2] = ptr.offset(ipred_edge_sz as isize * 128 * 2) as *mut DynPixel;
-        f.ipred_edge_sz = ipred_edge_sz;
-    }
+    // TODO: Fallible allocation
+    f.ipred_edge.resize(ipred_edge_sz as usize * 128 * 3, 0);
```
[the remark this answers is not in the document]
kkysen: The previous code didn't initialize it here, so no.

src/recon.rs
randomPoison: Yup, the original logic was doing `ipred_edge[0]`, which changed to `ipred_edge_off * 0`. I wanted to keep that in there for parity with the other branches.

src/recon.rs
randomPoison: The problem isn't `f.sb128w`, it's `sby - 1`. `sby` can sometimes be 0, which makes the entire expression negative. That's why I'm now passing the index into `rav1d_prepare_intra_edges`, so that a final non-negative offset can be calculated before indexing.

src/recon.rs
randomPoison: Doesn't work with `slice_as` methods.

src/decode.rs
randomPoison: I'd rather use `pxstride` here, I think it makes it more clear that we're going from a byte offset to a pixel offset.

randomPoison: Ah wait, I see what you're suggesting across the two comments here. Yeah that makes sense, will change.

randomPoison: Actually this change causes the size of the buffer to be wrong. I'm not exactly sure why, but I don't think it's worth digging into when the existing logic is correct.

src/recon.rs
randomPoison: Doesn't work with the `slice_as` methods.

src/ipred_prepare.rs
kkysen: Oh, okay.  Damnit.  But yeah it's fine as is then.

src/recon.rs
kkysen: Ohh

kkysen (approved):

PR #997, opened 2024-04-23
`struct Rav1dRef`: Remove as it's now unused

kkysen:


include/dav1d/dav1d.rs
```diff
@@ -4,14 +4,13 @@ use crate::src::error::Rav1dError;
...
 use crate::src::log::Rav1dLogger;
-use crate::src::r#ref::Rav1dRef;
 use bitflags::bitflags;
 use std::ffi::c_int;
 use std::ffi::c_uint;
 use strum::FromRepr;
 
 pub type Dav1dContext = Rav1dContext;
-pub type Dav1dRef = Rav1dRef;
+pub type Dav1dRef = ();
```
randomPoison, line 13: Do we still need `Dav1dRef` if `Rav1dRef` is gone? Is this part of the public API in some way where we can't just remove it?

randomPoison (approved):

PR #998, opened 2024-04-23
CI: Pin version of macos to avoid build failure

thedataking:
It seems `macos-latest` transitioned (is transitioning) from `x86` to `aarch64`. This may lead to unintended interactions with caching of the Rust toolchain. We've observed runs that failed because `cargo` was not found. My best guess is that we restored a `x86` toolchain onto a `aarch64` host but reproducing to make sure would take too much time. Requesting `macos-13` instead of `macos-latest` should avoid this issue altogether.

### DIFF
--- a/.github/workflows/build-and-test-x86.yml
+++ b/.github/workflows/build-and-test-x86.yml
@@ -78,8 +78,8 @@ jobs:
           name: meson-test-logs-${{ matrix.target }}-${{ matrix.build.name }}
           path: |
              ${{ github.workspace }}/build/meson-logs/testlog-*.txt
-  test-on-macos-latest:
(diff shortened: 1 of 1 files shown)


randomPoison (approved):

kkysen:
Can we add a test for `aarch64` macOS, too?

thedataking:
> Can we add a test for `aarch64` macOS, too?

that's already covered in another workflow file (`build-and-test-aarch64-darwin.yml`).

kkysen:
> > Can we add a test for `aarch64` macOS, too?
> 
> that's already covered in another workflow file (`build-and-test-aarch64-darwin.yml`).

Ah, okay. Awesome then!

PR #1000, opened 2024-04-24
`struct Rav1dContext`: Make frame contexts a `Box<[_]>`

rinon:
Replaces the C allocation and `fc` array pointer of `Rav1dFrameContext` structures with a Rust boxed slice.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -3888,7 +3888,7 @@ pub(crate) unsafe fn rav1d_decode_tile_sbrow(
         );
     }
 
-    if frame_hdr.frame_type.is_inter_or_switch() && c.n_fc > 1 {
+    if frame_hdr.frame_type.is_inter_or_switch() && c.fc.len() > 1 {
         let sby = t.b.y - ts.tiling.row_start >> f.sb_shift;
         *f.lowest_pixel_mem.index_mut(ts.lowest_pixel + sby as usize) = [[i32::MIN; 2]; 7];
     }
@@ -4096,7 +4096,7 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
 
     let n_ts = frame_hdr.tiling.cols * frame_hdr.tiling.rows;
     if n_ts != f.n_ts {
-        if c.n_fc > 1 {
+        if c.fc.len() > 1 {
             // TODO: Fallible allocation
             f.frame_thread.tile_start_off.resize(n_ts as usize, 0);
         }
@@ -4109,7 +4109,7 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
         f.n_ts = n_ts;
     }
 
-    let a_sz = f.sb128w * frame_hdr.tiling.rows * (1 + (c.n_fc > 1 && c.tc.len() > 1) as c_int);
+    let a_sz = f.sb128w * frame_hdr.tiling.rows * (1 + (c.fc.len() > 1 && c.tc.len() > 1) as c_int);
     // TODO: Fallible allocation
     f.a.resize_with(a_sz as usize, Default::default);
 
@@ -4117,7 +4117,7 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
     let size_mul = &ss_size_mul[f.cur.p.layout];
     let seq_hdr = &***f.seq_hdr.as_ref().unwrap();
     let hbd = (seq_hdr.hbd != 0) as c_int;
-    if c.n_fc > 1 {
+    if c.fc.len() > 1 {
         let mut tile_idx = 0;
         let sb_step4 = f.sb_step as u32 * 4;
         for tile_row in 0..frame_hdr.tiling.rows {
@@ -4284,7 +4284,7 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
     // index this from the level type and can thus over-read by up to 3 bytes.
     f.lf.level
         .resize(num_sb128 as usize * 32 * 32 + 1, [0u8; 4]); // TODO: Fallible allocation
-    if c.n_fc > 1 {
+    if c.fc.len() > 1 {
         // TODO: Fallible allocation
         f.frame_thread
             .b
@@ -4346,7 +4346,7 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
             &f.refrefpoc,
             &f.ref_mvs,
             c.tc.len() as u32,
-            c.n_fc,
+            c.fc.len() as u32,
         )?;
     }
 
@@ -4421,7 +4421,7 @@ pub(crate) unsafe fn rav1d_decode_frame_init_cdf(
         rav1d_cdf_thread_copy(&mut f.out_cdf.cdf_write(), in_cdf);
     }
 
-    let uses_2pass = c.n_fc > 1;
+    let uses_2pass = c.fc.len() > 1;
 
     let tiling = &frame_hdr.tiling;
 
@@ -4590,7 +4590,7 @@ pub(crate) unsafe fn rav1d_decode_frame_exit(
         task_thread.error.store(0, Ordering::Relaxed);
     }
     let cf = f.frame_thread.cf.get_mut();
-    if c.n_fc > 1 && retval.is_err() && !cf.is_empty() {
+    if c.fc.len() > 1 && retval.is_err() && !cf.is_empty() {
         cf.fill_with(Default::default);
     }
     // TODO(kkysen) use array::zip when stable
@@ -4626,7 +4626,7 @@ pub(crate) unsafe fn rav1d_decode_frame_exit(
 }
 
 pub(crate) unsafe fn rav1d_decode_frame(c: &Rav1dContext, fc: &Rav1dFrameContext) -> Rav1dResult {
-    assert!(c.n_fc == 1);
+    assert!(c.fc.len() == 1);
     // if.tc.len() > 1 (but n_fc == 1), we could run init/exit in the task
     // threads also. Not sure it makes a measurable difference.
     let mut res = rav1d_decode_frame_init(c, fc);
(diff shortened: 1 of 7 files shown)


src/decode.rs
```diff
@@ -4674,23 +4674,20 @@ fn get_upscale_x0(in_w: c_int, out_w: c_int, step: c_int) -> c_int {
...
     // wait for c->out_delayed[next] and move into c->out if visible
-    let (fc, out, _task_thread_lock) = if c.n_fc > 1 {
+    let (fc, out, _task_thread_lock) = if c.fc.len() > 1 {
         let mut task_thread_lock = c.task_thread.delayed_fg.lock().unwrap();
         let next = c.frame_thread.next;
-        c.frame_thread.next += 1;
-        if c.frame_thread.next == c.n_fc {
-            c.frame_thread.next = 0;
-        }
+        c.frame_thread.next = (c.frame_thread.next + 1).wrapping_rem(c.fc.len() as u32);
```
kkysen, line 4680:
```suggestion
        c.frame_thread.next = (c.frame_thread.next + 1) % c.fc.len() as u32;
```

`.wrapping_rem` is a no-op.

src/internal.rs
```diff
@@ -367,6 +379,15 @@ pub struct Rav1dContext {
     pub(crate) picture_pool: *mut Rav1dMemPool,
 }
 
+impl Rav1dContext {
+    /// Iterates over all frame contexts in the `fc` buffer, starting at a given
+    /// index. The iterator wraps around to the start of the buffer. so all
+    /// contexts will be included.
+    pub(crate) fn fc_iter(&self, start: usize) -> impl Iterator<Item = &Rav1dFrameContext> {
+        self.fc.iter().skip(start).chain(self.fc.iter().take(start))
+    }
```
kkysen, line 388: Maybe it should be a `VecDeque`?  This should be fine for now, though.

src/internal.rs
```diff
@@ -783,6 +804,26 @@ pub(crate) struct Rav1dFrameContext_task_thread {
...
+impl Default for Rav1dFrameContext_task_thread {
+    fn default() -> Self {
+        Self {
+            lock: Default::default(),
+            cond: Default::default(),
+            ttd: Default::default(),
+            tasks: Default::default(),
+            init_done: Default::default(),
+            done: Default::default(),
+            retval: Mutex::new(Ok(())),
```
kkysen, line 816: nit just cause this `impl` feels annoying: we could make `retval` an `Option<Rav1dError>`.  Fine for now, though.

src/internal.rs
```diff
@@ -798,13 +839,26 @@ pub(crate) struct Rav1dFrameContext_frame_thread_progress {
...
 }
 
 impl Rav1dFrameContext {
+    pub(crate) unsafe fn zeroed(index: usize) -> Self {
+        Self {
+            index,
+            data: RwLock::new(unsafe { Rav1dFrameData::zeroed() }),
+            in_cdf: RwLock::new(CdfThreadContext::default()),
+            task_thread: Rav1dFrameContext_task_thread::default(),
+            frame_thread_progress: Rav1dFrameContext_frame_thread_progress::default(),
```
kkysen, line 858:
```suggestion
            in_cdf: Default::default(),
            task_thread: Default::default(),
            frame_thread_progress: Default::default(),
```

Can we just do this?  Is there a reason you did it with the specific types?

src/lib.rs
```diff
@@ -249,23 +249,11 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
...
-    ) as *mut Rav1dFrameContext;
-    if ((*c).fc).is_null() {
-        return error(c, c_out);
-    }
-    memset(
-        (*c).fc as *mut c_void,
-        0 as c_int,
-        ::core::mem::size_of::<Rav1dFrameContext>().wrapping_mul((*c).n_fc as usize),
-    );
+    (*c).fc = (0..n_fc).map(|i| Rav1dFrameContext::zeroed(i)).collect();
```
kkysen, line 252:
```suggestion
    // TODO fallible allocation
    (*c).fc = (0..n_fc).map(|i| Rav1dFrameContext::zeroed(i)).collect();
```

src/lib.rs
```diff
@@ -800,11 +773,10 @@ impl Drop for Rav1dContext {
...
-            let mut n_1: c_uint = 0 as c_int as c_uint;
-            while !(self.fc).is_null() && n_1 < self.n_fc {
-                let fc = &mut (*(self.fc).offset(n_1 as isize));
+            let fc_len = self.fc.len();
+            for fc in self.fc.iter_mut() {
                 let f = fc.data.get_mut().unwrap();
-                if self.n_fc > 1 as c_uint {
+                if fc_len > 1 {
                     let _ = mem::take(&mut f.lowest_pixel_mem); // TODO: remove when context is owned
                 }
```
kkysen, line 781:
```suggestion
                let _ = mem::take(&mut f.lowest_pixel_mem); // TODO: remove when context is owned
```

We can avoid the separate `fc_len` variable by just unconditionally taking it.  When `fc.len() == 0`, `f.lowest_pixel_mem` is `Default::default()` already.

src/obu.rs
```diff
@@ -2537,12 +2537,9 @@ unsafe fn parse_obus(
                 let mut task_thread_lock = c.task_thread.delayed_fg.lock().unwrap();
                 // Need to append this to the frame output queue.
                 let next = c.frame_thread.next;
-                c.frame_thread.next += 1;
-                if c.frame_thread.next == c.n_fc {
-                    c.frame_thread.next = 0;
-                }
+                c.frame_thread.next = (c.frame_thread.next + 1).wrapping_rem(c.fc.len() as u32);
```
kkysen, line 2540:
```suggestion
                c.frame_thread.next = (c.frame_thread.next + 1) % c.fc.len() as u32;
```

src/thread_task.rs
```diff
@@ -56,14 +56,12 @@ pub const TILE_ERROR: i32 = i32::MAX - 1;
 /// this special case.
 #[inline]
 unsafe fn reset_task_cur(c: &Rav1dContext, ttd: &TaskThreadData, mut frame_idx: c_uint) -> c_int {
-    unsafe fn curr_found(c: &Rav1dContext, ttd: &TaskThreadData, first: u32) -> c_int {
-        let mut i: c_uint = ttd.cur.load(Ordering::Relaxed);
-        while i < c.n_fc {
-            (*(*(c.fc).offset(first.wrapping_add(i).wrapping_rem(c.n_fc) as isize))
+    unsafe fn curr_found(c: &Rav1dContext, ttd: &TaskThreadData, first: usize) -> c_int {
+        for i in ttd.cur.load(Ordering::Relaxed) as usize..c.fc.len() {
+            (*c.fc[(first + i).wrapping_rem(c.fc.len())]
```
kkysen, line 61:
```suggestion
            (*c.fc[(first + i) % c.fc.len()]
```

src/thread_task.rs
```diff
@@ -89,39 +87,35 @@ unsafe fn reset_task_cur(c: &Rav1dContext, ttd: &TaskThreadData, mut frame_idx:
...
-    while ttd.cur.load(Ordering::Relaxed) < c.n_fc {
-        if (*(*(c.fc).offset(
-            first
-                .wrapping_add(ttd.cur.load(Ordering::Relaxed))
-                .wrapping_rem(c.n_fc) as isize,
-        ))
-        .task_thread
-        .tasks())
+    while (ttd.cur.load(Ordering::Relaxed) as usize) < c.fc.len() {
+        if (*c.fc[((first + ttd.cur.load(Ordering::Relaxed)) as usize).wrapping_rem(c.fc.len())]
```
kkysen, line 108:
```suggestion
        if (*c.fc[((first + ttd.cur.load(Ordering::Relaxed)) as usize) % c.fc.len()]
```

src/thread_task.rs
```diff
@@ -804,13 +794,11 @@ pub unsafe fn rav1d_worker_task(c: &Rav1dContext, task_thread: Arc<Rav1dTaskCont
...
-                let fc = &mut *(c.fc).offset(
-                    first_0
-                        .wrapping_add(ttd.cur.load(Ordering::Relaxed))
-                        .wrapping_rem(c.n_fc) as isize,
-                );
+            while (ttd.cur.load(Ordering::Relaxed) as usize) < c.fc.len() {
+                let fc = c
+                    .fc_iter(ttd.first.load(Ordering::SeqCst) as usize)
+                    .nth(ttd.cur.load(Ordering::Relaxed) as usize)
+                    .unwrap();
```
kkysen, line 801: I think `%` is probably more efficient here than `.nth` on `Chain`, and we use `%` and indexing in the other cases.  `.fc_iter` is better when we iterate through all of them.

```suggestion
                let first = ttd.first.load(Ordering::SeqCst);
                let fc = &c.fc[(first + ttd.cur.load(Ordering::Relaxed)) as usize % fc.len()];
```

kkysen (changes requested):
Generally LGMT.  Only little things are that we should just use `%` instead of `.wrapping_rem`.  `.wrapping_rem` is only there since other operations have `.wrapping_*` versions, but this one does nothing extra.  Also, for the place where you did `.fc_iter` and `.nth`, I think we should do `%` and indexing like is done for the other places, and is done in C.

src/lib.rs
```diff
@@ -284,10 +272,9 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
     addr_of_mut!((*c).cache).write(Default::default());
     addr_of_mut!((*c).refs).write(Default::default());
     for n in 0..n_fc {
-        let fc = &mut *((*c).fc).offset(n as isize);
+        let fc = &mut (*c).fc[n];
         addr_of_mut!(fc.data).write(RwLock::new(Rav1dFrameData::zeroed()));
```
kkysen, line 276:
```suggestion
```

src/lib.rs
```diff
@@ -284,10 +272,9 @@ pub(crate) unsafe fn rav1d_open(c_out: &mut *mut Rav1dContext, s: &Rav1dSettings
...
         let f = fc.data.get_mut().unwrap();
-        f.index = n;
         addr_of_mut!(fc.in_cdf).write(RwLock::new(Default::default()));
         addr_of_mut!(fc.task_thread.tasks).write(UnsafeCell::new(Default::default()));
         addr_of_mut!(fc.task_thread.retval).write(Mutex::new(Ok(())));
```
kkysen, line 280:
```suggestion
```

I can't comment on the next lines, but all of the lines up to 302/289 can be removed, all of the ones initializing `fc` since that's already done on construction now.

src/internal.rs
rinon: Yeah, but not important for now.

src/internal.rs
rinon: I don't really want it to be a Vec at all, it's fixed-size after initialization. I think I'd rather have this bit of weirdness, we only wrap in a couple of places.

src/lib.rs
rinon: Eh, we're going to rewrite all of this when the context is fully owned.

src/thread_task.rs
rinon: Efficiency with very small numbers of elements here isn't going to matter, but sure.

src/internal.rs
```diff
@@ -879,9 +930,13 @@ pub(crate) struct Rav1dFrameData {
...
+        addr_of_mut!((*data.as_mut_ptr()).refp).write(Default::default());
+        addr_of_mut!((*data.as_mut_ptr()).cur).write(Default::default());
+        addr_of_mut!((*data.as_mut_ptr()).sr_cur).write(Default::default());
         addr_of_mut!((*data.as_mut_ptr()).out_cdf).write(Default::default());
         addr_of_mut!((*data.as_mut_ptr()).tiles).write(vec![]);
         addr_of_mut!((*data.as_mut_ptr()).a).write(vec![]);
+        addr_of_mut!((*data.as_mut_ptr()).rf).write(Default::default());
         addr_of_mut!((*data.as_mut_ptr()).lowest_pixel_mem).write(Default::default());
         addr_of_mut!((*data.as_mut_ptr()).frame_thread).write(Default::default());
         addr_of_mut!((*data.as_mut_ptr()).lf).write(Default::default());
```
kkysen, line 942: I'm not sure which PR added this, but this is the exact opposite of how we should be initializing things.  If done this way, we can trivially forget to initialize a field.  If we initialize directly, we can't forget a field.  The remaining individual fields should be `mem::zeroed()`.

kkysen (commented):
I think you might've missed a few of the changes.  The `.wrapping_rem`s are still there.

src/internal.rs
rinon: Sure, but I thought you were doing that initialization change so I didn't want to alter anything substantially here.

kkysen: Okay then that's fine. I can do it.

kkysen (approved):

PR #1001, opened 2024-04-24
Backport faster T.35 payload copy from dav1d 1.2.1

fbossen:


src/getbits.rs
```diff
@@ -197,4 +197,11 @@ impl<'a> GetBits<'a> {
     pub const fn has_pending_bits(&self) -> bool {
         self.state != 0 || self.bits_left != 0
     }
+
+    pub fn get_bytes(&mut self, n: usize) -> Box<[u8]> {
```
kkysen, line 201:
```suggestion
    pub fn get_bytes(&mut self, n: usize) -> &[u8] {
```

src/getbits.rs
```diff
@@ -197,4 +197,11 @@ impl<'a> GetBits<'a> {
...
+        assert_eq!(self.bits_left, 0);
+        let i = self.index;
+        self.index += n;
+        self.data[i..][..n].into() // TODO(kkysen) fallible allocation
```
kkysen, line 205:
```suggestion
        &self.data[i..][..n].into()
```

src/obu.rs
```diff
@@ -2462,7 +2462,7 @@ unsafe fn parse_obus(
                     } else {
                         let country_code = country_code as u8;
                         let country_code_extension_byte = country_code_extension_byte as u8;
-                        let payload = (0..payload_size).map(|_| gb.get_bits(8) as u8).collect(); // TODO(kkysen) fallible allocation
+                        let payload = gb.get_bytes(payload_size as usize);
```
kkysen, line 2465:
```suggestion
                        // TODO fallible allocation
                        let payload = gb.get_bytes(payload_size as usize).into();
```

src/getbits.rs
kkysen: I think it's better to leave it up to the caller if they want to allocate or not.

kkysen (changes requested):
I'd change `fn get_bytes` to not allocate and allocate in the caller instead, but otherwise LGTM.

src/getbits.rs
fbossen: Agreed

kkysen (approved):

PR #1002, opened 2024-04-24
Backport type changes for `time_scale`, `num_units_in_tick`, and `num_units_in_decoding_tick` from dav1d 1.3.0

fbossen:


include/dav1d/picture.rs
```diff
@@ -82,7 +82,7 @@ pub struct Dav1dPicture {
     pub mastering_display: Option<NonNull<Rav1dMasteringDisplay>>,
     pub itut_t35: Option<NonNull<Dav1dITUTT35>>,
     pub n_itut_t35: usize,
-    pub reserved: [uintptr_t; 3],
+    pub reserved: [uintptr_t; 4],
```
kkysen, line 85: I'm curious where this extra room came from.

kkysen (approved):

kkysen (commented):
Oh, actually could you add doc comments to each saying they're all `> 0`?  That's what their commit message says, and that'd be helpful to know later if we ever want to make them `NonZero*` types.

include/dav1d/picture.rs
fbossen: Don't know. This was reduced to 3 when `n_itut_t35` was added and now it's changed back to 4.

kkysen: Yeah I saw it was reduced then which made sense because the new field used up one of them, but the change here doesn't change the size.

kkysen (approved):

PR #1003, opened 2024-04-25
Backport parser overflow fix from dav1d 1.2.1

fbossen:


src/obu.rs
```diff
@@ -2362,18 +2320,17 @@ unsafe fn parse_obus(
                 // There's no trailing bit at the end to skip,
                 // but we do need to align to the next byte.
                 gb.bytealign();
-                parse_tile_grp(c, r#in, props, &mut gb, init_bit_pos, init_byte_pos, len)?;
+                parse_tile_grp(c, r#in, props, &mut gb)?;
```
kkysen, line 2323:
```suggestion
                parse_tile_grp(c, r#in, props, gb)?;
```

Now that `gb: &mut GetBits` instead of `gb: GetBits`, we can pass just `gb` instead of `&mut gb` or `&gb` everywhere we do in `fn parse_obus`.

src/obu.rs
```diff
@@ -2437,16 +2394,13 @@ unsafe fn parse_obus(
                     })); // TODO(kkysen) fallible allocation
                 }
                 Some(ObuMetaType::ItutT32) => {
-                    let mut payload_size = len as c_int;
+                    let mut payload_size = gb.get_len() - gb.pos() / 8;
```
kkysen, line 2397: Can we add a `fn remaining_len` for this?

src/obu.rs
```diff
@@ -2098,40 +2098,13 @@ fn parse_tile_hdr(tiling: &Rav1dFrameHeader_tiling, gb: &mut GetBits) -> Rav1dTi
...
-
-    0
-}
-
 unsafe fn parse_obus(
     c: &mut Rav1dContext,
     r#in: &CArc<[u8]>,
     props: &Rav1dDataProps,
+    mut gb: &mut GetBits,
 ) -> Rav1dResult<usize> {
```
kkysen, line 2106: Now that `gb` is from the caller, we could remove the `usize` from the return, as the caller can just call `gb.len()`.

src/obu.rs
```diff
@@ -2206,24 +2168,20 @@ unsafe fn parse_obus(
...
             return Err(EINVAL);
         }
 
         // The current bit position is a multiple of 8
         // (because we just aligned it) and less than `8 * pkt_bytelen`
         // because otherwise the overrun check would have fired.
-        let pkt_bytelen = init_byte_pos + len;
+        let pkt_bytelen = gb.get_len();
         let bit_pos = gb.pos();
         assert!(bit_pos & 7 == 0);
```
kkysen, line 2184: This is deleted in C now, right?

src/obu.rs
```diff
@@ -2206,24 +2168,20 @@ unsafe fn parse_obus(
...
-        assert!(pkt_bytelen >= bit_pos >> 3);
         let mut data = r#in.clone();
         data.slice_in_place(..pkt_bytelen);
         data.slice_in_place(bit_pos >> 3..);
```
kkysen, line 2187:
```suggestion
        data.slice_in_place(..gb.len());
        data.slice_in_place(gb.byte_pos()..);
```

src/getbits.rs
```diff
@@ -204,4 +208,19 @@ impl<'a> GetBits<'a> {
...
+
+    #[inline]
+    pub fn get_len(&self) -> usize {
+        self.data.len()
+    }
+
+    #[inline]
+    pub fn at(&self, offset: usize) -> u8 {
+        self.data[self.index + offset]
+    }
```
kkysen, line 225: Can this be an `impl Index for GetBits` instead?

kkysen (changes requested):

src/obu.rs
kkysen: Do you think slicing with the index/pos first and then remaining len is easier to understand?  I thought full length and then index was simpler, but if you think it's easier to understand this way, that's fine, too.

kkysen (approved):

src/obu.rs
fbossen: When slicing, I believe we most often have a pattern like `&slice[pos..][..len]`, which is what I replicated here. It might be easier to understand because it is a more common pattern. Let me know if you think otherwise.

kkysen: I had thought the other way around was simpler, but that's just me.  This way's fine, too, since you feel it's easier to understand.

Issue #505, opened 2023-10-01
Backport: Reduce the size of some public structs


kkysen:
We should backport https://code.videolan.org/videolan/dav1d/-/merge_requests/1540.  This reduces the size of a bunch of the public types, so our conversions will be faster/smaller, and it makes many of them more Rusty, to the point where we can make the `Rav1d*` and `Dav1d*` types the same for some of them (e.x. `Dav1dContentLightLevel`).

This also means we will target compatibility with [`DAV1D_API_VERSION`](https://github.com/memorysafety/rav1d/blob/05daedaa975ca87f39160566442a7199f0ff4836/meson.build#L33-L37) `7.0.0`.

For `Rav1d*` types that haven't been rustified yet, we should backport these changes to the `Rav1d*` types, too.

thedataking:
> This also means we will target compatibility with dav1d 7.0.0.

I'm confused. Isn't dav1d currently on version 1.2.1?

kkysen:
> > This also means we will target compatibility with dav1d 7.0.0.
> 
> I'm confused. Isn't dav1d currently on version 1.2.1?

Hmm, I might've misread that.  They updated the version of something.  It might be supposed to be `dav1d` 1.3.0 instead.

kkysen:
> > This also means we will target compatibility with dav1d 7.0.0.
> 
> I'm confused. Isn't dav1d currently on version 1.2.1?

@thedataking, I think there's a separate `DAV1D_API_VERSION` from the `dav1d` version.  We're currently on `6.8.0` (see [`meson.build`](https://github.com/memorysafety/rav1d/blob/05daedaa975ca87f39160566442a7199f0ff4836/meson.build#L33-L37))

PR #1004, opened 2024-04-25
`struct Dav1dSequenceHeader`: backport size reduction from dav1d 1.3.0

fbossen:
* Fixes `Dav1dSequenceHeader` of #505.

Shrinks `Rav1dSequenceHeader` from 1468 to 784 bytes.


src/picture.rs
```diff
@@ -160,7 +160,7 @@ impl Rav1dPicAllocator {
                 w,
                 h,
                 layout: seq_hdr.layout,
-                bpc: 8 + 2 * seq_hdr.hbd,
+                bpc: 8 + 2 * seq_hdr.hbd as c_int,
```
kkysen, line 163:
```suggestion
                bpc: 8 + 2 * seq_hdr.hbd,
```

We can change `Rav1dPictureParameters::bpc` to `u8` instead.

src/cdef_apply.rs
```diff
@@ -196,7 +196,7 @@ pub(crate) unsafe fn rav1d_cdef_brow<BD: BitDepth>(
     let uv_dir: &[u8; 8] = &UV_DIRS[(layout == Rav1dPixelLayout::I422) as usize];
 
     let have_tt = c.tc.len() > 1;
-    let sb128 = f.seq_hdr.as_ref().unwrap().sb128;
+    let sb128 = f.seq_hdr.as_ref().unwrap().sb128 as c_int;
```
kkysen, line 199:
```suggestion
    let sb128 = f.seq_hdr.as_ref().unwrap().sb128;
```

Can you move the `as c_int` casts to where it's needed?

src/decode.rs
```diff
@@ -4738,7 +4738,7 @@ pub unsafe fn rav1d_submit_frame(c: &mut Rav1dContext) -> Rav1dResult {
         f.task_thread.finished.store(true, Ordering::SeqCst);
     }
 
-    let bpc = 8 + 2 * seq_hdr.hbd;
+    let bpc = 8 + 2 * seq_hdr.hbd as c_int;
     match Rav1dDSPContext::get(bpc) {
```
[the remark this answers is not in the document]
fbossen: `Rav1dDSPContext::get` is called in other places where `bpc` is not derived from `seq_hdr.hbd` and is a `c_int`. I would rather not change `Rav1dDSPContext::get` in this PR.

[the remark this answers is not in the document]
kkysen: It should all work out if `Rav1dPictureParameters::bpc` is also made a `u8` as suggested here: https://github.com/memorysafety/rav1d/pull/1004#discussion_r1579285675.  I think it makes sense to propagate the type changes as far as possible in this PR, where the initial type changes occur.  I can do it if you'd prefer; it's not much of a change.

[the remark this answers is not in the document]
fbossen: Ok. I made the change in a separate commit

src/decode.rs
```diff
@@ -3992,7 +3992,7 @@ pub(crate) unsafe fn rav1d_decode_tile_sbrow(
                 let x1 = (4 * (t.b.x + sb_step) * d >> ss_hor) + rnd >> shift;
 
                 for x in x0..cmp::min(x1, n_units) {
-                    let px_x = x << unit_size_log2 + ss_hor;
+                    let px_x = x << unit_size_log2 + ss_hor as u8;
```
kkysen, line 3995:
```suggestion
                    let px_x = x << unit_size_log2 + ss_hor;
```

I think this is redundant since you already made `ss_hor` a `u8` above.

kkysen (approved):

src/decode.rs
fbossen: Must be a different `ss_hor`. The one here is a `c_int`.

PR #1007, opened 2024-04-27
`struct Dav1dFrameHeader`: backport memory size reduction from dav1d 1.3.0

fbossen:
* Fixes `Dav1dFrameHeader` of #505.

Shrinks `Rav1dFrameHeader` from 1664 to 1120 bytes.

### DIFF
--- a/include/dav1d/headers.rs
+++ b/include/dav1d/headers.rs
@@ -64,7 +64,7 @@ pub const DAV1D_MAX_TILE_COLS: usize = 64;
 pub const DAV1D_MAX_TILE_ROWS: usize = 64;
 pub const DAV1D_MAX_SEGMENTS: u8 = 8;
 pub const DAV1D_NUM_REF_FRAMES: usize = 8;
-pub const DAV1D_PRIMARY_REF_NONE: c_int = 7;
+pub const DAV1D_PRIMARY_REF_NONE: u8 = 7;
 pub const DAV1D_REFS_PER_FRAME: usize = 7;
 pub const DAV1D_TOTAL_REFS_PER_FRAME: usize = DAV1D_REFS_PER_FRAME + 1;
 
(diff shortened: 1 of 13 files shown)


src/obu.rs
```diff
@@ -704,8 +704,8 @@ unsafe fn parse_refidx(
                         .frame_hdr
                         .as_ref()
                         .ok_or(EINVAL)?
-                        .frame_offset,
-                    frame_offset,
+                        .frame_offset as c_int,
+                    frame_offset as c_int,
```
kkysen, line 708: We could make `fn get_poc_diff` and all of the `poc` fields/vars `u8`s.  I can add that if you want (I checked to see if it it's possible).

kkysen (changes requested):
Mostly good, I just have some small comments about casts.  And I think `sharpness` is supposed to be a `u8` like in C.

include/dav1d/headers.rs
```diff
@@ -2276,25 +2276,25 @@ impl From<Rav1dFrameHeader_delta> for Dav1dFrameHeader_delta {
...
-    pub mode_ref_delta_enabled: c_int,
-    pub mode_ref_delta_update: c_int,
+    pub level_y: [u8; 2],
+    pub level_u: u8,
+    pub level_v: u8,
+    pub mode_ref_delta_enabled: u8,
+    pub mode_ref_delta_update: u8,
     pub mode_ref_deltas: Rav1dLoopfilterModeRefDeltas,
-    pub sharpness: c_int,
+    pub sharpness: i8,
```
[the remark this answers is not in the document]
fbossen: To simplify comparing/assigning to `last_sharpness` which is `i8`

src/obu.rs
```diff
@@ -900,21 +898,21 @@ fn parse_tiling(
...
+        log2_rows = tile_log2(1, rows.into());
     }
     col_start_sb[cols as usize] = sbw as u16;
     row_start_sb[rows as usize] = sbh as u16;
     let update;
     let n_bytes;
     if log2_cols != 0 || log2_rows != 0 {
-        update = gb.get_bits(log2_cols + log2_rows) as c_int;
-        if update >= cols * rows {
+        update = gb.get_bits((log2_cols + log2_rows).into()) as u16;
```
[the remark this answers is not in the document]
fbossen: Yes. `log_2_{cols,rows}` should be `<= 6` because `RAV1D_MAX_TILE_{COLS,ROWS}` is 64.

src/obu.rs
fbossen: Yes, but that's really beyond the scope of this PR. I'll let you create a new one.

include/dav1d/headers.rs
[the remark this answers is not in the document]
kkysen: Hmm, maybe it should be an `Option`.  This is good for now, though.

kkysen (approved):

src/decode.rs
```diff
@@ -358,7 +359,7 @@ unsafe fn read_tx_tree(
     };
 }
 
-fn neg_deinterleave(diff: c_int, r#ref: c_int, max: c_int) -> c_int {
+fn neg_deinterleave(diff: u8, r#ref: u8, max: u8) -> u8 {
```
rinon, line 362: We can't change these to unsigned, `max - 1` may be negative.

fbossen: Does #1056 solve it?

PR #1018, opened 2024-05-02
`struct Av1BlockInterNd`: Make a `zerocopy` `"union"`

### LINKED ISSUE #854: `levels.rs`: Make unions safe
We define a handful of unions in `levels.rs` that are used as part of `Av1Block`. These should be made into a safe Rust equivalent, likely an enum, and the various helper methods for accessing specific fields should be made safe.

kkysen:
* Fixes #854.

I originally tried to make this a `union`, but that would change the size from 12 to 14 bytes, which wasn't great.  There should be enough niches available for this, but it'd require an `i14` type, which we can't define in stable Rust.  Instead, I made it a `zerocopy` "union".  To do this, I had to remove the padding and invalid bits in `Av1BlockInter1d`.  The invalid bits in `enum InterIntraPredMode` are done by storing it as a `MaskedInterIntraPredMode` (a `u8`) and then masking the `u8` and converting to an `InterIntraPredMode`, which is just an `& 3`.  For actual `zerocopy` "union", `Av1BlockInterNd`, I decided to store `Av1BlockInter1d` as the field instead of a byte array because 1) `one_d` is accessed much more often, so this is convenient, and 2) `Av1BlockInter{1,2}d` have identical sizes and alignments, so `FromBytes::ref_from(AsBytes::as_bytes(...)).unwrap()` is a no-op.  In contrast, a byte array would have a lower alignment.


src/levels.rs
```diff
@@ -316,27 +321,101 @@ pub struct Av1BlockIntra {
...
+/// Really an [`InterIntraPredMode`],
+/// but stored as a `u8` so there are no invalid bits.
+#[derive(Clone, Copy, PartialEq, Eq, FromZeroes, FromBytes, AsBytes)]
+#[repr(transparent)]
+pub struct MaskedInterIntraPredMode(u8);
+
+impl MaskedInterIntraPredMode {
+    pub const fn get(self) -> InterIntraPredMode {
+        // Should just be an `& 3`.
+        match InterIntraPredMode::from_repr(self.0 as usize % InterIntraPredMode::COUNT) {
```
randomPoison, line 333: Why do we need to do `% InterIntraPredMode::COUNT`? Isn't `self.0` just the integer representation for `InterIntraPredMode`?

src/decode.rs
```diff
@@ -2936,7 +2921,13 @@ unsafe fn decode_b(
                                 t.warpmv.matrix[4],
                                 t.warpmv.matrix[5] - 0x10000,
                             ]
-                            .map(|it| it as i16)
+                            .map(|coef| {
+                                let coef = coef as i16;
+                                // warped matrix coefs are at least `i14`s.
+                                debug_assert!(coef < (1 << 13));
+                                debug_assert!(coef >= -(1 << 13));
+                                coef
```
randomPoison, line 2929: I'm not sure I understand the `i14` thing. Why do we need these asserts now if they weren't necessary before?

randomPoison (approved):

src/levels.rs
kkysen: The `u8`, since it's `FromBytes`, could be anything.  The `%` truncates off any invalid values so we only use the lower 2 bits that are the `InterIntraPredMode`.  Plus, even if the other bits are all `0` (if we use the `zerocopy` `union` correctly and don't actually reinterpret bytes), the compiler doesn't know this so it'll have to include a check and panic here, but by wrapping/truncating it in bounds, the compiler can elide the check.  Thus, it's just an `& 3`, which is essentially free.

src/decode.rs
kkysen: We don't actually need this, but I had added it initially because I was looking for more niches so that I could make things an `enum` without adding extra bytes to these 12-byte types, and I saw in the spec that these warped matrix coefs have these bounds.  It turned out I couldn't exploit those niches, so I ended up going the `zerocopy` route instead, but I thought it could be useful doc-wise to leave these in since they're debug only.

PR #1027, opened 2024-05-02
`x86` asm: backport changes from dav1d 1.4.0

fbossen:


src/loopfilter.rs
```diff
@@ -1054,10 +1054,13 @@ impl Rav1dLoopFilterDSPContext {
...
 
-                    self.loop_filter_sb[0][0] = dav1d_lpf_h_sb_y_16bpc_avx512icl;
                     self.loop_filter_sb[0][1] = dav1d_lpf_v_sb_y_16bpc_avx512icl;
-                    self.loop_filter_sb[1][0] = dav1d_lpf_h_sb_uv_16bpc_avx512icl;
                     self.loop_filter_sb[1][1] = dav1d_lpf_v_sb_uv_16bpc_avx512icl;
+
+                    if !flags.contains(CpuFlags::SLOW_GATHER) {
+                        self.loop_filter_sb[0][0] = dav1d_lpf_h_sb_y_16bpc_avx512icl;
+                        self.loop_filter_sb[1][0] = dav1d_lpf_h_sb_uv_16bpc_avx512icl;
+                    }
```
kkysen, line 1063: We need to change this for `BPC8`, too, right?

kkysen (changes requested):

src/loopfilter.rs
fbossen: Indeed.

kkysen (approved):

PR #1030, opened 2024-05-03
`fn MsacContext::new`: Make `fn rav1d_msac_init` a method returning `Self` directly and make `data` a slice

kkysen:
I was trying to make `fn new` totally safe by adding a lifetime to `MsacContext`, but I'm not sure how to do that yet (it is a slice from a `Rav1dData`).  Since @randomPoison is working on making `ts: *mut Rav1dTileState` safe right now and `MsacContext` is a field of `Rav1dTileState`, I'll leave that part for later.

### DIFF
--- a/src/msac.rs
+++ b/src/msac.rs
@@ -7,6 +7,7 @@ use std::ffi::c_int;
 use std::ffi::c_uint;
 use std::mem;
 use std::ops::Range;
+use std::slice;
 
 #[cfg(all(feature = "asm", target_feature = "sse2"))]
 extern "C" {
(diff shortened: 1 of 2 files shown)


randomPoison (approved):

src/msac.rs
```diff
@@ -399,34 +395,32 @@ fn rav1d_msac_decode_hi_tok_rust(s: &mut MsacContext, cdf: &mut [u16; 4]) -> c_u
...
-            s.symbol_adapt16 = dsp.symbol_adapt16;
-        }
-        #[cfg(not(target_arch = "x86_64"))]
-        {
-            let _ = dsp.symbol_adapt16;
-        }
+impl MsacContext {
+    /// # Safety
+    ///
+    /// `data` must live longer than all of the other functions called on the returned [`Self`].
```
rinon, line 401: Why can't MsacContext just have a `PhantomData<&'a [u8]>` and a lifetime?

kkysen: Since it's stored in other types and would make lifetimes really complex to a level I'm not sure how to model.

PR #1031, opened 2024-05-03
`struct Rav1dTileStateContext`: Put `cdf` and `msac` in a `Mutex`

randomPoison:
This is the first step in giving `Rav1dTileState` interior mutability, which is necessary to make the `ts` allocation safe.

The context fields of `Rav1dTileState` (`msac` and `cdf`) are accessed exclusively from one thread at a time, so I've put them in a `Mutex` to allow us to still gain mutable access once the tile state is immutably shared between threads. This required reworking how the data is accessed, as the original logic was accessing the fields at multiple points in the callstacks, which would have resulted in lock conflicts when a function further down in the callstack would try to acquire a lock that's already held higher up. I've changed the code to pass down the mutable reference, ensuring that the lock is only held once. For a couple of functions this meant taking an `Option<&mut Rav1dTileStateContext>` in cases where the function only needs access to the context object in some code paths.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -69,6 +69,7 @@ use crate::src::internal::Rav1dFrameContext;
 use crate::src::internal::Rav1dFrameData;
 use crate::src::internal::Rav1dTaskContext;
 use crate::src::internal::Rav1dTileState;
+use crate::src::internal::Rav1dTileStateContext;
 use crate::src::internal::ScalableMotionParams;
 use crate::src::internal::ScratchPal;
 use crate::src::internal::TileStateRef;
@@ -212,29 +213,24 @@ fn init_quant_tables(
 }
 
 unsafe fn read_mv_component_diff(
-    t: &mut Rav1dTaskContext,
     f: &Rav1dFrameData,
+    msac: &mut MsacContext,
     mv_comp: &mut CdfMvComponent,
     have_fp: bool,
 ) -> c_int {
-    let ts = &mut *f.ts.offset(t.ts as isize);
     let have_hp = f.frame_hdr.as_ref().unwrap().hp;
-    let sign = rav1d_msac_decode_bool_adapt(&mut ts.msac, &mut mv_comp.sign.0);
-    let cl = rav1d_msac_decode_symbol_adapt16(&mut ts.msac, &mut mv_comp.classes.0, 10);
+    let sign = rav1d_msac_decode_bool_adapt(msac, &mut mv_comp.sign.0);
+    let cl = rav1d_msac_decode_symbol_adapt16(msac, &mut mv_comp.classes.0, 10);
     let mut up;
     let fp;
     let hp;
 
     if cl == 0 {
-        up = rav1d_msac_decode_bool_adapt(&mut ts.msac, &mut mv_comp.class0.0) as c_uint;
+        up = rav1d_msac_decode_bool_adapt(msac, &mut mv_comp.class0.0) as c_uint;
         if have_fp {
-            fp = rav1d_msac_decode_symbol_adapt4(
-                &mut ts.msac,
-                &mut mv_comp.class0_fp[up as usize],
-                3,
-            );
+            fp = rav1d_msac_decode_symbol_adapt4(msac, &mut mv_comp.class0_fp[up as usize], 3);
             hp = if have_hp {
-                rav1d_msac_decode_bool_adapt(&mut ts.msac, &mut mv_comp.class0_hp.0)
+                rav1d_msac_decode_bool_adapt(msac, &mut mv_comp.class0_hp.0)
             } else {
                 true
             };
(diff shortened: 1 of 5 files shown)


src/msac.rs
```diff
@@ -159,6 +160,22 @@ pub struct MsacContext {
...
+        Self {
+            buf_pos: ptr::null(),
+            buf_end: ptr::null(),
+            dif: Default::default(),
+            rng: Default::default(),
+            cnt: Default::default(),
+            allow_update_cdf: Default::default(),
+
+            #[cfg(all(feature = "asm", target_arch = "x86_64"))]
+            symbol_adapt16: rav1d_msac_decode_symbol_adapt_c,
```
kkysen, line 174:
```suggestion
            symbol_adapt16: Rav1dMsacDSPContext::default().symbol_adapt16,
```

Just to centralize the logic.

src/recon.rs
```diff
@@ -1957,6 +1985,7 @@ pub(crate) unsafe fn rav1d_read_coef_blocks<BD: BitDepth>(
                             read_coef_tree::<BD>(
                                 f,
                                 t,
+                                Some(ts_c),
```
kkysen, line 1988: It looks like we're only ever passing `Some` as I don't see any `None`.  Why does it need to be an `Option` then?

src/recon.rs
```diff
@@ -2752,7 +2785,10 @@ pub(crate) unsafe fn rav1d_recon_b_intra<BD: BitDepth>(
                             if debug_block_info!(f, t.b) {
                                 println!(
                                     "Post-y-cf-blk[tx={},txtp={},eob={}]: r={}",
-                                    intra.tx as c_int, txtp as c_uint, eob, ts.msac.rng,
+                                    intra.tx as c_int,
+                                    txtp as c_uint,
+                                    eob,
+                                    ts_c.as_deref().unwrap().msac.rng,
```
kkysen, line 2791: For all of these `.as_deref()`s and `.as_deref_mut()`s, I think they can just be `.as_ref()` and `.as_mut()`s (maybe).  `.as_deref()` is like `.as_ref()` but is also doing a deref, so it's helpful from going from `Option<String>` to `Option<&str>` instead of `Option<&String>`, but I don't think we need that extra deref here.

kkysen (changes requested):
Overall this is really good, and it's much nicer to pass this mostly on the callstack vs. a field, as it'll make adding a lifetime to `MsacContext` a whole lot easier.  I do have some questions, though, namely, why does it need to be an `Option` if we're only ever passing `Some`?

src/recon.rs
randomPoison: We do pass `None` in one specific case: Towards the beginning of `decode_b`, in the `if t.frame_thread.pass == 2` block where we call `recon_b_intra` and `recon_b_inter` [we pass `None` into those functions](https://github.com/memorysafety/rav1d/blob/b41ebcad9d5505eb55c6f788be17f479aa84594f/src/decode.rs#L1227). That's the specific case where the tile state is shared between threads, but only in one branch of execution needs mutable access to the context objects.

src/recon.rs
kkysen: Ahh, okay then your way should be fine.  Didn't realize we'd get the double reference with my way.

src/decode.rs
```diff
@@ -4267,6 +4351,12 @@ pub(crate) unsafe fn rav1d_decode_frame_init(
             return Err(ENOMEM);
         }
         f.n_ts = n_ts;
+
+        for index in 0..n_ts as usize {
+            let ts = &mut *f.ts.add(index);
+            // TODO: Safe initialization.
+            addr_of_mut!(ts.context).write(Mutex::new(Default::default()));
```
kkysen, line 4358: It's not sound to have the intermediate `&mut Rav1dTileState` here if it's uninitialized, but since this gets fixed right away in #1032, it's fine here.

src/decode.rs
```diff
@@ -4588,8 +4678,8 @@ pub(crate) unsafe fn rav1d_decode_frame_init_cdf(
         let end: usize = tile.hdr.end.try_into().unwrap();
 
         let mut data = tile.data.as_ref();
-        for (j, (ts, tile_start_off)) in iter::zip(
-            slice::from_raw_parts_mut(f.ts, end + 1),
+        for (j, (ts_idx, tile_start_off)) in iter::zip(
+            0..end + 1,
             if uses_2pass {
                 &f.frame_thread.tile_start_off[..end + 1]
             } else {
```
kkysen, line 4685: `j` is the same as `ts_idx` here, so we can simplify this a lot:

```rust
for (ts_idx, tile_start_off) in if uses_2pass {
    &f.frame_thread.tile_start_off
} else {
    &[]
}
.into_iter()
.map(|&it| it as usize)
.chain(iter::repeat(0))
.take(end + 1)
.enumerate()
.skip(start)
```

and then use `ts_idx` instead of `j`.

That said, why can't we keep the old code here?  We're still indexing into `f.ts` to get `ts` anyways.

src/decode.rs
```diff
@@ -3447,28 +3516,29 @@ unsafe fn decode_sb(
...
                 "poc={},y={},x={},bl={:?},r={}",
-                frame_hdr.frame_offset, t.b.y, t.b.x, bl, ts.msac.rng,
+                frame_hdr.frame_offset, t.b.y, t.b.x, bl, ts_c.msac.rng,
             );
         }
         bx8 = (t.b.x & 31) >> 1;
         by8 = (t.b.y & 31) >> 1;
-        ctx = get_partition_ctx(&f.a[t.a], &t.l, bl, by8, bx8);
-        Some(&mut ts.cdf.m.partition[bl as usize][ctx as usize])
+        Some(get_partition_ctx(&f.a[t.a], &t.l, bl, by8, bx8) as usize)
```
kkysen, line 3533:
```suggestion
        Some(get_partition_ctx(&f.a[t.a], &t.l, bl, by8, bx8))
```

I think it's a bit better to store it as an `Option<u8>` than an `Option<usize>` I think, and do the `as usize` casts at use site, since it's smaller.  It could be optimized, but this is a big function.

kkysen (approved):
I'm still not sure why we need the change for that `ts_idx` iterator.  Other than that, it all LGTM.  I realized I missed looking at all of `decode.rs` last time, which is why I missed that we passed `None` sometimes.

src/decode.rs
randomPoison: Ah, we can keep the original code, this was an accidental change that I had meant to revert. I had initially put the context data in a separate `DisjointMut` allocation to help with debugging, and forgot to revert this once I moved the data back into the `Rav1dTileState`. Reverted.

kkysen: Ah okay.

PR #1033, opened 2024-05-03
`DisjointMut`: Ensure the correct range is removed

randomPoison:
Fixes an issue where the wrong immutable range could be removed from `DisjointMut` if there are multiple borrows of the same range.


randomPoison:
Woops, put this up too soon.

PR #1034, opened 2024-05-03
`struct DisjointMut`: Ensure the correct range is removed

randomPoison:
Fixes an issue where the wrong immutable range can be removed if there are multiple borrows with the same range.  Previously, the wrong one could be removed, which removed an equivalent range, but one with the wrong location (for debugging purposes).

### DIFF
--- a/src/disjoint_mut.rs
+++ b/src/disjoint_mut.rs
@@ -72,7 +72,7 @@ pub struct DisjointMutGuard<'a, T: ?Sized + AsMutPtr, V: ?Sized> {
     #[cfg(debug_assertions)]
     parent: &'a DisjointMut<T>,
     #[cfg(debug_assertions)]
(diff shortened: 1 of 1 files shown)


src/disjoint_mut.rs
```diff
@@ -654,14 +654,35 @@ mod debug {
     use std::thread::ThreadId;
 
     #[derive(Debug)]
-    struct DisjointMutBounds {
+    pub(crate) struct DisjointMutBounds {
```
rinon, line 657: Nit: does this need to be crate public?

rinon (approved):

src/disjoint_mut.rs
randomPoison: It needs to be accessed from the parent module, so I guess it could be `pub(super)` instead? Will change.

PR #1038, opened 2024-05-06
`decode.rs`: Remove `unsafe` marker from safe functions

randomPoison:
Remove `unsafe` marker from the smaller functions in `decode.rs` that are already safe or mostly safe, and add `unsafe` blocks and safety comments where needed.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -217,7 +217,7 @@ fn init_quant_tables(
     }
 }
 
-unsafe fn read_mv_component_diff(
+fn read_mv_component_diff(
     f: &Rav1dFrameData,
     msac: &mut MsacContext,
     mv_comp: &mut CdfMvComponent,
@@ -276,7 +276,7 @@ enum MvCdfSelect {
     Dmv,
 }
 
-unsafe fn read_mv_residual(
+fn read_mv_residual(
     f: &Rav1dFrameData,
     ts_c: &mut Rav1dTileStateContext,
     ref_mv: &mut mv,
@@ -314,7 +314,7 @@ unsafe fn read_mv_residual(
     };
 }
 
-unsafe fn read_tx_tree(
+fn read_tx_tree(
     t: &mut Rav1dTaskContext,
     f: &Rav1dFrameData,
     ts_c: &mut Rav1dTileStateContext,
@@ -409,7 +409,8 @@ unsafe fn read_tx_tree(
             [(&t.l, txh), (&f.a[t.a], txw)],
             [t_dim.h as usize, t_dim.w as usize],
             [by4 as usize, bx4 as usize],
-            |case, (dir, val)| {
+            // SAFETY: Only one thread is accessing the written portion of the buffer at a time.
+            |case, (dir, val)| unsafe {
                 case.set_disjoint(&dir.tx, if is_split { TX_4X4 } else { val });
             },
         );
@@ -444,7 +445,7 @@ fn neg_deinterleave(diff: u8, r#ref: u8, max: u8) -> u8 {
     }
 }
 
-unsafe fn find_matching_ref(
+fn find_matching_ref(
     f: &Rav1dFrameData,
     t: &Rav1dTaskContext,
     intra_edge_flags: EdgeFlags,
@@ -552,7 +553,7 @@ unsafe fn find_matching_ref(
     }
 }
 
-unsafe fn derive_warpmv(
+fn derive_warpmv(
     r: &DisjointMut<AlignedVec64<refmvs_block>>,
     t: &Rav1dTaskContext,
     bw4: c_int,
@@ -751,7 +752,7 @@ fn order_palette(
     }
 }
 
-unsafe fn read_pal_indices(
+fn read_pal_indices(
     ts_c: &mut Rav1dTileStateContext,
     pal_dsp: &Rav1dPalDSPContext,
     scratch_pal: &mut ScratchPal,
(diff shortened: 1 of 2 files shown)


src/decode.rs
```diff
@@ -3959,7 +3969,8 @@ unsafe fn setup_tile(
     ts.last_qidx = AtomicU8::new(frame_hdr.quant.yac);
     ts.last_delta_lf.fill_with(Default::default);
 
-    ts_c.msac = MsacContext::new(data, frame_hdr.disable_cdf_update != 0, &c.dsp.msac);
+    // SAFETY: `data` comes from `f.tiles`, which will live at least as long as the created `MsacContext`.
+    ts_c.msac = unsafe { MsacContext::new(data, frame_hdr.disable_cdf_update != 0, &c.dsp.msac) };
```
kkysen, line 3973: Not sure if we should make this one safe yet.  The lifetimes are not super clear and it might be possible to make this safe.

src/decode.rs
```diff
@@ -790,14 +791,20 @@ unsafe fn read_pal_indices(
...
-    (pal_dsp.pal_idx_finish)(
-        pal_idx.unwrap_or(pal_tmp).as_mut_ptr(),
-        pal_tmp.as_ptr(),
-        bw4 as c_int * 4,
-        bh4 as c_int * 4,
-        w4 as c_int * 4,
-        h4 as c_int * 4,
-    );
+
+    // SAFETY: Unsafe asm call, checkasm tests have verified that the signature is
```
kkysen, line 795: Is that all `checkasm` does?

src/decode.rs
```diff
@@ -790,14 +791,20 @@ unsafe fn read_pal_indices(
...
+    // correct. `pal_idx` and `pal_tmp` are at least `(bw4 * 4) * (bh4 * 4)`
+    // elements long, which is how many elements `pal_idx_finish` will read/write.
```
kkysen, line 797: These kinds of assertions should be `debug_assert!`s at the very least, I think.

src/refmvs.rs
```diff
@@ -478,18 +478,28 @@ impl Rav1dRefmvsDSPContext {
...
         let bw4 = bw4 as _;
         let bh4 = bh4 as _;
-        (self.splat_mv)(rr.as_mut_ptr(), rmv, bx4, bw4, bh4, rr.len());
+
+        // SAFETY: Unsafe asm call, checkasm has verified the signature is correct. `rr` is at
+        // least `bh4` elements long, and each pointer in `rr` is non-null and points to at
+        // least `bx4 * bw4` elements, which is what will be accessed in `splat_mv`.
+        unsafe {
+            (self.splat_mv)(rr.as_mut_ptr(), rmv, bx4, bw4, bh4, rr.len());
+        }
```
kkysen, line 502: We pass `rr.len()` here, so the fallback `fn` uses that for the length.

kkysen (changes requested):

src/decode.rs
randomPoison: Naw, looks like the checkasm tests are full unit tests for the asm and C fallback routines, but since we're not running the checkasm tests against Rust the only thing it's verified for us is that the function signatures are correct. I'm open to better ways of wording this if you have suggestions.

src/decode.rs
randomPoison: I think marking `setup_tile` safe for now is fine, though if we can make `MsacContext::new` safe at a later point that'd be great.

src/decode.rs
kkysen: If we're not running those tests, signature checking isn't even happening either right?  Can we just say eventually will be checked by checkasm?

src/decode.rs
kkysen: Hmm, it doesn't matter too much I guess.  It's fine as is then.

src/decode.rs
randomPoison: Eh, I don't think it makes sense to say that will eventually be checked, but I can reword it to remove the reference to checkasm.

src/decode.rs
randomPoison: Added. I also tweaked the wording of the safety comment since the required length of the two arrays is slightly different.

src/refmvs.rs
randomPoison: Added that to the safety comment.

src/decode.rs
```diff
@@ -797,14 +798,28 @@ unsafe fn read_pal_indices(
...
-    );
+
+    if let Some(pal_idx) = &pal_idx {
+        let read_len = bw4 * 2 * bh4 * 4;
+        debug_assert!(pal_idx.len() >= read_len);
+    }
+    let read_len = bw4 * 4 * bh4 * 4;
+    debug_assert!(pal_tmp.len() >= read_len);
+
+    // SAFETY: Unsafe asm call. `pal_idx` is at least `(bw * 2) * (bh * 4)`
```
kkysen, line 809:
```suggestion
    // SAFETY: Unsafe asm call. `pal_idx` is at least `(bw4 * 2) * (bh4 * 4)`
```

That's what `read_len` above is; one of these is wrong, right?

kkysen (approved):
I think one of the lengths in the comment has a typo?  Otherwise LGTM now.

src/decode.rs
randomPoison: Ah yeah that should be `bw4` and `bh4`, will fix.

PR #1052, opened 2024-05-07
`mod obu`: backport trailing bits check from dav1d 1.2.1

fbossen:


src/obu.rs
```diff
@@ -122,6 +122,29 @@ impl Debug {
...
+
+    let mut size = gb.remaining_len();
+
+    while size > 0 && gb[size - 1] == 0 {
+        size -= 1;
+    }
+
+    if size != 0 {
+        return Err(EINVAL);
+    }
```
kkysen, line 143:
```suggestion
    if !gb.get_bytes(gb.remaining_len()).iter().all(|&b| b == 0) {
        return Err(EINVAL);
    }
```

src/obu.rs
```diff
@@ -122,6 +122,29 @@ impl Debug {
...
+
+    return Ok(());
```
kkysen, line 145:
```suggestion
    Ok(())
```

kkysen (approved):
I made a couple of suggestions to make `fn check_trailing_bits` easier to understand/better optimized for bounds checks, but otherwise it all LGTM.

src/obu.rs
fbossen: Thank you for the suggestion. Had to add a call to `gb.bytealign()` to avoid an assertion failure in `gb.get_bytes()`

kkysen (approved):

PR #1054, opened 2024-05-08
Block while trying to read frame data for `check_tile`

rinon:
We encountered a rare conflict between trying to read the frame data for `check_tile` and a concurrent write. This change just blocks so we can have access to the necessary frame data. We believe this will be rare and therefore not have a significant performance impact.

### DIFF
--- a/src/thread_task.rs
+++ b/src/thread_task.rs
@@ -813,14 +813,17 @@ pub unsafe fn rav1d_worker_task(c: &Rav1dContext, task_thread: Arc<Rav1dTaskCont
                             t.type_0,
                             TaskType::TileEntropy | TaskType::TileReconstruction
                         ) {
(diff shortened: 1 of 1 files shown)


kkysen (commented):
I wonder if we should leave this until we're almost done.  I've noticed the error happening more often lately (still usually just 1 or 2 test failures per test run, but more consistently every test run, whereas before it was a lot rarer).  Should we just leave this open for a while and then if it's still rare, merge it?  I don't want it to accidentally get more common after we do other changes, and then we don't notice the perf hit (as opposed to a more often failure).

rinon:
I've added a comment. I think we should land this, I'm seeing failures in CI on other unrelated PRs and that is going to waste a lot of time while we go check what the issue was and then have to re-run the job.

PR #1055, opened 2024-05-08
`struct Rav1dFrameContext_frame_thread_progress`: Block write to `copy_lpf`

rinon:
Writes to `copy_lpf` were hitting contention with reads during task selection in the argon tests. This seem rare enough that we should just block.

Fixes argon test failures on main.

### DIFF
--- a/src/thread_task.rs
+++ b/src/thread_task.rs
@@ -323,7 +323,9 @@ unsafe fn create_filter_sbrow(
         let mut frame = fc.frame_thread_progress.frame.try_write().unwrap();
         frame.clear();
         frame.resize_with(prog_sz, || AtomicU32::new(0));
-        let mut copy_lpf = fc.frame_thread_progress.copy_lpf.try_write().unwrap();
+        // copy_lpf is read during task selection, so we are seeing contention
+        // here. This seems rare enough that it is not worth optimizing.
+        let mut copy_lpf = fc.frame_thread_progress.copy_lpf.write().unwrap();
         copy_lpf.clear();
         copy_lpf.resize_with(prog_sz, || AtomicU32::new(0));
         fc.frame_thread_progress.deblock.store(0, Ordering::SeqCst);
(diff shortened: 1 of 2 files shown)


src/thread_task.rs
```diff
@@ -323,7 +323,8 @@ unsafe fn create_filter_sbrow(
         let mut frame = fc.frame_thread_progress.frame.try_write().unwrap();
         frame.clear();
         frame.resize_with(prog_sz, || AtomicU32::new(0));
-        let mut copy_lpf = fc.frame_thread_progress.copy_lpf.try_write().unwrap();
+        // copy_lpf is read during task selection, so we may be blocked here
```
kkysen, line 326: Could you add to the comment that it's rare that it actually blocks so it should be fine performance wise, and that potentially the blocking could be removed, but it's not currently worth it?

kkysen (commented):
Could you

src/thread_task.rs
kkysen: Basically what you have in the PR description.

rinon:
Oh great, now we're getting

```
thread 'rav1d-worker-0' panicked at src/decode.rs:423:24:
attempt to subtract with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

which is underflow at https://github.com/memorysafety/rav1d/blob/81800066a66fc3201c9b8d682cf56fd10adb0f55/src/decode.rs#L423

rinon:
underflow is unrelated, issue appeared in https://github.com/memorysafety/rav1d/pull/1007/commits/67be411f12f2aecaad1955e80c147451c8c4e861#r1594759213

kkysen:
> Oh great, now we're getting
> 
> ```
> thread 'rav1d-worker-0' panicked at src/decode.rs:423:24:
> attempt to subtract with overflow
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> which is underflow at
> 
> https://github.com/memorysafety/rav1d/blob/81800066a66fc3201c9b8d682cf56fd10adb0f55/src/decode.rs#L423

Can we just do `r#ref + 1 >= max`?

fbossen:
> Can we just do `r#ref + 1 >= max`?

Or `r#ref > max`

kkysen (approved):

PR #1056, opened 2024-05-08
Fix underflow introduced in #1007

fbossen:
`max` can be 0 and subtracting 1 causes underflow


kkysen (approved):
Solution LGTM but @rinon saw the failure, so let me know if this fix is fine.

rinon (approved):

rinon:
There are still underflows here, `max - diff - 1` is negative. I think we need to change the params of `neg_deinterleave` to be signed 32-bit integers as it was in the C code.

PR #1057, opened 2024-05-09
`fn rav1d_create_lf_mask_{inter,intra}`: Make safe

randomPoison:


src/lf_mask.rs
```diff
@@ -423,23 +423,21 @@ pub(crate) unsafe fn rav1d_create_lf_mask_intra(
...
+            for x in 0..bw4 {
+                let idx = level_cache_off + x;
+                assert!(idx < level_cache.len());
+                // SAFETY: The Y portion of this element (indices 0 and 1) is not
+                // concurrently accessed by any other threads and the assert above ensures
+                // that it is in bounds.
                 unsafe {
+                    let cur = level_cache.as_mut_ptr().add(idx);
                     (*cur)[0] = filter_level[0][0][0];
                     (*cur)[1] = filter_level[1][0][0];
```
rinon, line 437: Is it safe to `Deref` then `IndexMut` here? Are we creating a mutable reference to the same element in multiple places? Basically I'm asking if we need to do this as a pointer offset rather than index and then use a pointer write. For example:

```rust
let cur = level_cache.as_mut_ptr().add(idx) as *mut u8;
*cur = filter_level[0][0][0];
*cur.offset(1) = filter_level[1][0][0];
```

rinon (commented):
I'm fine with this approach if we have to pass the level_cache to assembly, where we are accessing it unsafely anyway.

src/lf_mask.rs
randomPoison: HRm yeah good point, I think it'd be safer to use a pointer write. Will update.

kkysen: I don't think arrays call `DerefMut` when you index into them.

randomPoison: I'm not sure if I'm correctly testing what we're doing here, but I created [this example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3e1683b8900eaa897065362f9a5cbb41) of using interleaved pointer writes to separate parts of the `[u8; 4]` and miri doesn't produce any errors. So maybe the original version was fine?

kkysen: Yeah, I think the previous was is fine if it passes that `miri`.  Arrays work differently.

randomPoison: Okay, I've rolled this back to the original version.

kkysen (approved):

PR #1059, opened 2024-05-09
`decode.rs`: Mark functions safe and document unsafe operations

randomPoison:
Mark `decode_b`, `decode_sb`, `rav1d_decode_tile_sbrow`, and `rav1d_decode_frame_main` as safe, putting any unsafe operations in `unsafe` blocks and documenting their safety invariants. For the `bd_fn` calls those should be safe since they're pure Rust calls and just haven't been marked safe. Once they're made safe we can clean up the unnecessary `unsafe` blocks. I also marked `load_tmvs` and `save_tmvs` safe since those wrap the unsafe asm call and handle converting the safe Rust args to the raw pointers that asm needs.

Part of #843.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -1204,12 +1204,10 @@ fn decode_b(
     let mut b_guard;
     let mut b_mem = Av1Block::default();
     let b = if t.frame_thread.pass != 0 {
-        // SAFETY: No other thread is accessing the borrowed element.
-        b_guard = unsafe {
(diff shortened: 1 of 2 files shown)


kkysen:
I was discussing this a bit with @rinon; I think we should make `index_mut` safe like we did `index`.  All of the safety comments are essentially the same and we're not really saying anything meaningful.

rinon (approved):

PR #1060, opened 2024-05-09
Fix `fn neg_deinterleave` to handle negative values correctly

rinon:
The return value from `neg_deinterleave` can be negative, so rather than
changing the parameters and return value to `u8`s, we should just keep
the `i32`s used in the C implementation and cast the result to `u8`
after we return it.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -417,7 +417,7 @@ fn read_tx_tree(
     };
 }
 
-fn neg_deinterleave(diff: u8, r#ref: u8, max: u8) -> u8 {
+fn neg_deinterleave(diff: i32, r#ref: i32, max: i32) -> i32 {
     if r#ref == 0 {
         diff
     } else if r#ref + 1 >= max {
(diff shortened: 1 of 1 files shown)


kkysen:
If all of the calls are cast to `u8`, what's happening with the negative return values?  I'd rather do the casts inside `neg_deinterleave`.

fbossen:
Something seems a bit off here. The AV1 specification states "It is a requirement of bitstream conformance that the postprocessed value of segment_id (i.e. the value returned by neg_deinterleave) is in the range 0 to LastActiveSegId (inclusive of endpoints)." So, if the `neg_deinterleave` returns a negative value, something is probably may not be right with the bitstream. @rinon Do you have an example of when the return value of `neg_deinterleave` is negative?

rinon:
It seems like this is fixed on the caller side, negative values are clamped to zero. If you look at the argon tests before this change you'll see the underflow.

fbossen:
> It seems like this is fixed on the caller side, negative values are clamped to zero. If you look at the argon tests before this change you'll see the underflow.

Thanks. I was able to reproduce the issue. I submitted #1061 where the offending subs are made wrapping to match the cast to `uint8_t` in C. Overall, I think this part of the code is still a bit fishy: even the C code has comments saying `error?`.

rinon:
Abandoning in favor of #1061

PR #1062, opened 2024-05-10
`fn DisjointMut::index_mut`: Mark safe

randomPoison:


kkysen (commented):
Could we also remove the `unsafe` blocks around calls to `index_mut` that we added?

randomPoison:
> Could we also remove the `unsafe` blocks around calls to `index_mut` that we added?

Yup, good idea. Those have been removed.

kkysen (approved):

PR #1065, opened 2024-05-11
`enum MaybeTempPixels`: Add a safer abstraction over `dst{8,16}` args

kkysen:


randomPoison (approved):
Nice solution, hopefully we can also get the non-temp version to be a safe slice as well.

kkysen:
> Nice solution, hopefully we can also get the non-temp version to be a safe slice as well.

It's likely going to be a `DisjointMut`-like thing, so I left it as a raw pointer for now.  I'm working on making the `Rav1dPictureData` ptrs safe now.

PR #1069, opened 2024-05-13
`mod qm`: backport memory optimization from dav1d 1.4.0

fbossen:


src/qm.rs
```diff
@@ -2976,16 +1681,19 @@ static qm_tbl_32x32_t: [[[u8; (32 * (32 + 1)) / 2]; 2]; 15] = [
...
 
-    const_for!(y in 0..sz => {
-        const_for!(x in 0..sz => {
-            dst[y * sz + x] = src[y * sz * step * step + x * step];
+    const_for!(y in 0..h/vstep => {
+        const_for!(x in 0..32/hstep => {
+            dst[i] = src[(y * vstep + (vstep - 1) / 2) * 32 + x * hstep + (hstep - 1) / 2];
+            i += 1;
         });
     });
```
kkysen, line 1698: I think it might be clearer now to write these loops like they are in C.  It'd be `for y in in (0..h).step_by(vstep)` normally, but we can't have `for` loops in `const fn`s, so we need to use a `while` loop, like how `c2rust transpile` outputs things.

kkysen (approved):
Looks correct, but I think it'd be clearer to rewrite that `const_for!` loop in `fn subsampled` to use a `while` and iteration logic like C does.

Also, unfortunately, we don't get these static/binary size savings because we calculate the tables at compile time, vs. C zero initializing them at compile time and initializing them at runtime.  It doesn't increase runtime memory though, and hopefully, the pages don't get mapped unless needed.  So we can look at making it runtime-initialized like C if binary size matters (which will incur a `OnceLock` to access), but we can do that later, and this change is good for simplification anyways.

src/qm.rs
fbossen: The `const_for!` macro can actually take a `step_by` parameter, so I used that

kkysen: Oh it can? Cool! Guess I forgot.  Yeah that works even better then.

src/qm.rs
```diff
@@ -2976,16 +1681,19 @@ static qm_tbl_32x32_t: [[[u8; (32 * (32 + 1)) / 2]; 2]; 15] = [
...
+    assert!(M * hstep * vstep == h * 32);
     let mut dst = [0; M];
+    let mut i = 0;
 
-    const_for!(y in 0..sz => {
-        const_for!(x in 0..sz => {
-            dst[y * sz + x] = src[y * sz * step * step + x * step];
+    const_for!(y in 0..h, step_by vstep => {
+        const_for!(x in 0..32, step_by hstep => {
+            dst[i] = src[(y + (vstep - 1) / 2) * 32 + x + (hstep - 1) / 2];
```
kkysen, line 1695:
```suggestion
            dst[i] = src[y * 32 + x];
```

Why can't we do it simpler like C is?

fbossen: The C code adds an offset to the `src` pointer before passing it to this function. I avoided doing so to simplify the array size check in this function. I could do `src = &src[(vstep - 1) / 2) * 32 + (hstep - 1) / 2..]` before entering this loop or add an offset parameter to this function. Any preference?

kkysen: Ahh.  I didn't think slicing works in a `const fn`, but I think storing a separate `offset` that we add to the index each time would be easier to read.

PR #1072, opened 2024-05-13
Backport frame verification fix from dav1d 1.4.0

fbossen:


src/decode.rs
```diff
@@ -4879,6 +4879,18 @@ pub(crate) fn rav1d_decode_frame_exit(
     if c.fc.len() > 1 && retval.is_err() {
         cf.fill_with(Default::default);
     }
+
+    if !retval.is_err() && c.fc.len() > 1 && c.strict_std_compliance {
+        if f.refp.iter().any(|rf| {
+            rf.p.frame_hdr.is_some()
+                && rf.progress.as_ref().unwrap()[1].load(Ordering::Relaxed) == FRAME_ERROR
```
kkysen, line 4886:
```suggestion
                && rf.progress.as_ref().unwrap()[1].load(Ordering::SeqCst) == FRAME_ERROR
```

A C `atomic_load` uses `SeqCst` by default.  We're only using `Relaxed` where C had no atomics and we only use atomics for the borrow checker.

src/decode.rs
```diff
@@ -4879,6 +4879,18 @@ pub(crate) fn rav1d_decode_frame_exit(
...
+        }) {
+            retval = Err(EINVAL);
+            fc.task_thread.error.store(1, Ordering::Relaxed);
+            f.sr_cur.progress.as_mut().unwrap()[1].store(FRAME_ERROR, Ordering::Relaxed);
```
kkysen, line 4890:
```suggestion
            fc.task_thread.error.store(1, Ordering::SeqCst);
            f.sr_cur.progress.as_mut().unwrap()[1].store(FRAME_ERROR, Ordering::SeqCst);
```

src/decode.rs
```diff
@@ -4879,6 +4879,18 @@ pub(crate) fn rav1d_decode_frame_exit(
     if c.fc.len() > 1 && retval.is_err() {
         cf.fill_with(Default::default);
     }
+
+    if !retval.is_err() && c.fc.len() > 1 && c.strict_std_compliance {
```
kkysen, line 4883:
```suggestion
    if retval.is_ok() && c.fc.len() > 1 && c.strict_std_compliance {
```

src/decode.rs
```diff
@@ -4954,7 +4966,11 @@ pub(crate) unsafe fn rav1d_decode_frame(c: &Rav1dContext, fc: &Rav1dFrameContext
         }
     }
     rav1d_decode_frame_exit(c, fc, res);
-    res
+    if let Some(err) = *fc.task_thread.retval.try_lock().unwrap() {
+        Err(err)
+    } else {
+        Ok(())
+    }
```
kkysen, line 4973:
```suggestion
    fc.task_thread.retval.try_lock().unwrap().err_or(())
```

I added an extension method `.err_or` in `src/extensions.rs` for this.  It's annoying `std` doesn't have this.

kkysen: Also, just because we have a lock around `retval` now, can we return `retval` from `fn rav1d_decode_frame_exit`, too, and then just return that value here?

kkysen (changes requested):

src/decode.rs
fbossen: we certainly can

src/decode.rs
fbossen: I kept `Ordering::Relaxed` for `fc.task_thread.error`. There is an existing `store` for that variable in the same function that uses `Ordering::Relaxed` and I am assuming it's correct.

kkysen: That might be different.  Did C explicitly do a relaxed store there?  I think it's better to do seq cst here and match what C is doing.  Unless you have an idea why specifically relaxed would be okay here, I think it's safer to just match C.

fbossen: You're right. I fixed it to match the C code.

kkysen (approved):

PR #1074, opened 2024-05-14
`fn MsacContext::buf`: Add separate `pub fn` creating the `buf` slice

kkysen:
* Useful for #1071.

### DIFF
--- a/src/msac.rs
+++ b/src/msac.rs
@@ -8,6 +8,7 @@ use std::ffi::c_uint;
 use std::mem;
 use std::ops::Range;
 use std::ptr;
+use std::slice;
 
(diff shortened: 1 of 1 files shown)


fbossen:
It's not immediately obvious to me how this would be used in #1071. Note that the buffer returned in #1071 uses a negative index w.r.t. `buf_pos`.

kkysen:
> It's not immediately obvious to me how this would be used in #1071. Note that the buffer returned in #1071 uses a negative index w.r.t. `buf_pos`.

Ohh, `n_bytes` is negative?

fbossen:
> Ohh, `n_bytes` is negative?

yes, 0 or negative

PR #1076, opened 2024-05-14
`struct MsacContext`: Make data safe

kkysen:
This stores a `CArc<[u8]>` in `MsacContext` to safely store the `buf_{pos,end}` lifetime.  The lifetime is tricky to do otherwise since it comes from a `Rav1dData` is passed to multiple threads, and the `Rav1dData` already has a `CArc<[u8]>`.  Thus, we pay `Arc::clone`s on the `CArc::split_at`s introduced here, but hopefully that shouldn't be too expensive.  Furthermore, this lets us keep the data origin in `MsacContext`, which allows us to make #1071 safe, which does negative indexing.  To make that simple, `fn data` and `fn buf_index` can be used.

Since `MsacContext` was passed to asm, it needed to be FFI-safe, but `CArc` and the `Arc` inside of it are not.  To get around this, I renamed it to `MsacAsmContext` and created a new wrapper non-FFI-safe `MsacContext` storing the `CArc`.

`fn MsacContext::{buf,buf_index,with_buf}` are now all safe with no `unsafe` at all.

### DIFF
--- a/src/msac.rs
+++ b/src/msac.rs
@@ -8,6 +8,7 @@ use std::ffi::c_uint;
 use std::mem;
 use std::ops::Range;
 use std::ptr;
+use std::slice;
 
 #[cfg(all(feature = "asm", target_feature = "sse2"))]
 extern "C" {
@@ -183,16 +184,22 @@ impl MsacContext {
         self.buf_end = end;
     }
 
-    fn with_buf(&mut self, mut f: impl FnMut(&[u8]) -> &[u8]) {
-        // # Safety
-        //
-        // [`Self::buf_pos`] and [`Self::buf_end`] are the start and end ptrs of the `buf` slice,
-        // and are only set in [`Self::set_buf`], which derives them from a valid slice.
-        let buf = unsafe {
+    /// # Safety
+    ///
+    /// Not actually `'static`, but the lifetime safety is
+    /// guaranteed by [`Self::new`]'s safety preconditions.
+    pub fn buf(&self) -> &'static [u8] {
+        // SAFETY: [`Self::buf_pos`] and [`Self::buf_end`] are the start and end ptrs of the `buf` slice,
+        // and are only set in [`Self::set_buf`] and [`Self::new`],
+        // which derive them from a valid slice.
+        unsafe {
             let len = self.buf_end.offset_from(self.buf_pos) as usize;
-            std::slice::from_raw_parts(self.buf_pos, len)
-        };
-        self.set_buf(f(buf));
+            slice::from_raw_parts(self.buf_pos, len)
+        }
+    }
+
+    fn with_buf(&mut self, mut f: impl FnMut(&[u8]) -> &[u8]) {
+        self.set_buf(f(self.buf()));
     }
 
     fn allow_update_cdf(&self) -> bool {
--- a/src/msac.rs
+++ b/src/msac.rs
@@ -186,12 +186,18 @@ impl MsacAsmContext {
     fn allow_update_cdf(&self) -> bool {
         self.allow_update_cdf != 0
     }
+
+    fn set_buf(&mut self, buf: &[u8]) {
+        let Range { start, end } = buf.as_ptr_range();
+        self.buf_pos = start;
+        self.buf_end = end;
+    }
 }
 
 #[derive(Default)]
 pub struct MsacContext {
     asm: MsacAsmContext,
-    _data: Option<CArc<[u8]>>,
+    data: Option<CArc<[u8]>>,
 }
 
 impl Deref for MsacContext {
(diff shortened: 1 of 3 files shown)


randomPoison:
> `fn MsacAsmContext::with_buf` is now ensured to be safe, even with lifetimes, by ensuring that it is only called when `fn MsacContext::with_asm` is higher on the callstack, which ensures that the data owner, `MsacContext::data`, is live.

This doesn't seem to be true from what I can tell? `with_buf` is only used from `ctx_refill`, which is called from `ctx_norm`, which is called e.g. from `rav1d_msac_decode_bool_rust` without going through `with_asm`. From what I've seen I can't find any cases where `with_buf` is called from within `with_asm`. All of the calls seem to go through the `DerefMut` impl to get the `MsacAsmContext` before calling `with_buf`. Am I missing something?

src/msac.rs
```diff
@@ -176,27 +182,59 @@ impl Default for MsacContext {
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.asm
+    }
+}
+
+impl MsacContext {
+    fn with_asm<R>(&mut self, mut f: impl FnMut(&mut MsacAsmContext) -> R) -> R {
+        let data = self.data.as_ref().unwrap().as_ref();
+        // Hold this reference across `f`, which ensures [`MsacAsmContext::with_buf`] is safe.
+        let retval = f(&mut self.asm);
```
randomPoison, line 231: How does holding the reference here make `with_buf` safe? I would think that the thing that makes it safe is that we're enforcing that `buf_pos` and `buf_end` stay within bounds of the source allocation.

src/msac.rs
```diff
@@ -176,27 +182,59 @@ impl Default for MsacContext {
...
+        let data = data.as_ptr_range();
+        debug_assert!(data.end == self.buf_end);
+        debug_assert!(data.start <= self.buf_pos && data.end >= self.buf_pos);
+        // SAFETY: The asm `fn`s, `fn *_c`s, and `fn *_rust`s must keep `buf_pos` valid.
+        self.data_offset = unsafe { self.buf_pos.offset_from(data.start) as usize };
```
randomPoison, line 236: The above asserts are checking that preconditions for `offset_from` are upheld right? We should mention that in the safety comment, even if they're debug-only asserts.

src/msac.rs
kkysen: Well we also need to make sure the data `buf_pos` can't be freed while we're accessing it.  Holding the reference is what guarantees that.

src/msac.rs
kkysen: Sure, I can.

kkysen:
> > `fn MsacAsmContext::with_buf` is now ensured to be safe, even with lifetimes, by ensuring that it is only called when `fn MsacContext::with_asm` is higher on the callstack, which ensures that the data owner, `MsacContext::data`, is live.
> 
> This doesn't seem to be true from what I can tell? `with_buf` is only used from `ctx_refill`, which is called from `ctx_norm`, which is called e.g. from `rav1d_msac_decode_bool_rust` without going through `with_asm`. From what I've seen I can't find any cases where `with_buf` is called from within `with_asm`. All of the calls seem to go through the `DerefMut` impl to get the `MsacAsmContext` before calling `with_buf`. Am I missing something?

Hmm, yeah, I think some things are missed after I had to rearrange things to fix a bug.  I'll look into it.  I think the functions where `with_asm` is called unconditionally should be doing what I wanted.  I'll fix the rest.

src/msac.rs
fbossen: Do we need to compute and store `data_offset` every time here? Couldn't we compute it when needed?

fbossen (commented):
Do we need to store `data_offset`?

src/msac.rs
randomPoison: How would `buf_pos` get freed in a way that holding onto `data` across the call to `f` would prevent? `f` is getting the `MsacAsmContext` which doesn't have access to `data`, which is what actually owns the buffer pointed to by `buf_pos` right? So I don't see how the call to `f` would be able to free the buffer other than doing something unsafe like passing `buf_pos` to `free`, which wouldn't even work right?

randomPoison: How? Isn't the data owned by the `CArc` which we're still holding onto? It doesn't seem like holding a reference to `self.data` is what's preventing it from being freed, it's the fact that we still own a copy of the `CArc` and hence are preventing the ref count from going to 0.

kkysen: Hmm, yeah that's true.

kkysen:
d217277 looks neutral on perf, while 35e0aa9 is a ~6% regression on `chimera_8b_1080p.ivf`, so I'm going to remove/rework that part.  The important `Arc` part in d217277 is fine, though, so that's good.

src/msac.rs
kkysen: No more `data_offset` anymore.

src/msac.rs
kkysen: Doing it a different way now.

kkysen:
@fbossen, can you check the perf here?  It might be a bit slower, but my measurements are noisy, so it wasn't always the case.  And if it is, I'm not sure where the slowdown is, as the `CArc` commit didn't seem to slow down things, and the commits after that pretty much just affect the fallbacks, except for the initial `ctx_refill` call in `fn new`, but that's not called from the hot `fn decode_coefs`.

fbossen:
> @fbossen, can you check the perf here? It might be a bit slower, but my measurements are noisy, so it wasn't always the case. And if it is, I'm not sure where the slowdown is, as the `CArc` commit didn't seem to slow down things, and the commits after that pretty much just affect the fallbacks, except for the initial `ctx_refill` call in `fn new`, but that's not called from the hot `fn decode_coefs`.

Perf looks ok. On average, things may run very slightly faster with the PR.

fbossen (approved):

PR #1077, opened 2024-05-14
`union BitDepthUnion`: Remove and replace with safe `enum`

randomPoison:
Removes `BitDepthUnion` and replaces the last usage of it with an enum. Split off from #1073 due to questions around lazy initialization. I've reworked this to eagerly initialize the `Grain` enum in `rav1d_task_delayed_fg` where we setup the rest of the delayed fg task.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -266,8 +266,6 @@ pub(crate) struct Rav1dContext_frame_thread {
 
 pub type GrainLut<Entry> = [[Entry; GRAIN_WIDTH]; GRAIN_HEIGHT + 1];
 
-#[derive(Clone, Copy)]
-#[repr(C)]
 pub struct GrainBD<BD: BitDepth> {
     pub grain_lut: Align16<[GrainLut<BD::Entry>; 3]>,
     // TODO(kkysen) can use `BD::SCALING_LEN` directly with `#![feature(generic_const_exprs)]` when stabilized
@@ -284,12 +282,25 @@ impl<BD: BitDepth> Default for GrainBD<BD> {
     }
 }
 
-#[derive(Default)]
 pub enum Grain {
-    #[default]
-    Uninit,
+    #[cfg(feature = "bitdepth_8")]
     Bpc8(GrainBD<BitDepth8>),
+    #[cfg(feature = "bitdepth_16")]
     Bpc16(GrainBD<BitDepth16>),
+}
+
+impl Default for Grain {
+    fn default() -> Self {
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "bitdepth_8")] {
+                Self::Bpc8(Default::default())
+            } else if #[cfg(feature = "bitdepth_16")] {
+                Self::Bpc16(Default::default())
+            } else {
+                compile_error!("No bitdepths enabled");
+            }
+        }
+    }
 }
 
 #[derive(Default)]
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -1,8 +1,6 @@
 use crate::include::common::bitdepth::BitDepth;
 use crate::include::common::bitdepth::BitDepth16;
 use crate::include::common::bitdepth::BitDepth8;
-use crate::include::common::bitdepth::BitDepthDependentType;
-use crate::include::common::bitdepth::BitDepthUnion;
 use crate::include::common::bitdepth::BPC;
 use crate::include::dav1d::common::Rav1dDataProps;
 use crate::include::dav1d::data::Rav1dData;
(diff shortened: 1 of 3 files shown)


src/thread_task.rs
```diff
@@ -442,6 +443,11 @@ pub(crate) fn rav1d_task_delayed_fg(
         delayed_fg.in_0 = in_0.clone();
         delayed_fg.out = out.clone();
         delayed_fg.type_0 = TaskType::FgPrep;
+        delayed_fg.grain = match out.p.bpc {
+            8 => Grain::Bpc8(Default::default()),
+            10 | 12 => Grain::Bpc16(Default::default()),
+            _ => unreachable!(),
+        }
```
kkysen, line 450:
```suggestion
        // This initialization is done once per call to [`dav1d_apply_grain`].
        // Nevertheless, it is a 48 KB zero initialization that C avoids,
        // so we avoid a few common cases here.
        match (out.p.bpc, &delayed_fg.grain) {
            (8, Grain::Bpc8(_)) | (10 | 12, Grain::Bpc16(_)) => {}
            (8, _) => delayed_fg.grain = Grain::Bpc8(Default::default()),
            (10 | 12, _) => delayed_fg.grain = Grain::Bpc16(Default::default()),
            _ => unreachable!(),
        }
```

src/thread_task.rs
```diff
@@ -16,6 +16,7 @@ use crate::src::error::Rav1dError::ENOMEM;
 use crate::src::error::Rav1dResult;
 use crate::src::fg_apply::rav1d_apply_grain_row;
 use crate::src::fg_apply::rav1d_prep_grain;
+use crate::src::internal::Grain;
```
kkysen, line 19: Can't comment higher, but the `BitDepth` import is unused now.

kkysen (changes requested):
This does reduce the 48 KB initialization to once per `dav1d_apply_grain` call (a `DAV1D_API`), which is a lot better.  It's still big, though, so I think we should profile/benchmark to make sure this doesn't slow things down.

src/internal.rs
```diff
@@ -286,10 +284,12 @@ impl<BD: BitDepth> Default for GrainBD<BD> {
...
 }
 
-pub struct Grain;
-
-impl BitDepthDependentType for Grain {
-    type T<BD: BitDepth> = GrainBD<BD>;
+#[derive(Default)]
+pub enum Grain {
+    #[default]
+    Uninit,
```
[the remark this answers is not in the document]
randomPoison: Removed.

src/thread_task.rs
```diff
@@ -623,20 +629,26 @@ unsafe fn delayed_fg_task<'l, 'ttd: 'l>(
...
+                    let Grain::Bpc16(grain) = &mut delayed_fg.grain else {
+                        unreachable!();
+                    };
                     rav1d_prep_grain::<BitDepth16>(
                         &Rav1dBitDepthDSPContext::get(bpc).as_ref().unwrap().fg,
                         &mut delayed_fg.out,
                         &delayed_fg.in_0,
-                        BitDepth16::select_mut(&mut delayed_fg.grain),
+                        grain,
                     );
```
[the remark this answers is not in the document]
randomPoison: Good idea, done.

src/thread_task.rs
```diff
@@ -443,9 +442,14 @@ pub(crate) fn rav1d_task_delayed_fg(
...
         delayed_fg.out = out.clone();
         delayed_fg.type_0 = TaskType::FgPrep;
-        delayed_fg.grain = match out.p.bpc {
-            8 => Grain::Bpc8(Default::default()),
-            10 | 12 => Grain::Bpc16(Default::default()),
+
+        // This initialization is done once per call to [`dav1d_apply_grain`].
+        // Nevertheless, it is a 48 KB zero initialization that C avoids,
+        // so we avoid reinitializing here if it already has the correct variant.
+        match (out.p.bpc, &delayed_fg.grain) {
```
kkysen, line 449:
```suggestion
        #[cfg(all(feature = "bitdepth_8", feature = "bitdepth_16"))]
        match (out.p.bpc, &delayed_fg.grain) {
```

kkysen (changes requested):
One little thing left.  Now that we `cfg`d the `Grain` bitdepth variants, we need to `cfg` the initialization `match`.

kkysen:
@rinon, @fbossen, could one of you check if this causes and slowdowns for filmgrain application?

fbossen:
> @rinon, @fbossen, could one of you check if this causes any slowdowns for filmgrain application?

Let me check

src/thread_task.rs
randomPoison: Ah yeah. Added that along with a note about why we `cfg` out the entire match when only one bitdepth is enabled, since that wasn't obvious to me initially.

kkysen (approved):

Issue #819, opened 2024-03-15
Cleanup c2rust output in `mc.rs`


randomPoison:
The fallback functions in `mc.rs` have already been partially cleaned up, but they still make heavy use of raw pointer operations, even in places where slices could be easily used.

* [x] Cleanup helper functions to take safe arguments where possible (may take raw pointers as arguments currently).
* [x] General cleanup of c2rust generated code in functions.
* [x] Restore comments from original C.

The following functions still need to be cleaned:

- [x] `put_rust`
- [x] `prep_rust`
- [x] `filter_8tap`
- [x] `put_8tap_rust`
- [x] `put_8tap_scaled_rust`
- [x] `prep_8tap_rust`
- [x] `prep_8tap_scaled_rust`
- [x] `filter_bilin`
- [x] `put_bilin_rust`
- [x] `put_bilin_scaled_rust`
- [x] `prep_bilin_rust`
- [x] `prep_bilin_scaled_rust`
- [x] `avg_rust`
- [x] `w_avg_rust`
- [x] `mask_rust`
- [x] `blend_rust`
- [x] `blend_v_rust`
- [x] `blen_h_rust`
- [x] `w_mask_rust`
- [x] `warp_affine_8x8_rust`
- [x] `warp_affine_8x8t_rust`
- [x] `emu_edge_rust`
- [x] `resize_rust`

randomPoison:
Finished as of https://github.com/memorysafety/rav1d/pull/1094.

PR #1078, opened 2024-05-14
`fn emu_edge_rust`: Cleanup and make mostly safe

randomPoison:
* Part of #819

### DIFF
--- a/src/mc.rs
+++ b/src/mc.rs
@@ -1016,35 +1016,38 @@ unsafe fn emu_edge_rust<BD: BitDepth>(
     let ref_stride = BD::pxstride(ref_stride);
 
     // find offset in reference of visible block to copy
-    r#ref = r#ref.offset(clip(y, 0, ih - 1) as isize * ref_stride + clip(x, 0, iw - 1) as isize);
+    r#ref = r#ref.offset(clip(y, 0, ih - 1) * ref_stride + clip(x, 0, iw - 1));
 
     // number of pixels to extend (left, right, top, bottom)
-    let left_ext = clip(-x, 0, bw - 1);
-    let right_ext = clip(x + bw - iw, 0, bw - 1);
+    let left_ext = clip(-x, 0, bw - 1) as usize;
+    let right_ext = clip(x + bw - iw, 0, bw - 1) as usize;
     assert!(((left_ext + right_ext) as isize) < bw);
-    let top_ext = clip(-y, 0, bh - 1);
-    let bottom_ext = clip(y + bh - ih, 0, bh - 1);
+    let top_ext = clip(-y, 0, bh - 1) as usize;
+    let bottom_ext = clip(y + bh - ih, 0, bh - 1) as usize;
     assert!(((top_ext + bottom_ext) as isize) < bh);
 
+    let bw = bw as usize;
+    let bh = bh as usize;
+
     // copy visible portion first
-    let mut blk = top_ext as usize * dst_stride;
+    let mut blk = top_ext * dst_stride;
     let center_w = bw - left_ext - right_ext;
     let center_h = bh - top_ext - bottom_ext;
     for _ in 0..center_h {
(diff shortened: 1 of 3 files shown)


src/mc.rs
```diff
@@ -1004,78 +1006,72 @@ unsafe fn emu_edge_rust<BD: BitDepth>(
...
         );
+        // extend left edge for this line
         if left_ext != 0 {
-            BD::pixel_set(
-                std::slice::from_raw_parts_mut(blk, left_ext as usize),
-                *blk.offset(left_ext as isize),
-                left_ext as usize,
-            );
+            let val = dst[blk + left_ext];
+            dst[blk..][..left_ext].fill(val);
```
kkysen, line 1045:
```suggestion
            BD::pixel_set(&dst[blk..], val, left_ext);
```

I know these do the same thing, but maybe we should stick with the naming C uses?  That said, this way is indeed easier to read, so your call.

src/mc.rs
```diff
@@ -1004,78 +1006,72 @@ unsafe fn emu_edge_rust<BD: BitDepth>(
...
-            BD::pixel_set(
-                std::slice::from_raw_parts_mut(
-                    blk.offset(left_ext as isize).offset(center_w as isize),
-                    right_ext as usize,
-                ),
-                *blk.offset((left_ext + center_w - 1) as isize),
-                right_ext as usize,
-            );
+            let val = dst[blk + left_ext + center_w - 1];
+            dst[blk + left_ext + center_w..][..right_ext].fill(val);
```
kkysen, line 1050:
```suggestion
            BD::pixel_set(&dst[blk + left_ext + center_w..], val, right_ext);
```

src/mc.rs
```diff
@@ -1004,78 +1006,72 @@ unsafe fn emu_edge_rust<BD: BitDepth>(
...
-            bw as usize,
-        );
-        dst = dst.offset(BD::pxstride(dst_stride));
-        y_1 += 1;
+
+    // copy top
+    let mut dst_off = 0;
+    let blk = top_ext * dst_stride;
+    for _ in 0..top_ext {
+        let (front, back) = dst.split_at_mut(blk);
```
kkysen, line 1060: This can be outside of the loop, right?

src/mc.rs
```diff
@@ -1004,78 +1006,72 @@ unsafe fn emu_edge_rust<BD: BitDepth>(
...
+        BD::pixel_copy(&mut front[dst_off..][..bw], &mut back[..bw], bw);
```
kkysen, line 1061:
```suggestion
        BD::pixel_copy(&mut front[dst_off..][..bw], &back[..bw], bw);
```

src/mc.rs
```diff
@@ -1004,78 +1006,72 @@ unsafe fn emu_edge_rust<BD: BitDepth>(
...
+    // copy bottom
+    dst_off += center_h * dst_stride;
+    for _ in 0..bottom_ext {
+        let (front, back) = dst.split_at_mut(dst_off);
         BD::pixel_copy(
-            std::slice::from_raw_parts_mut(dst, bw as usize),
-            std::slice::from_raw_parts(dst.offset(-BD::pxstride(dst_stride)), bw as usize),
-            bw as usize,
+            &mut back[..bw],
+            &mut front[dst_off - dst_stride..][..bw],
```
kkysen, line 1071:
```suggestion
            &front[dst_off - dst_stride..][..bw],
```

kkysen (changes requested):
Couple of very little things; rest all LGTM.

src/mc.rs
randomPoison: Eh yeah I can go back to using `pixel_set`, it doesn't matter much either way.

src/mc.rs
randomPoison: Yup, moved.

kkysen (approved):

PR #1081, opened 2024-05-15
`struct Rav1dPictureDataComponent`: Store pic data in a `DisjointMut`, but still unsafely use `.as_mut_ptr()` for now

kkysen:
This adds `Rav1dPictureDataComponent`, which stores the ptr and length of each of the YUV components of a `Rav1dPictureData`.  This is wrapped in `DisjointMut` so that it can be safely accessed disjointly mutably, but for now, access is still done unsafely through `.as_mut_ptr()`.  Changing that to safe/checked `.index(_mut)`s can be done, but I'll have to narrow/split up a bunch of the borrows, and there are a lot of uses, so I'm leaving that to my next PR.

### DIFF
--- a/include/dav1d/picture.rs
+++ b/include/dav1d/picture.rs
@@ -99,7 +99,7 @@ pub struct Dav1dPicture {
 }
 
 pub struct Rav1dPictureData {
-    data: [*mut c_void; 3],
+    data: [NonNull<c_void>; 3],
     pub(crate) allocator_data: Option<NonNull<c_void>>,
     pub(crate) allocator: Rav1dPicAllocator,
 }
@@ -118,7 +118,7 @@ impl Drop for Rav1dPictureData {
 impl Rav1dPictureData {
     #[inline(always)]
     pub fn data<BD: BitDepth>(&self) -> [*mut BD::Pixel; 3] {
-        self.data.map(|data| data.cast())
+        self.data.map(|data| data.as_ptr().cast())
     }
 }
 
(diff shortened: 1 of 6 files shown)


include/dav1d/picture.rs
```diff
@@ -394,3 +435,66 @@ impl From<Rav1dPicAllocator> for Dav1dPicAllocator {
...
+        let allocator_data = pic_c.allocator_data;
+        let mut pic = pic_c.to::<Rav1dPicture>();
+        let len = pic.p.pic_len(pic.stride);
+        // TODO fallible allocation
+        pic.data = Some(Arc::new(Rav1dPictureData {
+            // SAFETY: `MaybeUninit<u8>` should be safe for anything.
+            data: array::from_fn(|i| {
+                let ptr = data[i].unwrap().cast::<AlignedPixelChunk>();
+                let len = len[(i != 0) as usize];
+                DisjointMut::new(Rav1dPictureDataComponent { ptr, len })
```
randomPoison, line 475: Is the allocated data guaranteed to be in multiples of `RAV1D_PICTURE_ALIGNMENT` and aligned to `RAV1D_PICTURE_ALIGNMENT`? And `len` is the size in bytes, not the number of chunks, right? Should we have an assert confirming that's the case? Otherwise this cast to `AlignedPixelChunk` could be incorrect.

kkysen: Oh I thought I had an alignment assert here.  Maybe I accidentally deleted it as I refactored.

kkysen: Yes, `len` is the size in bytes.  And yes, the data is required to be aligned to `RAV1D_PICTURE_ALIGNMEMT` by the C API.  I'll add the assert back.

kkysen: Added the `assert!(ptr.is_aligned())` back, deleted it accidentally.

kkysen: Also I clarified that the length is in bytes, and asserted that the length is a multiple of alignment.

randomPoison (approved):

Issue #842, opened 2024-03-18
`mc.rs`: Unsafe cleanup


randomPoison:
There are ~800 unsafe operations happening in `mc.rs`.

- [ ] Cleanup unsafe operations happening within functions.
- [ ] Mark any functions that can be made fully safe as such.
- [ ] Add `unsafe` blocks around any remaining unsafe operations along with standard safety documentation.
- [ ] Add safety documentation for functions that need to remain unsafe.
- [ ] Add `#![deny(unsafe_op_in_unsafe_fn)]` to file.

randomPoison:
`mc.rs` is mostly Rust fallback functions, which require raw pointer operations and so can't generally be made fully safe. I've done a cleanup pass on them and made things as safe and idiomatic as possible, but I've left the functions unsafe. I've also opted not to put `deny(unsafe_op_in_unsafe_fn)` because I think it would add more noise than is worth it. But otherwise the file has been cleaned up and made as safe and idiomatic as possible.

PR #1083, opened 2024-05-15
`fn mct`: Pass `tmp` len for use in Rust fallback fns

randomPoison:
The `tmp` arg for the `mct` fns is always a slice, so I've added an extra arg to the end of the function calls to pass the slice length in so that we can safely reconstruct the slice in the Rust fallback fns.

* Part of #842.


kkysen (commented):
Would it make sense to instead do `&mut tmp[..w * h]` and use `w * h` as the recreated length?  Since the length calculation is very simple here, I think that might be better.

randomPoison:
Yeah, in this case that seems reasonable. I've updated the PR to use that approach.

kkysen (approved):

PR #1085, opened 2024-05-15
`fn warp_affine_8x8t`: Pass `tmp` len for use in Rust fallback

randomPoison:
The `tmp` arg for the the `warp_affine_8x8t` fns is always a slice, so I've added an extra arg to the end of the function calls to pass the slice length in so that we can safely reconstruct the slice in the Rust fallback fns.

* Part of #842.

### DIFF
--- a/src/mc.rs
+++ b/src/mc.rs
@@ -947,8 +947,8 @@ unsafe fn warp_affine_8x8_rust<BD: BitDepth>(
 }
 
 unsafe fn warp_affine_8x8t_rust<BD: BitDepth>(
-    mut tmp: *mut i16,
-    tmp_stride: ptrdiff_t,
+    mut tmp: &mut [i16],
+    tmp_stride: usize,
     mut src: *const BD::Pixel,
     src_stride: ptrdiff_t,
     abcd: &[i16; 4],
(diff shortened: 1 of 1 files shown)


src/mc.rs
```diff
@@ -1305,11 +1306,19 @@ impl warp8x8t::Fn {
         my: c_int,
         bd: BD,
     ) {
-        let tmp = tmp.as_mut_ptr();
-        let tmp_stride = tmp_stride as isize;
         let src = src.cast();
         let bd = bd.into_c();
-        self.get()(tmp, tmp_stride, src, src_stride, abcd, mx, my, bd)
+        self.get()(
+            tmp.as_mut_ptr(),
```
kkysen, line 1312: Btw, I've been preferring putting casts, etc. on their own lines as `let`s in these wrapper methods.  Just slightly easier to read IMO.

src/mc.rs
```diff
@@ -1920,14 +1929,16 @@ pub(crate) unsafe extern "C" fn warp_affine_8x8_c_erased<BD: BitDepth>(
...
+    tmp_stride: usize,
     src: *const DynPixel,
     src_stride: ptrdiff_t,
     abcd: &[i16; 4],
     mx: c_int,
     my: c_int,
     bitdepth_max: c_int,
+    tmp_len: usize,
 ) {
+    let tmp = std::slice::from_raw_parts_mut(tmp, tmp_len);
```
kkysen, line 1941: Also tiny little nit:

```suggestion
    let tmp = slice::from_raw_parts_mut(tmp, tmp_len);
```

kkysen (approved):

src/mc.rs
randomPoison: All the other instances of this in the file are doing `std::slice::from_raw_parts`, so I'm going to import `std::slice` and update all the instances to `slice::from_raw_parts`.

PR #1088, opened 2024-05-16
Remove unnecessary `unsafe` block

randomPoison:
This `unsafe` block is no longer necessary after #1082.


randomPoison:
Duplicate of https://github.com/memorysafety/rav1d/pull/1086

PR #1091, opened 2024-05-16
`mc.rs`: Restore support for negative strides in Rust fallback fns

randomPoison:
Fixes the places where we were incorrectly converting potentially-negative stride values to `usize`. In a few places this meant reverting slices and iterator operations back to raw pointer operations.

* Part of #636.

### DIFF
--- a/src/mc.rs
+++ b/src/mc.rs
@@ -58,18 +58,19 @@ unsafe fn put_rust<BD: BitDepth>(
 #[inline(never)]
 unsafe fn prep_rust<BD: BitDepth>(
     tmp: &mut [i16],
-    src: *const BD::Pixel,
-    src_stride: usize,
+    mut src: *const BD::Pixel,
+    src_stride: isize,
     w: usize,
     h: usize,
     bd: BD,
 ) {
-    let src = slice::from_raw_parts(src, if h == 0 { 0 } else { src_stride * (h - 1) + w });
     let intermediate_bits = bd.get_intermediate_bits();
-    for (tmp, src) in iter::zip(tmp.chunks_exact_mut(w), src.chunks(src_stride)).take(h) {
-        for (tmp, src) in iter::zip(tmp, &src[..w]) {
-            *tmp = (((*src).as_::<i32>() << intermediate_bits) - (BD::PREP_BIAS as i32)) as i16;
+    for tmp in tmp.chunks_exact_mut(w).take(h) {
+        for x in 0..w {
+            tmp[x] =
+                (((*src.add(x)).as_::<i32>() << intermediate_bits) - (BD::PREP_BIAS as i32)) as i16
         }
+        src = src.offset(src_stride);
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/mc.rs
```diff
@@ -36,46 +35,50 @@ use crate::include::common::bitdepth::{bpc_fn, BPC};
...
     let tmp = slice::from_raw_parts_mut(tmp, w * h);
-    let src = slice::from_raw_parts(src, if h == 0 { 0 } else { src_stride * (h - 1) + w });
     let intermediate_bits = bd.get_intermediate_bits();
-    for (tmp, src) in iter::zip(tmp.chunks_exact_mut(w), src.chunks(src_stride)).take(h) {
-        for (tmp, src) in iter::zip(tmp, &src[..w]) {
-            *tmp = (((*src).as_::<i32>() << intermediate_bits) - (BD::PREP_BIAS as i32)) as i16;
+    for tmp in tmp.chunks_exact_mut(w).take(h) {
+        for x in 0..w {
+            tmp[x] =
+                (((*src.add(x)).as_::<i32>() << intermediate_bits) - (BD::PREP_BIAS as i32)) as i16
```
kkysen, line 71: Can we still make `src` a slice for the inner loop?

src/mc.rs
```diff
@@ -846,17 +842,16 @@ unsafe fn w_mask_rust<BD: BitDepth>(
     let rnd = (32 << intermediate_bits) + i32::from(BD::PREP_BIAS) * 64;
     let mask_sh = bitdepth + intermediate_bits - 4;
     let mask_rnd = 1 << (mask_sh - 5);
-    for (h, ((tmp1, tmp2), dst)) in iter::zip(tmp1.chunks_exact(w), tmp2.chunks_exact(w))
-        .zip(dst.chunks_mut(dst_stride))
-        .enumerate()
-    {
+    let mut tmp1 = tmp1.as_slice();
```
kkysen, line 845: Can we keep the `tmp` iterator?

kkysen (changes requested):
Can we keep the iterators for `tmp`.  I saw you did for some but not for a few others. 

Also, inside the outer loop, can we make `dst`/`src` a slice?  Where we're just iterating over `0..h` or something?

src/mc.rs
randomPoison: I don't see much point to that, we're just replacing one unsafe operation with another.

src/mc.rs
randomPoison: Sure, done.

src/mc.rs
kkysen: It'll eventually be a `DisjointMut` in the outer loop, though.

randomPoison: Ah sure, if it'll make that change cleaner then it makes more sense. Done.

fbossen (approved):

src/mc.rs
```diff
@@ -178,7 +180,7 @@ unsafe fn put_8tap_rust<BD: BitDepth>(
             mid_ptr = &mut mid[128 * 3..];
             for _ in 0..h {
                 for x in 0..w {
-                    dst[x] = rav1d_filter_8tap_clip(
+                    *dst.add(x) = rav1d_filter_8tap_clip(
```
kkysen, line 183: Can you do these other `dst.add(x)` ones, too?

randomPoison: Done.

kkysen (approved):

PR #1092, opened 2024-05-16
Fix use of negative strides in loop filters

### LINKED ISSUE #636: Account for negative strides

fbossen:
Fixes slice construction and offset computation when negative strides are used.
* Partly fixes #636

### DIFF
--- a/src/recon.rs
+++ b/src/recon.rs
@@ -4324,7 +4324,7 @@ pub(crate) unsafe fn rav1d_filter_sbrow_deblock_cols<BD: BitDepth>(
         ];
         let p_offset = [
             (cmp::max(0, -y_span) + y as isize * y_stride) as usize,
-            (cmp::max(0, -uv_span) + y as isize * uv_stride >> ss_ver) as usize,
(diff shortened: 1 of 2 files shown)


src/recon.rs
```diff
@@ -4426,23 +4426,38 @@ pub(crate) unsafe fn rav1d_filter_sbrow_lr<BD: BitDepth>(
     let y = sby * f.sb_step * 4;
     let ss_ver = (f.cur.p.layout == Rav1dPixelLayout::I420) as c_int;
     let h = f.sr_cur.p.p.h + 127 & !127;
+    let pix_stride = f.sr_cur.p.stride.map(BD::pxstride);
```
kkysen, line 4429: Could we name this `pxstride` to be consistent with `BD::pxstride`.  `pix` sounds a little too much like pic IMO.

kkysen (approved):

src/recon.rs
fbossen: done

kkysen (approved):

PR #1096, opened 2024-05-18
`fn Rav1dPicAllocator::alloc_picture_data`: Allow null and replace with `NonNull::dangling`

kkysen:
@fbossen pointed out that the pic allocator can return null for the chroma data, but I had added an `Option<NonNull<_>>::unwrap`.  This changes that to use `NonNull::dangling()` instead, which is valid to use to construct 0-length slices.

### DIFF
--- a/include/dav1d/picture.rs
+++ b/include/dav1d/picture.rs
@@ -479,7 +479,10 @@ impl Rav1dPicAllocator {
         pic.data = Some(Arc::new(Rav1dPictureData {
             // SAFETY: `MaybeUninit<u8>` should be safe for anything.
             data: array::from_fn(|i| {
-                let ptr = data[i].unwrap().cast::<AlignedPixelChunk>();
+                let ptr = data[i]
+                    // Need to cast before `NonNull::dangling` to get the right alignment.
+                    .map(|ptr| ptr.cast::<AlignedPixelChunk>())
+                    .unwrap_or_else(NonNull::dangling);
                 let len = len[(i != 0) as usize];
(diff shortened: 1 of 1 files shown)


include/dav1d/picture.rs
```diff
@@ -479,7 +479,9 @@ impl Rav1dPicAllocator {
         pic.data = Some(Arc::new(Rav1dPictureData {
             // SAFETY: `MaybeUninit<u8>` should be safe for anything.
             data: array::from_fn(|i| {
-                let ptr = data[i].unwrap().cast::<AlignedPixelChunk>();
+                let ptr = data[i]
+                    .unwrap_or_else(NonNull::dangling)
```
fbossen, line 483: Looks like `NonNull::dangling` may not have the same alignment as `AlignedPixelChunk` and `Rav1dPictureDataComponent::new` fails a couple lines below.

kkysen: Uhh, okay. I need to do the cast first, then it should work.

kkysen: What tests are you running for this?

fbossen: Running a modified argon streams script that calls `dav1d` with `--negstride`

kkysen: Fixed it now, I think.  Let me know if it worked.  I tested it on a few things with `--negstride` and it seemed to work.

fbossen: Works for me

fbossen (approved):

PR #1097, opened 2024-05-18
`fn dav1d_default_picture_alloc`: Zero initialize pic data with `fn MemPool::pop_init`

kkysen:
`fn MemPool::pop_init` uses `vec![]` and thus can optimize to `alloc_zeroed`, which lets the OS handle zeroing out pages, which is far faster.  Thus, we can now safely initialize picture data without significant perf overhead, at least for the default pic allocator.

From my basic benchmarking, it appears this eliminates the perf overhead of initialization that I previously saw with `buf.resize(pic_cap, 0)`.  @fbossen and/or @rinon, could you check this more carefully?

### DIFF
--- a/src/mem.rs
+++ b/src/mem.rs
@@ -11,7 +11,7 @@ impl<T> MemPool<T> {
         }
     }
 
-    pub fn pop(&self, size: usize) -> Vec<T> {
+    pub fn _pop(&self, size: usize) -> Vec<T> {
         if let Some(mut buf) = self.bufs.lock().unwrap().pop() {
             if size > buf.capacity() {
                 // TODO fallible allocation
@@ -21,6 +21,24 @@ impl<T> MemPool<T> {
         }
         // TODO fallible allocation
         Vec::with_capacity(size)
+    }
+
+    /// A version of [`Self::pop`] that initializes the [`Vec`].
+    /// This allows it to use [`vec!`], which, if used with `0`,
+    /// calls [`alloc_zeroed`], and thus can leave zero initialization to the OS.
+    ///
+    /// [`alloc_zeroed`]: std::alloc::alloc_zeroed
+    pub fn pop_init(&self, size: usize, init_value: T) -> Vec<T>
+    where
+        T: Copy,
+    {
+        if let Some(mut buf) = self.bufs.lock().unwrap().pop() {
+            // TODO fallible allocation
+            buf.resize(size, init_value);
+            return buf;
+        }
+        // TODO fallible allocation
+        vec![init_value; size]
     }
 
     pub fn push(&self, buf: Vec<T>) {
--- a/src/mem.rs
+++ b/src/mem.rs
@@ -34,7 +34,9 @@ impl<T> MemPool<T> {
     {
         if let Some(mut buf) = self.bufs.lock().unwrap().pop() {
             // TODO fallible allocation
-            buf.resize(size, init_value);
+            if buf.len() < size {
+                buf.resize(size, init_value);
+            }
             return buf;
         }
         // TODO fallible allocation
--- a/src/mem.rs
+++ b/src/mem.rs
@@ -32,12 +32,10 @@ impl<T> MemPool<T> {
     where
         T: Copy,
     {
-        if let Some(mut buf) = self.bufs.lock().unwrap().pop() {
-            // TODO fallible allocation
-            if buf.len() < size {
-                buf.resize(size, init_value);
+        if let Some(buf) = self.bufs.lock().unwrap().pop() {
+            if size <= buf.len() {
+                return buf;
             }
-            return buf;
         }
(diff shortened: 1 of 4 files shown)


kkysen:
Also, the algorithm for finding a buffer that may be reused is very naive (it just pops the last one).  In the benchmarks we've been doing, all the buffers are the same size, so this doesn't matter at all, but a smarter algorithm (e.x. first look for a buffer that is big enough already) could be a lot better in more real-world scenarios where the videos have much more differing sizes.  This might be something we should look into more to improve performance, as little things here make a big difference.

src/mem.rs
```diff
@@ -23,6 +23,26 @@ impl<T> MemPool<T> {
         Vec::with_capacity(size)
     }
 
+    /// A version of [`Self::pop`] that initializes the [`Vec`].
+    /// This allows it to use [`vec!`], which, if used with `0`,
+    /// calls [`alloc_zeroed`], and thus can leave zero initialization to the OS.
+    ///
+    /// [`alloc_zeroed`]: std::alloc::alloc_zeroed
+    pub fn pop_init(&self, size: usize, init_value: T) -> Vec<T>
```
randomPoison, line 31: If the point of this function is to allow for zero-initialization, does it make sense to have the caller specify the init value? It seems like it'd be better to always zero-initialize.

randomPoison (commented):
This seems fine, but how much benefit are we getting from having the default pic allocator initialize the data when we still need to handle potentially uninitialized data from custom allocators? Does this make anything about our usage of the pic data safer?

src/mem.rs
rinon: What about `T: Copy + Default` and we use `buf.resize(size, Default::default())`?

rinon:
It looks to me like buffer reuse with a modified size is a slow path in the original C code, so I don't think zeroing that memory will be an issue. @fbossen do you see any issues with this in practice?

fbossen:
> From my basic benchmarking, it appears this eliminates the perf overhead of initialization that I previously saw with `buf.resize(pic_cap, 0)`. @fbossen and/or @rinon, could you check this more carefully?

What benchmarking did you do?

src/mem.rs
kkysen: > What about `T: Copy + Default` and we use `buf.resize(size, Default::default())`?

`buf.resize` doesn't have the optimization, `Vec::from_elem` (i.e. `vec!`) does.

> If the point of this function is to allow for zero-initialization, does it make sense to have the caller specify the init value? It seems like it'd be better to always zero-initialize.

It's easier this way with `MemPool` being generic.  I could also do something like `pop_init_default` and use `Default::default()`, but I don't really see much benefit in doing that instead (it just constrains the API more).

src/mem.rs
```diff
@@ -11,7 +11,7 @@ impl<T> MemPool<T> {
         }
     }
 
-    pub fn pop(&self, size: usize) -> Vec<T> {
+    pub fn _pop(&self, size: usize) -> Vec<T> {
```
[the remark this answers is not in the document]
kkysen: No, not now, but if we need to add back pooling for any of the other `Arc`s, we may want to use this (they're not necessarily zero-initialized I think), so I think it's good to keep it around.

fbossen:
> It looks to me like buffer reuse with a modified size is a slow path in the original C code, so I don't think zeroing that memory will be an issue. @fbossen do you see any issues with this in practice?

Generally one would avoid filling a large buffer with any default value because it tends to have a negative impact on caching

fbossen:
> @fbossen, do you know the usage patterns around this. Are `Dav1dContext`s reused for multiple videos. Do picture sizes change throughout videos at all? If they do, do they ever change back and forth quickly?

Picture sizes can change throughout videos. Generally, such changes wouldn't occur very often (e.g., once per second or less). In the worst case picture sizes could change every frame.

kkysen:
@fbossen and @rinon, I changed `pop_init` slightly in b7a1bf2 to avoid the slow `.resize`/`memset` path.  So we only reuse the buffer if it's totally free (no allocation, no `memset`).  Otherwise, we zero allocate a new one.  This more closely matches C, which reallocates if the size is different at all (`pop_init` allows the existing buffer to be larger).

rinon:
If we're not growing these buffers now, wouldn't it make more sense for them to be a `Box<[u8]>`? Growing a vector with a potential copy if there is not room to grow is different semantics than a fresh allocation.

kkysen:
> If we're not growing these buffers now, wouldn't it make more sense for them to be a `Box<[u8]>`? Growing a vector with a potential copy if there is not room to grow is different semantics than a fresh allocation.

Yeah, maybe.  Can I do that as a separate PR?

rinon:
Sure, makes sense

PR #1099, opened 2024-05-18
`fn DisjointMut::{,mut_}{slice,element}_as`: Replace `u8` with `U: AsBytes` for more flexibility

kkysen:
This way we can allow the underlying data to be larger than just bytes (e.x. `AlignedPixelChunk`), allowing us to avoid lots of alignment checks.


randomPoison (approved):

kkysen:
It's difficult to translate the ranges when doing this (I think it's incorrect here), and I figured out how to get around this a different way.

PR #1102, opened 2024-05-19
`trait TranslateRange`: Replace `fn Bounds::multiply` as `Bounds` uses `usize::MAX` for unknown ends, which can't be multiplied

kkysen:
Since `Bounds` represents `start..` and `..` with `usize::MAX` for the end bound, multiplying it is incorrect.  This adds `trait TranslateRange` to do the multiplication on the generic range before converting to `Bounds`.

### DIFF
--- a/src/disjoint_mut.rs
+++ b/src/disjoint_mut.rs
@@ -23,6 +23,7 @@ use std::ops::IndexMut;
 use std::ops::Range;
 use std::ops::RangeBounds;
 use std::ops::RangeFrom;
+use std::ops::RangeFull;
 use std::ops::RangeInclusive;
 use std::ops::RangeTo;
 use std::ops::RangeToInclusive;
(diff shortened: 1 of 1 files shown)


src/disjoint_mut.rs
```diff
@@ -496,12 +582,6 @@ impl Bounds {
         let b = &other.range;
         a.start < b.end && b.start < a.end
     }
-
-    fn multiply(self, multiple: usize) -> Bounds {
-        let start = self.range.start * multiple;
-        let end = self.range.end * multiple;
```
rinon, line 502: Why not just check range.end here and only multiply if not `usize::MAX`? Would be a lot simpler than all the other stuff here.

kkysen: Using `usize::MAX` also prohibits other optimizations since it conflates actual `usize::MAX`.  And this isn't complex code, just somewhat repetitive.  I can remove the division ones since I ended up not needing them, though.  I can try to see if I can rearrange things to only need `fn mul` on `Range<usize>`, but I don't think this code is very complex.

kkysen: I removed `fn div` in [`edf6ae0..36164b9`](https://github.com/memorysafety/rav1d/compare/edf6ae0621aaab9438635a2a5a1f8ac03af21a8e..36164b9dbf24c725b22bf3b9a54df86d05cd7978).  I can see about other simplifications, too, but I'd like to merge this as is and fix changes on top of the stack, since this PR is at the bottom.  Also, some extra lines of code here may be worth the performance, since `usize::MAX` checks may be in the hot path of whatever indexing we're doing.

PR #1103, opened 2024-05-19
`fn DisjointMut::index{,_mut}`: Optimize

kkysen:
This inlines all of the `fn index`-type methods, allowing the bounds checks to be in the caller so that they can be optimized with knowledge from the caller.  At the same time, the error paths are put in `#[inline(never)]` inner `fn`s so that they don't bloat the code at all.  Then, the `Bounds::from` calls in `fn DisjointMutIndex::get_mut` are replaced with `fn SliceBounds::to_range`, which uses `len` directly instead of `usize::MAX`.  `Bounds::from` first substitutes `usize::MAX` for `len`, and then converts it back, but this conflates a range given as `..usize::MAX`, and thus can't be optimized as well.  Finally, the `start < len` check is redundant when we're already checking that `start <= end` and `end <= len`.  I thought LLVM could optimize this out, but it can't, and `std` does it this way.

This should lead to far better optimization for `DisjointMut` indexing and much less code bloat as well (panicking code is quite bloated).

### DIFF
--- a/src/disjoint_mut.rs
+++ b/src/disjoint_mut.rs
@@ -8,6 +8,7 @@
 use crate::src::align::AlignedByteChunk;
 use crate::src::align::AlignedVec;
 use std::cell::UnsafeCell;
+use std::convert::Infallible;
 use std::fmt;
 use std::fmt::Debug;
 use std::fmt::Display;
@@ -273,6 +274,7 @@ impl<T: ?Sized + AsMutPtr> DisjointMut<T> {
     ///
     /// [`index`]: DisjointMut::index
     /// [`index_mut`]: DisjointMut::index_mut
+    #[inline] // Inline to see bounds checks in order to potentially elide them.
     #[cfg_attr(debug_assertions, track_caller)]
     pub fn index_mut<'a, I>(&'a self, index: I) -> DisjointMutGuard<'a, T, I::Output>
     where
@@ -308,6 +310,7 @@ impl<T: ?Sized + AsMutPtr> DisjointMut<T> {
     /// during the lifetime of the returned borrow.
     ///
     /// [`index_mut`]: DisjointMut::index_mut
+    #[inline] // Inline to see bounds checks in order to potentially elide them.
     #[cfg_attr(debug_assertions, track_caller)]
     pub fn index<'a, I>(&'a self, index: I) -> DisjointImmutGuard<'a, T, I::Output>
     where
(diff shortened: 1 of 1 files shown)


src/disjoint_mut.rs
```diff
@@ -273,6 +274,7 @@ impl<T: ?Sized + AsMutPtr> DisjointMut<T> {
     ///
     /// [`index`]: DisjointMut::index
     /// [`index_mut`]: DisjointMut::index_mut
+    #[inline(always)] // Need to inline to see bounds checks in order to potentially elide them.
```
rinon, line 277: I mentioned this on another review, but we should use `#[inline]` instead.

src/disjoint_mut.rs
```diff
@@ -644,39 +657,35 @@ impl From<RangeFull> for Bounds {
...
-            unsafe { (slice as *mut T).add(self) }
+            unsafe { (slice as *mut T).add(index) }
         } else {
-            panic!("index out of bounds: the len is {len} but the index is {self}");
+            #[inline(never)]
+            #[cfg_attr(debug_assertions, track_caller)]
+            fn out_of_bounds(index: usize, len: usize) -> Infallible {
+                panic!("index out of bounds: the len is {len} but the index is {index}")
+            }
+            match out_of_bounds(index, len) {} // Infallible
```
rinon, line 688: This seems like an unusual pattern, is this modeled after something?

src/disjoint_mut.rs
```diff
@@ -687,30 +696,35 @@ where
...
+            fn out_of_bounds(start: usize, end: usize, len: usize) -> Infallible {
+                if start > end {
+                    panic!("slice index starts at {start} but ends at {end}");
+                }
+                if end > len {
+                    panic!("range end index {end} out of range for slice of length {len}");
+                }
+                if start >= len {
+                    panic!("range start index {start} out of range for slice of length {len}")
+                }
```
rinon, line 724: You took this case out? (zero-length slice) That's fine, but do we need this panic then?

src/disjoint_mut.rs
kkysen: No.  I was hoping the compiler would understand `Infallible` better but it didn't.  In `std` they just use `!` but that's unstable. `match`ing on `Infallible` (a variantless-`enum`) seemed like it worked, so I did that.  If there's a better way I can definitely switch to it.

src/disjoint_mut.rs
kkysen: I don't think so?  I wasn't paying attention to this much since it's in the error path, but it does seem like this one is unreachable now, right?

src/disjoint_mut.rs
kkysen: Turns out you actually can use `!` stably as long as it's only the return type, so I switched to that in [`10f5165..e25ab1c`](https://github.com/memorysafety/rav1d/compare/10f5165a2d60c136c929a4df64e33475a8c78c2d..e25ab1c181c7bf6103ba2815937939ae70b37a94).

src/disjoint_mut.rs
kkysen: Fixed in [`e25ab1c..3b2e057`](https://github.com/memorysafety/rav1d/compare/e25ab1c181c7bf6103ba2815937939ae70b37a94..3b2e057496de15d60bc039e9af9c288224bfa1ae).

PR #1104, opened 2024-05-19
`fn backup4xU`: fix handling of negative strides

fbossen:
Limit the number of iterations before reversing the iterator.

Partly fixes #646


src/lr_apply.rs
```diff
@@ -137,6 +137,7 @@ fn backup4xU<BD: BitDepth>(
     if src_stride < 0 {
         for (src, dst) in src[src_offset - (u - 1) as usize * abs_px_stride..]
             .chunks(abs_px_stride)
+            .take(u as usize)
             .rev()
             .zip(&mut dst[..u as usize])
```
kkysen, line 142: Is the reason we need this here because the `.take(u)` needs to be before the `.rev()`?  Because otherwise the `.zip(&mut dst[..u])` should handle it.

fbossen: Yes, `.take(u)` needs to happen before `.rev()`.

kkysen (approved):

Issue #815, opened 2024-03-15
Cleanup c2rust output in `ipred.rs`


randomPoison:
* [ ] Cleanup helper functions to take safe arguments where possible (may take raw pointers as arguments currently).
* [ ] General cleanup of c2rust generated code in functions.
* [ ] Restore comments from original C.

The following functions still need to be cleaned up:

- [x] `splat_dc`
- [x] `cfl_pred`
- [x] `dc_gen_top`
- [x] `dc_gen_left`
- [x] `dc_gen`
- [x] `ipred_v_rust`
- [x] `ipred_h_rust`
- [x] `ipred_paeth_rust`
- [x] `ipred_smooth_rust`
- [x] `ipred_smooth_v_rust`
- [x] `ipred_smooth_h_rust`
- [x] `get_filter_strength`
- [x] `filter_edge`
- [x] `upsample_edge`
- [x] `ipred_z1_rust`
- [x] `ipred_z2_rust`
- [x] `ipred_z3_rust`
- [x] `filter_fn`
- [x] `ipred_filter_rust`
- [x] `cfl_ac_rust`
- [x] `pal_pred_rust`

randomPoison:
Done as of #1184

PR #1105, opened 2024-05-20
`fn splat_dc`: Cleanup and make safer

randomPoison:
Part of https://github.com/memorysafety/rav1d/issues/815.

### DIFF
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -187,16 +187,16 @@ unsafe fn splat_dc<BD: BitDepth>(
             if width > 4 {
                 let dcN = dc as u64 * 0x101010101010101;
                 for _ in 0..height {
-                    for x in (0..width).step_by(mem::size_of::<u64>()) {
-                        *(&mut *dst.offset(x as isize) as *mut BD::Pixel as *mut u64) = dcN;
+                    for x in (0..width as usize).step_by(mem::size_of::<u64>()) {
+                        *dst.add(x).cast::<u64>() = dcN;
                     }
                     dst = dst.offset(stride);
                 }
             } else {
                 let dcN = dc as u32 * 0x1010101;
                 for _ in 0..height {
-                    for x in (0..width).step_by(mem::size_of::<u32>()) {
-                        *(&mut *dst.offset(x as isize) as *mut BD::Pixel as *mut c_uint) = dcN;
+                    for x in (0..width as usize).step_by(mem::size_of::<u32>()) {
+                        *dst.add(x).cast::<u32>() = dcN;
                     }
                     dst = dst.offset(stride);
                 }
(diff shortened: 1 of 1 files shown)


src/ipred.rs
```diff
@@ -181,56 +180,39 @@ unsafe fn splat_dc<BD: BitDepth>(
...
 ) {
+    let stride = BD::pxstride(stride);
+    let width = width as usize;
     match BD::BPC {
         BPC::BPC8 => {
-            if !(dc <= 0xff as c_int) {
-                unreachable!();
-            }
+            assert!(dc <= 0xff);
             if width > 4 {
```
kkysen, line 188: @fbossen, is `width` a power of 2 here?  Otherwise, this check doesn't make sense, since `width = 5` would overwrite for an 8-byte `u64`.  If `width` is a power of 2, this makes sense, but `width >= 8` would be much clearer.

kkysen (approved):
The changes here look fine, but we can be safer.  The multiplication by `0x10...` or `0x1000...` are similar to the ones that we in `case_set` (they multiply each element by 1, and thus essentially do nothing).  `dc` here fits in one `BD::Pixel`, so what the code is doing is basically a `.fill` with `dc`, bt taking advantage of higher alignment to be more efficient (doing it as `u64`s or `u32`s).  We can do this safely with `zerocopy`, safely transmuting to `const N: usize = 
RAV1D_PICTURE_ALIGNMENT / mem::size_of::<uN>()` and `[uN; N]` (where `uN` is `u8` or `u16`) and calling `.fill([uN; dc])`.  This is assuming `dst` is from picture data and thus `RAV1D_PICTURE_ALIGNMENT`-aligned, but if it's smaller, we can do the same as the current code, just safely.

randomPoison:
> We can do this safely with `zerocopy`, safely transmuting to `const N: usize = RAV1D_PICTURE_ALIGNMENT / mem::size_of::<uN>()` and `[uN; N]` (where `uN` is `u8` or `u16`) and calling `.fill([uN; dc])`.

I'm not following here. Could you provide a code example of what you're thinking of? I see how we could e.g. in `bpc == 8`, `width > 4` case do `[dc as u8; 8]` instead of the multiplication with `dcN` to get the same effect, but I don't follow how we could use zerocopy to make this any safer than it already is.

kkysen:
> > We can do this safely with `zerocopy`, safely transmuting to `const N: usize = RAV1D_PICTURE_ALIGNMENT / mem::size_of::<uN>()` and `[uN; N]` (where `uN` is `u8` or `u16`) and calling `.fill([uN; dc])`.
> 
> I'm not following here. Could you provide a code example of what you're thinking of? I see how we could e.g. in `bpc == 8`, `width > 4` case do `[dc as u8; 8]` instead of the multiplication with `dcN` to get the same effect, but I don't follow how we could use zerocopy to make this any safer than it already is.

Sure.  Can we merge this first?

Issue #846, opened 2024-03-18
`ipred.rs`: Unsafe cleanup


randomPoison:
There are ~300 unsafe operations happening in `ipred.rs`.

- [ ] Cleanup unsafe operations happening within functions.
- [ ] Mark any functions that can be made fully safe as such.
- [ ] Add `unsafe` blocks around any remaining unsafe operations along with standard safety documentation.
- [ ] Add safety documentation for functions that need to remain unsafe.
- [ ] Add `#![deny(unsafe_op_in_unsafe_fn)]` to file.

PR #1106, opened 2024-05-20
`fn cfl_pred`: Cleanup and make safer

randomPoison:
* Part of #815.
* Part of #846.

### DIFF
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -248,12 +248,14 @@ unsafe fn cfl_pred<BD: BitDepth>(
     bd: BD,
 ) {
     let mut ac = ac.as_slice();
+    let width = width as usize;
     for _ in 0..height {
-        for x in 0..width {
-            let diff = alpha * ac[x as usize] as c_int;
-            *dst.offset(x as isize) = bd.iclip_pixel(dc + apply_sign(diff.abs() + 32 >> 6, diff));
+        let slice = slice::from_raw_parts_mut(dst, width);
+        for (x, dst) in slice.iter_mut().enumerate() {
+            let diff = alpha * ac[x] as c_int;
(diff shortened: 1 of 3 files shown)


src/ipred.rs
```diff
@@ -116,7 +116,7 @@ wrap_fn_ptr!(pub unsafe extern "C" fn cfl_pred(
     topleft: *const DynPixel,
     width: c_int,
     height: c_int,
-    ac: *const i16,
+    ac: &[i16; 1024],
```
kkysen, line 119: Could we write these `1024`s as `32 * 32` as they are in `int16_t ac[32 * 32];` in C?

src/ipred.rs
```diff
@@ -243,21 +243,20 @@ unsafe fn cfl_pred<BD: BitDepth>(
...
     bd: BD,
 ) {
-    let mut y = 0;
-    while y < height {
-        let mut x = 0;
-        while x < width {
-            let diff = alpha * *ac.offset(x as isize) as c_int;
-            *dst.offset(x as isize) = bd.iclip_pixel(dc + apply_sign(diff.abs() + 32 >> 6, diff));
-            x += 1;
+    let mut ac = ac.as_slice();
```
kkysen, line 250:
```suggestion
    let mut ac = &ac[..width * height];
```

This should do all the bounds checks on `ac` at once and elide them elsewhere.

kkysen (approved):
I left a couple of small comments, but overall it LGTM.

src/ipred.rs
randomPoison: I'd rather introduce a constant rather than repeating the same expression everywhere, will add that to the PR before merging.

PR #1108, opened 2024-05-21
`fn ipred_{v,h}_rust`: Cleanup and make safe

randomPoison:
* Part of #815.
* Part of #846.

### DIFF
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -478,7 +478,7 @@ unsafe fn ipred_v_rust<BD: BitDepth>(
     width: c_int,
     height: c_int,
 ) {
-    let width = width.try_into().unwrap();
+    let width = width as usize;
 
     for _ in 0..height {
         BD::pixel_copy(
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -364,6 +364,29 @@ impl DcGen {
     }
 }
 
+/// Reconstructs the reference to the topleft edge array from a pointer into the
+/// array and an offset from the start of the array.
+///
+/// The topleft pointer passed to asm is always a pointer into a buffer of
+/// length [`SCRATCH_EDGE_LEN`]. For the Rust fallbacks we also pass in the
+/// offset from the front of the buffer so that we can reconstruct the original
+/// array reference in order to use safe array operations within the fallbacks.
+///
+/// # Safety
+///
+/// `topleft_ptr` must be a pointer into an array of length [`SCRATCH_EDGE_LEN`]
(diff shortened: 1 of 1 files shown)


src/ipred.rs
```diff
@@ -364,6 +364,29 @@ impl DcGen {
...
+/// The topleft pointer passed to asm is always a pointer into a buffer of
+/// length [`SCRATCH_EDGE_LEN`]. For the Rust fallbacks we also pass in the
+/// offset from the front of the buffer so that we can reconstruct the original
+/// array reference in order to use safe array operations within the fallbacks.
+///
+/// # Safety
+///
+/// `topleft_ptr` must be a pointer into an array of length [`SCRATCH_EDGE_LEN`]
+/// and is `topleft_off` elements from the beginning of the array. This should
+/// be guaranteed by the logic in `angular_ipred::call`.
```
kkysen, line 379:
```suggestion
/// be guaranteed by the logic in [`angular_ipred::Fn::call`].
```

src/ipred.rs
```diff
@@ -364,6 +364,29 @@ impl DcGen {
...
+unsafe fn reconstruct_topleft<'a, BD: BitDepth>(
```
kkysen, line 380: What does this lifetime do if it's not used in any argument?

kkysen (approved):

src/ipred.rs
randomPoison: It's the lifetime of the returned reference. Since we're creating a reference from a raw pointer we need to explicitly annotate the output lifetime since there's no corresponding input lifetime.

src/ipred.rs
randomPoison: Ugh, I accidently merged this before pushing the change that fixed this, sorry. Can we live with this not being a link in the doc comment? I can put up a separate PR for that, but tbh that doesn't seem worth the effort to me.

kkysen: Yeah, sure, or you can just tack it onto some other commit.

src/ipred.rs
kkysen: But we don't give it an explicit lifetime we can call it.  I'm not sure how this works exactly.

randomPoison: It works the same way that dereferencing a raw pointer works. When you dereference a pointer Rust materializes an arbitrary lifetime based on how the returned reference is used. So here the output lifetime is effectively unconstrained, and will last as long as you use the reference since there's no source lifetime.

PR #1120, opened 2024-05-24
`fn pal_pred`: Make safe

kkysen:
This makes the rest of `fn pal_pred` safe.  `dst` is made safe by adding an extra `FFISafe<Rav1dPictureDataComponent>` arg.  `idx` is made safe by recreating the slice, with the length checked in `fn call`.  Thus, the fallback `fn pal_pred_rust` is now safe, so I marked `fn pal_pred::Fn::call` safe, too (just asserting the asm is safe).

### DIFF
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -191,7 +191,7 @@ impl pal_pred::Fn {
         let dst_ptr = dst.as_mut_ptr::<BD>().add(dst_offset).cast();
         let stride = dst.stride();
         let pal = pal.as_ptr().cast();
-        let idx = idx.as_ptr();
+        let idx = idx[..(w * h) as usize / 2].as_ptr();
         let dst = FFISafe::new(dst);
         self.get()(dst_ptr, stride, pal, idx, w, h, dst)
     }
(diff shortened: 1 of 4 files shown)


src/ipred.rs
```diff
@@ -1461,39 +1469,51 @@ unsafe extern "C" fn cfl_ac_c_erased<BD: BitDepth, const IS_SS_HOR: bool, const
...
-    pal_pred_rust::<BD>(dst.cast(), stride, &*pal.cast(), idx, w, h);
+    let dst_ptr = dst_ptr.cast::<BD::Pixel>();
+    // SAFETY: Was passed as `FFISafe::new(dst)` in `pal_pred::Fn::call`.
+    let dst = unsafe { FFISafe::get(dst) };
+    // SAFETY: Reverse of what was done in `pal_pred::Fn::call`.
+    let dst_offset = unsafe { dst_ptr.offset_from(dst.as_mut_ptr::<BD>()) } as usize;
+    // SAFETY: Undoing dyn cast in `pal_pred::Fn::call`.
+    let pal = unsafe { &*pal.cast() };
+    // SAFETY: Length sliced in `pal_pred::Fn::call`.
+    let idx = slice::from_raw_parts(idx, (w * h) as usize / 2);
```
randomPoison, line 1515: Why no `unsafe` block around `from_raw_parts` when you have one around the other unsafe ops?

randomPoison (approved):

PR #1124, opened 2024-05-26
`fn fg{y,uv}_32x32xn::Fn::call`: Make args safe `Rav1dPictureDataComponent`s

kkysen:
This replaces the ptr args of `fn fg{y,uv}_32x32xn::Fn::call` with `&Rav1dPictureDataComponent`s, doing the ptr offset calculations inside the `fn`.  This prepares things for making the `fn` ptr call pass extra `FFISafe<Rav1dPictureDataComponent>` args to make the fallbacks safe.

Note also that the `src` and `dst` strides have already been asserted to be equal before these `fn`s are called.

### DIFF
--- a/src/fg_apply.rs
+++ b/src/fg_apply.rs
@@ -188,21 +188,19 @@ pub(crate) unsafe fn rav1d_apply_grain_row<BD: BitDepth>(
     }
 
     let layout = r#in.p.layout.try_into().unwrap();
-    let uv_off = (row * BLOCK_SIZE) as isize * BD::pxstride(out.stride[1]) >> ss_y;
     if data.chroma_scaling_from_luma {
         for pl in 0..2 {
             dsp.fguv_32x32xn[layout].call(
-                out_data[1 + pl].as_strided_mut_ptr::<BD>().offset(uv_off),
-                in_data[1 + pl].as_strided_ptr::<BD>().offset(uv_off),
-                r#in.stride[1],
+                layout,
+                &out_data[1 + pl],
+                &in_data[1 + pl],
                 data,
                 cpw,
                 &scaling[0],
                 &grain_lut[1 + pl],
                 bh,
                 row,
-                luma_src,
-                r#in.stride[0],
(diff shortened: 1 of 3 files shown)


src/fg_apply.rs
```diff
@@ -160,11 +160,8 @@ pub(crate) unsafe fn rav1d_apply_grain_row<BD: BitDepth>(
     if data.num_y_points != 0 {
         let bh = cmp::min(h - row * BLOCK_SIZE, BLOCK_SIZE);
         dsp.fgy_32x32xn.call(
-            out_data[0]
-                .as_strided_mut_ptr::<BD>()
-                .offset((row * BLOCK_SIZE) as isize * BD::pxstride(out.stride[0])),
-            luma_src,
```
rinon, line 166: We're now computing `luma_src` in two places, how will we ensure they stay in sync?

rinon (commented):
Why are we doing this? Is there no Rav1dPictureComponenent with an offset? It seems like a bad idea add recomputing these offsets in multiple places (I know the existing code recomputes offsets, but we're adding more here).

kkysen:
> Why are we doing this? Is there no Rav1dPictureComponenent with an offset? It seems like a bad idea add recomputing these offsets in multiple places (I know the existing code recomputes offsets, but we're adding more here).

I'm mostly just moving the offset computation into `fn call`.  It's not just being duplicated.

kkysen:
> Is there no Rav1dPictureComponenent with an offset?

No, there's not yet.  I could add one, though.  We didn't add types with the offset included for other places we do similar things, though.

src/fg_apply.rs
rinon: I don't really buy that as a good practice, but this is ok for now.

rinon:
It might be nice to have picture with offset, just so we can ensure that these offsets are only computed once to avoid bugs. That said, that should be separate once we have most of the safety work done.

rinon (approved):

PR #1125, opened 2024-05-26
`fn fg{y,uv}_32x32xn_rust`: Use slices where possible

kkysen:
This prepares things for when the raw ptrs becomes `Rav1dPictureDataComponent`s, and thus fully safe.


rinon (approved):
I don't entirely understand why we need this but LGTM

PR #1126, opened 2024-05-26
`fn fg{y,uv}_32x32xn_rust`: Make safe

kkysen:
This adds extra `FFISafe<Rav1dPictureDataComponent>` args to the `fn` ptr to make the fallback safe.

### DIFF
--- a/src/filmgrain.rs
+++ b/src/filmgrain.rs
@@ -16,6 +16,7 @@ use crate::src::cpu::CpuFlags;
 use crate::src::enum_map::enum_map;
 use crate::src::enum_map::enum_map_ty;
 use crate::src::enum_map::DefaultValue;
+use crate::src::ffi_safe::FFISafe;
 use crate::src::internal::GrainLut;
 use crate::src::tables::dav1d_gaussian_sequence;
 use crate::src::wrap_fn_ptr::wrap_fn_ptr;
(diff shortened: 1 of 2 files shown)


src/filmgrain.rs
```diff
@@ -133,9 +136,14 @@ impl fgy_32x32xn::Fn {
...
         let bd = bd.into_c();
-        self.get()(
-            dst_row, src_row, stride, data, pw, scaling, grain_lut, bh, row_num, bd,
-        )
+        let dst = FFISafe::new(dst);
+        let src = FFISafe::new(src);
+        // SAFETY: Fallback `fn fgy_32x32xn_rust` is safe; asm is supposed to do the same.
+        unsafe {
+            self.get()(
+                dst_row, src_row, stride, data, pw, scaling, grain_lut, bh, row_num, bd, dst, src,
```
rinon, line 144: I really don't like the idea of adding extra arguments to ASM calls. This will add unnecessary stack manipulation to add the params before the call. Please be careful with this and benchmark this change.

kkysen: I haven't noticed any changes in the benchmarks.  I just doubt an extra `push` or `str` or two is going to have noticeable perf effects.

rinon: Ok, just wanted to highlight this as a thing worth checking.

rinon (approved):

src/filmgrain.rs
kkysen: Okay, I'll be sure to check it each time I add an extra arg.

PR #1135, opened 2024-05-27
`fn loopfilter_sb`: Use `wrap_fn_ptr!` and make call safer

kkysen:
The only thing I'm not sure about here is that I changed `mask` from a `&[u32; 4]` to a `&[u32; 3]` in 0a553a8 for the `lpf_{h,v}_y` `fn`s.  The 4th element is unused in the fallbacks, appears to be unused in the asm, and is always 0.  I had thought maybe it was there for some sort of `u32x4` OR, but I don't see any in the asm.

Other alternatives to this are still creating the `[u32; 4]`s for the `y` calls, but slicing it to a `&[u32; 3]`, or having different `fn` ptr definitions for `y` and `uv`.  They are accessed by constant index currently, so this would be possible.

@fbossen, does this sound okay?


randomPoison (approved):
Changes look fine by me but wait for Frank to confirm if changing the length of the mask array would cause any problems. If it's unclear then the safer thing to do would be to leave it with the original length.

PR #1136, opened 2024-05-27
Replace std library locks with parking_lot

rinon:
Replaces all locks with the versions from parking_lot. Since we need parking_lot for the task queue lock, we might as well use its versions everywhere.


rinon:
I've tested both on linux x86 AVX512 and OS X ARM64 and there's no meaningful difference on Chimera 10b.

PR #1141, opened 2024-05-29
`fn filter_edge`,`fn upsample_edge`: Cleanup and make safe

randomPoison:
This PR makes `filter_edge` and `upsample_edge` safe, though doing so also involved threading the safe array refs through `ipred_z*_rust`. The remainder of the cleanup for those functions will come in followup PRs.

Part of https://github.com/memorysafety/rav1d/issues/846 and https://github.com/memorysafety/rav1d/issues/815.

### DIFF
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -785,7 +785,7 @@ fn get_upsample(wh: c_int, angle: c_int, is_sm: bool) -> bool {
 
 #[inline(never)]
 unsafe fn upsample_edge<BD: BitDepth>(
-    out: *mut BD::Pixel,
+    out: &mut [BD::Pixel],
     hsz: c_int,
     in_0: *const BD::Pixel,
     from: c_int,
@@ -796,7 +796,7 @@ unsafe fn upsample_edge<BD: BitDepth>(
     let mut i;
     i = 0 as c_int;
     while i < hsz - 1 {
-        *out.offset((i * 2) as isize) = *in_0.offset(iclip(i, from, to - 1) as isize);
+        out[(i * 2) as usize] = *in_0.offset(iclip(i, from, to - 1) as isize);
         let mut s = 0;
         let mut j = 0;
         while j < 4 {
@@ -804,11 +804,11 @@ unsafe fn upsample_edge<BD: BitDepth>(
                 * kernel[j as usize] as c_int;
             j += 1;
         }
-        *out.offset((i * 2 + 1) as isize) =
+        out[(i * 2 + 1) as usize] =
             iclip(s + 8 >> 4, 0 as c_int, bd.bitdepth_max().as_::<c_int>()).as_::<BD::Pixel>();
         i += 1;
     }
-    *out.offset((i * 2) as isize) = *in_0.offset(iclip(i, from, to - 1) as isize);
+    out[(i * 2) as usize] = *in_0.offset(iclip(i, from, to - 1) as isize);
 }
 
 unsafe fn ipred_z1_rust<BD: BitDepth>(
@@ -839,7 +839,7 @@ unsafe fn ipred_z1_rust<BD: BitDepth>(
     };
     if upsample_above {
         upsample_edge::<BD>(
-            top_out.as_mut_ptr(),
+            &mut top_out,
             width + height,
             &*topleft_in.offset(1),
             -(1 as c_int),
@@ -938,7 +938,7 @@ unsafe fn ipred_z2_rust<BD: BitDepth>(
     let topleft = 64;
     if upsample_above {
         upsample_edge::<BD>(
-            edge[topleft..].as_mut_ptr(),
+            &mut edge[topleft..],
             width + 1,
             topleft_in,
             0 as c_int,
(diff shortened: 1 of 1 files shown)


src/ipred.rs
```diff
@@ -739,38 +739,38 @@ fn get_filter_strength(wh: c_int, angle: c_int, is_sm: bool) -> c_int {
 }
 
 #[inline(never)]
-unsafe fn filter_edge<BD: BitDepth>(
-    out: *mut BD::Pixel,
+fn filter_edge<BD: BitDepth>(
+    out: &mut [BD::Pixel],
+    out_off: usize,
```
kkysen, line 744: If we're not indexing backwards (we're always doing `out_off + _ as usize`, unless that's a mistake), why can't we just slice `out` and not have a separate `out_off` arg?

src/ipred.rs
```diff
@@ -781,37 +781,36 @@ fn get_upsample(wh: c_int, angle: c_int, is_sm: bool) -> bool {
...
     to: c_int,
     bd: BD,
 ) {
     static kernel: [i8; 4] = [-1, 9, 9, -1];
-    let mut i;
-    i = 0 as c_int;
-    while i < hsz - 1 {
-        *out.offset((i * 2) as isize) = *in_0.offset(iclip(i, from, to - 1) as isize);
+    for i in 0..hsz - 1 {
+        out[(i * 2) as usize] = r#in[in_off + iclip(i, from, to - 1) as usize];
```
kkysen, line 795: If `from` and `to - 1` are both `usize`s, can we cast to `usize` for them separately, outside of the loop?

src/ipred.rs
```diff
@@ -930,10 +933,18 @@ unsafe fn ipred_z2_rust<BD: BitDepth>(
     } else {
         false
     };
-    let mut edge: [BD::Pixel; 129] = [0.into(); 129];
-    let topleft: *mut BD::Pixel = &mut *edge.as_mut_ptr().offset(64) as *mut BD::Pixel;
+    let mut edge = [0.into(); 129];
```
kkysen, line 936:
```suggestion
    let mut edge = [0.into(); 64 + 64 + 1];
```

src/ipred.rs
```diff
@@ -979,33 +993,28 @@ unsafe fn ipred_z2_rust<BD: BitDepth>(
         };
         if filter_strength_0 != 0 {
             filter_edge::<BD>(
-                &mut *topleft.offset(-height as isize),
+                &mut edge,
+                topleft - height as usize,
```
kkysen, line 997: If `height` is unsigned, can we cast to `usize` at the beginning of the `fn`?  Same for `width` (and get rid of the `width.try_into().unwrap()`).

src/ipred.rs
```diff
@@ -979,33 +993,28 @@ unsafe fn ipred_z2_rust<BD: BitDepth>(
...
-                    topleft.offset(-height as isize),
-                    height.try_into().unwrap(),
-                ),
-                slice::from_raw_parts(
-                    topleft_in.offset(-height as isize),
-                    height.try_into().unwrap(),
-                ),
+                &mut edge[topleft - height as usize..][..height as usize],
+                &topleft_in[topleft_in_off - height as usize..][..height as usize],
                 height.try_into().unwrap(),
```
kkysen, line 1011:
```suggestion
                height as usize,
```

src/ipred.rs
```diff
@@ -979,33 +993,28 @@ unsafe fn ipred_z2_rust<BD: BitDepth>(
...
             );
         }
     }
-    *topleft = *topleft_in;
+    edge[topleft] = topleft_in[topleft_in_off];
     let base_inc_x = 1 + upsample_above as c_int;
-    let left: *const BD::Pixel =
-        &mut *topleft.offset(-(1 + upsample_left as isize)) as *mut BD::Pixel;
+    let left = edge[topleft - (1 + upsample_left as usize)..].as_mut_ptr();
```
kkysen, line 1017: I would try to make `left` safe as part of this PR, as it could reveal borrowck issues that influence the design of the rest of this PR.

src/ipred.rs
randomPoison: Ah you're right we can do that. I initially thought we were doing negative indexing on `out`, but I later realized that wasn't the case and didn't think to remove `out_off`. Changed.

src/ipred.rs
randomPoison: `from` can't be a `usize` because it's -1 for one of the calls, and so it doesn't make sense to cast `to` to a `usize` because we need `from` and `to` to be the same type for `clip`.

src/ipred.rs
randomPoison: This PR isn't focused on cleaning up the `ipred_z*_rust` functions, I'm going to do that in a separate PR.

src/ipred.rs
randomPoison: `width` and `height` are both unsigned I think, but they're used heavily as `c_int`, so casting them to `usize` at the start of the function is going to require that we cast them back to `c_int` in a bunch of places, which I think is just going to make the code noisier.

src/ipred.rs
randomPoison: Will do in a followup PR.

randomPoison:
@kkysen I don't think it makes sense to shift around the casts. I know some of these values could be `usize`s, but I don't think changing them will make the code any cleaner since we'd end up casting to `c_int` about as many times. Can I merge this as is? The main focus was to get `filter_edge` and `upsample_edge` safe, and I having changes building on top of this that I don't want to have a rebase a bunch of times.

kkysen (approved):

PR #1142, opened 2024-05-29
CI: Test with negative strides

thedataking:
Add option (`-n`) to test with negative strides similar to dav1d's test configuration. See issue #636 for details.

We test negative strides with the native and optimized x86 and arm64 builds as they take the least time to complete.

### DIFF
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -8,8 +8,9 @@ rust_test_path=
 seek_stress_test_rust_path=
 debug_opt=
 frame_delay=
+negative_stride=
 wrapper=
-while getopts t:r:d:s:f:w: flag
+while getopts t:r:ds:f:nw: flag
 do
     case "${flag}" in
         t) timeout_multiplier=${OPTARG};;
@@ -17,6 +18,7 @@ do
         s) seek_stress_test_rust_path="-Dseek_stress_test_rust_path=${OPTARG}";;
         d) debug_opt="-Ddebug=true";;
         f) frame_delay=${OPTARG};;
+        n) negative_stride=1;;
         w) wrapper=${OPTARG};;
     esac
 done
@@ -63,10 +65,17 @@ else
     test_args+=(--suite testdata_seek-stress)
 fi
 
-if [[ -n $frame_delay ]]; then
+if [[ -n $frame_delay && -n $negative_stride ]]; then
+    # Frame delay is tested with two threads; negative strides with one.
+    echo "Error: frame_delay and negative_stride options can't be used together. Exiting."
+    exit 1
+elif [[ -n $frame_delay ]]; then
     # These test args override the args from test-data, resulting in 2 threads
     # and a frame delay
     test_args+=(--test-args "--threads 2 --framedelay $frame_delay")
+elif [[ -n $negative_stride ]]; then
+    # Run all tests with negative strides and multi-threading disabled
+    test_args+=(--test-args "--threads 1 --negstride")
 fi
 
 if [[ -n $wrapper ]]; then
--- a/src/filmgrain.rs
+++ b/src/filmgrain.rs
@@ -689,13 +689,11 @@ fn fgy_32x32xn_rust<BD: BitDepth>(
         static W: [[c_int; 2]; 2] = [[27, 17], [17, 27]];
 
         let src_row_y = |y| {
-            let row = (src.pixel_offset::<BD>() + src_row_offset)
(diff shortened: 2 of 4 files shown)


.github/workflows/test.sh
```diff
@@ -63,10 +65,17 @@ else
     test_args+=(--suite testdata_seek-stress)
 fi
 
-if [[ -n $frame_delay ]]; then
+if [[ -n $frame_delay && -n $negative_stride ]]; then
+    # Frame delay is tested with two threads; negative strides with one.
+    echo "Error: frame_delay and negative_stride options can't be used together. Exiting."
```
kkysen, line 70: Can frame delays not be used with negative strides?  Or we're just testing it this one way?

src/filmgrain.rs
```diff
@@ -835,18 +833,16 @@ fn fguv_32x32xn_rust<BD: BitDepth>(
         static W: [[[c_int; 2]; 2 /* off */]; 2 /* sub */] = [[[27, 17], [17, 27]], [[23, 22], [0; 2]]];
 
         let luma_row_uv = |y| {
-            let row = (luma.pixel_offset::<BD>() + luma_row_offset)
+            let row = (luma_row_offset)
```
kkysen, line 836:
```suggestion
            let row = luma_row_offset
```

src/filmgrain.rs
```diff
@@ -835,18 +833,16 @@ fn fguv_32x32xn_rust<BD: BitDepth>(
...
                 .wrapping_add_signed((y << sy) as isize * luma.pixel_stride::<BD>());
             luma.slice::<BD, _>((row + (bx << sx).., ..bw << sx))
         };
         let src_row_uv = |y| {
-            let row = (src.pixel_offset::<BD>() + src_row_offset)
-                .wrapping_add_signed(y as isize * src.pixel_stride::<BD>());
+            let row = (src_row_offset).wrapping_add_signed(y as isize * src.pixel_stride::<BD>());
```
kkysen, line 841:
```suggestion
            let row = src_row_offset.wrapping_add_signed(y as isize * src.pixel_stride::<BD>());
```

src/filmgrain.rs
```diff
@@ -835,18 +833,16 @@ fn fguv_32x32xn_rust<BD: BitDepth>(
...
             src.slice::<BD, _>((row + bx.., ..bw))
         };
         let dst_row_uv = |y| {
-            let row = (dst.pixel_offset::<BD>() + dst_row_offset)
-                .wrapping_add_signed(y as isize * dst.pixel_stride::<BD>());
+            let row = (dst_row_offset).wrapping_add_signed(y as isize * dst.pixel_stride::<BD>());
```
kkysen, line 845:
```suggestion
            let row = dst_row_offset.wrapping_add_signed(y as isize * dst.pixel_stride::<BD>());
```

src/filmgrain.rs
```diff
@@ -708,7 +706,7 @@ fn fgy_32x32xn_rust<BD: BitDepth>(
         };
 
         for y in ystart..bh {
-            let src = &*src_row_y(y);
+            let src: &[<BD as BitDepth>::Pixel] = &*src_row_y(y);
```
kkysen, line 709: Why do we need this?

kkysen (commented):
Oh whoops about the `fg{y,uv}_32x32xn_rust` bug; I missed that.

For the tests, from your PR description, it sounds like we're not running it on disabled asm builds.  I'm not sure, though.  If we're not, we should.

src/filmgrain.rs
thedataking: We don't. I think I accidentally double clicked the type annotation in vscode which inserts the inferred type... sry 😬

.github/workflows/test.sh
thedataking: I don't know enough about how these features interact so so the safest and simplest solution I could come up with was to support the test flows that dav1d has and make sure they don't interfere negatively. We can always come back and do something more sophisticated when there is a clear need.

thedataking:
> For the tests, from your PR description, it sounds like we're not running it on disabled asm builds. I'm not sure, though. If we're not, we should.

We are for the x86 variants. I added a `release-no-asm` variant for macOS/arm64.

kkysen:
> We are for the x86 variants. I added a `release-no-asm` variant for macOS/arm64.

Okay, that sounds good.

kkysen (approved):

PR #1153, opened 2024-05-31
`fn Rav1dRefmvsDSPContext::splat_mv`: Optimize

### LINKED ISSUE #1152: `fn Rav1dRefmvsDSPContext::splat_mv` is slow
When processing the "chimera 8-bit" bitstream, the decoder spends about 3% of its time in `fn Rav1dRefmvsDSPContext::splat_mv`. There is no equivalent C function. Also, this is much more time than what is spent in the `dav1d_splat_mv_*` assembly functions that do the actual data processing.

kkysen:
* Fixes #1152.

35dd0e9 avoids calculations on `r`s that are not actually passed to the `fn` ptr.  We still create the array of 37 elements, both for the guards and for the raw ptrs, as we can't use `alloca`.  This requires extra zero initialization, but otherwise avoids unneeded calculations.

752ec1d then optimizes this further by making the arrays `MaybeUninit`s.  So now the stack allocation is just a subtraction and no initialization is necessary.  This also means we must call destructors manually, but we can use `mem::needs_drop` to skip this when the guard has no `impl Drop` in release mode (i.e. `!cfg(debug_assertions)`).

In 90a6eb8, I also realized we don't need the extra `rr_len: usize` arg, as `bh4` is the same.

I benchmarked this with `cargo build --release && hyperfine './target/release/dav1d -i ./tests/large/chimera_8b_1080p.ivf -o /dev/null'`, and it looks like we do get a large amount of performance back.  @fbossen, let me know if you can reproduce this and if this eliminates the overhead you were seeing.

| Commit  | Time (s) | Speedup |
| ------- | -------- | ------- |
| 97b3184 | 18.566   | 0.00 %  |
| 35dd0e9 | 17.579   | 5.61 %  |
| 752ec1d | 17.185   | 8.04 %  |

### DIFF
--- a/src/refmvs.rs
+++ b/src/refmvs.rs
@@ -8,6 +8,7 @@ use crate::src::align::AlignedVec64;
 use crate::src::cpu::CpuFlags;
 use crate::src::disjoint_mut::DisjointMut;
 use crate::src::disjoint_mut::DisjointMutArcSlice;
+use crate::src::disjoint_mut::DisjointMutGuard;
 use crate::src::env::fix_mv_precision;
 use crate::src::env::get_gmv_2d;
 use crate::src::env::get_poc_diff;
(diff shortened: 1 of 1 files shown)


fbossen (approved):
This resolves the performance issue I was observing

PR #1169, opened 2024-06-04
`fn ipred_z2_rust`: Cleanup and make mostly safe

randomPoison:
* Part of #846.
* Part of #815.

### DIFF
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -944,9 +944,7 @@ unsafe fn ipred_z2_rust<BD: BitDepth>(
     let is_sm = (angle >> 9) & 1 != 0;
     let enable_intra_edge_filter = angle >> 10;
     angle &= 511 as c_int;
-    if !(angle > 90 && angle < 180) {
-        unreachable!();
-    }
+    assert!(angle > 90 && angle < 180);
     let mut dy = dav1d_dr_intra_derivative[(angle - 90 >> 1) as usize] as c_int;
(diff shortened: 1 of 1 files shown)


src/ipred.rs
```diff
@@ -1024,50 +1023,47 @@ unsafe fn ipred_z2_rust<BD: BitDepth>(
...
+        let dst_slice = slice::from_raw_parts_mut(
+            dst.offset(BD::pxstride(stride) * y as isize),
+            width as usize,
+        );
+        for (x, dst) in dst_slice.iter_mut().enumerate() {
+            let ypos = (y << 6 + upsample_left as c_int) - (dy * (x + 1) as c_int);
+            let base_x = base_x + (base_inc_x * x) as c_int;
+            let v = if base_x >= 0 {
+                edge[topleft + base_x as usize].as_::<c_int>() * (64 - frac_x)
+                    + edge[topleft + base_x as usize + 1].as_::<c_int>() * frac_x
```
kkysen, line 1057: Can you pre-slice `&edge[topleft..]`?

kkysen (approved):

src/ipred.rs
randomPoison: I don't think that makes this any cleaner, I think I prefer the consistency of using `base + offset` since we have to do the same thing in the other branch with `left`.

kkysen: It is a bit more efficient this way, though.

PR #1175, opened 2024-06-05
`fn ipred_filter_rust`: Cleanup and make safer

randomPoison:
This cleans up `ipred_filter_rust` and makes the helper function `filter_fn` safe. I left most of the pointers within the function as pointers because they will point into `dst` at some point (e.g. it's initialized to `topleft_in`, but in a later loop iteration is changed to point to `dst`), so they couldn't be trivially made safe. It may be possible to make this fully safe once `dst` has been made into a `Rav1dPictureDataComponent` ref.

### DIFF
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -1187,8 +1187,8 @@ unsafe extern "C" fn ipred_z_c_erased<BD: BitDepth, const Z: usize>(
     )
 }
 
-unsafe fn filter_fn(
-    flt_ptr: *const i8,
+fn filter_fn(
+    flt_ptr: &[i8],
     p0: c_int,
     p1: c_int,
     p2: c_int,
@@ -1198,21 +1198,21 @@ unsafe fn filter_fn(
     p6: c_int,
 ) -> c_int {
     if cfg!(any(target_arch = "x86", target_arch = "x86_64")) {
-        *flt_ptr.offset(0) as c_int * p0
-            + *flt_ptr.offset(1) as c_int * p1
-            + *flt_ptr.offset(16) as c_int * p2
-            + *flt_ptr.offset(17) as c_int * p3
-            + *flt_ptr.offset(32) as c_int * p4
-            + *flt_ptr.offset(33) as c_int * p5
-            + *flt_ptr.offset(48) as c_int * p6
+        flt_ptr[0] as c_int * p0
+            + flt_ptr[1] as c_int * p1
+            + flt_ptr[16] as c_int * p2
+            + flt_ptr[17] as c_int * p3
+            + flt_ptr[32] as c_int * p4
+            + flt_ptr[33] as c_int * p5
+            + flt_ptr[48] as c_int * p6
     } else {
-        *flt_ptr.offset(0) as c_int * p0
-            + *flt_ptr.offset(8) as c_int * p1
-            + *flt_ptr.offset(16) as c_int * p2
-            + *flt_ptr.offset(24) as c_int * p3
-            + *flt_ptr.offset(32) as c_int * p4
-            + *flt_ptr.offset(40) as c_int * p5
-            + *flt_ptr.offset(48) as c_int * p6
+        flt_ptr[0] as c_int * p0
+            + flt_ptr[8] as c_int * p1
+            + flt_ptr[16] as c_int * p2
+            + flt_ptr[24] as c_int * p3
+            + flt_ptr[32] as c_int * p4
+            + flt_ptr[40] as c_int * p5
+            + flt_ptr[48] as c_int * p6
     }
 }
 
@@ -1262,7 +1262,7 @@ unsafe fn ipred_filter_rust<BD: BitDepth>(
             while yy < 2 {
                 let mut xx = 0;
                 while xx < 4 {
-                    let acc = filter_fn(flt_ptr.as_ptr(), p0, p1, p2, p3, p4, p5, p6);
+                    let acc = filter_fn(flt_ptr, p0, p1, p2, p3, p4, p5, p6);
                     *ptr.offset(xx as isize) = bd.iclip_pixel(acc + 8 >> 4);
                     xx += 1;
                     flt_ptr = &flt_ptr[FLT_INCR..];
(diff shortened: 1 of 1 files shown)


src/ipred.rs
```diff
@@ -1198,87 +1198,82 @@ unsafe fn filter_fn(
...
-            + *flt_ptr.offset(32) as c_int * p4
-            + *flt_ptr.offset(33) as c_int * p5
-            + *flt_ptr.offset(48) as c_int * p6
+        flt_ptr[0] as c_int * p0
+            + flt_ptr[1] as c_int * p1
+            + flt_ptr[16] as c_int * p2
+            + flt_ptr[17] as c_int * p3
+            + flt_ptr[32] as c_int * p4
+            + flt_ptr[33] as c_int * p5
+            + flt_ptr[48] as c_int * p6
```
CrazyboyQCD, line 1207: In this branch, the length of slice in the final loop is 48, which cannot be indexed at 48, is this a bug with dav1d?
https://godbolt.org/z/T4GhhnxfE

randomPoison: I don't think this is an issue or we'd be seeing panics in the tests now that we're using safe slice indexing. In your example you're doing `flt_ptr = &flt_ptr[FLT_INC..]` before the call to `filter_fn`, which is incorrect. After the final loop iteration the length will be 48, but we don't pass `flt_ptr` to `filter_fn` after that point. See [this playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9e57720ec14d120877fc5569acbb55c2) for a demonstration of what happens. If you move the `dbg!` call to after doing `&flt_ptr[flt_incr..]` then it panics.

CrazyboyQCD: > I don't think this is an issue or we'd be seeing panics in the tests now that we're using safe slice indexing. In your example you're doing `flt_ptr = &flt_ptr[FLT_INC..]` before the call to `filter_fn`, which is incorrect. After the final loop iteration the length will be 48, but we don't pass `flt_ptr` to `filter_fn` after that point. See [this playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9e57720ec14d120877fc5569acbb55c2) for a demonstration of what happens. If you move the `dbg!` call to after doing `&flt_ptr[flt_incr..]` then it panics.

Thanks for reply, it's my fault, I did not notice that it calls before slicing, which is fine.

src/ipred.rs
```diff
@@ -1198,87 +1198,82 @@ unsafe fn filter_fn(
     p6: c_int,
 ) -> c_int {
     if cfg!(any(target_arch = "x86", target_arch = "x86_64")) {
```
kkysen, line 1200: To elide a bunch of bounds checks.

```suggestion
    let flt_ptr = &flt_ptr[..48 + 1];
    if cfg!(any(target_arch = "x86", target_arch = "x86_64")) {
```

src/ipred.rs
```diff
@@ -1198,87 +1198,82 @@ unsafe fn filter_fn(
...
 
 cfg_if! {
     if #[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {
-        const FLT_INCR: isize = 2;
+        const FLT_INCR: usize = 2;
     } else {
-        const FLT_INCR: isize = 1;
+        const FLT_INCR: usize = 1;
     }
 }
```
kkysen, line 1225:
```suggestion
const FLT_INCR: usize = if cfg!(any(target_arch = "x86", target_arch = "x86_64")) {
    2
} else {
    1
};
```

src/ipred.rs
```diff
@@ -1198,87 +1198,82 @@ unsafe fn filter_fn(
...
+            let mut ptr = dst.add(x);
+            let mut flt_ptr = filter.as_slice();
+
+            for _yy in 0..2 {
+                for xx in 0..4 {
                     let acc = filter_fn(flt_ptr, p0, p1, p2, p3, p4, p5, p6);
-                    *ptr.offset(xx as isize) = bd.iclip_pixel(acc + 8 >> 4);
-                    xx += 1;
-                    flt_ptr = flt_ptr.offset(FLT_INCR);
+                    *ptr.add(xx) = bd.iclip_pixel(acc + 8 >> 4);
```
kkysen, line 1265: Can you make `ptr` a slice for this inner `xx` loop?

src/ipred.rs
```diff
@@ -1198,87 +1198,82 @@ unsafe fn filter_fn(
...
-        let mut x = 0;
-        while x < width {
+    let width = width as usize;
+    let height = height as usize;
+    let stride = BD::pxstride(stride);
+    let filt_idx = filt_idx & 511;
+    assert!(filt_idx < 5);
+
+    let filter = &dav1d_filter_intra_taps[filt_idx as usize];
+    let mut top = topleft_in.as_ptr().add(topleft_off + 1);
```
kkysen, line 1246:
```suggestion
    let mut top = topleft_in[topleft_off + 1..].as_ptr();
```

src/ipred.rs
```diff
@@ -1198,87 +1198,82 @@ unsafe fn filter_fn(
...
+    for y in (0..height).step_by(2) {
+        let mut topleft = topleft_in.as_ptr().add(topleft_off - y);
```
kkysen, line 1248:
```suggestion
        let top = slice::from_raw_parts_mut(top, 4);
        let mut topleft = topleft_in.as_ptr().add(topleft_off - y);
```

src/ipred.rs
```diff
@@ -1198,87 +1198,82 @@ unsafe fn filter_fn(
...
-            while yy < 2 {
-                let mut xx = 0;
-                while xx < 4 {
+            let p5 = (*left.offset(0 * left_stride)).as_::<c_int>();
+            let p6 = (*left.offset(1 * left_stride)).as_::<c_int>();
+            let mut ptr = dst.add(x);
+            let mut flt_ptr = filter.as_slice();
+
+            for _yy in 0..2 {
+                for xx in 0..4 {
```
kkysen, line 1263:
```suggestion
                let ptr_slice = slice::from_raw_parts_mut(ptr, 4);
                for ptr in ptr_slice {
```

kkysen (changes requested):

src/ipred.rs
randomPoison: I had to create the slice from within the `for x` loop because `top` is updated at the end of the loop (not just the `for y` loop), but otherwise this works 👍

kkysen (approved):

PR #1176, opened 2024-06-05
`fn rav1d_open::error`: Remove

kkysen:
The null check is removed since the allocation is now infallible.

The "default allocator now having a cookie" check is moved up to use `validate_input!` instead.
I also changed it to return `EINVAL` instead of `ENOMEM` (different from C), since it's an argument being checked, and there's no out-of-memory error.

Since both are before the `Rav1dContext` exists or is (non-default) initialized,
we don't need to call `close_internal` here, which frees up some other API designs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,7 +25,6 @@ use crate::src::error::Dav1dResult;
 use crate::src::error::Rav1dError::EGeneric;
 use crate::src::error::Rav1dError::EAGAIN;
 use crate::src::error::Rav1dError::EINVAL;
-use crate::src::error::Rav1dError::ENOMEM;
 use crate::src::error::Rav1dResult;
 use crate::src::extensions::OptionError as _;
 use crate::src::fg_apply;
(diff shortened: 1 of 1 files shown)


src/cdef.rs
```diff
@@ -197,7 +197,7 @@ unsafe fn padding<BD: BitDepth>(
     for y in 0..h {
         let tmp = &mut tmp[(y + 2) * TMP_STRIDE..];
         let src = src + (y as isize * stride);
-        let src = &*src.data.slice::<BD, _>((src.offset.., ..x_end));
+        let src = &*src.data.slice::<BD, _>((src.offset.., ..x_end - 2));
```
randomPoison, line 200: Is this change meant to be part of this PR? It seems unrelated to the rest of the changes.

kkysen: Whoops, no, I forgot to change the base branch.  That's #1178.

randomPoison (approved):

PR #1181, opened 2024-06-06
`mod refmvs`: Backport fix for buffer overread from dav1d 1.4.2

fbossen:
Fix buffer overread in `save_tmvs()` asm
    
The `refmvs_block` struct is only 12 bytes large but it is accessed using 16-byte unaligned loads in asm.
    
In order to avoid reading past the end of the allocated buffer, it is made slightly larger.

### DIFF
--- a/src/refmvs.rs
+++ b/src/refmvs.rs
@@ -192,6 +192,8 @@ pub(crate) struct refmvs_frame<'a> {
     pub n_frame_threads: i32,
 }
 
+const R_PAD: usize = 1; // number of elements added to RefMvsFrame.r to avoid overread
+
 #[derive(Default)]
 pub struct RefMvsFrame {
     pub iw4: i32,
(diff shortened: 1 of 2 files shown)


randomPoison (approved):

src/refmvs.rs
```diff
@@ -428,7 +430,7 @@ impl Rav1dRefmvsDSPContext {
         // For the Rust fallback fn, the extra args `&rf.r` and `ri`
         // are passed to allow for disjointedness checking.
         let rr = &ri.map(|ri| {
-            if ri > rf.r.len() {
+            if ri > rf.r.len() - R_PAD {
                 return ptr::null();
             }
             // SAFETY: `.add` is in-bounds; checked above.
```
kkysen, line 436:
```suggestion
            // SAFETY: `.add` is in-bounds; checked above.
            // Also note that asm may read 12-byte `refmvs_block`s in 16-byte chunks.
            // This is safe because we allocate `rf.r` with an extra `R_PAD` (1) elements.
            // These ptrs are only read, so these overlapping reads are safe
            // (only read is only checked in the fallback Rust `fn`).
            // Furthermore, this is provenance safe because
            // we derive the ptrs from `rf.r.as_mut_ptr()`,
            // as opposed to materializing intermediate references.
            const _: () = assert!(mem::size_of::<refmvs_block>() * (1 + R_PAD) > 16);
```

kkysen (changes requested):
Could you add this more thorough safety comment?

kkysen (approved):

PR #1182, opened 2024-06-06
Fixes to build on Android

thedataking:


tools/compat/errno.rs
```diff
@@ -10,5 +10,9 @@ cfg_if::cfg_if! {
         pub unsafe fn errno_location() -> *mut c_int {
             libc::__error()
         }
+    } else if #[cfg(target_os = "android")] {
```
kkysen, line 13: Can you just fold this into the `macos` definition above?

kkysen (commented):
Is this still a draft?  Are we going to add CI for Android?  How do we do that?  Also, I think the CI failure is from a breakage I caused (fixed now in #1178.

tools/compat/errno.rs
thedataking: `macos` calls `__error`
`anroid` calls `__errno`

thedataking:
> Are we going to add CI for Android? How do we do that?

I've added a build for Android now. Stephen helped me resolve a linker issue; other than that, it wasn't too difficult since all the NDK tools were already installed. 

There's an android emulator we can use to test the output binaries. I haven't looked into the details of this yet. We can either merge what we have on this PR to detect build breakage or I can add to it until we also detect functional breakage.

tools/compat/errno.rs
kkysen: Oh, oops, missed that.

.github/workflows/build-and-test-aarch64-android.yml
```diff
@@ -0,0 +1,38 @@
...
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: aarch64-android-cargo-and-target-${{ hashFiles('**/Cargo.lock') }}
+      - name: cargo build for aarch64-apple-darwin
+        run:  |
+            rustup target add aarch64-linux-android
```
kkysen, line 32: Could we just add this to `rust-toolchain.toml`?

kkysen:
> > Are we going to add CI for Android? How do we do that?
> 
> I've added a build for Android now. Stephen helped me resolve a linker issue; other than that, it wasn't too difficult since all the NDK tools were already installed.
> 
> There's an android emulator we can use to test the output binaries. I haven't looked into the details of this yet. We can either merge what we have on this PR to detect build breakage or I can add to it until we also detect functional breakage.

I think build only CI is good enough for this PR.  We can add testing later.

.github/workflows/build-and-test-aarch64-android.yml
thedataking: > Could we just add this to rust-toolchain.toml?

I think that would make us download this target everywhere. I checked the Android build and it downloads all the targets in `rust-toolchain.toml` which is just slowing us down. We're also running `rustup target add $TARGET` for the QEMU builds so what we're doing here is consistent with the existing workflows.

From the Android CI run with the current `rust-toolchain.toml`:

```
info: syncing channel updates for 'nightly-2024-06-13-x86_64-unknown-linux-gnu'
info: latest update on 2024-06-13, rust version 1.81.0-nightly (8337ba918 2024-06-12)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std' for 'aarch64-unknown-linux-gnu'
info: downloading component 'rust-std' for 'armv7-unknown-linux-gnueabihf'
info: downloading component 'rust-std' for 'i686-unknown-linux-gnu'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
```

What do you think about taking the `targets` out of `rust-toolchain.toml` so we avoid wasting CI bandwidth?

kkysen: Hmm, I guess maybe.  They are useful locally.  I thought they were being cached altogether in CI so that it wouldn't be a big deal.  I didn't realize they were being redownloaded every time.  Could we just cache them?

thedataking: > Could we just cache them?

It is possible but not ideal. We want the caches to contain what we need for reach run, not what any run could need.

Over time, the caches would grow as the host images update their toolchains, as we do the same. The size of each toolchain update would be multiplied by the number of `targets` in `rust-toolchain.toml`.

> Hmm, I guess maybe. They are useful locally. 

I chatted with @rinon and his take was that if you're cross compiling, you should be able to add the toolchains (Rust ones plus the host tools for compiling, assembling, and linking the non-Rust bits). I think that's a fair point. 

What do you think about having a separate PR to document cross-compiling? We could suggest how to add the `targets` to `rust-toolchain.toml` along with the other steps required.

thedataking: decided during offline discussion to comment out `targets`. update cross-compiling instructions in README.

rust-toolchain.toml
```diff
@@ -1,9 +1,16 @@
...
-]
+
+# Commented out targets to avoid downloading or caching all these during CI
+# runs. See discussion on rav1d PR #1174. Uncomment to cross-compile locally.
+# targets = [
+#     "aarch64-unknown-linux-gnu",
+#     "aarch64-apple-darwin",
+#     "aarch64-linux-android",
+#     "armv7-unknown-linux-gnueabihf",
+#     "armv7-unknown-linux-gnueabihf",
```
kkysen, line 11: Duplicate

bench_results.py
kkysen, line 1: What's this?

thedataking: derp, removed, thanks for the catch!

rust-toolchain.toml
```diff
@@ -1,9 +1,15 @@
...
+
+# Commented out targets to avoid downloading or caching all these during CI
+# runs. See discussion on rav1d PR #1174. Uncomment to cross-compile locally.
+# targets = [
+#     "aarch64-unknown-linux-gnu",
+#     "aarch64-apple-darwin",
+#     "aarch64-linux-android",
+#     "armv7-unknown-linux-gnueabihf",
+#     "i686-unknown-linux-gnu",
+#     "riscv64gc-unknown-linux-gnu",
```
kkysen, line 12: Do we support riscv32, too?

kkysen (approved):

rust-toolchain.toml
thedataking: I haven't tested it. Took a quick look, short answer is no, dav1d doesn't seem to support riscv32 either. Notes:

The `riscv64gc-unknown-linux-gnu` target is tier 2 whereas `riscv32gc-unknown-linux-gnu` is tier 3 and the nightly we're on doesn't have a build for that target. I also wasn't able to find ubuntu packages with a suitable C compiler and linker for riscv32.

Issue #817, opened 2024-03-15
Cleanup c2rust output in `itx.rs`


randomPoison:
Note that this should wait until after #683 is merged in order to not duplicate work between the 8bpc and 16bpc versions.

* [ ] Cleanup helper functions to take safe arguments where possible (may take raw pointers as arguments currently).
* [ ] General cleanup of c2rust generated code in functions.
* [ ] Restore comments from original C.

The following functions still need to be cleaned:

- [x] `inv_txfm_add_wht_wht_4x4_rust`

kkysen:
> I'm working on deduplicating and cleaning up this module in PR #683 so I can do this one unless someone else wants to.

@thedataking, have you actually started this?  I can do this.  I started doing `itx_1d` but they overlap a lot

randomPoison:
Looks like this was finished in #1042, and the module has been made fully safe as of #1143, so I'm going to close this.

PR #1183, opened 2024-06-07
`itx.rs`: Replace C types with Rust equivalents

randomPoison:
* Last bit of trivial cleanup for #817.


kkysen (approved):
LGTM but I'm not sure if it's too important to do this.  That said, if we are, it makes sense to replace `libc::ptrdiff_t` with `isize`, too, I think.

kkysen (approved):

PR #1184, opened 2024-06-08
`fn ipred_z*_neon`: Cleanup and make safer

randomPoison:
* Part of #846.
* Part of #815.

### DIFF
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -1594,7 +1594,7 @@ mod neon {
     }
 
     unsafe fn rav1d_ipred_pixel_set_neon<BD: BitDepth>(
-        out: *mut BD::Pixel,
+        out: &mut [BD::Pixel],
         px: BD::Pixel,
         n: c_int,
     ) {
@@ -1618,7 +1618,7 @@ mod neon {
             );
         }
 
-        let out = out.cast();
+        let out = out.as_mut_ptr().cast();
         match BD::BPC {
             BPC::BPC8 => dav1d_ipred_pixel_set_8bpc_neon(
                 out,
@@ -1688,9 +1688,10 @@ mod neon {
         }
         let base_inc = 1 + upsample_above as c_int;
         let pad_pixels = width + 15;
+        let px = top_out[max_base_x as usize];
         rav1d_ipred_pixel_set_neon::<BD>(
-            top_out.as_mut_ptr().offset((max_base_x + 1) as isize),
-            top_out[max_base_x as usize],
+            &mut top_out[max_base_x as usize + 1..],
+            px,
             pad_pixels * base_inc,
         );
         if upsample_above {
@@ -1955,9 +1956,10 @@ mod neon {
         }
         let base_inc = 1 + upsample_left as c_int;
         let pad_pixels = cmp::max(64 - max_base_y - 1, height + 15);
+        let px = left_out[max_base_y as usize];
         rav1d_ipred_pixel_set_neon::<BD>(
-            left_out.as_mut_ptr().offset((max_base_y + 1) as isize),
-            left_out[max_base_y as usize],
+            &mut left_out[max_base_y as usize + 1..],
+            px,
             pad_pixels * base_inc,
         );
         if upsample_left {
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -1893,7 +1893,7 @@ mod neon {
         _max_height: c_int,
         bd: BD,
     ) {
-        let topleft_in = topleft_in.as_ptr().add(topleft_off);
+        let topleft_in = &topleft_in[topleft_off..];
         let is_sm = (angle >> 9) & 1 != 0;
         let enable_intra_edge_filter = angle >> 10;
         angle &= 511;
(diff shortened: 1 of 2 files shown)


src/ipred.rs
```diff
@@ -1941,13 +1928,13 @@ mod neon {
...
                 get_filter_strength(width + height, angle - 180, is_sm)
             } else {
-                0 as c_int
+                0
             };
             if filter_strength != 0 {
-                flipped[0] = *topleft_in.offset(0);
+                flipped[0] = topleft_in[0];
                 bd_fn!(reverse::decl_fn, BD, ipred_reverse, neon).call::<BD>(
                     flipped.as_mut_ptr().offset(1),
```
kkysen, line 1936:
```suggestion
                    flipped[1..].as_mut_ptr(),
```

src/ipred.rs
```diff
@@ -2007,12 +1995,15 @@ mod neon {
         max_width: c_int,
         max_height: c_int,
         bitdepth_max: c_int,
-        _topleft_off: usize,
+        topleft_off: usize,
     ) {
+        // SAFETY: Reconstructed from args passed by `angular_ipred::Fn::call`.
```
kkysen, line 2000: Can you also add the `# Safety` comment on `fn ipred_z_neon_erased` saying it must be called by ``[`angular_ipred::Fn::call`]``?

kkysen (changes requested):
Can you also move the `.as_ptr()`/`.as_mut_ptr()`-type calls to the `fn call`s instead of repeating them at the call sites?

src/ipred.rs
```diff
@@ -1985,6 +1984,7 @@ mod neon {
         };
     }
 
+    /// SAFETY: Must be called from [`angular_ipred::Fn::call`].
```
kkysen, line 1987:
```suggestion
    /// # Safety
    ///
    /// Must be called from [`angular_ipred::Fn::call`].
```

`# Safety` is used for docs, vs. `// SAFETY` for comments for `unsafe {}` blocks.

kkysen (changes requested):
Other new changes look good, but you can do this, too? https://github.com/memorysafety/rav1d/pull/1184#pullrequestreview-2106061876

randomPoison:
> Can you also move the `.as_ptr()`/`.as_mut_ptr()`-type calls to the `fn call`s instead of repeating them at the call sites?

I'm not sure what you're asking for here. Are you referring to the places in the `ipred_z*_neon` fns where we pass a pointer to a buffer into the asm routines? So e.g. instead of:

```rust
let buf = [0; N];
some_asm_call(buf.as_ptr());
another_asm_call(buf.as_ptr());
```

we'd do:

```rust
let buf = [0; N];
let buf_ptr = buf.as_ptr();
some_asm_call(buf_ptr);
another_asm_call(buf_ptr);
```

Is that what you had in mind? That doesn't really seem any cleaner to me, it seems better to convert to a pointer at the call site. Am I misunderstanding what you're suggesting?

kkysen:
> > Can you also move the `.as_ptr()`/`.as_mut_ptr()`-type calls to the `fn call`s instead of repeating them at the call sites?
> 
> I'm not sure what you're asking for here. Are you referring to the places in the `ipred_z*_neon` fns where we pass a pointer to a buffer into the asm routines? So e.g. instead of:
> 
> ```rust
> let buf = [0; N];
> some_asm_call(buf.as_ptr());
> another_asm_call(buf.as_ptr());
> ```
> 
> we'd do:
> 
> ```rust
> let buf = [0; N];
> let buf_ptr = buf.as_ptr();
> some_asm_call(buf_ptr);
> another_asm_call(buf_ptr);
> ```
> 
> Is that what you had in mind? That doesn't really seem any cleaner to me, it seems better to convert to a pointer at the call site. Am I misunderstanding what you're suggesting?

No, I meant, we're doing things like `.call::<BD>(buf[..n].as_ptr())` repeatedly instead of doing `.call::<BD(&buf[..n])` and doing the `let buf = buf.as_ptr()` inside of the `fn call`.  The approach we've been doing for most of the `fn call`s wrapping `fn` ptrs.

randomPoison:
Ah, okay that makes sense. I've made that change.

src/ipred.rs
```diff
@@ -1820,10 +1820,11 @@ mod neon {
                     topleft_in.as_ptr(),
                     height,
                 );
+                let (front, back) = buf.split_at_mut(1 + left_offset);
```
kkysen, line 1810: Maybe `out` and `in_0` are better names, corresponding to the arg names.

src/ipred.rs
```diff
@@ -1786,10 +1786,11 @@ mod neon {
                 topleft_in,
                 height,
             );
+            let (front, back) = buf.split_at_mut(left_offset);
```
kkysen, line 1789: Same here.

kkysen (changes requested):
One little thing left: for the places where you did `let (front, back) = _.split_at_mut(...);`, can you name them `in`, `out` or `src`, `dst` instead, for clarity, and matching the arg names and var names used elsewhere?

randomPoison:
Done, I went with `src` and `dst`.

kkysen (approved):

PR #1185, opened 2024-06-10
`fn rav1d_open`: Make safe by sending an `Arc<Rav1dContext>` to worker threads

### LINKED ISSUE #705: Cleanup c2rust output in lib.rs
This file contains the boundary between the C and Rust APIs, so we may still need to make heavy use of unsafe operations in the cleaned up code.

* [x] #707 
* [ ] Cleanup helper functions to take safe arguments where possible (may take raw  pointers as arguments currently).

### LINKED ISSUE #862: Make struct initialization logic safe
Currently the logic for creating the various context structs either leaves them uninitialized or unsafely zero-initializes their fields. The struct fields are then initialized field-by-field. This logic was correct in C, but can lead to undefined behavior in rust if we zero-initialize a type like `Vec` or leave a field uninitialized before using the struct.

To solve this, we need to invert the initialization logic to only construct the final struct with valid values for all of the fields, i.e. rather than allocating the struct and then initializing the fields one-by-one, we create the value for the fields and then initialize the outer struct using Rust's normal struct initialization syntax. In some cases, we may just need a `Default` impl so that we can safely default-initialize the struct before performing the rest of the initialization logic.

In order to identify places this is happening, look for any of the following operations:

* Calls to `mem::zeroed` or `MaybeUninit::zeroed`.
* Calls to `malloc` or `rav1d_alloc_aligned`.

The following types/fields are being unsafely initialized:

- [x] `Rav1dContext` - Allocated and initialized in `rav1d_open`.
- [x] `Rav1dFrameData` - Allocated and initialized in `rav1d_open`.
- [x] `TaskThreadData_delayed_fg` - Being zero-initialized in `rav1d_open`.
- [x] `Rav1dFrameContext_lf` - Part of `Rav1dContext`, initialized in `rav1d_open`.
- [x] `Rav1dPicture` - Zero-initialized in `dav1d_get_picture` and `dav1d_apply_grain`.
- [x] `Rav1dTaskContext` - `new` method initializes fields with `mem::zeroed`. Called from `rav1d_open`.
- [ ] `Rav1dTileState` - Allocated in `rav1d_decode_frame_init`.
- [x] `BlockContext` - Allocated in `rav1d_decode_frame_init`.
- [ ] `Rav1dFrameHeader` - Being partially zero-initialized in `parse_frame_hdr`.
- [x] `refmvs_block` - Allocated in `rav1d_refmvs_init_frame`.
- [x] `refmvs_temporal_block` - Allocated in `rav1d_refmvs_init_frame`.

We have the following `*_init` functions that take a mutable pointer to some struct and initialize its fields. These will have to likewise be "inverted" where possible to directly return the initialized struct instead of taking a pointer to a struct to initialize. In cases where the the initialization logic makes this impossible, we need to at least make sure that the struct being initialized with default values when it's first created.

- [x] `rav1d_cdf_thread_init_static`
- [ ] `rav1d_decode_frame_init`
- [ ] `rav1d_decode_frame_init_cdf`
- [ ] `setup_tile` (called from `rav1d_decode_frame_init_cdf`)
- [x] `rav1d_refmvs_init`
- [x] `rav1d_refmvs_init_frame`
- [x] `rav1d_refmvs_tile_sbrow_init`

The following are functions for initializing function pointer structs. These are generally the leaf nodes of the initialization logic and should be straightforward to "invert". The DSP structs are currently being zero-initialized, which is UB for function pointers.

- [x] `rav1d_cdef_dsp_init`
- [x] `rav1d_intra_pred_dsp_init`
- [x] `rav1d_itx_dsp_init`
- [x] `rav1d_loop_filter_dsp_init`
- [x] `rav1d_loop_restoration_dsp_init`
- [x] `rav1d_mc_dsp_init`
- [x] `rav1d_msac_init`
- [x] `rav1d_refmvs_dsp_init`

kkysen:
* Fixes #862 (`Rav1dContext` initialization, which is the last one).
* Fixes #705.

This does a few things to make `fn rav1d_open`, the construction of `Rav1dContext`, and its usage from the main and worker threads, and its `unsafe` conversion from `Dav1dContext`s all safe or as safe as possible (full `// SAFETY` comments for the latter incoming shortly).

Specifically, this first makes the initialization of `Rav1dContext` in `fn rav1d_open` safe (not through a raw ptr).  This necessitates sending a `&Rav1dContext` to the worker threads (as we `unsafe`ly used to do through a raw ptr).  Since the worker threads may outlive both `fn rav1d_open` and `fn rav1d_close`, we change the `Box<Rav1dContext>` to `Arc<Rav1dContext>`.

Since we still need to set `c.tc` after spawning the worker threads, but we also need to send the `Arc<Rav1dContext>` to the worker threads, we have to do this indirectly and after `Rav1dContext` is fully constructed (otherwise we can't set `c.tc` unless we add a `Mutex` to it, which we don't want to do, since it's used often).  This is done by having the worker threads immediately `thread::park()`.  Then the main thread (in `fn rav1d_open`) set a field (with a `Mutex`) to contain an `Arc<Rav1dContext>`.  Then finally all of the worker threads are `unpark`ed, whereupon they `.take()` that `Arc<Rav1dContext>`, owning it now, and are able to use it as a `&Rav1dContext`.

Now that each thread owns an `Arc<Rav1dContext>`, `fn Rav1dContext::drop` won't be called until the main thread calls `fn rav1d_close` and all of the worker threads exit.  But `drop` is where we notify the worker threads that they should die and then joins them.  So we move that notification logic to `fn Rav1dContext::tell_worker_threads_to_die`, which is now explicitly called in `fn rav1d_close`.  Actually joining the worker threads isn't necessary anymore since `drop` will already be called once the last worker thread is exiting.

Then we update all of the `DAV1D_API`s that use `Dav1dContext` to be safer.  `Dav1dContext`, an opaque type, is changed from an alias of `Rav1dContext` to a ptr to it, specifically an `RawArc<Rav1dContext>`.  Thus, this changes usages of `*const Dav1dContext` to `Dav1dContext`, and `Option` can be used where `None` may be accepted/checked for.  And `RawArc` provides safer (still `unsafe`, but much safer) APIs for getting a `&Rav1dContext` or consuming the `Arc<Rav1dContext>`.  I'm going to add `// SAFETY` and `# Safety` comments to all of these `DAV1D_API`s next, but wanted to do it at the same time that I make all of the other arguments safe, too.

Finally, I cleaned up `fn rav1d_open` some more to directly initialize the `Rav1dContext` instead of first default initializing it.

### DIFF
--- a/include/dav1d/dav1d.rs
+++ b/include/dav1d/dav1d.rs
@@ -1,5 +1,6 @@
 use crate::include::dav1d::picture::Dav1dPicAllocator;
 use crate::include::dav1d::picture::Rav1dPicAllocator;
+use crate::src::c_arc::RawArc;
 use crate::src::error::Rav1dError;
 use crate::src::internal::Rav1dContext;
 pub use crate::src::log::Dav1dLogger;
@@ -9,7 +10,8 @@ use std::ffi::c_int;
 use std::ffi::c_uint;
 use strum::FromRepr;
 
-pub type Dav1dContext = Rav1dContext;
+pub type Dav1dContext = RawArc<Rav1dContext>;
+
 pub type Dav1dRef = ();
 
 pub type Dav1dInloopFilterType = c_uint;
--- a/src/c_arc.rs
+++ b/src/c_arc.rs
@@ -142,6 +142,24 @@ impl<T: ?Sized> CArc<T> {
 #[repr(transparent)]
 pub struct RawArc<T>(NonNull<PhantomData<T>>);
 
+/// We need a manual `impl` since we don't require `T: Clone`.
+///
+/// # Safety
+///
+/// Note that this [`RawArc::clone`] does not call [`Arc::clone`],
+/// since implicit clones/copies are expected to be done outside of Rust,
+/// for which there is no way to force [`RawArc::clone`] to be called.
+/// Instead, [`RawArc::as_ref`] and [`RawArc::into_arc`] are `unsafe`,
+/// and require [`RawArc::clone`]s (actual explicit calls
+/// or implicit ones outside of Rust) to respect the rules of [`Arc`].
+impl<T> Clone for RawArc<T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<T> Copy for RawArc<T> {}
+
 impl<T> RawArc<T> {
     pub fn from_arc(arc: Arc<T>) -> Self {
         Self(arc_into_raw(arc).cast())
--- a/src/c_arc.rs
+++ b/src/c_arc.rs
@@ -149,10 +149,28 @@ impl<T> RawArc<T> {
 
     /// # Safety
     ///
-    /// The [`RawArc`] must be originally from [`Self::from_arc`].F
+    /// The [`RawArc`] must be originally from [`Self::from_arc`].
+    pub unsafe fn as_ref(&self) -> &T {
+        // SAFETY: `self` must be from `Self::from_arc`,
+        // which calls `Arc::into_raw`,
+        // which returns a ptr to its `T`.
+        // `Arc` allows us to get a `&T` from it,
+        // so this is allowed (unlike `&mut T`).
+        // We don't call `Self::into_arc` since that's consuming,
+        // so we'd have to `mem::forget` the `Arc`
+        // and also do a redundant dereference.
+        unsafe { self.0.cast().as_ref() }
+    }
+
+    /// # Safety
+    ///
+    /// The [`RawArc`] must be originally from [`Self::from_arc`].
     pub unsafe fn into_arc(self) -> Arc<T> {
         let raw = self.0.cast().as_ptr();
-        Arc::from_raw(raw)
+        // SAFETY: `self` must be from `Self::from_arc`,
+        // which calls `Arc::into_raw`.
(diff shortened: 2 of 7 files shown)


randomPoison:
Can you elaborate on why joining on the worker threads is no longer necessary? What was it doing previously that we no longer need? My guess would be ensuring that the `Rav1dContext` doesn't get freed until the worker threads referencing it have died, is that accurate?

randomPoison (approved):
Code changes look good, I just had the one question about joining on the worker threads.

kkysen:
> Can you elaborate on why joining on the worker threads is no longer necessary? What was it doing previously that we no longer need? My guess would be ensuring that the `Rav1dContext` doesn't get freed until the worker threads referencing it have died, is that accurate?

Yeah, basically `drop` previously freed `Rav1dContext`s fields, so it had to make sure the worker threads were done and no longer holding references to the `Rav1dContext` for it to be safe to do this.  But now that the `Rav1dContext` is in an `Arc`, the `Arc` ensures it won't be `drop`ped until the last reference to it is `drop`ped, so we don't have to ensure the threads are joined first.  It's now enough just to tell the threads to die, and `drop` will run on the last worker thread to exit, rather than us joining all of the threads and then `drop`ping all of the fields on the main thread.

PR #1188, opened 2024-06-10
Backport `lvl` buffer alignment from dav1d 1.4.2

fbossen:
Ensures that SIMD stores performed by memset() are aligned.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -724,8 +724,8 @@ pub struct Rav1dFrameContext_lf {
     pub mask: Vec<Av1Filter>, /* len = w*h */
     pub lr_mask: Vec<Av1Restoration>,
     pub lim_lut: Align16<Av1FilterLUT>,
+    pub lvl: [Align16<[[[u8; 2]; 8]; 4]>; 8], /* [8 seg_id][4 dir][8 ref][2 is_gmv] */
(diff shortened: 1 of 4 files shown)


src/internal.rs
```diff
@@ -724,8 +724,8 @@ pub struct Rav1dFrameContext_lf {
     pub mask: Vec<Av1Filter>, /* len = w*h */
     pub lr_mask: Vec<Av1Restoration>,
     pub lim_lut: Align16<Av1FilterLUT>,
+    pub lvl: Align16<[[[[u8; 2]; 8]; 4]; 8]>, /* [8 seg_id][4 dir][8 ref][2 is_gmv] */
```
kkysen, line 727: A couple things:

Can you pass the type including the `Align16` to `fn`s operating on it?  Giving that alignment information to them will let them be optimized better.  And since we pass elements of this to `fn rav1d_create_lf_mask_inter`, could you move the `Align16` one dimension inside, like this?

```suggestion
    pub lvl: [Align16<[[[u8; 2]; 8]; 4]>; 8], /* [8 seg_id][4 dir][8 ref][2 is_gmv] */
```

kkysen (changes requested):

kkysen (approved):

PR #1189, opened 2024-06-10
`fn decode_coefs`: Cleanup `current_block` control flow

randomPoison:
The `goto` here is a bit more complex than just a cleanup block at the end of the function, so I think the generated state machine control flow is pretty much the most reasonable way to represent the necessary control flow. So I've just replaced the `current_block` variable with a cleaner enum that does the same thing.

### DIFF
--- a/src/recon.rs
+++ b/src/recon.rs
@@ -1354,15 +1354,11 @@ fn decode_coefs<BD: BitDepth>(
         NoQm,
     }
 
-    let mut ac = None;
+    let ac;
     if dc_tok == 0 {
         cul_level = 0;
         dc_sign_level = 1 << 6;
-        if qm_tbl.is_some() {
-            ac = Some(Ac::Qm);
-        } else {
-            ac = Some(Ac::NoQm);
-        }
+        ac = Some(if qm_tbl.is_some() { Ac::Qm } else { Ac::NoQm });
     } else {
         dc_sign_ctx = get_dc_sign_ctx(tx, a, l) as c_int;
         let dc_sign_cdf = &mut ts_c.cdf.coef.dc_sign[chroma][dc_sign_ctx as usize];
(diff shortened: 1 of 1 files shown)


src/recon.rs
```diff
@@ -1350,15 +1349,19 @@ fn decode_coefs<BD: BitDepth>(
     let mut cul_level: c_uint;
     let dc_sign_level: c_uint;
 
+    enum Ac {
+        Qm,
+        NoQm,
+    }
+
+    let mut ac = None;
```
kkysen, line 1357:
```suggestion
    let ac;
```

src/recon.rs
```diff
@@ -1350,15 +1349,19 @@ fn decode_coefs<BD: BitDepth>(
...
         dc_sign_level = 1 << 6;
         if qm_tbl.is_some() {
-            // goto ac_qm;
-            current_block = 1669574575799829731;
+            ac = Some(Ac::Qm);
         } else {
-            // goto ac_noqm;
-            current_block = 2404388531445638768;
+            ac = Some(Ac::NoQm);
         }
```
kkysen, line 1365:
```suggestion
        ac = Some(if qm_tbl.is_some() { Ac::Qm } else { Ac::NoQm });
```

src/recon.rs
```diff
@@ -1405,9 +1408,7 @@ fn decode_coefs<BD: BitDepth>(
             );
 
             if rc != 0 {
-                current_block = 1669574575799829731;
-            } else {
-                current_block = 15494703142406051947;
+                ac = Some(Ac::Qm);
             }
```
kkysen, line 1412:
```suggestion
            ac = if rc != 0 { Some(Ac::Qm) } else { None };
```

src/recon.rs
```diff
@@ -1439,15 +1440,12 @@ fn decode_coefs<BD: BitDepth>(
             );
 
             if rc != 0 {
-                current_block = 2404388531445638768;
-            } else {
-                current_block = 15494703142406051947;
+                ac = Some(Ac::NoQm);
             }
```
kkysen, line 1444:
```suggestion
            ac = if rc != 0 { Some(Ac::NoQm) } else { None };
```

kkysen (approved):
Generally LGTM, though we can make `ac` non-`mut` as I suggested below.  As for perf, I wouldn't think this would be that bad, as it's not in a loop.  And the asm appears to compile down to `goto`s essentially (the `if dc_tok == 0` branch basically does the `match ac`, while the other fallthrough ones are direct jumps).

randomPoison:
Applied the suggestions for making `ac` not mutable 👍

PR #1192, opened 2024-06-11
`struct RelaxedAtomic`: Replace `Atomic*`s that use only relaxed ops with `RelaxedAtomic`

### LINKED ISSUE #1173: Slow access to `struct Av1Filter` data
The `struct Av1Filter` contains `Atomic` elements. Access to these elements, in particular writes in `fn mask_edges_inter`, appear to be much slower than in the C code. 

`struct Rav1dFrameContext_lf` contains a vector of `Av1Filter`. Whereas multiple threads may read/write different elements of this vector, any single `Av1Filter` element would not be written to by a thread at the same time as another thread accesses the same `Av1Filter` element.

Use of non-`Atomic` elements in `struct Av1Filter` should be considered for performance reasons.

kkysen:
* Fixes #1173.

`RelaxedAtomic` requires `Ordering::Relaxed` loads and stores, disallowing other relaxed ops like `.fetch_or`.  The former are always plain loads and stores, while the latter are contended `lock *` instructions, so this ensures `RelaxedAtomic`s have little to no overhead.  It also makes clear which types are meant to be fully relaxed.

This should fix #1173 because it removes all relaxed `.fetch_*`es and replaces them with `.update`s, which do separate relaxed loads and stores, which don't have overhead.  This may replace an `add` with a load, `add`, and store, for example, but it gets rid of the `lock add`, which can cause contention, so it should generally have little to no overhead over no atomics at all.

### DIFF
--- a/include/dav1d/headers.rs
+++ b/include/dav1d/headers.rs
@@ -1,5 +1,5 @@
 use crate::src::enum_map::EnumKey;
-use atomig::Atomic;
+use crate::src::relaxed_atomic::RelaxedAtomic;
 use parking_lot::Mutex;
 use std::ffi::c_int;
 use std::ffi::c_uint;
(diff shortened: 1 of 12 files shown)


fbossen (approved):
With these changes, the time spent in `fn mask_edges_inter` is more consistent with the C code.

PR #1199, opened 2024-06-12
Backport simplification of `fn read_mv_residual` from dav1d 1.4.2

fbossen:


src/decode.rs
```diff
@@ -272,38 +264,22 @@ fn read_mv_component_diff(
...
-    let mv_joint = MVJoint::from_repr(rav1d_msac_decode_symbol_adapt4(
+fn read_mv_residual(ts_c: &mut Rav1dTileStateContext, ref_mv: &mut mv, mv_prec: i32) {
+    let mv_joint = MVJoint::from_bits(rav1d_msac_decode_symbol_adapt4(
         &mut ts_c.msac,
         &mut ts_c.cdf.mv.joint.0,
-        MVJoint::COUNT - 1,
-    ) as usize)
+        MVJoint::all().bits().into(),
+    ) as u8)
     .expect("valid variant");
```
kkysen, line 273: You can use `from_bits_truncate` and remove the `.expect("valid variant")`.

src/decode.rs
```diff
@@ -2550,7 +2526,11 @@ fn decode_b(
                 }
                 NEWMV => {
                     let mut mv1d = mvstack[drl_idx as usize].mv.mv[i];
-                    read_mv_residual(f, ts_c, &mut mv1d, !frame_hdr.force_integer_mv);
+                    read_mv_residual(
+                        ts_c,
+                        &mut mv1d,
+                        frame_hdr.hp as i32 - frame_hdr.force_integer_mv as i32,
```
kkysen, line 2532: Could you put this in a closure (or var, but with a closure, you only have to repeat it once), since C puts it in a separate var?

kkysen (changes requested):
Generally LGTM, but I made a couple small suggestions.

src/decode.rs
```diff
@@ -2272,6 +2247,8 @@ fn decode_b(
         }
     } else {
         // inter-specific mode/mv coding
+        let mv_prec = || -> i32 { frame_hdr.hp as i32 - frame_hdr.force_integer_mv as i32 };
```
kkysen, line 2250: Nit: I think this should work, too.

```suggestion
        let mv_prec = || frame_hdr.hp as i32 - frame_hdr.force_integer_mv as i32;
```

kkysen (approved):
LGTM, though I had one little nit.

kkysen (approved):

PR #1204, opened 2024-06-14
`looprestoration.rs`: Cleanup in aarch64 functions

randomPoison:
The aarch64 functions were already basically clean, minus some redundant type annotations and other minor issues. This PR cleans those up.

### DIFF
--- a/src/looprestoration.rs
+++ b/src/looprestoration.rs
@@ -2080,30 +2080,30 @@ unsafe fn sgr_filter_3x3_neon<BD: BitDepth>(
 
     const BUF_STRIDE: usize = 384 + 16;
 
-    let mut sumsq_buf: Align16<[i32; BUF_STRIDE * 3 + 16]> = Align16([0; BUF_STRIDE * 3 + 16]);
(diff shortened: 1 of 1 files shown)


kkysen (approved):
The changes here all good, but we should also make the `ptr::null_mut()`s into safe slices.  I'm not sure if you're already doing that or this is the final cleanup for these `neon` `fn`s.

randomPoison:
> we should also make the `ptr::null_mut()`s into safe slices.

I considered that, but I don't think it's worth doing. The only usage of those pointers is to pass to asm routines, they're never dereferenced or indexed into from Rust code. So I don't think we gain anything by making them slices instead of pointers, since we would just have to convert them back to pointers before passing to asm. I think the code as-is is as clean and safe as we can reasonably make it.

kkysen:
> > we should also make the `ptr::null_mut()`s into safe slices.
> 
> I considered that, but I don't think it's worth doing. The only usage of those pointers is to pass to asm routines, they're never dereferenced or indexed into from Rust code. So I don't think we gain anything by making them slices instead of pointers, since we would just have to convert them back to pointers before passing to asm. I think the code as-is is as clean and safe as we can reasonably make it.

But we do do `.split_at_mut` type operations.  I think that would be better in Rust to ensure we don't have mutable aliasing, even if they're just being passed to asm.

PR #1206, opened 2024-06-14
`ipred.rs`: Replace C types with Rust equivalents

### LINKED ISSUE #815: Cleanup c2rust output in `ipred.rs`

randomPoison:
* Closes #815

### DIFF
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -32,10 +32,7 @@ use crate::src::tables::dav1d_dr_intra_derivative;
 use crate::src::tables::dav1d_filter_intra_taps;
 use crate::src::tables::dav1d_sm_weights;
 use crate::src::wrap_fn_ptr::wrap_fn_ptr;
-use libc::ptrdiff_t;
 use std::cmp;
-use std::ffi::c_int;
-use std::ffi::c_uint;
 use std::mem;
 use std::slice;
 use strum::FromRepr;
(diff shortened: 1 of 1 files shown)


kkysen (commented):
I think we should stick to C types for FFI.  Uses inside functions we can change, though, and we can remove `ptrdiff_t` as you did.

randomPoison:
Why should we stick to C types in FFI? I think it'd be better to use the concrete Rust types rather than `c_int` and friends.

kkysen:
> Why should we stick to C types in FFI? I think it'd be better to use the concrete Rust types rather than `c_int` and friends.

Because it's `extern "C"`.  I think `c_int` is always `i32`, but others like `c_long` can be different, and I think it's better to be consistent with making the integer types in `extern "C" fn`s use C types.

thedataking:
> Because it's extern "C". I think c_int is always i32, but others like c_long can be different, and I think it's better to be consistent with making the integer types in extern "C" fns use C types.

I agree. A key reason why 32-bit builds were broken after transpiling for a 64-bit host was that `c_ulong` is 

```
pub type c_ulong = u64;
```

for `x86_64-unknown-linux-gnu` but

```
pub type c_ulong = u32;
```

for `i686-unknown-linux-gnu`. I spent a lot of time getting that unbungled.

randomPoison:
Ah okay noted. I'll close this for now then, I threw this together quickly because it was easy, going through and cleaning up C types within functions probably isn't worth the time right now.

PR #1208, opened 2024-06-16
`fn angular_ipred`: Make safe w/ `Rav1dPictureDataComponentOffset`

kkysen:
This makes `fn angular_ipred` and all of its fallbacks fully safe.

For most of these, this was pretty simple, but `fn ipred_filter_rust` was more complex since it uses normal buffers for the first iteration and then switches to `dst: Rav1dPictureDataComponentOffset`.  For the inner of these, I switched to an `if`, which as I explain in the commit 14dde20, I think should be fine perf-wise.

As for the call sites, sometimes and tmp/scratch buffer is passed in instead of an existing `Rav1dPictureDataComponent`, so I added `fn wrap_buf` for this, which safely wraps a `&mut buf`.  Since this is a unique reference, subdividing it further with the `DisjointMut` is sound (normally the buf is owned by the `DisjointMut`, and so is unique that way).

### DIFF
--- a/include/dav1d/picture.rs
+++ b/include/dav1d/picture.rs
@@ -175,6 +175,20 @@ impl Rav1dPictureDataComponentInner {
         assert!(len % RAV1D_PICTURE_MULTIPLE == 0);
         Self { ptr, len, stride }
     }
+
+    /// # Safety
+    ///
+    /// As opposed to [`Self::new`], this is safe because `buf` is a `&mut` and thus unique,
+    /// so it is sound to further subdivide it into disjoint `&mut`s.
+    pub fn wrap_buf<BD: BitDepth>(buf: &mut [BD::Pixel], stride: usize) -> Self {
+        let buf = AsBytes::as_bytes_mut(buf);
+        let ptr = NonNull::new(buf.as_mut_ptr()).unwrap();
+        assert!(ptr.cast::<AlignedPixelChunk>().is_aligned());
+        let len = buf.len();
+        assert!(len % RAV1D_PICTURE_MULTIPLE == 0);
+        let stride = (stride * mem::size_of::<BD::Pixel>()) as isize;
+        Self { ptr, len, stride }
+    }
 }
 
 // SAFETY: We only store the raw pointer, so we never materialize a `&mut`.
@@ -207,6 +221,12 @@ unsafe impl AsMutPtr for Rav1dPictureDataComponentInner {
 pub struct Rav1dPictureDataComponent(DisjointMut<Rav1dPictureDataComponentInner>);
 
 impl Rav1dPictureDataComponent {
+    pub fn wrap_buf<BD: BitDepth>(buf: &mut [BD::Pixel], stride: usize) -> Self {
+        Self(DisjointMut::new(
+            Rav1dPictureDataComponentInner::wrap_buf::<BD>(buf, stride),
+        ))
+    }
+
     /// Length in number of [`u8`] bytes.
     fn len(&self) -> usize {
         self.0.len()
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -1207,11 +1207,12 @@ unsafe fn ipred_filter_rust<BD: BitDepth>(
     let filter = &dav1d_filter_intra_taps[filt_idx];
     let mut top = &topleft_in[topleft_off + 1..][..width];
     for y in (0..height).step_by(2) {
-        let mut topleft = topleft_in.as_ptr().add(topleft_off - y);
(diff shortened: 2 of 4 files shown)


include/dav1d/picture.rs
```diff
@@ -207,6 +221,12 @@ unsafe impl AsMutPtr for Rav1dPictureDataComponentInner {
 pub struct Rav1dPictureDataComponent(DisjointMut<Rav1dPictureDataComponentInner>);
 
 impl Rav1dPictureDataComponent {
+    pub fn wrap_buf<BD: BitDepth>(buf: &mut [BD::Pixel], stride: usize) -> Self {
```
randomPoison, line 224: We're not tying the lifetime of the new `Rav1dPictureDataComponent{Inner}` to the lifetime of the buffer, right? Could we add a lifetime to `Rav1dPictureDataComponent` so that we can ensure that it doesn't outlive the borrowed buffer? The lifetime could be `'static` for owned picture data components. I suspect that'd make the type much more cumbersome to work with though, since we'd need to add lifetime annotations pretty much everywhere this is used.

If that won't work, should we at least add a safety comment here saying that the `Rav1dPictureDataComponent` can't outlive `buf`? I feel like that should make this function `unsafe` as well, since it creates a borrow that's not tracked by the compiler.

src/ipred.rs
```diff
@@ -2206,6 +2282,9 @@ impl Rav1dIntraPredDSPContext {
     }
 
     pub const fn new<BD: BitDepth>(flags: CpuFlags) -> Self {
-        Self::default::<BD>().init::<BD>(flags)
+        let default = Self::default::<BD>();
+        let mut c = Self::default::<BD>().init::<BD>(flags);
+        c.intra_pred[FILTER_PRED as usize] = default.intra_pred[FILTER_PRED as usize];
```
randomPoison, line 2287: What's this change for? It seems like it's unrelated to the rest of the changes in the PR, and it's unclear why this is even necessary. This seems like it's overriding the value of `c.intra_pred[FILTER_PRED]` set in `init` with the default value, i.e. the Rust fallback. Is that the correct thing to do? Isn't that going to override the asm routine that was set in `init`?

include/dav1d/picture.rs
kkysen: Yeah, the lifetime is incorrect/unchecked right now.  @rinon brought that up, too.  I think we can add a lifetime that defaults to `'static`.  Can I add that as a PR on top of everything, though, since there's a lot on top of this?

src/ipred.rs
kkysen: Whoops, I added this for debugging/checking I didn't break anything while running tests faster.  I did not mean to commit this.

src/ipred.rs
```diff
@@ -1200,7 +1200,8 @@ unsafe fn ipred_filter_rust<BD: BitDepth>(
     for y in (0..height).step_by(2) {
         let topleft_off = topleft_off - y;
         let mut topleft = topleft_in[topleft_off];
-        for x in (0..width).step_by(4) {
+        for x in 0..width / 4 {
+            let x = x * 4;
```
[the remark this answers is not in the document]
kkysen: Hmm, I might be totally wrong.  I forgot exactly what I was trying to optimize.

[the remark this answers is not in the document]
kkysen: Oh yeah, I think I wanted to make sure `width` is divisible by 4, so that `top[x..][..4]` where `top.len() == width` doesn't need a check.  If `width % 4 != 0`, then `x` could be `width - 1`, but if `width % 4 == 0`, then it guarantees that the max `x` is `width - 4`.  Let me fix this.

include/dav1d/picture.rs
randomPoison: Yeah, I'm fine doing that in a separate PR 👍

src/ipred.rs
kkysen: Fixed in [`c8c8921..cdc3c5e`](https://github.com/memorysafety/rav1d/compare/c8c89218a1ffb4537d12bf56e463d5f04c5b039d..cdc3c5e32b4e84bd0da00dca204bea2b740d1a60).

src/ipred.rs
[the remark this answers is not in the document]
kkysen: Fixed in [`cdc3c5e..59ab627`](https://github.com/memorysafety/rav1d/compare/cdc3c5e32b4e84bd0da00dca204bea2b740d1a60..59ab62729478ac84d58b19b8460ed41612dc3694).

randomPoison (approved):

PR #1209, opened 2024-06-16
`fn read_coef_tree`: Make `y_dst` arg safe w/ `Rav1dPictureDataComponentOffset`

kkysen:


src/recon.rs
```diff
@@ -3935,7 +3927,7 @@ pub(crate) unsafe fn rav1d_recon_b_inter<BD: BitDepth>(
                 y += ytx.h as c_int;
                 y_off += 1;
             }
-            y_dst += y_dst.pixel_stride::<BD>() * 4 * y as isize;
+            y_dst -= y_dst.pixel_stride::<BD>() * 4 * y as isize;
```
randomPoison, line 3930: Why did this change from `+=` to `-=`? This looks incorrect to me, but I might be missing something.

kkysen: It was a bug.

kkysen: You can see in the `dst.offset` version below that there's `-`.

randomPoison (approved):

PR #1211, opened 2024-06-16
`mod ipred`: Mark `#![deny(unsafe_op_in_unsafe_fn)]`

### LINKED ISSUE #846: `ipred.rs`: Unsafe cleanup

kkysen:
* Fixes #846.

### DIFF
--- a/src/ipred.rs
+++ b/src/ipred.rs
@@ -1473,7 +1473,6 @@ unsafe extern "C" fn pal_pred_c_erased<BD: BitDepth>(
     pal_pred_rust::<BD>(dst, pal, idx, w, h)
 }
 
-#[allow(unsafe_op_in_unsafe_fn)]
 #[cfg(all(feature = "asm", target_arch = "aarch64"))]
 mod neon {
     use super::*;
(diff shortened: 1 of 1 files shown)


src/ipred.rs
```diff
@@ -1504,7 +1509,8 @@ mod neon {
         ) {
             let dst = dst.cast();
             let topleft = topleft.as_ptr().cast();
-            self.get()(dst, stride, topleft, width, height, dxy, max_base_xy)
+            // SAFETY: We're assuming the asm is actually correct and safe.
```
randomPoison, line 1512: Nit: I prefer the wording you've used previously that specifically notes that the Rust fallback is fully safe and that the asm routine is supposed to do the same. Not super important, I just think it better clarified why we think the function pointer call is safe.

randomPoison (approved):

src/ipred.rs
kkysen: But there's no fallback for this.

PR #1212, opened 2024-06-16
`static dav1d_filter_intra_taps`: Rewrite `f!` and simplify as `const fn gen_filter{,s}`, as well as `const FLT_INCR` and `fn filter_fn`

kkysen:


randomPoison (approved):
The changes look okay, but I don't think this shouldn't have been prioritized over safety cleanup. Does this unblock any safety work? If not you should probably have waited to do this.

kkysen:
> The changes look okay, but I don't think this shouldn't have been prioritized over safety cleanup. Does this unblock any safety work? If not you should probably have waited to do this.

No, not safety, but I thought it was the general cleanup stuff and it was pretty simple to do.

PR #1214, opened 2024-06-16
`mod align`: Make safe

kkysen:
This makes all `unsafe` ops in `mod align` justified.  The remaining one, `fn <Align* as AsMutPtr>::as_mut_ptr`, I added a `// SAFETY` comment and fixed the impl to avoid creating a temporary `&mut`.

I also adjusted a few of the other comments, and also made the temporary `&mut` more apparent in one of them.

### DIFF
--- a/src/align.rs
+++ b/src/align.rs
@@ -104,7 +104,9 @@ macro_rules! def_align {
             type Target = V;
 
             unsafe fn as_mut_ptr(ptr: *mut Self) -> *mut V {
-                (*ptr).0.as_mut_ptr()
+                // SAFETY: `$name` (`Align*`) is a `#[repr(C)]` aligned wrapper around `[V; N]`,
+                // so a `*mut Self` is the same as a `*mut V` to the first `V`.
+                ptr.cast()
             }
 
             fn len(&self) -> usize {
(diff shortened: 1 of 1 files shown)


src/align.rs
```diff
@@ -148,7 +150,7 @@ impl<T: Copy, C: AlignedByteChunk> AlignedVec<T, C> {
         }
     }
 
-    /// Returns the number of elements in the vector.
+    /// Return the number of elements in the vector.
```
randomPoison, line 153: Nit: Convention for doc comments is to use present tense, so this should be "Returns" instead of "Return" (the way it was before). See for example [the `slice::len` method in the std docs](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.len). Not really an important thing to clean up, but we should stick with the more conventional wording in our comments.

randomPoison (approved):

src/align.rs
```diff
@@ -161,17 +163,17 @@ impl<T: Copy, C: AlignedByteChunk> AlignedVec<T, C> {
         self.inner.as_mut_ptr().cast()
     }
 
-    /// Extracts a slice containing the entire vector.
+    /// Extract a slice containing the entire vector.
     pub fn as_slice(&self) -> &[T] {
-        // Safety: The first `len` elements have been initialized to `T`s in
-        // `Self::resize_with`.
+        // Safety: The first `len` elements have been
+        // initialized to `T`s in `Self::resize_with`.
```
randomPoison, line 169: Another minor nit that I'll mention here because I've seen it in a few of your PRs is that you tend to wrap comments a bit too aggressively. [The style guide](https://doc.rust-lang.org/nightly/style-guide/#comments) indicates that comments should be wrapped at 80 characters wide, unless they'd pass the 100 character line limit. So in this PR you've taken some comments that were fine at their previous width and wrapped them shorter. Like I said, not important and I wouldn't normally mention it, but I bring it up here since most of this PR is updating comments.

src/align.rs
kkysen: Oh, oops, I thought it was the other way around.

src/align.rs
kkysen: Hmm, I didn't know the style guide said that.  100 is a bit more forgiving than just 80, so this one could fit all on one line.  I'll keep that in mind.  I do read code a lot on my phone, though, which is quite narrow.  In general, though, I prefer semantic line wrapping rather than any hard wrapping, since it reduces diffs and is IMO easier to read.

randomPoison: What do you mean by "semantic line wrapping"?

kkysen: I mean wrapping lines at semantic breaks, like punctuation, end of phrases, other places you'd pause while reading it aloud, etc.  That way if you change one word in a line, it generally would only affect that line, as opposed to with hard wrapping, it might cascade down to all subsequent lines.

PR #1223, opened 2024-06-17
Backport `mod msac` improvements from dav1d 1.4.1

fbossen:
Skip the overhead of shifting in ones into the LSB in the common case, that's only required in the EOB padding. In practice this means we only have to invert bits once during the refill process instead of twice in every call to `msac` functions.
    
Also make some improvements to the refill asm, mainly involving keeping partially inserted bytes at the end instead of clearing them.

NOTE: the original dav1d commit also applies changes to `loongarch` asm that aren't included here.

### DIFF
--- a/src/arm/32/msac.S
+++ b/src/arm/32/msac.S
@@ -279,60 +279,67 @@ L(renorm):
         sub             r4,  r4,  r3           // rng = u - v
         clz             r5,  r4                // clz(rng)
         eor             r5,  r5,  #16          // d = clz(rng) ^ 16
-        mvn             r7,  r7                // ~dif
-        add             r7,  r7,  r3, lsl #16  // ~dif + (v << 16)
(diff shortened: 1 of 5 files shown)


kkysen (approved):
LGTM, but could we include the `longaarch` asm, too?  Even if we're not using it, we already have the existing asm checked in already, so I think it's best to keep it all in sync in case we ever add support in the future.

fbossen:
> LGTM, but could we include the `loongarch` asm, too? Even if we're not using it, we already have the existing asm checked in already, so I think it's best to keep it all in sync in case we ever add support in the future.

AFAICT, we don't have any `loongarch` asm code yet.

kkysen:
> AFAICT, we don't have any `loongarch` asm code yet.

Oh if we don't have any yet then it's fine.

PR #1226, opened 2024-06-17
Backport extended arm feature detection from dav1d 1.4.1

fbossen:
Add run-time CPU feature detection for DotProd, i8mm, SVE and SVE2. SVE and SVE2 are AArch64-only features.


kkysen (approved):
LGTM.  I'd been wondering why there wasn't any SVE asm.

PR #1231, opened 2024-06-20
`struct LossLess`: Make `Rav1dFrameHeader_segmentation::lossless` a bit array

### LINKED ISSUE #1180: `fn decode_coefs` is slow
`fn decode_coefs` has been observed to be about 40% slower than its C equivalent. The main contributor to the speed difference appears to be the numerous bound checks that Rust is performing. For example, bounds checks in `get_lo_ctx` when accessing accessing `levels` may be superfluous as the maximum offset into `levels` is `2 * stride` and we know that an additional `2 * stride` values have been cleared in `levels` before: `levels[..(stride * (4 * sw as isize + 2)) as usize].fill(0);` (note the `+ 2`).
It may sufficient to check for `y < 4 * sh` (note that `stride = 4 * sh`) and `x < 4 * sw` and forgo all other bound checks inside the `while i > 0` loop.

kkysen:
* Part of #1180.

I'm working on trying to fix #1180; this is one of a bunch of steps.


include/dav1d/headers.rs
```diff
@@ -2096,6 +2096,48 @@ pub struct Dav1dFrameHeader_segmentation {
...
+
+    pub fn from_array(a: [bool; RAV1D_MAX_SEGMENTS as usize]) -> Self {
+        let mut this = Self::empty();
+        let mut i = 0;
+        while i < RAV1D_MAX_SEGMENTS {
+            this.bits |= (a[i as usize] as u8) << i;
+            i += 1;
+        }
+        this
+    }
```
CrazyboyQCD, line 2138: These two methods can be replaced with instructions `pdep` and `pext` (slow on `Zen` processor).
And there are other [alternatives](https://stackoverflow.com/questions/8461126/how-to-create-a-byte-out-of-8-bool-values-and-vice-versa#answer-51750902), just for reference, since they are unsafe operations.

randomPoison (commented):
If I'm understanding this correctly, it seems like the main optimization here is using a bitmask operations (`LossLess::get`) over an array index operation. It's not clear to me that this would necessarily be more efficient though. Have you profiled this and observed a clear performance improvement?

include/dav1d/headers.rs
```diff
@@ -2096,6 +2096,48 @@ pub struct Dav1dFrameHeader_segmentation {
     pub qidx: [u8; DAV1D_MAX_SEGMENTS as usize],
 }
 
+/// Like a `[bool; RAV1D_MAX_SEGMENTS as usize]`,
+/// but the [`bool`]s are compressed into the bits of the [`u8`].
+#[derive(Clone, Copy, Default)]
+pub struct LossLess {
```
randomPoison, line 2102: Nit: This should be `Lossless` (one word), instead of "Loss Less" (two words).

include/dav1d/headers.rs
kkysen: Interesting... Thanks for letting me know!  These `fn`s are only called a few times in `DRav1d` conversions, and in `mod obu`, though, none of which are hot, so I'm not sure if it's worth it to make this optimization.  Definitely not asm ones.  For the safe multiplication one, maybe, but I'm not sure it's worth the clarity if the `fn` is not hot at all.  Maybe I'll do it.

include/dav1d/headers.rs
kkysen: But `LossLess` looks cooler 🥺.

kkysen:
> If I'm understanding this correctly, it seems like the main optimization here is using a bitmask operations (`LossLess::get`) over an array index operation. It's not clear to me that this would necessarily be more efficient though. Have you profiled this and observed a clear performance improvement?

No, I didn't profile this one individually yet.  It does remove the bounds checks, though I guess that's also solved by `SegmentId` and `InRange` later on.  It does also use less memory, which I think is useful as well.

randomPoison:
> It does remove the bounds checks, though I guess that's also solved by `SegmentId` and `InRange` later on. It does also use less memory, which I think is useful as well.

Given that you have other changes that target bounds checks for the `lossless` array, the only real advantage here is the memory reduction. That may still be a nice optimization to have, depending on how many of these are expected to be in memory at once, i.e. how many frame headers there are at a given time. I'm assuming not many, so this doesn't really seem like a worthwhile change to me, but I could be wrong about that. @fbossen might be the best person to weigh in on if this is a reasonable optimization.

include/dav1d/headers.rs
randomPoison: I would say stick with the simpler/clearer loop-based approach for now. If we decide to go with this change (which we may not), we can profile and use a more optimized approach if it proves to be on the hot path.

kkysen: Left a comment in 2720efd explaining this for if we do want to optimize it in the future.

kkysen:
> Given that you have other changes that target bounds checks for the `lossless` array, the only real advantage here is the memory reduction. That may still be a nice optimization to have, depending on how many of these are expected to be in memory at once, i.e. how many frame headers there are at a given time. I'm assuming not many, so this doesn't really seem like a worthwhile change to me, but I could be wrong about that. @fbossen might be the best person to weigh in on if this is a reasonable optimization.

It's not just about less overall memory, for which memory reductions would be quite small, but about what could be cached.

@fbossen, what do you think?

fbossen:
I've run a bunch of tests and it seems this PR leads to a small performance regression. The potential memory saving here is too small to have a positive impact.

PR #1235, opened 2024-06-20
`fn get_skip_ctx`: Optimize

kkysen:
* Part of #1180.

Optimize and remove all of the bounds checks by checking `dir.len()` directly instead of `tx`.


randomPoison (approved):
The changes look fine, have you profiled this to verify that it's having a positive performance impact?

PR #1237, opened 2024-06-20
aarch64: Check for assembler support for various aarch64 extensions

fbossen:
First check if the assembler supports the ".arch" directive, and what architecture levels are supported.

In principle, we'd only need to check for support for ".arch armv8.2-a", since that's enough for enabling the i8mm and sve2 extensions.

However, recent Clang versions (before version 17) wasn't able to enable the dotprod and i8mm extensions via the ".arch_extension" directives, so check for support for armv8.4-a and armv8.6-a as well, which enable dotprod and i8mm implicitly.

This allows assembling these instructions on most commonly available GCC and Clang based toolchains, while still allowing toggling support for the instruction sets on and off within the source files.

Within assembly, we disable these extensions by default, so that instructions enabled within these extension sets can't be used by accident in unintended functions. Code meaning to use these extensions can be assembled like this:

    #if HAVE_SVE
    ENABLE_SVE
    // code
    DISABLE_SVE
    #endif

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -369,6 +369,66 @@ if (is_asm_enabled and
     if cc.compiles(check_pic_code)
         cdata.set('PIC', '3')
     endif
+
+    if host_machine.cpu_family() == 'aarch64'
+        have_as_arch = cc.compiles('''__asm__ (".arch armv8-a");''')
+        cdata.set10('HAVE_AS_ARCH_DIRECTIVE', have_as_arch)
+        as_arch_str = ''
+        if have_as_arch
+            as_arch_level = 'armv8-a'
+            # Check what .arch levels are supported. In principle, we only
+            # want to detect up to armv8.2-a here (binutils requires that
+            # in order to enable i8mm). However, older Clang versions
+            # (before Clang 17, and Xcode versions up to and including 15.0)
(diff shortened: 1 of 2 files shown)


meson.build
kkysen, line 1: @thedataking, @randomPoison, do we need to port any of this to `build.rs`?

randomPoison: Yeah, it looks like we'd need additional `build.rs` support in order to take advantage of any assembly functionality that relies on the `ENABLE_*`/`DISABLE_*` macros. If I'm understanding this correctly, without setting the expected defines any such functionality will be disabled by default, meaning we're going to have potentially less efficient assembly on aarch64. We might want to open a ticket to track the work to port the build logic over.

fbossen: Note that I am working on a PR that includes asm with aarch64 extensions. I mainly need this commit to define macros such as `ENABLE_DOTPROD` to avoid compilation failures.

fbossen:
> @fbossen should we just land this then add support in build.rs or were you planning to add support to define `HAVE_*` to build.rs?

Yes, it would be good to land this. I have changes to `build.rs` in dependent PRs that I haven't pushed yet.

rinon (approved):

PR #1239, opened 2024-06-21
`fn cdef`: Make safe

### LINKED ISSUE #850: `cdef.rs`: Unsafe cleanup
There are ~170 unsafe operations happening in `cdef.rs`.

- [ ] Cleanup unsafe operations happening within functions.
- [ ] Mark any functions that can be made fully safe as such.
- [ ] Add `unsafe` blocks around any remaining unsafe operations along with standard safety documentation.
- [ ] Add safety documentation for functions that need to remain unsafe.
- [ ] Add `#![deny(unsafe_op_in_unsafe_fn)]` to file.

### LINKED ISSUE #851: `cdef_apply.rs`: Unsafe cleanup
There are ~60 unsafe operations happening in `cdef_apply.rs`.

- [ ] Cleanup unsafe operations happening within functions.
- [ ] Mark any functions that can be made fully safe as such.
- [ ] Add `unsafe` blocks around any remaining unsafe operations along with standard safety documentation.
- [ ] Add safety documentation for functions that need to remain unsafe.
- [ ] Add `#![deny(unsafe_op_in_unsafe_fn)]` to file.

randomPoison:
Makes the remaining Rust fallback in `cdef.rs` safe by passing in additional arguments.

* Closes #850.
* Closes #851.

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -10,9 +10,9 @@ use crate::src::cpu::CpuFlags;
 use crate::src::disjoint_mut::DisjointMut;
 use crate::src::ffi_safe::FFISafe;
 use crate::src::pic_or_buf::PicOrBuf;
-use crate::src::pixels::Pixels;
 use crate::src::strided::Strided as _;
 use crate::src::tables::dav1d_cdef_directions;
+use crate::src::with_offset::WithOffset;
 use crate::src::wrap_fn_ptr::wrap_fn_ptr;
 use bitflags::bitflags;
 use libc::ptrdiff_t;
@@ -54,11 +54,12 @@ wrap_fn_ptr!(pub unsafe extern "C" fn cdef(
     edges: CdefEdgeFlags,
     bitdepth_max: c_int,
     _dst: *const FFISafe<Rav1dPictureDataComponentOffset>,
-    _top: *const FFISafe<DisjointMut<AlignedVec64<u8>>>,
+    _top: *const FFISafe<CdefTop<'_>>,
     _bottom: *const FFISafe<CdefBottom<'_>>,
 ) -> ());
 
-pub type CdefBottom<'a> = PicOrBuf<'a, AlignedVec64<u8>>;
+pub type CdefTop<'a> = WithOffset<&'a DisjointMut<AlignedVec64<u8>>>;
+pub type CdefBottom<'a> = WithOffset<PicOrBuf<'a, AlignedVec64<u8>>>;
 
 impl cdef::Fn {
     /// CDEF operates entirely on pre-filter data.
@@ -70,10 +71,8 @@ impl cdef::Fn {
         &self,
         dst: Rav1dPictureDataComponentOffset,
         left: &[LeftPixelRow2px<BD::Pixel>; 8],
-        top: &DisjointMut<AlignedVec64<u8>>,
-        top_off: usize,
+        top: CdefTop<'_>,
         bottom: CdefBottom<'_>,
-        bottom_off: usize,
         pri_strength: c_int,
         sec_strength: u8,
         dir: c_int,
@@ -84,12 +83,9 @@ impl cdef::Fn {
         let dst_ptr = dst.as_mut_ptr::<BD>().cast();
         let stride = dst.stride();
         let left = ptr::from_ref(left).cast();
-        let top_ptr = (&*top.element_as(top_off) as *const BD::Pixel).cast();
-        let bottom_ptr = match bottom {
-            PicOrBuf::Pic(pic) => pic.as_ptr_at::<BD>(bottom_off).cast(),
-            PicOrBuf::Buf(buf) => (&*buf.element_as(bottom_off) as *const BD::Pixel).cast(),
-        };
-        let top = FFISafe::new(top);
+        let top_ptr = top.as_ptr::<BD>().cast();
+        let bottom_ptr = bottom.as_ptr::<BD>().cast();
+        let top = FFISafe::new(&top);
         let bottom = FFISafe::new(&bottom);
         let sec_strength = sec_strength as c_int;
         let damping = damping as c_int;
@@ -172,16 +168,14 @@ fn padding<BD: BitDepth>(
     tmp: &mut [i16; TMP_STRIDE * TMP_STRIDE],
     src: Rav1dPictureDataComponentOffset,
     left: &[LeftPixelRow2px<BD::Pixel>; 8],
-    top: &DisjointMut<AlignedVec64<u8>>,
-    top_off: usize,
+    top: CdefTop<'_>,
     bottom: CdefBottom<'_>,
-    bottom_off: usize,
     w: usize,
     h: usize,
     edges: CdefEdgeFlags,
 ) {
-    let top_off = top_off - 2;
-    let bottom_off = bottom_off - 2;
+    let top = top - 2_usize;
+    let bottom = bottom - 2_usize;
     let stride = src.pixel_stride::<BD>();
 
     // Fill extended input buffer.
@@ -207,8 +201,8 @@ fn padding<BD: BitDepth>(
     }
 
     for (i, y) in (y_start..2).enumerate() {
-        let offset = top_off.wrapping_add_signed(i as isize * stride);
-        let top = top.slice_as::<_, BD::Pixel>((offset.., ..x_end));
+        let top = top + i as isize * stride;
+        let top = top.data.slice_as::<_, BD::Pixel>((top.offset.., ..x_end));
         for x in x_start..x_end {
             tmp[x + y * TMP_STRIDE] = top[x].as_::<i16>();
         }
@@ -228,12 +222,12 @@ fn padding<BD: BitDepth>(
     }
     for (i, y) in (h + 2..y_end).enumerate() {
         let tmp = &mut tmp[y * TMP_STRIDE..];
-        let bottom_off = bottom_off.wrapping_add_signed(i as isize * stride);
+        let bottom = bottom + i as isize * stride;
         // This is a fallback `fn`, so perf is not as important here, so an extra branch
         // here should be okay.
-        let bottom = match bottom {
-            PicOrBuf::Pic(pic) => &*pic.slice::<BD, _>((bottom_off.., ..x_end)),
-            PicOrBuf::Buf(buf) => &*buf.slice_as((bottom_off.., ..x_end)),
+        let bottom = match bottom.data {
+            PicOrBuf::Pic(pic) => &*pic.slice::<BD, _>((bottom.offset.., ..x_end)),
+            PicOrBuf::Buf(buf) => &*buf.slice_as((bottom.offset.., ..x_end)),
         };
         for x in x_start..x_end {
             tmp[x] = bottom[x].as_::<i16>();
@@ -245,10 +239,8 @@ fn padding<BD: BitDepth>(
 fn cdef_filter_block_rust<BD: BitDepth>(
     dst: Rav1dPictureDataComponentOffset,
     left: &[LeftPixelRow2px<BD::Pixel>; 8],
-    top: &DisjointMut<AlignedVec64<u8>>,
-    top_off: usize,
+    top: CdefTop<'_>,
     bottom: CdefBottom<'_>,
-    bottom_off: usize,
     pri_strength: c_int,
     sec_strength: c_int,
     dir: c_int,
@@ -263,9 +255,7 @@ fn cdef_filter_block_rust<BD: BitDepth>(
     assert!((w == 4 || w == 8) && (h == 4 || h == 8));
     let mut tmp = [0; TMP_STRIDE * TMP_STRIDE]; // `12 * 12` is the maximum value of `TMP_STRIDE * (h + 4)`.
 
-    padding::<BD>(
-        &mut tmp, dst, left, top, top_off, bottom, bottom_off, w, h, edges,
-    );
+    padding::<BD>(&mut tmp, dst, left, top, bottom, w, h, edges);
 
     let tmp = tmp;
     let tmp_offset = 2 * TMP_STRIDE + 2;
@@ -385,8 +375,8 @@ unsafe extern "C" fn cdef_filter_block_c_erased<BD: BitDepth, const W: usize, co
     _dst_ptr: *mut DynPixel,
     _stride: ptrdiff_t,
     left: *const [LeftPixelRow2px<DynPixel>; 8],
-    top_ptr: *const DynPixel,
-    bottom_ptr: *const DynPixel,
+    _top_ptr: *const DynPixel,
+    _bottom_ptr: *const DynPixel,
     pri_strength: c_int,
     sec_strength: c_int,
     dir: c_int,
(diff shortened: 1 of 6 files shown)


src/cdef.rs
```diff
@@ -41,29 +43,37 @@ wrap_fn_ptr!(pub unsafe extern "C" fn cdef(
...
+    top_ptr: *const DynPixel,
+    bottom_ptr: *const DynPixel,
     pri_strength: c_int,
     sec_strength: c_int,
     dir: c_int,
     damping: c_int,
     edges: CdefEdgeFlags,
     bitdepth_max: c_int,
     _dst: *const FFISafe<Rav1dPictureDataComponentOffset>,
+    _top: *const FFISafe<DisjointMut<AlignedVec<u8, Align64<[u8; 64]>>>>,
```
kkysen, line 55: I think we have an `AlignedVec64` alias for this.

src/cdef.rs
```diff
@@ -41,29 +43,37 @@ wrap_fn_ptr!(pub unsafe extern "C" fn cdef(
...
+    _bottom: *const FFISafe<CdefBottom<'_>>,
 ) -> ());
 
+#[derive(Clone, Copy)]
+pub enum CdefBottom<'a> {
+    Pic(Rav1dPictureDataComponentOffset<'a>),
+    LineBuf((&'a DisjointMut<AlignedVec<u8, Align64<[u8; 64]>>>, usize)),
```
kkysen, line 62:
```suggestion
    Line {
        buf: &'a DisjointMut<AlignedVec<u8, Align64<[u8; 64]>>>,
        offset: usize,
    },
```

src/cdef.rs
```diff
@@ -41,29 +43,37 @@ wrap_fn_ptr!(pub unsafe extern "C" fn cdef(
...
     /// However, the edge pixels above `dst` may be post-filter,
     /// so in order to get access to pre-filter top pixels, use `top`.
-    pub unsafe fn call<BD: BitDepth>(
+    pub fn call<BD: BitDepth>(
         &self,
         dst: Rav1dPictureDataComponentOffset,
         left: &[LeftPixelRow2px<BD::Pixel>; 8],
-        top: *const BD::Pixel,
-        bottom: *const BD::Pixel,
+        (top, top_off): (&DisjointMut<AlignedVec<u8, Align64<[u8; 64]>>>, usize),
```
kkysen, line 75: Can these be separate args?  No real reason to combine them into a tuple, right?

src/cdef.rs
```diff
@@ -74,26 +84,38 @@ impl cdef::Fn {
         let dst_ptr = dst.as_mut_ptr::<BD>().cast();
         let stride = dst.stride();
         let left = ptr::from_ref(left).cast();
-        let top = top.cast();
-        let bottom = bottom.cast();
+        let top_ptr = (&*top.element_as(top_off) as *const BD::Pixel).cast();
```
kkysen, line 87: Provenance wise, `.element_as` is incorrect here since it will turn into a slice in the fallbacks, and the bounds check is only checking that one element.  What I did for `.as_ptr` for `Rav1dPictureDataComponentOffset` is just check that the offset is `<=` the length, corresponding to slicing/`.add` being safe.  This is called `.as_ptr_at` (and a `mut` variant) for `Rav1dPictureDataComponent`, so we could add it to `DisjointMut`, too, I think.  Internally, it calls `.as_mut_ptr().add(offset)` after checking `offset` is in bounds (and see that `fn` for how the out of bounds path should be a non-generic `#[inline(never)]` call).

src/cdef.rs
```diff
@@ -346,31 +380,40 @@ unsafe fn cdef_filter_block_rust<BD: BitDepth>(
...
     // SAFETY: Was passed as `FFISafe::new(_)` in `cdef_dir::Fn::call`.
     let dst = *unsafe { FFISafe::get(dst) };
     // SAFETY: Reverse of cast in `cdef::Fn::call`.
     let left = unsafe { &*left.cast() };
-    let top = top.cast();
-    let bottom = bottom.cast();
+    // SAFETY: Was passed as `FFISafe::new(_)` in `cdef::Fn::call`.
+    let top = unsafe { FFISafe::get(top) };
+    let top_base = top.as_mut_ptr().cast::<BD::Pixel>().cast_const();
+    // SAFETY: Reverse of what was done ine `cdef::Fn::call`. `top_ptr` is
```
kkysen, line 407:
```suggestion
    // SAFETY: Reverse of what was done in `cdef::Fn::call`. `top_ptr` is
```

src/cdef_apply.rs
```diff
@@ -260,66 +263,66 @@ pub(crate) unsafe fn rav1d_cdef_brow<BD: BitDepth>(
...
-                                    + (bx * 4) as isize;
-                                top = &*f
-                                    .lf
-                                    .lr_line_buf
-                                    .element_as((f.lf.lr_lpf_line[0] as isize + offset) as usize);
-                            }
+                                (
+                                    &f.lf.lr_line_buf,
+                                    f.lf.lr_lpf_line[0].wrapping_add_signed(
+                                        (sby * ((4 as c_int) << sb128) - 4) as isize * y_stride
```
kkysen, line 280: nit:

```suggestion
                                        (sby * (4 << sb128) - 4) as isize * y_stride
```

src/cdef_apply.rs
```diff
@@ -260,66 +263,66 @@ pub(crate) unsafe fn rav1d_cdef_brow<BD: BitDepth>(
...
-                                    .element_as((f.lf.cdef_lpf_line[0] as isize + offset) as usize);
+                            let bottom = if resize {
+                                (
+                                    &f.lf.cdef_line_buf,
+                                    f.lf.cdef_lpf_line[0].wrapping_add_signed(
+                                        (sby * 4 + 2) as isize * y_stride + (bx * 4) as isize,
+                                    ),
+                                )
                             } else {
                                 let line = sby * ((4 as c_int) << sb128) + 4 * sb128 as c_int + 2;
```
kkysen, line 302:
```suggestion
                                let line = sby * (4 << sb128) + 4 * sb128 as c_int + 2;
```

src/cdef.rs
```diff
@@ -204,19 +231,26 @@ unsafe fn padding<BD: BitDepth>(
     }
     for (i, y) in (h + 2..y_end).enumerate() {
         let tmp = &mut tmp[y * TMP_STRIDE..];
-        let bottom = slice::from_raw_parts(bottom.offset(i as isize * stride), x_end);
+        let offset = i as isize * stride;
+        let bottom = match bottom {
```
kkysen, line 235:
```suggestion
        // This is a fallback `fn`, so perf is not as important here,
        // so an extra branch here should be okay.
        let bottom = match bottom {
```

src/cdef_apply.rs
```diff
@@ -361,66 +364,72 @@ pub(crate) unsafe fn rav1d_cdef_brow<BD: BitDepth>(
...
+                                        )
                                     } else {
-                                        let line_0 = sby * ((4 as c_int) << sb128) - 4;
-                                        offset = line_0 as isize * uv_stride
-                                            + (bx * 4 >> ss_hor) as isize;
-                                        top = &*f.lf.lr_line_buf.element_as(
-                                            (f.lf.lr_lpf_line[pl] as isize + offset) as usize,
-                                        );
-                                    }
+                                        let line = sby * ((4 as c_int) << sb128) - 4;
```
kkysen, line 379:
```suggestion
                                        let line = sby * (4 << sb128) - 4;
```

src/cdef_apply.rs
```diff
@@ -361,66 +364,72 @@ pub(crate) unsafe fn rav1d_cdef_brow<BD: BitDepth>(
...
+                                        (
+                                            &f.lf.cdef_line_buf,
+                                            f.lf.cdef_lpf_line[pl].wrapping_add_signed(
+                                                (sby * 4 + 2) as isize * uv_stride
+                                                    + (bx * 4 >> ss_hor) as isize,
+                                            ),
+                                        )
                                     } else {
                                         let line =
                                             sby * ((4 as c_int) << sb128) + 4 * sb128 as c_int + 2;
```
kkysen, line 408:
```suggestion
                                            sby * (4 << sb128) + 4 * sb128 as c_int + 2;
```

src/cdef.rs
kkysen: Also, would it be simpler to put the `enum` only around the `DisjointMut<AlignedVec64<u8>>` and `Rav1dPictureDataComponent`, since the `offset` is the same for both variants?

kkysen (changes requested):

src/cdef.rs
randomPoison: Yeah, that's reasonable and unifies the offset logic. I made that change.

src/cdef.rs
randomPoison: Sure, done.

src/cdef.rs
randomPoison: > Provenance wise, `.element_as` is incorrect here since it will turn into a slice in the fallbacks, and the bounds check is only checking that one element.

I don't understand what you mean by "turn into a slice in the fallback" here. We're not using the top pointer to construct a slice in the fallback, we use the top buffer directly. So any index operations in the fallback will be bounds checked. So I think `element_as` is reasonable here because it handles both the bounds check and using pixel offsets.

kkysen: Oh, true, you're only using this to recalculate the offset, so provenance-wise it's fine.  That said, I think the check should be `<= len` (`.slice_as`) rather than `< len` (`.element_as`).

randomPoison: That doesn't really make sense to me, either. `slice_as` requires a range to slice, but in this case we don't know the range up front, we only know the initial offset. Also I don't see how `element_as`'s bounds check would be different from `slice_as`? Don't both require that `offset < len`?

kkysen: Sorry, I think I got a bit confused here.  Forget about the slicing.

What `Rav1dPictureDataComponentOffset::as_ptr` and `Rav1dPictureDataComponent::as_ptr_at` do is effectively a checked `.add` on the raw ptr.  No intermediary reference or guard is created.  This is provably safe to do because we check the bounds, and thus `.add` is safe.  This makes no guarantees on how the ptr will be used later, which is correct, other than that it is in bounds of course.

In contrast, using something like `.element_as` creates an intermediary guard and reference that we then cast to a ptr and drop.  This has very different semantics.  The creation of this intermediary `&` asserts that there is no other `&mut` concurrently, which we do not know, because we don't know (in general, though we may look at `fn`s case-by-case) that uses of this ptr, or the `DisjointMut` in the fallback, will access this element.  They may first offset it by some amount and access different elements, while another thread may access that first element.  That would be UB.

kkysen:
Wait how did 093a17d close this PR?

randomPoison:
> Wait how did [093a17d](https://github.com/memorysafety/rav1d/commit/093a17d9d091c4f24d18d1621a1a62714efc358a) close this PR?

"Resolve" is a keyword to trigger closing a PR.

kkysen:
> > Wait how did [093a17d](https://github.com/memorysafety/rav1d/commit/093a17d9d091c4f24d18d1621a1a62714efc358a) close this PR?
> 
> "Resolve" is a keyword to trigger closing a PR.

Ugh, but I only linked a comment on a PR, not the PR itself.  GitHub should be smarter.

src/cdef.rs
randomPoison: This is done now that I'm using `PicOrBuf`.

randomPoison:
@kkysen I've reworked this to use `PicOrBuf` and `WithOffset`, please give this another look.

src/cdef_apply.rs
```diff
@@ -362,66 +376,82 @@ pub(crate) unsafe fn rav1d_cdef_brow<BD: BitDepth>(
...
-                                            (f.lf.cdef_lpf_line[pl] as isize + offset) as usize,
-                                        );
+                                    let top = if resize {
+                                        WithOffset {
+                                            data: &f.lf.cdef_line_buf,
+                                            offset: f.lf.cdef_lpf_line[pl].wrapping_add_signed(
+                                                ((sby - 1) * 4) as isize * uv_stride
+                                                    + (bx * 4 >> ss_hor) as isize,
+                                            ),
+                                        }
```
kkysen, line 389:
```suggestion
                                        WithOffset {
                                            data: &f.lf.cdef_line_buf,
                                            offset: f.lf.cdef_lpf_line[pl],
                                        } + ((sby - 1) * 4) as isize * uv_stride + (bx * 4 >> ss_hor) as isize
```

src/cdef_apply.rs
```diff
@@ -362,66 +376,82 @@ pub(crate) unsafe fn rav1d_cdef_brow<BD: BitDepth>(
...
-                                        );
-                                    }
+                                        let line = sby * (4 << sb128) - 4;
+                                        WithOffset {
+                                            data: &f.lf.lr_line_buf,
+                                            offset: f.lf.lr_lpf_line[pl].wrapping_add_signed(
+                                                line as isize * uv_stride
+                                                    + (bx * 4 >> ss_hor) as isize,
+                                            ),
+                                        }
```
kkysen, line 398:
```suggestion
                                        WithOffset {
                                            data: &f.lf.lr_line_buf,
                                            offset: f.lf.lr_lpf_line[pl],
                                        } + line as isize * uv_stride + (bx * 4 >> ss_hor) as isize
```

src/cdef_apply.rs
```diff
@@ -362,66 +376,82 @@ pub(crate) unsafe fn rav1d_cdef_brow<BD: BitDepth>(
...
-                                        );
+                                    let top = WithOffset {
+                                        data: &f.lf.cdef_line_buf,
+                                        offset: f.lf.cdef_line[tf as usize][pl]
+                                            .wrapping_add_signed(
+                                                (sby * 8) as isize * uv_stride
+                                                    + (bx * 4 >> ss_hor) as isize,
+                                            ),
+                                    };
+                                    let (buf, offset) = if resize {
```
kkysen, line 411: Can this be a `WithOffset`?

src/cdef_apply.rs
```diff
@@ -362,66 +376,82 @@ pub(crate) unsafe fn rav1d_cdef_brow<BD: BitDepth>(
...
-                                            (f.lf.cdef_lpf_line[pl] as isize + offset) as usize,
-                                        );
+                                    let top = WithOffset {
+                                        data: &f.lf.cdef_line_buf,
+                                        offset: f.lf.cdef_line[tf as usize][pl]
+                                            .wrapping_add_signed(
+                                                (sby * 8) as isize * uv_stride
+                                                    + (bx * 4 >> ss_hor) as isize,
+                                            ),
+                                    };
```
kkysen, line 410:
```suggestion
                                    let top = WithOffset {
                                        data: &f.lf.cdef_line_buf,
                                        offset: f.lf.cdef_line[tf as usize][pl],
                                    } + (sby * 8) as isize * uv_stride + (bx * 4 >> ss_hor) as isize;
```

src/cdef_apply.rs
```diff
@@ -362,66 +376,82 @@ pub(crate) unsafe fn rav1d_cdef_brow<BD: BitDepth>(
...
+
+                                let (top, bot) = top_bot.unwrap_or_else(|| {
+                                    let top = WithOffset {
+                                        data: &f.lf.cdef_line_buf,
+                                        offset: f.lf.cdef_line[tf as usize][pl]
+                                            .wrapping_add_signed(
+                                                have_tt as isize * (sby * 8) as isize * uv_stride
+                                                    + (bx * 4 >> ss_hor) as isize,
+                                            ),
+                                    };
```
kkysen, line 451:
```suggestion
                                    let top = WithOffset {
                                        data: &f.lf.cdef_line_buf,
                                        offset: f.lf.cdef_line[tf as usize][pl],
                                    } + have_tt as isize * (sby * 8) as isize * uv_stride + (bx * 4 >> ss_hor) as isize;
```

kkysen (changes requested):

kkysen (approved):
My previous comments were just little nitpicks, so I fixed them myself and I'll merge now.  I also squashed one of the commits that was broken with its fix.

PR #1241, opened 2024-06-24
`fn get_lo_ctx`: Optimize

kkysen:
* Part of #1180.

This optimizes `fn get_lo_ctx`.

9d00ae6 optimizes it enough that it is now inlined (it was already `#[inline]`), which enables other optimizations.

a5c6209 makes `stride` a `u8`, which it is in the caller `fn decode_coefs`, and which tells LLVM that `_ * stride` can't overflow, and thus `2 * stride` being in bounds means that `0 * stride` is in bounds, for example.

242065b then pre-indexes `levels` so that there is only one bounds check per branch.  There is still actually another since LLVM does not know that `stride != 0`.  This can be optimized out with more effort in `fn decode_coefs`, as `stride` is ultimately derived from `t_dim.h`, which comes from `static`/`const` data.

### DIFF
--- a/src/levels.rs
+++ b/src/levels.rs
@@ -145,7 +145,7 @@ pub const DCT_ADST: TxfmType = 2;
 pub const ADST_DCT: TxfmType = 1;
 pub const DCT_DCT: TxfmType = 0;
 
-#[derive(Clone, Copy, PartialEq, Eq)]
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub enum TxClass {
     TwoD,
     H,
--- a/src/recon.rs
+++ b/src/recon.rs
@@ -468,28 +468,28 @@ fn get_dc_sign_ctx(tx: TxfmSize, a: &[u8], l: &[u8]) -> c_uint {
 fn get_lo_ctx(
     levels: &[u8],
     tx_class: TxClass,
-    hi_mag: &mut c_uint,
+    hi_mag: &mut u32,
     ctx_offsets: Option<&[[u8; 5]; 5]>,
     x: u32,
     y: u32,
     stride: u8,
 ) -> u8 {
     let stride = stride as usize;
-    let level = |y, x| levels[y * stride + x] as usize;
+    let level = |y, x| levels[y * stride + x] as u32;
 
     let mut mag = level(0, 1) + level(1, 0);
     let offset = match ctx_offsets {
         Some(ctx_offsets) => {
             debug_assert_matches!(tx_class, TxClass::TwoD);
             mag += level(1, 1);
-            *hi_mag = mag as c_uint;
+            *hi_mag = mag;
             mag += level(0, 2) + level(2, 0);
             ctx_offsets[cmp::min(y as usize, 4)][cmp::min(x as usize, 4)]
         }
         None => {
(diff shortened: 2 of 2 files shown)


randomPoison (approved):

fbossen:
I am seeing some strange behaviour here. On `x86_64` the first commit leads to a decrease in the number of instructions and branches, but also to an increase in the number of branch misses (~10%) and cycles (~4%). This appears to be mainly because of the inlining of `fn get_lo_ctx`. Subsequent commits do not overcome this initial performance regression.

kkysen:
> I am seeing some strange behaviour here. On `x86_64` the first commit leads to a decrease in the number of instructions and branches, but also to an increase in the number of branch misses (~10%) and cycles (~4%). This appears to be mainly because of the inlining of `fn get_lo_ctx`. Subsequent commits do not overcome this initial performance regression.

Oof, let me look into that.  I thought inlining would help.  Can you check if you see the same thing by #1244?

fbossen:
> Can you check if you see the same thing by #1244?

#1244 has fewer instructions and branches than this PR, but the number of branch misses and cycles increases further, although not by as much...

fbossen:
Looks like most branch misses happen here:
```
                        if tok != 0 {
                            rc = rc_i;
                        }
```
This is somewhat confusing as the corresponding asm code looks quite similar with and without this PR.

fbossen:
> Looks like most branch misses happen here:
> 
> ```
>                         if tok != 0 {
>                             rc = rc_i;
>                         }
> ```
> 
> This is somewhat confusing as the corresponding asm code looks quite similar with and without this PR.

Looks like computing `tok != 0` before multiplying `tok` by `0x17ff41` takes care of the branch miss issue

kkysen:
That's weird.  That's not compiled to a `cmov`?

fbossen:
> That's weird. That's not compiled to a `cmov`?

No, which as you say is weird

kkysen:
> Looks like computing `tok != 0` before multiplying `tok` by `0x17ff41` takes care of the branch miss issue

I guess this does seem to pass all the tests.  I'm not quite sure yet why this is a valid transform, but the C looks heavily optimized here by the comments left, so if this fixes the branch miss, that's great.

I think the reason a `cmov` isn't emitted is that LLVM is trying to avoid more stack spills.  I tried forcing it with `rc = [rc, rc_i][(tok != 0) as usize]`, which results in a `setne`, but that increases the stack usage in `fn decode_coefs` from `296` to `312`.

fbossen:
> > Looks like computing `tok != 0` before multiplying `tok` by `0x17ff41` takes care of the branch miss issue
> 
> I guess this does seem to pass all the tests. I'm not quite sure yet why this is a valid transform, but the C looks heavily optimized here by the comments left, so if this fixes the branch miss, that's great.

I believe `tok` can only be 0, 1, or 2 when entering this section of the code.

kkysen:
@fbossen, storing the `let tok_non_zero = tok != 0` before the multiplication results in the same extra 16 bytes of stack spills as `rc = [rc, rc_i][(tok != 0) as usize]` does.

fbossen (approved):
Branch issue fixed in #1257

PR #1246, opened 2024-06-25
`fn decode_coefs_class`: Fix branch misprediction

kkysen:
* Part of #1180.

Move the `tok != 0` check before the `*= 0x17ff41` so that `cmov` is used, not a mispredicted branch.

### DIFF
--- a/src/recon.rs
+++ b/src/recon.rs
@@ -929,7 +929,7 @@ fn decode_coefs<BD: BitDepth>(
                 } else {
                     0
                 };
-                tok = (tok >> 9) & (rc as u32).wrapping_add(!0x7ff);
+                tok = (tok >> 9) & (rc as u32 + !0x7ff);
                 debug_assert!(tok == tok_check as u32);
                 debug_assert!(tok_non_zero == (tok != 0));
                 if tok_non_zero {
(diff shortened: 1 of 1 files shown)


kkysen:
How's the perf now here, @fbossen?

My benchmarks are kinda noisy when I `cargo b -r && hyperfine './target/release/dav1d -i ./tests/large/chimera_8b_1080p.ivf -o /dev/null'`.  To me it seems like this recovers some perf (~16.3-16.4 s) vs `main` (~16.0-16.5 s), definitely a lot better than before this PR (~17.0-17.3 s).

randomPoison:
What is `tok_check` here checking? Seems like it's trying to do the same thing as `(tok >> 9) & (rc as u32 + !0x7ff)` since we're checking that it produces the same result?

kkysen:
> What is `tok_check` here checking? Seems like it's trying to do the same thing as `(tok >> 9) & (rc as u32 + !0x7ff)` since we're checking that it produces the same result?

Yeah I just replaced the comment with actual code, just debug only.

randomPoison (approved):

fbossen (approved):
Branch issue is fully addressed in #1257

PR #1251, opened 2024-06-25
`fn output_image`: Make safe

### LINKED ISSUE #1023: `fn output_image`: Make `r#in: *mut Rav1dThreadPicture` var safe
This one's a bit tricky with borrowck stuff.

randomPoison:
* Fixes #1023.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,7 +47,6 @@ use crate::src::obu::rav1d_parse_obus;
 use crate::src::obu::rav1d_parse_sequence_header;
 use crate::src::picture::rav1d_picture_alloc_copy;
 use crate::src::picture::PictureFlags;
-use crate::src::picture::Rav1dThreadPicture;
 use crate::src::thread_task::rav1d_task_delayed_fg;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -379,28 +378,23 @@ impl Rav1dPicture {
...
+    if !c.apply_grain || !r#in.p.has_grain() {
+        *out = mem::take(&mut r#in.p);
     } else {
-        res = rav1d_apply_grain(c, out, &(*r#in).p);
+        res = rav1d_apply_grain(c, out, &r#in.p);
     }
-    let _ = mem::take(&mut *r#in);
+    let _ = mem::take(r#in);
 
     if !c.all_layers && state.max_spatial_id != 0 && state.out.p.data.is_some() {
```
kkysen, line 396: Nit: Could you pull out this expr, `!c.all_layers && state.max_spatial_id != 0` into a variable like `use_cache` and then also use `use_cache` where we initialize `r#in` above?

kkysen (approved):
Just one small nit about pulling out a variable to make things a bit clearer.

PR #1257, opened 2024-06-27
`fn decode_coefs_class`: Replace simple branch w/ branchless logic

kkysen:
Although the previous code would usually emit branchless `cmov`s, this was not guaranteed and a branch was emitted in one the 16bpc `TxClass`es, causing mispredictions and slowdowns.  This changes the logic to use only bit operations.

This effectively replaces a `test` and `cmov` with a `xor`, `and`, `xor`, which should be similarly efficient.  In the branched case, it is far more performant.

During benchmarking, @fbossen found that this helps on x86 and doesn't hurt on arm.

### DIFF
--- a/src/recon.rs
+++ b/src/recon.rs
@@ -947,25 +947,23 @@ fn decode_coefs<BD: BitDepth>(
                 rc = rc_i;
             } else {
                 // `0x1` for `tok`, `0x7ff` as bitmask for `rc`, `0x41` for `level_tok`.
-
-                // If we do this after the `tok *= 0x17ff41`,
-                // it uses a mispredicted branch instead of `cmov`.
-                let tok_non_zero = tok != 0;
-
-                let mut tok = tok as u32;
-                tok *= 0x17ff41;
+                let tok = tok as u32 * 0x17ff41;
                 level[0] = tok as u8;
+
                 let tok_check = if tok != 0 {
                     ((tok as u16) << 11) | rc
                 } else {
                     0
                 };
-                tok = (tok >> 9) & (rc as u32 + !0x7ff);
+
+                // This is optimized differently from C to avoid branches,
+                // as simple branches are not always optimized to branchless `cmov`s.
+                let mask = tok >> 9;
+                let tok = mask & (rc as u32 + !0x7ff);
+                let mask = mask as u16;
+                rc = (rc_i & mask) | (rc & !mask);
+
                 debug_assert!(tok == tok_check as u32);
-                debug_assert!(tok_non_zero == (tok != 0));
-                if tok_non_zero {
-                    rc = rc_i;
-                }
                 cf.set(rc_i, tok);
             }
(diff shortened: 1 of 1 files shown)


fbossen (approved):
This definitely helps on x86 and doesn't appear to hurt on arm.

thedataking:
Are we sure we need this?
```
                let tok_check = if tok != 0 {
                    ((tok as u16) << 11) | rc
                } else {
                    0
                };
...
                debug_assert!(tok == tok_check as u32);
```

It is unclear to me what this accomplishes. More importantly, what happens when rustc changes or LLVM changes?

Along the same lines, this seems to assume a certain compiler or processor. Dav1d supports x86, ARM, RISC-V, longsoon, PPC.
 
 `// as simple branches are not always optimized to branchless `cmov`s`

Specializing for one compiler and ISA seems very brittle. I'm not sure what is a better way to proceed though. One step would be to reflect in the commit message what compiler, CPU, OS you're seeing results on so we can rip this out if/when this breaks.

kkysen:
> Are we sure we need this?
> 
> ```rust
>                 let tok_check = if tok != 0 {
>                     ((tok as u16) << 11) | rc
>                 } else {
>                     0
>                 };
> ...
>                 debug_assert!(tok == tok_check as u32);
> ```

We don't really need this.  It is there as a comment, and was a comment in the C.  I thought it would be better as a debug check, though, and since it's a `debug_assert!`, it's entirely optimized out in release builds.  So I think it's fine.

> It is unclear to me what this accomplishes. More importantly, what happens when rustc changes or LLVM changes?
> 
> Along the same lines, this seems to assume a certain compiler or processor. Dav1d supports x86, ARM, RISC-V, longsoon, PPC.
> 
> `// as simple branches are not always optimized to branchless `cmov`s`
> 
> Specializing for one compiler and ISA seems very brittle. I'm not sure what is a better way to proceed though. One step would be to reflect in the commit message what compiler, CPU, OS you're seeing results on so we have a way to reproduce the optimization and detect if it breaks.

The new code I would think is a lot more architecture-independent.  The original, naive code is the one that is `debug_assert!`-checked.  `dav1d` then optimized it further.  This PR then changes the optimization to not use any branches in the source code, just bitwise operations.  It replaces the `if condition { rc = rc_i; }` with a `xor`, `and`, and `xor`.  I guess it is possible that a certain LLVM backend could optimize the `xor`, `and`, `xor` back into a branch, but I think this is quite unlikely, and definitely more unlikely than it emitting branches for the code before this PR.

As for actual performance results being dependent on the CPU and arch, I added @fbossen's comment saying this helps on x86 and doesn't hurt on arm to the PR description.  I can't be certain how it will perform on other CPUs, but in general, bitwise operations are some of the fastest instructions, including on older CPUs IIUC (in contrast to branches and `cmov`s), so I would think it would be fine.

PR #1258, opened 2024-06-28
`enum CfSelect`: Optimize some checks and make offsets `u32`s

kkysen:
* Part of #1180.

I forgot to open this PR earlier.  This optimizes some `CfSelect` uses and makes its offset, as well as the offsets in `Rav1dTileState_frame_thread` `u32`s instead of `usize`s, saving space, prohibiting overflows on multiplications, and reducing stack usage a bit.


src/recon.rs
```diff
@@ -1562,7 +1551,9 @@ pub(crate) fn rav1d_read_coef_blocks<BD: BitDepth>(
                             let a_len = t_dim.w as usize;
                             let l_start = by4 + y as usize;
                             let l_len = t_dim.h as usize;
-                            let cf_idx = ts.frame_thread[1].cf.get();
+                            let cf = CfSelect::Frame(ts.frame_thread[1].cf.get_update(|i| {
+                                i + cmp::min(t_dim.w, 8) as u32 * cmp::min(t_dim.h, 8) as u32 * 16
+                            }));
```
rinon, line 1556: This changes the ordering of the set, why is that ok?

src/recon.rs
```diff
@@ -1236,24 +1231,19 @@ fn decode_coefs<BD: BitDepth>(
...
 impl CfSelect {
-    fn set<BD: BitDepth>(
-        self,
-        f: &Rav1dFrameData,
-        task_cf: &mut Cf,
-        index: usize,
-        value: BD::Coef,
-    ) {
+    fn set<BD: BitDepth>(self, f: &Rav1dFrameData, task_cf: &mut Cf, index: u16, value: BD::Coef) {
+        let index = index as usize % 1024; // Mask in-bounds.
```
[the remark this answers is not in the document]
rinon: Ah, should this be taking `index` as a `Scan`? It's not clear where this came from.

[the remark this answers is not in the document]
kkysen: In one branch it comes from a `Scan`, but in other it doesn't.

The `% 1024`/`& 1023` replaces the `CfSelect::Task` array bounds check.

src/recon.rs
kkysen: This is the pattern used in most other places in C, and when I moved it up, all of the tests still passed so I assumed it was okay.

rinon: Ok looking through I also don't see any uses of `cf` in the `decode_coefs` call stack. Maybe put a comment in that we changed this and it should not be accessed inside `decode_coefs`.

kkysen: Comment added in [`7065142..6155d6b`](https://github.com/memorysafety/rav1d/compare/706514243a4f422029fa4798e1177c26572cf0b7..6155d6b0915ab8d4cef5692f20e5fbf95b47a645).

rinon (approved):

PR #1259, opened 2024-06-28
Add `release-with-debug` profile and make `test_rust` default false

rinon:


meson_options.txt
```diff
@@ -66,7 +66,7 @@ option('trim_dsp',
 
 option('test_rust',
     type: 'boolean',
-    value: true,
+    value: false,
```
kkysen, line 69: Does this change how we should be running `test.sh` locally?

rinon: You're passing it the path to the Rust binaries explicitly, right? If so, this shouldn't change anything. It just was defaulting to building the C CLI tools against librav1d.a instead of libdav1d.so.

kkysen: > You're passing it the path to the Rust binaries explicitly, right?

Yeah.

kkysen (approved):

thedataking (approved):

PR #1263, opened 2024-06-28
`struct Rav1dTileStateContext`: Extract locking higher in the call stack

rinon:


src/decode.rs
```diff
@@ -5215,10 +5235,7 @@ pub fn rav1d_submit_frame(c: &Rav1dContext, state: &mut Rav1dState) -> Rav1dResu
         }
 
         f.cur_segmap = Some(
-            match (
-                frame_hdr.segmentation.update_map != 0,
-                f.prev_segmap.as_mut(),
-            ) {
+            match (frame_hdr.segmentation.update_map != 0, f.prev_segmap.as_mut()) {
```
kkysen, line 5238: What's the diff here?

rinon: Formatting mistake

src/decode.rs
```diff
@@ -3462,21 +3466,21 @@ fn decode_sb(
     let ctx = if t.frame_thread.pass == 2 {
         None
     } else {
+        // Available if t.frame_thread.pass != 2
+        let ts_c = &mut **ts_c.as_mut().unwrap();
```
kkysen, line 3470: These two checks can be combined.

kkysen: We could use a `ts_c.as_mut().map(|ts_c| {})`.

src/decode.rs
```diff
@@ -1317,7 +1318,8 @@ fn decode_b(
         return Ok(());
     }
 
-    let ts_c = &mut *ts.context.try_lock().unwrap();
+    // Available if t.frame_thread.pass != 2
+    let ts_c = &mut **ts_c.as_mut().unwrap();
```
kkysen, line 1322: This can be combined with the `if t.frame_thread.pass == 2 {` that is for the block ending in the `return Ok(())`.

kkysen (changes requested):
Both of the `.unwrap()`s here can be folded into the `f t.frame_thread.pass != 2` checks.

Also, did you find that `TryLock` is more efficient than `parking_lot::Mutex` here?

rinon:
These _could_ be combined, but I don't see a clear and easy way to do that without changing a lot. The branch isn't an issue here, taking the lock was.

`TryLock` seems to be about the same as `parking_lot::Mutex::try_lock`, so I'm not sure this is actually worthwhile. I'll benchmark this on more systems and see, but I suspect it makes no difference and we should stick with parking lot, just to simplify dependencies.

kkysen:
> These _could_ be combined, but I don't see a clear and easy way to do that without changing a lot. The branch isn't an issue here, taking the lock was.

It's a pretty simple change, though:

```diff
diff --git a/src/decode.rs b/src/decode.rs
index ac71faac6..445c6e880 100644
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -1188,7 +1188,8 @@ fn decode_b(
         && (bh4 > ss_ver || t.b.y & 1 != 0);
     let frame_type = f.frame_hdr.as_ref().unwrap().frame_type;

-    if t.frame_thread.pass == 2 {
+    let Some(ts_c) = ts_c else {
+        debug_assert!(t.frame_thread.pass == 2);
         match &b.ii {
             Av1BlockIntraInter::Intra(intra) => {
                 (bd_fn.recon_b_intra)(f, t, None, bs, intra_edge_flags, b, intra);
@@ -1316,10 +1317,8 @@ fn decode_b(
         }

         return Ok(());
-    }
-
-    // Available if t.frame_thread.pass != 2
-    let ts_c = &mut **ts_c.as_mut().unwrap();
+    };
+    let ts_c = &mut **ts_c;

     let cw4 = w4 + ss_hor >> ss_hor;
     let ch4 = h4 + ss_ver >> ss_ver;
@@ -3463,11 +3462,9 @@ fn decode_sb(
     let bp;
     let mut bx8 = 0;
     let mut by8 = 0;
-    let ctx = if t.frame_thread.pass == 2 {
-        None
-    } else {
-        // Available if t.frame_thread.pass != 2
-        let ts_c = &mut **ts_c.as_mut().unwrap();
+    let ctx = ts_c.as_mut().map(|ts_c| {
+        debug_assert!(t.frame_thread.pass != 2);
+        let ts_c = &mut **ts_c;
         if false && bl == BlockLevel::Bl64x64 {
             println!(
                 "poc={},y={},x={},bl={:?},r={}",
@@ -3476,8 +3473,8 @@ fn decode_sb(
         }
         bx8 = (t.b.x & 31) >> 1;
         by8 = (t.b.y & 31) >> 1;
-        Some((get_partition_ctx(&f.a[t.a], &t.l, bl, by8, bx8), ts_c))
-    };
+        (get_partition_ctx(&f.a[t.a], &t.l, bl, by8, bx8), ts_c)
+    });

     if have_h_split && have_v_split {
         if let Some((ctx, ts_c)) = ctx {
```

> `TryLock` seems to be about the same as `parking_lot::Mutex::try_lock`, so I'm not sure this is actually worthwhile. I'll benchmark this on more systems and see, but I suspect it makes no difference and we should stick with parking lot, just to simplify dependencies.

Yeah, that's what I thought you had said earlier.

rinon:
Ok, yes if you use this weird new lock param to replace the pass check, however that loses the semantic information about WHY we are taking this code path. It's not because we have the lock, it's because of a logical distinction in passes. The lock is a side effect of that. I didn't really like that, which is why I wasn't going with it. My fancier idea was to pass a new enum instead of option that carries both the pass number (as the variant) and the lock, if available, but that's a pain, so I didn't. At the end of the day, I'm not sure it matters.

kkysen:
> Ok, yes if you use this weird new lock param to replace the pass check, however that loses the semantic information about WHY we are taking this code path. It's not because we have the lock, it's because of a logical distinction in passes. The lock is a side effect of that. I didn't really like that, which is why I wasn't going with it. My fancier idea was to pass a new enum instead of option that carries both the pass number (as the variant) and the lock, if available, but that's a pain, so I didn't. At the end of the day, I'm not sure it matters.

I think it's clear enough with the `debug_assert!`s, though.

src/decode.rs
```diff
@@ -4224,7 +4252,14 @@ pub(crate) fn rav1d_decode_tile_sbrow(
                 read_restoration_info(ts, &mut lr, p, frame_type, debug_block_info!(f, t.b));
             }
         }
-        decode_sb(c, t, f, root_bl, EdgeIndex::root())?;
+        decode_sb(
+            c,
+            t,
+            f,
+            &mut &mut FrameThreadPassState::First(&mut f.ts[t.ts].context.try_lock().unwrap()),
```
kkysen, line 4259: Does this need two `&mut`s?

kkysen (approved):
LGTM but I'm not sure if those two consecutive `&mut`s are needed.  If they are, then it's fine.

src/decode.rs
rinon: Nope, typo, thanks for catching that.

PR #1264, opened 2024-06-28
Simplify get_lo_ctx

rinon:
get_lo_ctx being inlined is bad for performance, and this change makes it small enough to inline with the hint, so this change also removes the `#[inline]` hint.

### DIFF
--- a/src/recon.rs
+++ b/src/recon.rs
@@ -463,10 +463,8 @@ fn get_dc_sign_ctx(tx: TxfmSize, a: &[u8], l: &[u8]) -> c_uint {
     (s != 0) as c_uint + (s > 0) as c_uint
 }
 
-#[inline]
 fn get_lo_ctx(
     levels: &[u8],
-    tx_class: TxClass,
     hi_mag: &mut c_uint,
     ctx_offsets: Option<&[[u8; 5]; 5]>,
     x: usize,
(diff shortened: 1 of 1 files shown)


src/recon.rs
```diff
@@ -476,19 +474,16 @@ fn get_lo_ctx(
...
-            *hi_mag = mag as c_uint;
-            mag += level(0, 2) + level(2, 0);
-            ctx_offsets.unwrap()[cmp::min(y, 4)][cmp::min(x, 4)] as usize
-        }
-        TxClass::H | TxClass::V => {
-            mag += level(0, 2);
-            *hi_mag = mag as c_uint;
-            mag += level(0, 3) + level(0, 4);
-            26 + if y > 1 { 10 } else { y * 5 }
-        }
```
kkysen, line 491: I already did this optimization in #1241.  As for the `#[inline]`, could you check if it's still bad for perf on top of #1257?

kkysen (changes requested):

src/recon.rs
rinon: It looks like even on top of main, inlining is now ok here. I'll close this PR.

kkysen: Okay, that's good.  @fbossen and I spent a while optimizing it.

PR #1271, opened 2024-06-29
`fn filter_plane_cols_{y,uv}`: avoid division when eliding bounds check

fbossen:
Keep size of `mask` as 32 instead of shortening to `w` such that `mask.len()` is a known power of 2 and `% mask.len()` is easily computable using an `and` operation.


kkysen (commented):
Oh, I may have messed up here.  If `mask.len() == w`, then the `% mask.len()` should not be needed at all.  Does that work also?  That would be preferable I think.

fbossen:
> Oh, I may have messed up here. If `mask.len() == w`, then the `% mask.len()` should not be needed at all. Does that work also? That would be preferable I think.

That works too. See #1276.

fbossen:
Superseded by #1276

PR #1272, opened 2024-06-29
Remove inlining of `read_vartx_tree` and `rav1d_create_lf_mask_intra`

rinon:
These functions are not inlined in C and removing `read_vartx_tree` inlining speeds up performance on Chimera 8-bit on my 7700X by about 0.7%.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -781,6 +781,8 @@ struct VarTx {
     tx_split1: u16,
 }
 
+// not inlined in C and inlining in Rust degrades performance slightly
+#[inline(never)]
 fn read_vartx_tree(
     t: &mut Rav1dTaskContext,
     f: &Rav1dFrameData,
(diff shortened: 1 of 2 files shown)


kkysen (commented):
Could you comment the inline nevers saying they're for perf?  Some of them are for binary size I think, so it's helpful to know generally which are which.

kkysen:
Are these functions inlined in C?  I wonder if inlining being bad here indicates that we're missing optimizations inside of these functions.

rinon:
They are not inlined by Clang which is why I tried un-inlining them.

kkysen:
> They are not inlined by Clang which is why I tried un-inlining them.

Ahh, they're not inlined in C.  That's good then.

randomPoison (approved):
Just the last two commits are supposed to be part of this, and it just needs a rebasing right?

Issue #1177, opened 2024-06-05
Remove `unsafe impl {Send,Sync} for TaskThreadData_delayed_fg`


kkysen:
These were `TODO`s and are unjustified/unsound.

PR #1277, opened 2024-07-01
`struct ITUTT35PayloadPtr`: Newtype `Dav1dITUTT35::payload` to `unsafe impl Send + Sync` on it

kkysen:
* Part of #1177.

This makes `Dav1dITUTT35: Send + Sync`.

### DIFF
--- a/include/dav1d/headers.rs
+++ b/include/dav1d/headers.rs
@@ -774,18 +774,26 @@ pub struct Rav1dMasteringDisplay {
 
 pub type Dav1dMasteringDisplay = Rav1dMasteringDisplay;
 
+/// An immutable ptr to [`Rav1dITUTT35::payload`].
+///
+/// [`Rav1dITUTT35::payload`] is a [`Box`], so it doesn't move,
+/// and [`Self::payload`]'s lifetime is that of the [`Rav1dITUTT35`],
+/// which is itself stored in a [`Box`] as returned from [`Rav1dITUTT35::to_immut`].
+#[repr(transparent)]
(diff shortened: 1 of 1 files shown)


randomPoison (approved):

rinon:
Why do we need Send and Sync for a C interface type? I don't think we should impl anything on this type, we should implement it for the container that contains one of these.

rinon (changes requested):
It looks like this is used in a `DRav1d`? Maybe we should impl Send and Sync if the safe rav1d side of the `DRav1d` impls them.

kkysen:
> It looks like this is used in a `DRav1d`? Maybe we should impl Send and Sync if the safe rav1d side of the `DRav1d` impls them.

It'd be unsound to do this since there are not sufficient constraints on the generics in `DRav1d<R, D>` to guarantee such an `unsafe impl` would be sound.

rinon:
How so exactly? We should never access and use the C type from Rust, right? So we can't enforce anything about it. I'm basically thinking of the `D` part of that structure as opaque from Rust's perspective, is that not the case?

kkysen:
> How so exactly? We should never access and use the C type from Rust, right? So we can't enforce anything about it. I'm basically thinking of the `D` part of that structure as opaque from Rust's perspective, is that not the case?

The whole type is public, so C callers can do whatever they want with the `D` type.  I think it makes sense to limit the safety constraints on it to one particular type rather than a blanket impl that can hide a ton of stuff on the many other types that don't need it.

PR #1278, opened 2024-07-01
`struct StableRef`: Refactor out from `CArc::{,base_}stable_ref` and `unsafe impl Send + Sync` on it

kkysen:
* Part of #1177.

This helps make `CArc<T: Send + Sync>: Send + Sync` (still need to fix `CBox` in the next PR).

### DIFF
--- a/src/c_arc.rs
+++ b/src/c_arc.rs
@@ -42,10 +42,32 @@ pub struct CArc<T: ?Sized> {
 
     /// The same as [`Self::stable_ref`] but it never changes.
     #[cfg(debug_assertions)]
-    base_stable_ref: NonNull<T>,
+    base_stable_ref: StableRef<T>,
 
-    stable_ref: NonNull<T>,
+    stable_ref: StableRef<T>,
 }
+
+/// A stable reference, stored as a raw ptr.
+///
+/// It is stable (e.x. from a [`Box`] or a [`CBox`]),
+/// so it can be stored relative to its owner.
+/// Otherwise, it is essentially a `&T` (so immutable).
+#[derive(Debug)]
+struct StableRef<T: ?Sized>(NonNull<T>);
+
+impl<T: ?Sized> Clone for StableRef<T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<T: ?Sized> Copy for StableRef<T> {}
+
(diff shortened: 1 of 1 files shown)


randomPoison (approved):

rinon (changes requested):
Why can't you just impl `Send+Sync` for `CArc` if `T` implements them?

src/c_arc.rs
```diff
@@ -85,7 +108,7 @@ impl<T: ?Sized> AsRef<T> for CArc<T> {
         // The [`CBox`] is [`Pin`]ned and
         // [`Self::stable_Ref`] is always updated on writes to [`Self::owner`],
```
kkysen, line 109:
```suggestion
        // [`Self::stable_ref`] is always updated on writes to [`Self::owner`],
```

src/c_arc.rs
```diff
@@ -42,11 +42,33 @@ pub struct CArc<T: ?Sized> {
...
-    base_stable_ref: NonNull<T>,
+    base_stable_ref: StableRef<T>,
 
-    stable_ref: NonNull<T>,
+    stable_ref: StableRef<T>,
 }
 
+/// A stable reference, stored as a raw ptr.
+///
+/// It is stable (e.x. from a [`Box`] or a [`CBox`]),
```
rinon, line 52: Shouldn't this be a `Pin<Box<T>`? We don't want the object to move out of its allocation.

Can we put a SAFETY comment here that any pointer stored here MUST not move for its lifetime?

src/c_arc.rs
```diff
@@ -42,11 +42,33 @@ pub struct CArc<T: ?Sized> {
...
+
+impl<T: ?Sized> Clone for StableRef<T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<T: ?Sized> Copy for StableRef<T> {}
+
+/// SAFETY: [`StableRef`]`<T>` is essentially `&T`, which is [`Send`] if `T: `[`Send`]`.
```
rinon, line 66: This is only valid if the invariants on what is put into the ref is held, we should at least mention that here.

rinon (commented):
We talked offline about this, it's ok, just a few small doc comments now.

PR #1283, opened 2024-07-03
`struct CaseSet`: Optimize by `match`ing on `len` once per `set_ctx`

kkysen:
In C, `case_set` works by `switch`ing once on `len`, once per `set_ctx`, but my `CaseSet` implementation in Rust accidentally switched to `match`ing on `buf.len()` each time. This goes back to the original, optimal behavior.

To do this, `set_ctx` is made a rank-2 polymorphic "closure". This does not exist in Rust, so it's emulated through a generic trait with a generic method. This inner generic over `trait CaseSetter` is what allows `fn CaseSet::one` to select the correct `CaseSetterN` at compile time.

However, this means that closures can't be used anymore, which is very annoying. To partially remedy this, I added the `set_ctx!` macro, which emulates the `set_ctx` closure as much as possible. All captures (up vars in `rustc`) and their types must be declared.

I didn't actually benchmark this or look at the asm yet, though, if anyone wants to do that.

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -52,38 +52,55 @@ use std::iter::zip;
 /// This optimizes for the common cases where `buf.len()` is a small power of 2,
 /// where the array write is optimized as few and large stores as possible.
 #[inline]
-pub fn small_memset<T: Clone + Copy, const UP_TO: usize, const WITH_DEFAULT: bool>(
+pub fn small_memset<T: Clone + Copy, const N: usize, const WITH_DEFAULT: bool>(
     buf: &mut [T],
     val: T,
 ) {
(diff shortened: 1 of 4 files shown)


rinon:
I'm working on validating performance for this. It's not making as much of a difference as expected, and I think some checks are not getting elided.

kkysen:
> I'm working on validating performance for this. It's not making as much of a difference as expected, and I think some checks are not getting elided.

😭 I didn't get a chance to look at the asm yet, though.  Let me know if there's anything simple we can do there.  I can look a bit too to see which things aren't elided.

rinon:
I have a very different way of doing this that is a lot less complicated, I think, so don't spend time on it for now.

kkysen:
> I have a very different way of doing this that is a lot less complicated, I think, so don't spend time on it for now.

Oh interesting, hope it works!

PR #1286, opened 2024-07-04
`mod looprestoration::neon{,_erased}`: Move `neon` `fn`s to their own `mod`

kkysen:
I wanted to split this out into its own PR since it's diff heavy.

### DIFF
--- a/src/looprestoration.rs
+++ b/src/looprestoration.rs
@@ -26,15 +26,6 @@ use zerocopy::AsBytes;
 use zerocopy::FromBytes;
 use zerocopy::FromZeroes;
 
-#[cfg(all(feature = "asm", any(target_arch = "arm", target_arch = "aarch64")))]
-use ::{std::ffi::c_void, std::ptr};
-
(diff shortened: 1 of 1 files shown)


src/looprestoration.rs
```diff
@@ -3300,199 +3135,362 @@ unsafe fn sgr_filter_mix_neon<BD: BitDepth>(
...
+        let h = h as usize;
+        // SAFETY: Length sliced in `loop_restoration_filter::Fn::call`.
+        let left = unsafe { slice::from_raw_parts(left, h) };
+        neon::sgr_filter_5x5_neon(p, left, lpf, w, h, params, edges, bd)
+    }
+
+    /// # Safety
+    ///
+    /// Must be called by [`loop_restoration_filter::Fn::call`].
+    #[cfg(all(feature = "asm", any(target_arch = "arm", target_arch = "aarch64")))]
```
randomPoison, line 3441: This cfg is redundant I think.

randomPoison (approved):

src/looprestoration.rs
kkysen: Good catch

PR #1289, opened 2024-07-05
`Cargo.toml`: Add `profile.checked-release`

kkysen:
Add a `--profile checked-release`, which is mostly `release`, but with all checks enabled (e.x. `debug-assertions` and `overflow-checks`).  This doesn't need to absolute highest performance, so `opt-level = 2` is done instead of `opt-level = 3`, and `codegen-units = 16` is used as well (the normal `release` default, overridden by us to `codegen-units = 1`).

This is especially helpful when running tests locally under qemu, which can be very slow if running the `opt-dev` profile.  This could also be used in CI instead of `opt-dev`, as `opt-level = 2` should help those run quite a bit faster while still having all checks enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,3 +62,9 @@ panic = "abort"
 [profile.release-with-debug]
 inherits = "release"
 debug = "line-tables-only"
+
+[profile.checked-release]
+# Close to a release build, but with all checks enabled.
+inherits = "dev"
+opt-level = 2
+codegen-units = 16


randomPoison:
I feel like it'd make more sense to use `opt-level = 3` here since the point is for it to be performant. What's the advantage of using `opt-level = 2` here?

kkysen:
> I feel like it'd make more sense to use `opt-level = 3` here since the point is for it to be performant. What's the advantage of using `opt-level = 2` here?

2 and 3 generally have near identical performance.  3 is more often a little faster, but sometimes 2 is faster.  But 3 usually takes far longer to compile.

randomPoison (approved):

PR #1292, opened 2024-07-06
[WIP] case_set implementation comparison

rinon:
This PR compares the performance of our latest CaseSet on main, the implementation from #1283, and a new macro implementation. The implementation in #1283 is slightly faster than main for multiple case sets in one dispatch, but significantly slower for set_disjoints. The new macro implementation is faster across the board, but does contain an unsafe block to hoist bounds checking outside of the match.

### DIFF
--- a/src/ctx_macro.rs
+++ b/src/ctx_macro.rs
@@ -1,0 +1,239 @@
+//! The [`case_set!`] macro is a safe and simplified version of the `case_set*` macros in `ctx.h`.
+//!
+//! The `case_set*` macros themselves replaced `memset`s in order to further optimize them
+//! (in e3b5d4d044506f9e0e95e79b3de42fd94386cc61,
+//! which performed the following optimizations:
+//! 1. larger, inlinable writes for small power of 2 lengths
+//! 2. aligned writes for the above
+//! 3. 8-byte aligned fields [`BlockContext`]
+//!   * allows 8-byte aligned writes
+//!   * better cache boundary alignment
+//!
+//! (1) is easy to preserve in Rust, but (2) is difficult to do so without overhead,
+//! as unaligned writes are UB, and so we'd need to check at runtime if they're aligned
+//! (a runtime-determined `off`set is used, so we can't reasonably ensure this at compile-time).
+//!
+//! 
+//! TODO: benchmark new implementation against old one?
+//! 
+//! To more thoroughly check this, I ran the same benchmarks done in
+//! e3b5d4d044506f9e0e95e79b3de42fd94386cc61, which introduced the `case_set*` macros:
+//!
+//! ```sh
+//! cargo build --release && hyperfine './target/release/dav1d -i ./tests/large/chimera_8b_1080p.ivf -l 1000 -o /dev/null'
+//! ```
+//!
+//! for 3 implementations:
+//! 1. the original `case_set*` macros translated directly to `unsafe` Rust `fn`s
+//! 2. the safe [`CaseSet`] implementation below using [`small_memset`] with its small powers of 2 optimization
+//! 3. a safe [`CaseSet`] implementation using [`slice::fill`]/`memset` only
+//!
+//! The [`small_memset`] version was ~1.27% faster than the `case_set*` one,
+//! and ~3.26% faster than the `memset` one.
+//! The `case_set*` macros were also faster than `memset` in C by a similar margin,
+//! meaning the `memset` option is the slowest in both C and Rust,
+//! and since it was replaced with `case_set*` in C, we shouldn't use it in Rust.
+//! Thus, the [`small_memset`] implementation seems optimal, as it:
+//! * is the fastest of the Rust implementations
+//! * is completely safe
+//! * employs the same small powers of 2 optimization the `case_set*` implementation did
+//! * is far simpler than the `case_set*` implementation, consisting of a `match` and array writes
+//!
+//! [`BlockContext`]: crate::src::env::BlockContext
+
+/// Fill small ranges of buffers with a value.
+/// 
+/// This is effectively a specialized version [`slice::fill`] for small
+/// power-of-two sized ranges of buffers.
+/// 
+/// `$UP_TO` is the maximum length that will be optimized, with powers of two up
+/// to 64 supported.
+/// 
+/// If `$WITH_DEFAULT` is `true`, then the implementation falls back to
+/// [`slice::fill`] for buffer lengths not a power of two or greater than
+/// `$UP_TO`. Otherwise no operation is performed in these cases.
+/// 
+/// # Examples
+/// 
+/// ```
+/// # use rav1d::case_set;
+/// let mut buf = [0u8; 32];
+/// let len = 16;
+/// for offset in [0, 16] {
+///     case_set!(32, len, offset, {
+///         set!(&mut buf, 1u8);
+///     });
+/// }
+/// ```
+/// 
+/// In the simplest case, `$len` is the length of the buffer range to fill
+/// starting from `$offset`. The `$body` block is executed with `len` and
+/// `offset` identifiers set to the given length and offset values. Within the
+/// body a `set!` macro is available and must be called to set each buffer range
+/// to a value. `set!` takes a buffer and a value and sets the range
+/// `buf[offset..][..len]` to the value.
+/// ```
+/// # macro_rules! set {
+/// #     ($buf:expr, $val:expr) => {};
+/// # }
+/// set!(buf, value);
+/// ```
+/// 
+/// ## Naming parameters
+/// 
+/// The identifier for either or both of `len` and `offset` can be overridden by
+/// specifying `identifer=value` for those parameters:
+/// ```
+/// # use rav1d::case_set;
+/// let mut buf = [0u8; 32];
+/// let outer_len = 16;
+/// for outer_offset in [0, 16] {
+///     case_set!(32, len=outer_len, offset=outer_offset, {
+///         set!(&mut buf, (offset+len) as u8);
+///     });
+/// }
+/// ```
+/// 
+/// ## `DisjointMut` buffers
+/// 
+/// [`DisjointMut`] buffers can be used in basically the same way as normal
+/// buffers but using the `set_disjoint!` macro instead of `set!`.
+/// ```
+/// # use rav1d::case_set;
+/// # use rav1d::src::disjoint_mut::DisjointMut;
+/// let mut buf = DisjointMut::new([0u8; 32]);
+/// let len = 16;
+/// for offset in [0, 16] {
+///     case_set!(32, len, offset, {
+///         set_disjoint!(&mut buf, 1u8);
+///     });
+/// }
+/// ```
+/// 
+/// ## Multiple buffer ranges
+/// 
+/// Multiple buffers with different lengths and offsets can be filled with the
+/// same body statements. In the following example, two buffers with different
+/// sizes are initialized by quarters.
+/// ```
+/// # use rav1d::case_set;
+/// let mut buf1 = [0u8; 32];
+/// let mut buf2 = [0u8; 64];
+/// for offset in [0, 8, 16, 24] {
+///     case_set!(16, buf=[&mut buf1[..], &mut buf2[..]], len=[8, 16], offset=[offset, offset*2], {
+///         set!(buf, len as u8 >> 3);
+///     });
+/// }
+/// ```
+/// 
+/// A more realistic example of filling multiple buffers with the same value is
+/// initializing different struct fields at the same time (from
+/// `src/decode.rs`):
+/// ```ignore
+/// case_set!(32, ctx=[(&t.l, 1), (&f.a[t.a], 0)], len=[bh4, bw4], offset=[by4, bx4], {
+///      let (dir, dir_index) = ctx;
+///      set_disjoint!(dir.seg_pred, seg_pred.into());
+///      set_disjoint!(dir.skip_mode, b.skip_mode);
+///      set_disjoint!(dir.intra, 0);
+///      set_disjoint!(dir.skip, b.skip);
+///      set_disjoint!(dir.pal_sz, 0);
+/// });
+/// ```
+/// 
+/// [`DisjointMut`]: crate::src::disjoint_mut::DisjointMut
+#[macro_export]
+macro_rules! case_set {
+    ($UP_TO:literal, $($WITH_DEFAULT:literal,)? $ctx:ident=[$($ctx_expr:expr),*], $len:ident=[$($len_expr:expr),*], $offset:ident=[$($offset_expr:expr),*], $body:block) => {
+        for ($ctx, ($len, $offset)) in core::iter::zip([$($ctx_expr,)*], core::iter::zip([$($len_expr,)*], [$($offset_expr,)*])) {
+            case_set!($UP_TO, $($WITH_DEFAULT,)? $ctx=$ctx, $len=$len, $offset=$offset, $body);
+        }
+    };
+    ($UP_TO:literal, $($WITH_DEFAULT:literal,)? $len:ident, $offset:ident, $body:block) => {
+        case_set!($UP_TO, $($WITH_DEFAULT,)? ctx=(), $len=$len, $offset=$offset, $body);
+    };
+    ($UP_TO:literal, $($WITH_DEFAULT:literal,)? $len:ident=$len_expr:expr, $offset:ident, $body:block) => {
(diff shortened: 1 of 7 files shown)


rinon:
Measurements from my 7700X (i7-1260p was similar):

Rank2 (#1283):
```
     Running benches/case_set.rs (target/release/deps/case_set-e476f33a0d7986a5)
case_set_one/len=1, offset=0
                        time:   [2.2400 ns 2.2424 ns 2.2449 ns]
                        change: [-1.8507% -0.6324% +0.6952%] (p = 0.38 > 0.05)
                        No change in performance detected.
Found 33 outliers among 100 measurements (33.00%)
  13 (13.00%) low severe
  2 (2.00%) low mild
  1 (1.00%) high mild
  17 (17.00%) high severe
case_set_one/len=8, offset=0
                        time:   [2.2565 ns 2.2578 ns 2.2591 ns]
                        change: [-0.3228% -0.2908% -0.2588%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 25 outliers among 100 measurements (25.00%)
  10 (10.00%) low severe
  3 (3.00%) low mild
  7 (7.00%) high mild
  5 (5.00%) high severe
case_set_one/len=32, offset=0
                        time:   [2.4410 ns 2.4423 ns 2.4435 ns]
                        change: [-0.3766% -0.3293% -0.2802%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

case_set_multiple/len=1, offset=0
                        time:   [4.7789 ns 4.7819 ns 4.7847 ns]
                        change: [-1.7647% -1.6945% -1.6316%] (p = 0.00 < 0.05)
                        Performance has improved.
case_set_multiple/len=8, offset=0
                        time:   [4.7634 ns 4.7662 ns 4.7694 ns]
                        change: [-1.3016% -1.2461% -1.1913%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
case_set_multiple/len=32, offset=0
                        time:   [3.4152 ns 3.4168 ns 3.4186 ns]
                        change: [-6.0929% -5.8711% -5.7244%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  1 (1.00%) high severe

case_set_disjoint_multiple/len=1, offset=0
                        time:   [4.1365 ns 4.1375 ns 4.1389 ns]
                        change: [+19.292% +19.412% +19.532%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
case_set_disjoint_multiple/len=8, offset=0
                        time:   [3.3176 ns 3.3212 ns 3.3248 ns]
                        change: [+6.8295% +6.9520% +7.0832%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 10 outliers among 100 measurements (10.00%)
  10 (10.00%) high mild
case_set_disjoint_multiple/len=32, offset=0
                        time:   [3.3819 ns 3.3833 ns 3.3848 ns]
                        change: [+0.4226% +0.5032% +0.5784%] (p = 0.00 < 0.05)
                        Change within noise threshold.
```

Macro implementation:
```
     Running benches/case_set.rs (target/release/deps/case_set-e476f33a0d7986a5)
case_set_one/len=1, offset=0
                        time:   [2.0460 ns 2.0473 ns 2.0488 ns]
                        change: [-10.497% -9.6214% -9.1096%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
case_set_one/len=8, offset=0
                        time:   [2.0718 ns 2.0721 ns 2.0724 ns]
                        change: [-8.5562% -8.5301% -8.5003%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 100 measurements (16.00%)
  5 (5.00%) high mild
  11 (11.00%) high severe
case_set_one/len=32, offset=0
                        time:   [2.2465 ns 2.2475 ns 2.2487 ns]
                        change: [-8.2221% -8.1723% -8.1220%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  7 (7.00%) high mild
  2 (2.00%) high severe

case_set_multiple/len=1, offset=0
                        time:   [4.5780 ns 4.5798 ns 4.5815 ns]
                        change: [-5.8970% -5.8619% -5.8257%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 22 outliers among 100 measurements (22.00%)
  9 (9.00%) low severe
  2 (2.00%) low mild
  11 (11.00%) high severe
case_set_multiple/len=8, offset=0
                        time:   [4.5038 ns 4.5058 ns 4.5083 ns]
                        change: [-6.5875% -6.5363% -6.4806%] (p = 0.00 < 0.05)
                        Performance has improved.
case_set_multiple/len=32, offset=0
                        time:   [3.3860 ns 3.3866 ns 3.3872 ns]
                        change: [-6.9543% -6.7518% -6.6278%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  13 (13.00%) high mild
  2 (2.00%) high severe

case_set_disjoint_multiple/len=1, offset=0
                        time:   [3.4651 ns 3.4679 ns 3.4712 ns]
                        change: [+0.0199% +0.1649% +0.3159%] (p = 0.03 < 0.05)
                        Change within noise threshold.
Found 18 outliers among 100 measurements (18.00%)
  16 (16.00%) high mild
  2 (2.00%) high severe
case_set_disjoint_multiple/len=8, offset=0
                        time:   [3.0519 ns 3.0542 ns 3.0563 ns]
                        change: [-1.9387% -1.7971% -1.6571%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  5 (5.00%) low severe
  1 (1.00%) low mild
case_set_disjoint_multiple/len=32, offset=0
                        time:   [3.3722 ns 3.3731 ns 3.3740 ns]
                        change: [+0.0807% +0.1628% +0.2379%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) low mild
  1 (1.00%) high severe
```

src/ctx_macro.rs
```diff
@@ -0,0 +1,239 @@
...
+        case_set!($UP_TO, $($WITH_DEFAULT,)? ctx=(), $len=$len, $offset=$offset_expr, $body);
+    };
+    ($UP_TO:literal, $($WITH_DEFAULT:literal,)? $len:ident=$len_expr:expr, $offset:ident=$offset_expr:expr, $body:block) => {
+        case_set!($UP_TO, $($WITH_DEFAULT,)? ctx=(), $len=$len_expr, $offset=$offset_expr, $body);
+    };
+    ($UP_TO:literal, $($WITH_DEFAULT:literal,)? $ctx:ident=$ctx_expr:expr, $len:ident=$len_expr:expr, $offset:ident=$offset_expr:expr, $body:block) => {
+        let $ctx = $ctx_expr;
+        let $len = $len_expr;
+        let $offset = $offset_expr;
+        {
```
kkysen, line 167: Why the extra `{}` block here?

src/ctx_macro.rs
```diff
@@ -0,0 +1,239 @@
...
+                            *<&mut [_; $N]>::try_from(buf_range).unwrap() = [$val; $N];
+                        };
+                    }
+                    #[allow(unused_macros)]
+                    macro_rules! set_disjoint {
+                        ($buf:expr, $val:expr) => {
+                            // SAFETY: The offset and length are checked by the
+                            // assert outside of the match.
+                            let mut buf_range = unsafe {
+                                $buf.index_mut_unchecked(($offset.., ..$N))
```
kkysen, line 202: Instead of adding `*_unchecked` methods, we can just call `assume` before it.  This works for normal slices as well.  And it feels "safer" since we can have essentially the same condition in the `assume` call here as the `assert!` outside of the `match`, so it's very obvious that it's sound.

src/ctx_macro.rs
```diff
@@ -0,0 +1,239 @@
...
+    ($UP_TO:literal, $($WITH_DEFAULT:literal,)? $len:ident=$len_expr:expr, $offset:ident, $body:block) => {
+        case_set!($UP_TO, $($WITH_DEFAULT,)? ctx=(), $len=$len_expr, $offset=$offset, { $(($buf, $val)),* });
+    };
+    ($UP_TO:literal, $($WITH_DEFAULT:literal,)? $len:ident, $offset:ident=$offset_expr:expr, $body:block) => {
+        case_set!($UP_TO, $($WITH_DEFAULT,)? ctx=(), $len=$len, $offset=$offset_expr, $body);
+    };
+    ($UP_TO:literal, $($WITH_DEFAULT:literal,)? $len:ident=$len_expr:expr, $offset:ident=$offset_expr:expr, $body:block) => {
+        case_set!($UP_TO, $($WITH_DEFAULT,)? ctx=(), $len=$len_expr, $offset=$offset_expr, $body);
+    };
+    ($UP_TO:literal, $($WITH_DEFAULT:literal,)? $ctx:ident=$ctx_expr:expr, $len:ident=$len_expr:expr, $offset:ident=$offset_expr:expr, $body:block) => {
```
kkysen, line 163: A few nits:

1. `UP_TO` and `WITH_DEFAULT` should probably be named; it makes it a lot clearer I think, especially for optional args (`WITH_DEFAULT`).
2. `UP_TO` and `WITH_DEFAULT` can just be `:expr`s, since they can be `const` `:expr`s.  And we can use inline `const {}`s to ensure they are computed and computable at `const` time.
3. We can assign them to variables, which I think makes things easier to read.

```suggestion
    (up_to=$up_to:expr, $(with_default=$with_default:expr,)? $ctx:ident=$ctx_expr:expr, $len:ident=$len_expr:expr, $offset:ident=$offset_expr:expr, $body:block) => {
    let up_to = const { $up_to };
    let with_default: bool = const { $($with_default ||)? false };
```

src/ctx_macro.rs
```diff
@@ -0,0 +1,239 @@
...
+/// });
+/// ```
+/// 
+/// [`DisjointMut`]: crate::src::disjoint_mut::DisjointMut
+#[macro_export]
+macro_rules! case_set {
+    ($UP_TO:literal, $($WITH_DEFAULT:literal,)? $ctx:ident=[$($ctx_expr:expr),*], $len:ident=[$($len_expr:expr),*], $offset:ident=[$($offset_expr:expr),*], $body:block) => {
+        for ($ctx, ($len, $offset)) in core::iter::zip([$($ctx_expr,)*], core::iter::zip([$($len_expr,)*], [$($offset_expr,)*])) {
+            case_set!($UP_TO, $($WITH_DEFAULT,)? $ctx=$ctx, $len=$len, $offset=$offset, $body);
+        }
```
kkysen, line 149: I wouldn't use `iter::zip` here.  It's not always totally zero cost, and it won't catch errors where the array lengths are different.  The clearest I think would be defining a const generic `fn` with a single `<const N: usize>` generic for all 3 arrays.  This should make error messages a lot clearer, too.

```suggestion
        fn f<const N: usize, T>(ctxs: [T; N], lens: [usize; N], offsets: [usize; N]) {
            for i in 0..N {
                case_set!($UP_TO, $($WITH_DEFAULT,)? $ctx=ctxs[i], $len=lens[i], $offset=offsets[i], $body);
            }
        }
        
        f([$($ctx_expr,)*], [$($len_expr,)*], [$($offset_expr,)*]);
```

kkysen: Also, given `[]` arrays are also `:expr`s, I'm not sure why this macro isn't ambiguous (does it just choose the first matching one).  I think it would be clearer if we just had a single macro arm, required `[]`s for single elements, and `()` for no `ctx`.

src/ctx_macro.rs
```diff
@@ -0,0 +1,239 @@
...
+    };
+    ($UP_TO:literal, $($WITH_DEFAULT:literal,)? $len:ident, $offset:ident, $body:block) => {
+        case_set!($UP_TO, $($WITH_DEFAULT,)? ctx=(), $len=$len, $offset=$offset, $body);
+    };
+    ($UP_TO:literal, $($WITH_DEFAULT:literal,)? $len:ident=$len_expr:expr, $offset:ident, $body:block) => {
+        case_set!($UP_TO, $($WITH_DEFAULT,)? ctx=(), $len=$len_expr, $offset=$offset, { $(($buf, $val)),* });
```
kkysen, line 155: What is `$val` here?

Also, I think `$offset=$offset` is incorrect (is this macro arm never used?).

src/ctx_macro.rs
```diff
@@ -0,0 +1,239 @@
...
+                ($buf:expr, $val:expr) => {
+                    assert!($offset <= $buf.len() && $offset + $len <= $buf.len());
+                };
+            }
+            #[allow(unused_macros)]
+            macro_rules! set_disjoint {
+                ($buf:expr, $val:expr) => {
+                    assert!($offset <= $buf.len() && $offset + $len <= $buf.len());
+                };
+            }
```
kkysen, line 179:
```suggestion
            use set as set_disjoint;
```

Since they're identical.

src/ctx_macro.rs
kkysen: It's also easy to make soundness mistakes this way, vs. the `assume` way, which just results in unoptimized code if we make a mistake.  I don't think we do this here, but it would be reasonable to call something like this on `Pixels` (so the elements are `BD::Pixel`s), in which case there will be a `* 2` for the index and length calculation.  If one of these overflows, the `unsafe` becomes unsound.  With an `assume`, it just becomes a bounds check.

src/ctx_macro.rs
```diff
@@ -0,0 +1,239 @@
...
+                        ($buf:expr, $val:expr) => {
+                            $buf[$offset..][..$len].fill($val);
+                        };
+                    }
+                    #[allow(unused_macros)]
+                    macro_rules! set_disjoint {
+                        ($buf:expr, $val:expr) => {
+                            $buf.index_mut($offset..$offset+$len).fill($val);
+                        };
+                    }
```
kkysen, line 232: Should we have the same `*_unchecked()` indexing/`assume`s here, too?

kkysen (changes requested):
Your `case_set!` solution generally LGTM, though I had a bunch of comments on the details.  If it's better for perf, we should choose it.  Have you looked at the asm for them?  Also, while the micro benchmarks are helpful since they magnify differences a lot better and there's less noise, how do the implementations fare on real uses?

src/ctx_macro.rs
```diff
@@ -0,0 +1,239 @@
...
+///      set_disjoint!(dir.seg_pred, seg_pred.into());
+///      set_disjoint!(dir.skip_mode, b.skip_mode);
+///      set_disjoint!(dir.intra, 0);
+///      set_disjoint!(dir.skip, b.skip);
+///      set_disjoint!(dir.pal_sz, 0);
+/// });
+/// ```
+/// 
+/// [`DisjointMut`]: crate::src::disjoint_mut::DisjointMut
+#[macro_export]
```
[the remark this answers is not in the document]
rinon: Just for benchmarking

src/ctx_macro.rs
rinon: When `$block` is a single statement, we need to make it into an expr block or we get a warning.

src/ctx_macro.rs
rinon: I strongly disagree with this. `assume` is not stable, first, but more importantly even if it was we have a simple API that does exactly what we want (don't insert a check) rather than relying on the optimizer to do something for us.

src/ctx_macro.rs
rinon: We have no instances where we need to make `$UP_TO` an expression, and I don't want to generalize this unless we need to. I don't think any of this needs to be done with my revised version that provides a separate `case_set_default!` macro.

src/ctx_macro.rs
rinon: It just chooses the first one. I don't think we should require `[]` in general, that makes call sites more confusing.

We can't create a function here without knowing the type of `T` because we want to call methods on `T` inside the body. I've added asserts to ensure that all three are the same length.

src/ctx_macro.rs
rinon: removed

src/ctx_macro.rs
rinon: :shrug: I'll test it and see if it makes a difference. That should be the slow path anyway.

PR #1293, opened 2024-07-09
Re-implement case set as a macro

rinon:


rinon:
Still need to address some comments from #1292

kkysen:
> Still need to address some comments from #1292

Let me know when this is ready to review.

rinon:
Ready. Performance is neutral, unfortunately this whole thing didn't make anything faster even though the micro-benchmarks showed improvement. We only need to land this if it is an improvement to readability (which I think it is, personally).

src/ctx.rs
```diff
@@ -1,147 +1,240 @@
...
-        08 if UP_TO >= 08 => *as_array(buf) = [val; 08],
-        16 if UP_TO >= 16 => *as_array(buf) = [val; 16],
-        32 if UP_TO >= 32 => *as_array(buf) = [val; 32],
-        64 if UP_TO >= 64 => *as_array(buf) = [val; 64],
-        _ => {
-            if WITH_DEFAULT {
-                buf.fill(val)
+/// [`DisjointMut`]: crate::src::disjoint_mut::DisjointMut
+macro_rules! case_set {
+    ($UP_TO:literal, $(@DEFAULT=$WITH_DEFAULT:literal,)? $ctx:ident=[$($ctx_expr:expr),* $(,)?], $len:ident=[$($len_expr:expr),* $(,)?], $offset:ident=[$($offset_expr:expr),* $(,)?], $body:block) => {
```
kkysen, line 133: What's the `@` in `@DEFAULT` for?  And can we just name it `WITH_DEFAULT`?

src/decode.rs
```diff
@@ -3174,10 +3192,11 @@ fn decode_b(
         let b4_stride = usize::try_from(f.b4_stride).unwrap();
         let cur_segmap = &f.cur_segmap.as_ref().unwrap().inner;
         let offset = by * b4_stride + bx;
-        CaseSet::<32, false>::one((), bw4, 0, |case, ()| {
+        case_set!(32, len = bw4, case_offset = 0, {
```
kkysen, line 3195: Why `case_offset` instead of `offset` here?

src/recon.rs
```diff
@@ -1395,22 +1396,23 @@ fn read_coef_tree<BD: BitDepth>(
                     ytx, txtp, eob, ts_c.msac.rng,
                 );
             }
-            CaseSet::<16, true>::many(
-                [&t.l.lcoef, &f.a[t.a].lcoef],
-                [
+            case_set_default!(
+                16,
+                dir = [&t.l.lcoef, &f.a[t.a].lcoef],
```
kkysen, line 1401:
```suggestion
                dir = [&t.l, &f.a[t.a]],
```

I did this change in my version, since I think it's something I accidentally did during initial safety cleanup.  Same for the couple of other places this happens.

src/recon.rs
```diff
@@ -1395,22 +1396,23 @@ fn read_coef_tree<BD: BitDepth>(
...
                     cmp::min(txh as c_int, f.bh - t.b.y) as usize,
                     cmp::min(txw as c_int, f.bw - t.b.x) as usize,
                 ],
-                [by4, bx4],
-                |case, dir| {
-                    case.set_disjoint(dir, cf_ctx);
-                },
+                offset = [by4 as usize, bx4 as usize],
+                {
+                    set_disjoint!(dir, cf_ctx);
```
kkysen, line 1408:
```suggestion
                    set_disjoint!(&dir.lcoef, cf_ctx);
```

src/decode.rs
rinon: `case` would shadow a variable here, which we don't want.

kkysen: What `case` variable?

kkysen: Oh, you mean `offset` variable.

rinon: Sorry, yep

src/ctx.rs
```diff
@@ -1,147 +1,240 @@
-//! The [`CaseSet`] API below is a safe and simplified version of the `case_set*` macros in `ctx.h`.
+//! The [`case_set!`] macro is a safe and simplified version of the `case_set*`
+//! macros in `ctx.h`.
 //!
-//! The `case_set*` macros themselves replaced `memset`s in order to further optimize them
-//! (in e3b5d4d044506f9e0e95e79b3de42fd94386cc61,
-//! which performed the following optimizations:
+//! The `case_set*` macros themselves replaced `memset`s in order to further
+//! optimize them (in e3b5d4d044506f9e0e95e79b3de42fd94386cc61, which performed
+//! the following optimizations:
```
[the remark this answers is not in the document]
rinon: Sure. Out of curiosity, why did they get broken at 100 instead of 80 in the first place?

src/ctx.rs
```diff
@@ -1,147 +1,240 @@
...
-//! (a runtime-determined `off`set is used, so we can't reasonably ensure this at compile-time).
+//! (1) is easy to preserve in Rust, but (2) is difficult to do so without
+//! overhead, as unaligned writes are UB, and so we'd need to check at runtime
+//! if they're aligned (a runtime-determined `off`set is used, so we can't
+//! reasonably ensure this at compile-time).
 //!
-//! To more thoroughly check this, I ran the same benchmarks done in
-//! e3b5d4d044506f9e0e95e79b3de42fd94386cc61, which introduced the `case_set*` macros:
+//! We also want to avoid multiple switches when setting a group of buffers,
+//! which was implemented in https://github.com/memorysafety/rav1d/pull/1293.
```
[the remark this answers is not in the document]
rinon: Eh, we're arguing about small things here and I'm not sure it's worthwhile. The list specifically lists what e3b5d4d implemented, which should not include this, so I'd have to re-write that as well.

src/ctx.rs
```diff
@@ -1,147 +1,240 @@
...
+        case_set!($UP_TO, $(@DEFAULT=$WITH_DEFAULT,)? _ctx=(), $len=$len, $offset=$offset, $body);
+    };
+    ($UP_TO:literal, $(@DEFAULT=$WITH_DEFAULT:literal,)? $len:ident=$len_expr:expr, $offset:ident=$offset_expr:expr, $body:block) => {
+        case_set!($UP_TO, $(@DEFAULT=$WITH_DEFAULT,)? _ctx=(), $len=$len_expr, $offset=$offset_expr, $body);
+    };
+    ($UP_TO:literal, $(@DEFAULT=$WITH_DEFAULT:literal,)? $ctx:ident=$ctx_expr:expr, $len:ident=$len_expr:expr, $offset:ident=$offset_expr:expr, $body:block) => {
+        #[allow(unused_mut)]
+        let mut $ctx = $ctx_expr;
+        let $len = $len_expr;
+        let $offset = $offset_expr;
```
kkysen, line 153: Do these need to be variable names set by the caller?  I don't think they're accessed in any `$body:block`.

kkysen: Then we don't have to worry about name collisions like `offset` and `case_offset`.

kkysen: Only `$ctx` is accessed by the caller IIUC.

rinon: I see this now, good point. I figured somewhere we access it somewhere from looking at the original C macro, but I guess we don't. It's a bit confusing that the body can access the context variable but not the others. I think I'd prefer to keep it consistent even if we don't use it right now.

I guess we could use `let ctx = ...` instead to indicate that it creates a new variable name, if we really want to make them different.

src/ctx.rs
rinon: The `@` distinguishes it from `$ctx:ident` (`@` is not a valid identifier, or anything else in prefix position for that matter).

rinon: What I really wanted was a way to pass in the boolean to unify the implementations but have two different macros for this (`case_set!` and `case_set_with_default!` or whatever).

src/decode.rs
```diff
@@ -368,19 +368,21 @@ fn read_tx_tree(
...
-        CaseSet::<16, false>::many(
-            [(&t.l, txh), (&f.a[t.a], txw)],
-            [t_dim.h as usize, t_dim.w as usize],
-            [by4 as usize, bx4 as usize],
-            |case, (dir, val)| {
+        case_set!(
+            16,
+            ctx = [(&t.l, txh), (&f.a[t.a], txw)],
+            len = [t_dim.h as usize, t_dim.w as usize],
+            offset = [by4 as usize, bx4 as usize],
```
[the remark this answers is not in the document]
rinon: Done.

src/ctx.rs
kkysen: In the old API, though, `ctx` was explicitly the only extra variable passed to the closure.  I don't think it's confusing that the others aren't accessible.  `ctx` is special in this way.  It's more confusing to me that there's all these other variables the caller defines.

src/ctx.rs
kkysen: It might work if you put it at the end of the args?  Especially if `len` and `offset` are not caller named.  This way is okay, too, though, just funny looking I guess.

PR #1295, opened 2024-07-09
Initial changes to build on Windows

thedataking:
DONE:
- [x] fix build with assembly enabled.
- [x] Clean up `cfg`ed `use`s
- [x] Add GitHub Actions workflow

Deferred to follow up PR:
- [ ] Enable caching in workflow
- [ ] Run tests in workflow


tests/seek_stress.rs
```diff
@@ -69,19 +71,28 @@ use std::ffi::c_double;
 use std::ffi::c_float;
 use std::ffi::c_int;
 use std::ffi::c_uint;
+#[cfg(not(target_os = "windows"))]
 use std::ffi::c_ulonglong;
```
kkysen, line 75: For `cfg`ed `use`s, I put them separate from all of the other `use`s since they mess up `rustfmt` sorting the `use`s.  Not too important I guess since this is in the tools, but it's nicer this way generally.

rinon:
Can you rebase this on top of #1307 ? Removing `snprintf` is non-trivial, so maybe we should just keep that part for now. Unfortunately the CLI allows dynamic filename formatting with C-style format strings. Maybe we can use https://crates.io/crates/sprintf or similar at some point? We should split the CLI tools into a separate crate in order to not add an extra dependency for the library in that case.

Cargo.toml
```diff
@@ -36,9 +36,16 @@ strum = { version = "0.25.0", features = ["derive"] }
...
+windows-sys = { version = "0.52", features = [
+    "Win32_Foundation",
+    "Win32_System_Performance",
+    "Win32_System_SystemInformation",
+] }
+
 [build-dependencies]
 cc = "1.0.79"
-nasm-rs = "0.2.4"
+nasm-rs = { version = "0.2.4", features = ["parallel"] }
```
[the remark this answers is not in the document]
kkysen: If it makes assembly build in parallel and it wasn't doing so before, that would be great, since it takes a long time to compile.

kkysen (changes requested):
For the `#[cfg]` imports, it's definitely cleaner when separate, but these are the tools, so I don't really care.  I do think the best would be to have the imports not `#[cfg]`ed and export from `compat` platform-independent interfaces (e.x. `pub use libc::snprintf` in `compat::stdio`), but again it's the tools, so it's probably not important.

I had a few other little comments, though, and could you rebase it on #1308 (or just wait for that to merge, probably tomorrow)?

Cargo.toml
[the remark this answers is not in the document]
thedataking: That's the intent, yes.

build.rs
```diff
@@ -308,4 +310,39 @@ fn main() {
...
+            println!("cargo:rustc-link-lib=static=oldnames");
+            let tool = windows_registry::find_tool(&target, "cl.exe")
+                .expect("couldn't find cl.exe; are the Visual Studio C++ tools installed?");
+            let lib_paths = &tool
+                .env()
+                .iter()
+                .find(|(key, _val)| key == "LIB")
+                .expect("LIB path not found")
+                .1;
+            for path in lib_paths.to_str().map(|s| s.split(';')).unwrap() {
```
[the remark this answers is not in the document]
thedataking: Thanks!

build.rs
```diff
@@ -308,4 +310,39 @@ fn main() {
     {
         asm::main();
     }
+
+    #[cfg(all(target_os = "windows", target_env = "msvc"))]
```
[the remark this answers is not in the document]
thedataking: Correct. 

We rely on Microsoft compilers and libraries that are only available on Windows as far as I know.

tools/compat/errno.rs
```diff
@@ -14,5 +14,11 @@ cfg_if::cfg_if! {
         pub unsafe fn errno_location() -> *mut c_int {
             libc::__errno()
         }
+    } else if #[cfg(target_os = "windows")] {
+        extern "C" { pub fn _errno() -> *mut c_int; }
```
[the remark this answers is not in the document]
thedataking: LMK if I didn't resolve this the way you intended.

thedataking:
> I do think the best would be to have the imports not `#[cfg]`ed and export from `compat` platform-independent interfaces (e.x. `pub use libc::snprintf` in `compat::stdio`), but again it's the tools, so it's probably not important.

Hadn't thought of that; makes it easier to rip out when we get rid of `snprintf` down the line.

tools/compat/getopt.rs
```diff
@@ -0,0 +1,28 @@
+cfg_if::cfg_if! {
+    if #[cfg(target_os = "windows")] {
```
rinon, line 2: Why is this `cfg_if`? It could just be a `mod` with a `cfg` on it and then re-exports.

tools/dav1d.rs
```diff
@@ -91,28 +92,38 @@ use std::ffi::c_uint;
...
-unsafe fn get_time_nanos() -> u64 {
-    let mut ts: libc::timespec = libc::timespec {
-        tv_sec: 0,
-        tv_nsec: 0,
-    };
-    libc::clock_gettime(1, &mut ts);
-    return (1000000000 as c_ulonglong)
-        .wrapping_mul(ts.tv_sec as c_ulonglong)
-        .wrapping_add(ts.tv_nsec as c_ulonglong) as u64;
+cfg_if::cfg_if! {
```
rinon, line 97: Why are we using `cfg_if` instead of just putting a `cfg` on the function itself?

tools/seek_stress.rs
```diff
@@ -62,19 +63,28 @@ use std::ffi::c_double;
...
-unsafe fn get_seed() -> c_uint {
-    let mut ts: libc::timespec = libc::timespec {
-        tv_sec: 0,
-        tv_nsec: 0,
-    };
-    libc::clock_gettime(1, &mut ts);
-    return (1000000000 as c_ulonglong)
-        .wrapping_mul(ts.tv_sec as c_ulonglong)
-        .wrapping_add(ts.tv_nsec as c_ulonglong) as c_uint;
+cfg_if::cfg_if! {
```
rinon, line 69: Same, why more `cfg_if`?

tools/compat/getopt.rs
rinon: Hm, I see the other compat modules do this. It's fine, just seemed weird.

tools/input/annexb.rs
```diff
@@ -254,9 +257,15 @@ unsafe extern "C" fn annexb_open(
         if res < 0 {
             break;
         }
+        #[cfg(target_os = "windows")]
+        fseek((*c).f, len as libc::c_long, 1 as c_int);
```
rinon, line 261: Shouldn't you just use the `fseeko` from compat here?

tools/input/ivf.rs
```diff
@@ -1,10 +1,16 @@
...
+use crate::compat::stdio::ftello;
 use crate::compat::stdio::stderr;
 use libc::fclose;
 use libc::fopen;
 use libc::fprintf;
 use libc::fread;
+#[cfg(not(target_os = "windows"))]
 use libc::fseeko;
+#[cfg(not(target_os = "windows"))]
 use libc::ftello;
```
rinon, line 14: nit: Could add these as re-exports from compat, like snprintf.

rinon (approved):
A few small nitpicks, but if you deem them unworthy of fixing, this LGTM. Builds fine on my windows 11 machine.

kkysen (approved):
LGTM besides Stephen's comments.

PR #1296, opened 2024-07-09
`mod looprestoration::neon`: (arm64) Make safe

kkysen:


src/looprestoration.rs
```diff
@@ -1549,148 +1549,103 @@ mod neon {
...
-        );
+            edges: LrEdgeFlags,
+            bd: BD,
+        ) {
+            let left = left
+                .map(|left| left.as_ptr().cast())
+                .unwrap_or_else(ptr::null);
+            let src = src.cast();
+            let bd = bd.into_c();
+            // SAFETY: asm should be safe.
```
rinon, line 1613: Not sure what this means. _Should_ be safe? According to what? We seem to have this safety comment in a lot of places. I think this should just be something like "asm call."

rinon (approved):

src/looprestoration.rs
kkysen: Yeah this is the safety comment I've been using for unconditionally asm (no fallback) calls.  We depend on the asm being safe for soundness which is why I used "should", but I'm open to different wording, though I'd want to change them all en masse if I do.

rinon: Alright, let's discuss, "should" isn't clear here. Agreed we should do that in a separate change.

PR #1297, opened 2024-07-10
`mod looprestoration`: Mark safe

### LINKED ISSUE #835: Cleanup remaining unsafe operations
For all files in the project we need to go through and cleanup remaining unsafe operations. This depends on making relevant struct fields safe, since we can't make a function fully safe if it depends on data that is behind a raw pointer. See https://github.com/memorysafety/rav1d/issues/706 for the remaining struct field cleanup work.

Once a file has been cleaned up, we should add `#![deny(unsafe_op_in_unsafe_fn)]` to it in order to lint that we've either cleaned up all unsafe operations or have put them within `unsafe` blocks with appropriate safety documentation. Functions that can be fully made safe should have the `unsafe` marker removed.

- [x] #836
- [x] #841
- [x] #842
- [x] #843
- [x] #844
- [x] #845
- [x] #846
- [x] #847
- [x] #848
- [x] #849
- [x] #850
- [x] #851
- [x] #852
- [x] #854
- [x] #855
- [x] #856
- [x] #857
- [x] #858
- [x] #859
- [x] #860
- [x] #863

kkysen:
* Fixes #835.

### DIFF
--- a/src/looprestoration.rs
+++ b/src/looprestoration.rs
@@ -3276,6 +3276,7 @@ mod neon {
     }
 }
 
+#[deny(unsafe_op_in_unsafe_fn)]
 #[cfg(all(feature = "asm", any(target_arch = "arm", target_arch = "aarch64")))]
 mod neon_erased {
     use super::*;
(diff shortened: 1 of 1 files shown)


src/looprestoration.rs
```diff
@@ -3276,13 +3278,15 @@ mod neon {
     }
 }
 
+#[deny(unsafe_op_in_unsafe_fn)]
```
rinon, line 3281: This doesn't propagate from the module level? Thought it did.

rinon (approved):

src/looprestoration.rs
kkysen: It does, but I've put them on all the `unsafe extern "C"` fallback `fn`s before making the entire `mod` safe, so I wanted to be consistent.

PR #1298, opened 2024-07-10
`mod looprestoration::neon`: Final cleanup

### LINKED ISSUE #839: Cleanup remaining c2rust output
These files still contain raw or partially-cleaned c2rust output. A cleanup pass needs to be done on each of them to make the code cleaner, safer, and better adhere to Rust idioms. Comments from the original C should also be added in to the corresponding Rust code. Where trivial to do so arguments to the function should be made safe.

Note that these files are also covered by https://github.com/memorysafety/rav1d/issues/835, which tracks the work to do a safety cleanup pass on the code. The safety cleanup pass should generally be done after an initial pass to cleanup/simplify the code and restore comments, as safety issues will be easier to address once the code has been simplified.

- [x] https://github.com/memorysafety/rav1d/issues/705
- [x] https://github.com/memorysafety/rav1d/issues/703
- [x] https://github.com/memorysafety/rav1d/issues/701
- [x] https://github.com/memorysafety/rav1d/issues/700
- [x] https://github.com/memorysafety/rav1d/issues/814
- [x] https://github.com/memorysafety/rav1d/issues/815
- [x] https://github.com/memorysafety/rav1d/issues/816
- [x] https://github.com/memorysafety/rav1d/issues/817
- [x] https://github.com/memorysafety/rav1d/issues/818
- [x] https://github.com/memorysafety/rav1d/issues/819
- [x] https://github.com/memorysafety/rav1d/issues/820

kkysen:
* Fixes #839.

### DIFF
--- a/src/looprestoration.rs
+++ b/src/looprestoration.rs
@@ -2015,7 +2015,7 @@ mod neon {
         // `lpf` may be negatively out of bounds.
         let mut lpf_bottom = lpf.wrapping_offset(6 * stride);
 
-        #[derive(PartialEq)]
+        #[derive(PartialEq, Eq)]
         enum Track {
             Main,
             Vert1,
@@ -2297,7 +2297,7 @@ mod neon {
         // `lpf` may be negatively out of bounds.
         let mut lpf_bottom = lpf.wrapping_offset(6 * stride);
 
-        #[derive(PartialEq)]
+        #[derive(PartialEq, Eq)]
         enum Track {
             Main,
             Vert1,
@@ -2726,7 +2726,7 @@ mod neon {
         // `lpf` may be negatively out of bounds.
         let mut lpf_bottom = lpf.wrapping_offset(6 * stride);
 
-        #[derive(PartialEq)]
+        #[derive(PartialEq, Eq)]
         enum Track {
             Main,
             Vert1,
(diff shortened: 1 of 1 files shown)


src/looprestoration.rs
```diff
@@ -722,44 +721,46 @@ fn selfguided_filter<BD: BitDepth>(
...
-            A += REST_UNIT_STRIDE;
+            b += REST_UNIT_STRIDE;
+            a += REST_UNIT_STRIDE;
             for i in 0..w {
-                let a = B[i].as_::<c_int>() * 6 + (B[i as isize - 1] + B[i + 1]).as_::<c_int>() * 5;
-                let b = A[i] * 6 + (A[i as isize - 1] + A[i + 1]) * 5;
+                let (a, b) = (
+                    b[i].as_::<c_int>() * 6 + (b[i as isize - 1] + b[i + 1]).as_::<c_int>() * 5,
+                    a[i] * 6 + (a[i as isize - 1] + a[i + 1]) * 5,
+                );
```
rinon, line 735: Nit for future changes, this wasn't necessary and made the diff harder to review (it's fine to leave here, just makes the review take longer).

rinon (approved):

src/looprestoration.rs
kkysen: It was necessary, since `a` and `b` are reassigned.  If you do it one at a time, then the new expression for `b` uses the new definition of `a` instead of the old one.

rinon: Ahhhh missed that you were using them inside each other now.

PR #1301, opened 2024-07-10
Fix camel case warnings

rinon:


tools/dav1d_cli_parse.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(non_camel_case_types)]
```
kkysen, line 1: We have this is `dav1d.rs` and `seek_stress.rs` already.

src/recon.rs
```diff
@@ -1,4 +1,5 @@
 #![deny(unsafe_op_in_unsafe_fn)]
+#![allow(non_camel_case_types)]
```
kkysen, line 2: We already have this in `lib.rs`; it's not needed here.  If you're seeing `rust-analyzer` warnings, that's a `rust-analyzer` issue.  It got fixed last time we upgraded our `nughtly` but then it came back at some point, so I'm not sure what the underlying issue is.

kkysen (changes requested):
I don't think this is needed; we already have them in the root modules. I think it's a rust-analyzer bug.

rinon:
I'll check with stable, but is there any real reason not to just work around this? It's been super annoying for development.

kkysen:
> I'll check with stable, but is there any real reason not to just work around this? It's been super annoying for development.

Can I just make them camel case?  That fixes it too and it's not very hard.

kkysen:
It's been very annoying for me too

rinon:
Closing in favor of #1302

Issue #1299, opened 2024-07-10
Fix and remove `#[allow(non_*)]`s


kkysen:
* [x] `#![allow(non_camel_case_types)]`
* [x] `#![allow(non_snake_case)]`
* [ ] `#![allow(non_upper_case_globals)]`

Would be nice to make things more idiomatically Rusty.  A bunch of renames, but simple to do.

PR #1302, opened 2024-07-10
`#![allow(non_camel_case_types)]`: Fix and remove

kkysen:
* Fixes `#![allow(non_camel_case_types)]` of #1299.

Supersedes #1301 by just fixing the warnings instead of papering over what appears to be a `rust-analyzer` false positive bug.

### DIFF
--- a/src/refmvs.rs
+++ b/src/refmvs.rs
@@ -19,8 +19,8 @@ use crate::src::error::Rav1dResult;
 use crate::src::ffi_safe::FFISafe;
 use crate::src::internal::Bxy;
 use crate::src::intra_edge::EdgeFlags;
-use crate::src::levels::mv;
 use crate::src::levels::BlockSize;
+use crate::src::levels::Mv;
 use crate::src::wrap_fn_ptr::wrap_fn_ptr;
 use std::cmp;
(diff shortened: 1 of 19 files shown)


src/refmvs.rs
```diff
@@ -32,44 +32,44 @@ use zerocopy::FromZeroes;
...
 }
 
 // In C, this is packed and is 12 bytes.
 // In Rust, being packed and aligned is tricky
-const _: () = assert!(mem::size_of::<refmvs_block>() == 12);
+const _: () = assert!(mem::size_of::<RefMvsBlock>() == 12);
 
 #[repr(C)]
-pub(crate) struct refmvs_frame<'a> {
+pub(crate) struct AsmRefMvsFrame<'a> {
```
rinon, line 72: Did `RefMvsFrame` conflict with another type? This seems to be the only one with `Asm`?

rinon (approved):

src/refmvs.rs
kkysen: Yeah, we have a safe non-asm version of this called `RefMvsFrame` already, and this is the version to pass to asm.

PR #1305, opened 2024-07-10
`#![deny(clippy::undocumented_unsafe_blocks)]`: Add and fix existing safety comments

kkysen:
* Part of #1270.

This adds `#![deny(clippy::undocumented_unsafe_blocks)]` to the `librav1d` crate.  Since we don't yet run `clippy` in CI, this doesn't do anything yet, but that's good, since we can turn that on once we fix all of the errors.  This PR only turns on the lint and fixes the existing safety comments into ones that the lint recognizes.

### DIFF
--- a/src/refmvs.rs
+++ b/src/refmvs.rs
@@ -318,6 +318,8 @@ impl save_tmvs::Fn {
             if ri > rf.r.len() - R_PAD {
                 return ptr::null();
             }
+
+            const _: () = assert!(mem::size_of::<RefMvsBlock>() * (1 + R_PAD) > 16);
             // SAFETY: `.add` is in-bounds; checked above.
             // Also note that asm may read 12-byte `refmvs_block`s in 16-byte chunks.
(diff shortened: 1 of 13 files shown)


src/refmvs.rs
```diff
@@ -318,6 +318,8 @@ impl save_tmvs::Fn {
             if ri > rf.r.len() - R_PAD {
                 return ptr::null();
             }
+
+            const _: () = assert!(mem::size_of::<RefMvsBlock>() * (1 + R_PAD) > 16);
```
rinon, line 322: This shouldn't be in this PR, should it?

rinon (changes requested):

src/refmvs.rs
kkysen: It should. There can't be anything in between the `unsafe` block and the `// SAFETY` comment, so I moved this assert to right before the safety comment.

rinon (approved):

PR #1308, opened 2024-07-11
Split CLI tools into a separate package

rinon:


randomPoison:
Looks like you need to add features for `asm`, `bitdepth_8`, and `bitdepth_16` to the new tools crate to be able to build with the same feature flags that we have on the main rav1d crate. You can add the following to `tools/Cargo.toml` to set this up:

```toml
[features]
default = ["asm", "bitdepth_8", "bitdepth_16"]
asm = ["rav1d/asm"]
bitdepth_8 = ["rav1d/bitdepth_8"]
bitdepth_16 = ["rav1d/bitdepth_16"]
```

Cargo.toml
```diff
@@ -1,28 +1,18 @@
 [workspace]
-members = []
+members = ["tools"]
```
kkysen, line 2: Don't you need to make `rav1d` the library a member, too?  That was my understanding of how workspaces work at least.

randomPoison: Yeah I think this needs to be

```toml
[workspace]
members = [".", "tools"]
default-members = [".", "tools"]
```

Otherwise the root rav1d library doesn't get built (or only gets built as a dependency of the CLI tools)

randomPoison: Adding this may also mean you don't need to add the feature flags I suggested in [this comment](https://github.com/memorysafety/rav1d/pull/1308#issuecomment-2223740522), I'm not 100% sure how dependency features interact with workspaces in this case.

tools/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+name = "dav1d"
+
+[[bin]]
+path = "seek_stress.rs"
+name = "seek_stress"
+
+[dependencies]
+cfg-if = "1.0.0"
+libc = "0.2"
+rav1d = { path = "../", version = "0.2.0" }
```
kkysen, line 20: Can we add a trailing newline?

tools/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "rav1d_cli"
+authors = ["C2Rust"]
+version = "0.2.0"
+publish = false
+edition = "2021"
+default-run = "dav1d"
+
+[[bin]]
+path = "dav1d.rs"
```
kkysen, line 10: Can we put them in the normal location now?  `src/bin/dav1d.rs` I think.

Cargo.toml
kkysen: I think it'd be better to do some restructuring of the files at this point.  Move `src/`, `include`, and `lib.rs` under `rav1d`.  Having the top level be a package that contains other packages which depend on it is IMO very confusing.

rinon: Not in this PR, and we have other higher priority things to worry about first. We need to do a general re-org and separate and document the C API, but not this week.

tools/Cargo.toml
rinon: We don't need to right now, no.

rinon:
It looks like fixing the workspace fixed the features, as they can now be used when building. I don't think we need to explicitly pass them from the binaries unless we want a different set of defaults.

kkysen (approved):

randomPoison (approved):

kkysen:
> It looks like fixing the workspace fixed the features, as they can now be used when building. I don't think we need to explicitly pass them from the binaries unless we want a different set of defaults.

@rinon, I don't think this is fixed.  @fbossen says that building with `--no-default-features --features bitdepth_8,bitdepth_16` no longer disables asm.  `rav1d-cli` is probably depending on `rav1d` with default features, whereas we need to forward them.  Not sure exactly how to do that.

rinon:
`cargo clean && cargo build  --no-default-features --features=bitdepth_8,bitdepth_16` results in a build with no assembly. I'll investigate to see what's going on.

PR #1309, opened 2024-07-11
Update toolchain to 1.79 stable

rinon:


rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly-2024-06-13"
+channel = "stable"
```
kkysen, line 2: We need to be careful since `arm` and `riscv*` still need nightly.  Not sure how to set that up to work right, though.

src/in_range.rs
```diff
@@ -71,11 +71,13 @@ macro_rules! impl_const_new {
         }
 
         impl<const MIN: u128, const MAX: u128> InRange<$T, MIN, MAX> {
+            #[allow(unused)]
```
kkysen, line 74: I thought these are used.  Or is it just that they're unused for some `$T`?

rinon: Most `T` are unused.

rust-toolchain.toml
rinon: I see this, and I'm not sure the best way either. We might need to keep the rust-toolchain pinned and add `+stable` to the non-arm and riscv CI builds. I don't love that solution though... Maybe we should rewrite this to not use the `std` feature detection.

src/levels.rs
```diff
@@ -55,7 +55,6 @@ pub enum TxfmSize {
 
 impl TxfmSize {
     pub const NUM_SQUARE: usize = Self::S64x64 as usize + 1;
-    pub const NUM_RECT: usize = Self::COUNT;
```
kkysen, line 58: Can we leave this with just a `_` prefix? Since a similar constant is in C, so it could be used later.  These are also things I think we'll make public for the crate if we add a Rust API, so I don't think we want to just completely delete it.

rust-toolchain.toml
kkysen: It's pretty obscure and bespoke to do it without `std`s feature detection, though.  It didn't feel very stable.

The ones that are stable already do correspond to the tier 1 targets.  Not sure what to do about it either.  I wish `rust-toolchain.toml` could be set per target.

rinon: I'm going to leave nightly for now and switch CI to stable.

thedataking (approved):

PR #1310, opened 2024-07-13
`struct *::cookie`: Make `*mut c_void`s into `Option<NonNull<c_void>>`s

kkysen:
This helps simplify some upcoming changes for `Send + Sync` safety.


rinon:
I'd like to see the changes this facilitates before reviewing. It looks fine I'm just not sure what this is for. I'm sure it'll make sense with the other PR(s) on top.

rinon (approved):
This is ok, I haven't thoroughly looked at what it enables.

kkysen:
> This is ok, I haven't thoroughly looked at what it enables.

It just makes #1312 a bit easier, and so I don't have to specify all of the nullability stuff in the safety docs every time.

PR #1311, opened 2024-07-13
`struct MsacAsmContextBuf`: Refactor out and `unsafe impl Send + Sync`

kkysen:
This refactors out `MsacAsmContext::buf_{pos,end}` to `MsacAsmContextBuf` and then `unsafe impl Send + Sync` on it since it stores pointers from the `Send + Sync` `Option<CArc<[u8]>>` in `MsacContext::asm`.  This makes `MsacContext: Send + Sync`.

### DIFF
--- a/src/msac.rs
+++ b/src/msac.rs
@@ -150,9 +150,30 @@ impl Default for Rav1dMsacDSPContext {
 pub type EcWin = usize;
 
 #[repr(C)]
+struct MsacAsmContextBuf {
+    pos: *const u8,
+    end: *const u8,
+}
+
+impl Default for MsacAsmContextBuf {
+    fn default() -> Self {
+        Self {
+            pos: ptr::null(),
+            end: ptr::null(),
+        }
+    }
+}
+
+impl From<&[u8]> for MsacAsmContextBuf {
+    fn from(value: &[u8]) -> Self {
+        let Range { start, end } = value.as_ptr_range();
+        Self { pos: start, end }
+    }
+}
+
+#[repr(C)]
 pub struct MsacAsmContext {
-    buf_pos: *const u8,
-    buf_end: *const u8,
+    buf: MsacAsmContextBuf,
     pub dif: EcWin,
     pub rng: c_uint,
     pub cnt: c_int,
@@ -169,8 +190,7 @@ pub struct MsacAsmContext {
 impl Default for MsacAsmContext {
     fn default() -> Self {
         Self {
-            buf_pos: ptr::null(),
-            buf_end: ptr::null(),
+            buf: Default::default(),
             dif: Default::default(),
             rng: Default::default(),
             cnt: Default::default(),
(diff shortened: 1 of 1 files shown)


src/msac.rs
```diff
@@ -149,10 +149,43 @@ impl Default for Rav1dMsacDSPContext {
...
 pub type EcWin = usize;
 
+#[repr(C)]
+struct MsacAsmContextBuf {
+    pos: *const u8,
+    end: *const u8,
+}
+
+/// SAFETY: [`MsacAsmContextBuf`] is always contained in [`MsacAsmContext::buf`],
+/// which is always contained in [`MsacContext::asm`], whos [`MsacContext::data`] field
```
rinon, line 159:
```suggestion
/// which is always contained in [`MsacContext::asm`], whose [`MsacContext::data`] field
```

src/msac.rs
```diff
@@ -149,10 +149,43 @@ impl Default for Rav1dMsacDSPContext {
...
+/// is what is stored in [`MsacAsmContextBuf::pos`] and [`MsacAsmContextBuf::end`].
+/// Since [`MsacContext::data`] is [`Send`], [`MsacAsmContextBuf`] is also [`Send`].
+unsafe impl Send for MsacAsmContextBuf {}
+
+/// SAFETY: [`MsacAsmContextBuf`] is always contained in [`MsacAsmContext::buf`],
+/// which is always contained in [`MsacContext::asm`], whos [`MsacContext::data`] field
```
rinon, line 165:
```suggestion
/// which is always contained in [`MsacContext::asm`], whose [`MsacContext::data`] field
```

src/msac.rs
```diff
@@ -149,10 +149,43 @@ impl Default for Rav1dMsacDSPContext {
 
 pub type EcWin = usize;
 
+#[repr(C)]
+struct MsacAsmContextBuf {
```
rinon, line 153: Document what this must point to hold the unsafe impl invariants. Either these must both be null or they must point into the `data` field of a `MsacContext` that also contains this structure.

src/msac.rs
```diff
@@ -149,10 +149,49 @@ impl Default for Rav1dMsacDSPContext {
 
 pub type EcWin = usize;
 
+/// # Safety
+///
+/// [`Self`] must be the first field of [`MsacAsmContext`] for asm layout purposes.
+/// And [`Self::pos`] and [`Self::end`] must be either [`ptr::null`],
+/// or [`Self::pos`] must point into (or the end of) [`MsacContext::data`],
```
rinon, line 156: I think it's important to point out that it must be the containing `MsacContext`, not some unrelated buffer.

rinon (approved):

src/msac.rs
kkysen: Okay I'll point that out.

kkysen: Fixed in [`1ec6eab..6856860`](https://github.com/memorysafety/rav1d/compare/1ec6eab4e00770b6022a66294f7f3c91b3967257..6856860f24d841e011e8a69508e417699b1042c0).

rinon (approved):

PR #1312, opened 2024-07-14
`struct SendSyncNonNull`: Add a `Send + Sync` `NonNull` type

kkysen:
`SendSyncNonNull` is sound through `unsafe` methods, and we use it here to make the `cookie` fields `Send + Sync`.

### DIFF
--- a/src/send_sync_non_null.rs
+++ b/src/send_sync_non_null.rs
@@ -1,0 +1,92 @@
+use std::ptr::NonNull;
+
+/// A [`NonNull`] that is [`Send`]` + `[`Sync`].
+///
+/// Since [`Send`]/[`Sync`] safety can't be guaranteed by the type system
+/// due to the raw [`NonNull`] pointers and [`Self::cast`],
+/// [`Send`]/[`Sync`] safety is instead guaranteed by
+/// the `unsafe`ty of [`Self::new_unchecked`].
+///
+/// Safe wrappers are provided around this, namely
+/// * [`Self::from_ref`]
+/// * [`Self::from_box`]
+///
+/// This can be reversed with the corresponding methods:
+/// * [`Self::as_ref`]
+/// * [`Self::into_box`]
+///
+/// though these are `unsafe` due to the raw pointers
+/// and potential [`Self::cast`] called in between.
+#[derive(Debug)]
+#[repr(transparent)]
+pub struct SendSyncNonNull<T: ?Sized>(NonNull<T>);
+
+// `Self: Clone` doesn't depend on `T: Clone`.
+impl<T: ?Sized> Clone for SendSyncNonNull<T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<T: ?Sized> Copy for SendSyncNonNull<T> {}
+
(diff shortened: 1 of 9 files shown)


src/send_sync_non_null.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// # Safety
+    ///
+    /// `ptr` must be derived from a type that is [`Send`]` + `[`Sync`],
+    /// and must only be used in a [`Send`]` + `[`Sync`] way.
+    pub const unsafe fn new_unchecked(ptr: NonNull<T>) -> Self {
+        Self(ptr)
+    }
+
+    pub fn from_ref(r#ref: &T) -> Self {
+        require_send_sync(&r#ref);
```
randomPoison, line 54: What is this `require_send_sync` function doing? Isn't `T: Send + Sync` already enforced by the trait bounds on the `impl` block?

kkysen: This machine checks it for `&T`, too.  Since the call has `&r#ref`.  Unfortunately subtle, but I didn't know how else to do it.

randomPoison: I'm not sure this is actually checking anything, though. It's the `T: Send + Sync` that means that `&T: Send + Sync`. There's no way to pass a value into the function that would cause `require_send_sync` to be the point that generates a compiler error.

kkysen: I'm machine checking that T: Send + Sync` implies `&T: Send + Sync`.  It's always true, but I just added it as a sanity check.

randomPoison: I don't think that's necessary to include, but I don't think it needs to block this PR either.

randomPoison (approved):

src/send_sync_non_null.rs
rinon: > I'm machine checking that `T: Send + Sync` implies `&T: Send + Sync`. It's always true, but I just added it as a sanity check.

This doesn't seem true. if the function took a `T` then I think it would do that, but I agree that this should be removed. That's a fundamental property of Rust references.

kkysen: > > I'm machine checking that `T: Send + Sync` implies `&T: Send + Sync`. It's always true, but I just added it as a sanity check.
> 
> This doesn't seem true.

`T: Send + Sync` doesn't imply `&T: Send + Sync`?

if the function took a `T` then I think it would do that,

The `fn` takes `&T`, and I'm passing it `&&T` here.

> but I agree that this should be removed.  That's a fundamental property of Rust references.

I can remove it, but I want to make sure I'm understanding it correctly.  I added this check since I wasn't super familiar with `Send` and `Sync` auto `impl` rules, so I thought it'd be easier to just machine check it, even if trivial.  Same with `Box<T>`.

rinon: IF `&T: !Sync` (which it can't be if `T` is `Send + Sync`) then wouldn't that function auto-ref so that the param was actually `&T`? What I was trying to say is that this function is doing nothing. Please remove it.

PR #1317, opened 2024-07-14
`fn Rav1dPictureDataComponentInner::wrap_buf`: Attempt to add lifetimes

kkysen:
I'm struggling to figure out the lifetimes here, even though I think they should work.  The `'buf` lifetime, i.e. the lifetime of `Rav1dPictureDataComponentInner::buf`, should outlive `'a`, where `'a` is in `&'a Rav1dPictureDataComponent<'buf>` when stored in `Rav1dPictureDataComponentOffset::data`.  But things aren't working out.  Can anyone else help and figure this out?

### DIFF
--- a/include/common/bitdepth.rs
+++ b/include/common/bitdepth.rs
@@ -135,7 +135,8 @@ pub trait BitDepth: Clone + Copy {
     const BPC: BPC;
     const BITDEPTH: u8 = Self::BPC.bitdepth();
 
-    type Pixel: Copy
+    type Pixel: 'static
+        + Copy
         + Ord
         + Add<Output = Self::Pixel>
         + Mul<Output = Self::Pixel>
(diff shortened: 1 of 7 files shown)


rinon:
It's not that it needs to outlive `'a` there, it's that you're assigning a `Rav1dPictureComponent<'buf>` and a `Rav1dPictureComponent<'static>` to the same variable, so the lifetimes in those two types must match (because `DisjointMut` is invariant).

kkysen:
> It's not that it needs to outlive `'a` there, it's that you're assigning a `Rav1dPictureComponent<'buf>` and a `Rav1dPictureComponent<'static>` to the same variable, so the lifetimes in those two types must match (because `DisjointMut` is invariant).
> 
> Sorry I don't have a solution tonight, gonna get to bed, but I'll think on it.

Yeah, I realized that.  But I don't know how to make the lifetimes exactly the same, or how to not have the lifetime be invariant.  But as far as I can tell, the `UnsafeCell<T>` forces it to be invariant since `&mut T`s can be made from it and `&mut T` is invariant over `T`.  That's my current understanding at least.

PR #1319, opened 2024-07-15
Enable full LTO

rinon:


kkysen (approved):
How much does this help perf?  Does it put us in the negative vs. C as @lqd saw, or was that mostly from `target-cpu=native`?

lqd:
The overhead doesn’t change much. My results were mostly because of target-cpu + fat LTO + rebuilding std/alloc/core with them.

rinon:
In my measurements this was a very small, if anything, speedup. I'll measure more carefully and post numbers.

rinon:
This is a 0.2% speedup on my Ryzen 7700X. It's consistent and slightly above measurement noise, with only a small increase in build time. I'm going to put this in.

PR #1320, opened 2024-07-16
`fn decode_b`: Performance improvements

rinon:
Inlines `backup2x8` which is inlined in C and improves performance slightly.

The `f.a[t.a]` block context reference is constant
throughout `decode_b`, but it appears that the
function is too complex for the optimizer to not
recompute this reference. Making it a local
improves performance measurably (~1% on a Ryzen
7700X for 8-bit Chimera).

### DIFF
--- a/src/cdef_apply.rs
+++ b/src/cdef_apply.rs
@@ -80,9 +80,10 @@ fn backup2lines<BD: BitDepth>(
     }
 }
 
+#[inline(always)]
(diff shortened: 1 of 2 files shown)


kkysen (approved):
Oh, very nice!  This is quite a simple change for a decent perf improvement.  Can we do the `ta` thing in any of other hot functions?  Or is `decode_b` especially worse since it's so massive?

PR #1324, opened 2024-07-17
Backport buffer over-read prevention from dav1d 1.4.3

fbossen:


tools/input/section5.rs
```diff
@@ -104,6 +104,9 @@ unsafe fn parse_obu_header(
     buf_size -= 1;
     if extension_flag != 0 {
         buf = buf.offset(1);
```
kkysen, line 106: I think the check is supposed to go before the `buf++`.

kkysen (changes requested):

tools/input/section5.rs
fbossen: oops

rinon (approved):

kkysen (approved):

PR #1327, opened 2024-07-18
`struct Unique`: Copy more from `std` and use for `Rav1dPictureDataComponentInner::ptr` to make it `Send + Sync`

kkysen:


include/dav1d/picture.rs
```diff
@@ -189,7 +192,7 @@ impl Rav1dPictureDataComponentInner {
     /// so it is sound to further subdivide it into disjoint `&mut`s.
     pub fn wrap_buf<BD: BitDepth>(buf: &mut [BD::Pixel], stride: usize) -> Self {
         let buf = AsBytes::as_bytes_mut(buf);
-        let ptr = NonNull::new(buf.as_mut_ptr()).unwrap();
+        let ptr = Unique::from_ref_mut(buf).cast();
```
rinon, line 195: How is this unique in the wrap_buf case? We're borrowing a reference to this data, right? So should we really use unique here?

include/dav1d/picture.rs
```diff
@@ -292,11 +295,14 @@ impl Rav1dPictureDataComponent {
...
 
-    fn as_dav1d(&self) -> Option<NonNull<c_void>> {
+    fn as_dav1d(&self) -> Option<Unique<c_void>> {
         if self.byte_len() == 0 {
             None
         } else {
-            NonNull::new(self.as_strided_byte_mut_ptr().cast())
+            let ptr = NonNull::new(self.as_strided_byte_mut_ptr()).unwrap();
+            // SAFETY: The `ptr` originally comes from a `Unique` in `Rav1dPictureDataComponentInner::ptr`.
+            let ptr = unsafe { Unique::new(ptr) };
```
rinon, line 304: Again, how is this unique? Can this method return a NonNull instead?

src/unique.rs
```diff
@@ -0,0 +1,190 @@
+//! Largely copied from [`rust-lang/rust/library/core/src/ptr/unique.rs`](https://github.com/rust-lang/rust/blob/e35364a521372ce682e4bd4a5850d97ea33b0eab/library/core/src/ptr/unique.rs#L58).
```
rinon, line 1: License? Copyright?

include/dav1d/picture.rs
kkysen: A `&mut` reference is always unique.  So it can be `&mut` or owned.  `core::ptr::Unique` `impl`s `From<&mut T>` even.

include/dav1d/picture.rs
kkysen: Because `self.as_strided_byte_mut_ptr()` is derived from `Rav1dPictureDataComponentInner::ptr`, which is a `Unique`.  I can say that more explicitly if that's clearer.

src/unique.rs
kkysen: Umm I don't know how to do that.  What do I write?

rinon:
```suggestion
//! Largely copied from [`rust-lang/rust/library/core/src/ptr/unique.rs`](https://github.com/rust-lang/rust/blob/e35364a521372ce682e4bd4a5850d97ea33b0eab/library/core/src/ptr/unique.rs#L58).
//!
//! Used under the MIT license <LICENSE-MIT> or <http://opensource.org/licenses/MIT>.
//! Copyright is retained by the original Rust contributors.
//! For full authorship information, see the Rust compiler version control history or
//! https://thanks.rust-lang.org 
```

include/dav1d/picture.rs
rinon: e.g. can I just call this twice and get two `Unique`s? Doesn't that violate the invariants?

rinon:
To follow up on the discussion of "should `DisjointMut` require `T: Sync`," `DisjointMut` is effectively a ranged `RwLock`, which _does_ require `T: Sync` to allow concurrent reads from multiple threads (which is what `Sync` means). I think we should impl `Sync` for this data pointer but I don't think it can be `Unique`.

include/dav1d/picture.rs
kkysen: Ugh, yeah, you're right. 

Why does `core::ptr::Unique: Copy + Clone` then?

rinon: No idea.

Issue #459, opened 2023-09-12
Pre-release cleanup


thedataking:
The repository contains artifacts that are helpful for migration but are not necessary to build or maintain the final deliverable. This includes:

- The dav1d C headers and sources
- Scripts added during migration (e.g. `retranspile.sh`)
- Changes to `.gitignore` to facilitate migration
- The dav1d build systems

After some discussion, the plan of record is to store these artifacts on a side branch and delete them from the main branch. Removing the meson build files is nice to have since it is contingent on tests being run by cargo, not meson.

We also need to make sure the README is updated to accurately reflect maturity, build steps,  prerequisites, etc.

PR #1330, opened 2024-07-18
Prepare for 1.0 release

rinon:


THANKS.md
```diff
@@ -1,3 +1,15 @@
+# The rav1d project would like to thank
+
+## Prossimo
+The Internet Safety Research Group and its donors for funding this Rust port.
+
+## Contributors
+
+The rav1d Rust port authors:
+
+Khyber Sen, Nicole L, Per Larsen, Stephen Crane, Folkert de Vries
```
randomPoison, line 10: Should we also list @fbossen here?

randomPoison (approved):

THANKS.md
rinon: Oh my! Yes, so sorry, I went through the top contributors and missed Frank!

Cargo.toml
```diff
@@ -23,13 +25,13 @@ libc = "0.2"
 parking_lot = "0.12.2"
 paste = "1.0.14"
 raw-cpuid = "11.0.1"
-strum = { version = "0.25.0", features = ["derive"] }
+strum = { version = "0.26", features = ["derive"] }
```
kkysen, line 28: Why are we switching to no patch version?

rinon: Why not? Do we mind if dependencies update patch levels?

kkysen: Aren't those compatible so that's still allowed?  I'd just been doing what `cargo add` defaults to, and all the other dependencies are still like that.

rinon: The patch level would get updated with `cargo update` but we don't get updates to the next minor version (I had to manually edit this to update). I'd prefer we get minor version updates when running `cargo update`, so eventually I'd like to switch to "x.y" for all deps unless there are other issues. But that doesn't need to be done right now.

kkysen: Okay if that's what we intend to do later then this is fine now.

kkysen (approved):

thedataking (approved):
LGTM. 

We had an issue for pre-release cleanup https://github.com/memorysafety/rav1d/issues/459 that we should probably close given that we've decided not to take action on any of those.

PR #1331, opened 2024-07-19
`struct Av1Block`: Replace Rust enum with separate tag and union

rinon:
This saves 4 bytes x number of Av1Block allocations by making `Av1Block` 32 bytes instead of 36. We can't just make the previous `ii` field a Rust enum with a 1-byte discriminant because the contents must be 2-byte aligned due to `Mv` needing that alignment.

On Chimera 8-bit this saves about 2% of the total heap usage.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -1194,8 +1194,9 @@ fn decode_b(
     let frame_type = f.frame_hdr.as_ref().unwrap().frame_type;
 
     let FrameThreadPassState::First(ts_c) = pass else {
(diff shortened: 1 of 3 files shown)


kkysen (changes requested):
We can also make `Av1Block` 32 bytes simply by removing the `#[repr(C)]` on `Av1BlockIntraInter`.  I'll open a PR for that.  I'm not sure how to measure the memory usage, but I assume that will have similar/the same memory savings.

PR #1334, opened 2024-08-05
Unify and rename `struct WithOffset`

rinon:


kkysen (commented):
I'm not sure if `Cursor` is a better name than `WithOffset`.  `WithOffset` isn't a smooth but it's descriptive, because it just adds an offset to anything (which could even be `()`).  And it matches the similar `WithStride`, which just adds a stride to anything.

We can unify the existing `Cursor` with `WithOffset`, though.  That part seems good.

PR #1335, opened 2024-08-13
Make dotprod and i8mm configurable features

rinon:
Adds the `asm_arm64_dotprod` and `asm_arm64_i8mm` features. We previously assumed that dotprod and i8mm were always enabled when building for aarch64 and then disabled at runtime if not supported. We also want to support the configuration where we do not even build the code, so these added features (enabled by default) allow this to be configurable.


kkysen (approved):
What environments is this for?

rinon:
Android does not build with dotprod or i8mm support.

PR #1338, opened 2024-09-04
Make logging match dav1d

rinon:
Reverts default logging back to stderr and fixes a logging print to match dav1d.


kkysen (approved):
These were both mistakes we made in porting, right?

rinon:
Yes, I believe so. This just reverts behavior back to match dav1d.

Issue #1356, opened 2024-09-10
Overflows can unsoundly taint length state in `AlignedVec`


HeroicKatora:
Relates to: https://github.com/memorysafety/rav1d/blob/7d7240943d519288fdc9f2b9532b750bd494bf2f/src/align.rs#L185-L191

The following test _must_ fail as it demonstrates memory unsafety, accessing an array way beyond its possible bounds.

```rust
#[test]
#[should_panic]
fn align_vec_fails() {
    let mut v = AlignedVec::<u16, Align8<[u8; 8]>>::new();
    v.resize(isize::MAX as usize + 2, 0u16);
    assert_eq!(v.as_slice()[isize::MAX as usize], 0);
}
```

The problem occurs in release mode, where no overflow checks are enabled. Code must not depend on overflow for correctness. The Safety comment is just incorrect:

            // SAFETY: We've allocated enough space to write
            // up to `new_len` elements into the buffer.

This is untrue. If the calculation for `new_bytes` overflows, the allocation which actually occurs is much smaller. The fix should be quite small, enforce a limit to the newly requested length such that it never overflows. The case of a zero-sized `T` can be ignored, this will never allocate at all.

I'm intending to create a PR soonish.

HeroicKatora:
Expanding on the memory safety issue by stepping through the test: After resizing `v` the `inner` vector has a length of `1`. This is because `(isize::MAX as usize + 2) * 2` wraps to the result `2usize`. Consequently, the new byte length is calculated as `1`. The following assert succeeds at this point:

```rust
    assert_eq!(v.inner.len(), 1);
```

The allocation consequently has a size of 8 bytes. The method nevertheless continues and the initialization loop offsets the base pointer of the allocation _way_ past those 8 bytes and writes the requested number into each memory. This is already UB. Note that the later access in `as_slice()[isize::MAX as usize]` will *also* wrap in memory but the safety requirement of slices has been violated in the call to `from_raw_parts` so this is not necessarily detected. In any case, sometimes this test produces:

```
Caused by:
  process didn't exit successfully: `/tmp/rav1d/target/release/deps/rav1d-ddebe794449c1390` (signal: 11, SIGSEGV: invalid memory reference)
```

Lokathor:
One possible "quick fix" is the classic `size_of::<T>().checked_mul(new_len).unwrap()`

PR #1357, opened 2024-09-10
`fn AlignedVec::resize`: Validate safety requirements, specifically overflow

HeroicKatora:
This is a necessary check for soundness, as demonstrated by the test which can SIGSEGV without the check. Before the check, an overflow in the underlying buffer calculation can create incoherent state where the vector believes in an impossibly large buffer of the item type which is not actually backed by a correctly sized buffer of chunks.

* Closes: #1356

### DIFF
--- a/src/align.rs
+++ b/src/align.rs
@@ -141,12 +141,12 @@ pub struct AlignedVec<T: Copy, C: AlignedByteChunk> {
 }
 
 impl<T: Copy, C: AlignedByteChunk> AlignedVec<T, C> {
-    // Note that in Rust, no single allocation can exceed `isize::MAX` __bytes_.
+    // Note that in Rust, no single allocation can exceed `isize::MAX` _bytes_.
     const MAX_LEN: usize = {
-        if core::mem::size_of::<T>() == 0 {
+        if mem::size_of::<T>() == 0 {
             usize::MAX
         } else {
-            (isize::MAX as usize) / core::mem::size_of::<T>()
+            (isize::MAX as usize) / mem::size_of::<T>()
         }
     };
 
@@ -292,7 +292,7 @@ fn align_vec_fails() {
     // This resize must fail. Otherwise, the code below creates a very small actual allocation, and
     // consequently a slice reference that points to memory outside the buffer.
     v.resize(isize::MAX as usize + 2, 0u16);
-    // Note that in Rust, no single allocation can exceed `isize::MAX` __bytes_. _Meaning it is
+    // Note that in Rust, no single allocation can exceed `isize::MAX` _bytes_. Meaning it is
     // impossible to soundly create a slice of `u16` with `isize::MAX` elements. If we got to this
     // point, everything is broken already. The indexing will the probably also wrap and appear to
     // work.
--- a/src/align.rs
+++ b/src/align.rs
@@ -141,15 +141,6 @@ pub struct AlignedVec<T: Copy, C: AlignedByteChunk> {
 }
 
 impl<T: Copy, C: AlignedByteChunk> AlignedVec<T, C> {
-    // Note that in Rust, no single allocation can exceed `isize::MAX` _bytes_.
-    const MAX_LEN: usize = {
-        if mem::size_of::<T>() == 0 {
-            usize::MAX
-        } else {
-            (isize::MAX as usize) / mem::size_of::<T>()
-        }
-    };
-
     /// Must check in all constructors.
     const fn check_byte_chunk_type_is_aligned() {
         assert!(mem::size_of::<C>() == mem::align_of::<C>());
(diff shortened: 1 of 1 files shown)


src/align.rs
```diff
@@ -141,6 +141,15 @@ pub struct AlignedVec<T: Copy, C: AlignedByteChunk> {
 }
 
 impl<T: Copy, C: AlignedByteChunk> AlignedVec<T, C> {
+    // Note that in Rust, no single allocation can exceed `isize::MAX` __bytes_.
+    const MAX_LEN: usize = {
+        if core::mem::size_of::<T>() == 0 {
```
kkysen, line 146:
```suggestion
        if mem::size_of::<T>() == 0 {
```

We already have `use std::mem` and use `mem::size_of` of elsewhere.

src/align.rs
```diff
@@ -141,6 +141,15 @@ pub struct AlignedVec<T: Copy, C: AlignedByteChunk> {
...
+            usize::MAX
+        } else {
+            (isize::MAX as usize) / core::mem::size_of::<T>()
```
kkysen, line 149:
```suggestion
            (isize::MAX as usize) / mem::size_of::<T>()
```

src/align.rs
```diff
@@ -141,6 +141,15 @@ pub struct AlignedVec<T: Copy, C: AlignedByteChunk> {
 }
 
 impl<T: Copy, C: AlignedByteChunk> AlignedVec<T, C> {
+    // Note that in Rust, no single allocation can exceed `isize::MAX` __bytes_.
```
kkysen, line 144: What is the `__` and `_` surrounding `bytes`?  Is that supposed to be Markdown?

src/align.rs
```diff
@@ -183,6 +192,14 @@ impl<T: Copy, C: AlignedByteChunk> AlignedVec<T, C> {
     }
 
     pub fn resize(&mut self, new_len: usize, value: T) {
+        // In addition to the obvious effect, this verifies the wrapping behavior of the
+        // `new_bytes` calculation. That can not overflow as the length limit does not overflow
+        // when multiplied with the size of `T`. Note that we one can still pass ludicrous
+        // requested buffer lengths, just not unsound ones.
+        assert!(
+            new_len <= Self::MAX_LEN,
```
kkysen, line 200: We only use it here and it'd be nicer to see the logic to compute `MAX_LEN` here, so could you move it to this `fn`?

kkysen: Is there any advantage of doing this `<=` check rather than `let new_bytes = new_len.checked_mul(mem::size_of::<T>()).expect("Resizing would overflow the underlying aligned buffer");`?  The latter seems more obviously correct and direct in what it's preventing: overflow.

kkysen (changes requested):
Thanks for finding this bug and the PR!

src/align.rs
HeroicKatora: That was meant to underscore that this is a size in terms of bytes, not units. Apparently that emphasis didn't achieve its goal. I can just remove it?

HeroicKatora:
```suggestion
    // Note that in Rust, no single allocation can exceed `isize::MAX` bytes.
```

src/align.rs
HeroicKatora: Unfortunately, the constant can not be moved into the function. Rust prevents the local constant from using the outer item's generic parameter. The main advantage relates to the open comment of customizing the panic message. I was unsure if the `MAX_LEN` value should be mentioned for instance. Neither prevents the other, if `checked_mul(..).unwrap()` was used after the manual assert, the bounds check should get removed by the compiler but conversely `expect` can not incorporate the constant in its message.

As an additional consideration, verifying arguments to the function as early as possible. Since this is a precondition rather than an internal logic bug, it might also be helpful to have the constant available for other modules to independently check against some lengths they calculate for future use.

src/align.rs
kkysen: If that's supposed to be markdown for bold, that's fine, I was just confused by the 2 underscores before and only 1 after.  Could you just do 1 before and after? And also for the other place you did the same thing.

src/align.rs
kkysen: I think a normal `.unwrap()`/``.expect()` would probably be fine here as this function shouldn't be very hot.  But we've definitely run into issues where moving the panic and the formatting code it uses to an `#[inline(never)]` (we didn't mark them `#[cold]`, though we probably should) was important, so feel free to do that here as well.

If you'd like to do the `<= MAX_LEN` check first, that's fine, if you pair it with a `.checked_mul(...).unwrap()` that's optimized out so that it's clear which operation exactly we're checking for overflow.  Or you can do the `let Some(_) else` approach you also suggested.

src/align.rs
```diff
@@ -186,15 +186,22 @@ impl<T: Copy, C: AlignedByteChunk> AlignedVec<T, C> {
...
         // Resize the underlying vector to have enough chunks for the new length.
-        //
-        // NOTE: We don't need to `drop` any elements if the `Vec` is truncated since `T: Copy`.
-        let new_bytes = mem::size_of::<T>() * new_len;
+        // SAFETY: The `new_bytes` calculation must not overflow, ensuring a mathematical match
+        // with the underlying `inner` buffer size. NOTE: one can still pass ludicrous requested
+        // buffer lengths, just not unsound ones.
+        let Some(new_bytes) = mem::size_of::<T>().checked_mul(new_len) else {
+            panic!("Resizing would overflow the underlying aligned buffer");
+        };
```
kkysen, line 194:
```suggestion
        let new_bytes = mem::size_of::<T>().checked_mul(new_len)
            .expect("Resizing would overflow the underlying aligned buffer");
```

If we aren't gonna do any fancy `#[inline(never)]`/`#[cold]` inner `fn` stuff for the panic, I'd rather just use `.expect` here.

(Also not sure if the formatting is right).

kkysen (changes requested):
LGTM now except for that one little comment.

src/align.rs
HeroicKatora: Feel free to push that change to the fork.

kkysen (approved):
LGTM now.  @HeroicKatora, I'll merge it now assuming you're all done with it.  Thanks again!

PR #1362, opened 2024-10-03
Copy `dav1d` as the Rust API

### LINKED ISSUE #1252: Provide Rust API
Currently we primarily expose a C API that matches dav1d's API in order for rav1d to be a drop-in replacement for dav1d. However we also want rav1d to be usable in pure Rust projects, so we should also expose a Rust API.

For the most part the `dav1d_*` functions already call into equivalent `rav1d_*` Rust functions, e.g. `dav1d_open` simply wraps and calls into `rav1d_open`. But the Rust functions are not currently exported publicly (they're currently `pub(crate)`) and do not have any documentation.

emilk:
* Closes https://github.com/memorysafety/rav1d/issues/1252
* Used in https://github.com/rerun-io/rerun/pull/7557

This basically copy-pastes https://github.com/rust-av/dav1d-rs/blob/master/src/lib.rs (actually via https://github.com/rerun-io/rav1d-rs/blob/master/src/lib.rs) into rav1d to give rav1d a proper Rust interface.

It is a bit silly to wrap the C FFI, but it works.

The upshot is that this makes `rav1d` a drop-in replacement for https://crates.io/crates/dav1d

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,12 +20,14 @@ crate-type = ["staticlib", "rlib"]
 [dependencies]
 assert_matches = "1.5.0"
 atomig = { version = "0.4.0", features = ["derive"] }
+av-data = "0.4.2"
 bitflags = "2.4.0"
 cfg-if = "1.0.0"
 libc = "0.2"
 parking_lot = "0.12.2"
 paste = "1.0.14"
 raw-cpuid = "11.0.1"
+static_assertions = "1"
 strum = { version = "0.26", features = ["derive"] }
 to_method = "1.1.0"
 zerocopy = { version = "0.7.32", features = ["derive"] }
--- a/include/dav1d/headers.rs
+++ b/include/dav1d/headers.rs
@@ -671,7 +671,7 @@ pub const DAV1D_MC_SMPTE240: Dav1dMatrixCoefficients = Rav1dMatrixCoefficients::
 pub const DAV1D_MC_SMPTE_YCGCO: Dav1dMatrixCoefficients =
     Rav1dMatrixCoefficients::SMPTE_YCGCO.to_dav1d();
 pub const DAV1D_MC_BT2020_NCL: Dav1dMatrixCoefficients =
-    Rav1dMatrixCoefficients::BT2020_CL.to_dav1d();
+    Rav1dMatrixCoefficients::BT2020_NCL.to_dav1d();
 pub const DAV1D_MC_BT2020_CL: Dav1dMatrixCoefficients =
     Rav1dMatrixCoefficients::BT2020_CL.to_dav1d();
 pub const DAV1D_MC_SMPTE2085: Dav1dMatrixCoefficients =
--- a/lib.rs
+++ b/lib.rs
@@ -96,8 +96,40 @@ pub use src::error::Dav1dResult;
 
 // ---------------------------------------------------------------------------------------
 
-/// Public Rust API, mostly copied from <https://github.com/rust-av/dav1d-rs/blob/master/src/lib.rs>
+/// Public Rust API.
+///
+/// This is more or less the same API as <https://crates.io/crates/dav1d>,
+/// and is indeed a fork of that work.
 pub mod dav1d {
+    // This whole module was originally copied from https://github.com/rust-av/dav1d-rs/
+    // (specifically https://github.com/rust-av/dav1d-rs/blob/94b1deaa1e25bf29c77bb5cc8a08ddaf7663eede/src/lib.rs)
+    // with some modifications.
+    // `dav1d-rs` is under the MIT license, replicated here:
+
+    // MIT License
+    //
+    // Copyright (c) 2018 Luca Barbato
+    //
+    // Permission is hereby granted, free of charge, to any person obtaining a copy
+    // of this software and associated documentation files (the "Software"), to deal
+    // in the Software without restriction, including without limitation the rights
+    // to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+    // copies of the Software, and to permit persons to whom the Software is
+    // furnished to do so, subject to the following conditions:
+    //
+    // The above copyright notice and this permission notice shall be included in all
+    // copies or substantial portions of the Software.
+    //
+    // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    // IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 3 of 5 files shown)


include/dav1d/headers.rs
```diff
@@ -671,7 +671,7 @@ pub const DAV1D_MC_SMPTE240: Dav1dMatrixCoefficients = Rav1dMatrixCoefficients::
 pub const DAV1D_MC_SMPTE_YCGCO: Dav1dMatrixCoefficients =
     Rav1dMatrixCoefficients::SMPTE_YCGCO.to_dav1d();
 pub const DAV1D_MC_BT2020_NCL: Dav1dMatrixCoefficients =
-    Rav1dMatrixCoefficients::BT2020_CL.to_dav1d();
+    Rav1dMatrixCoefficients::BT2020_NCL.to_dav1d();
```
kkysen, line 674: Thanks for finding this! This does look like a bug. Can you put this in its own PR? We can merge it quickly on its own.

emilk: Sure thing:
* https://github.com/memorysafety/rav1d/pull/1363

Cargo.toml
```diff
@@ -20,12 +20,14 @@ crate-type = ["staticlib", "rlib"]
...
 assert_matches = "1.5.0"
 atomig = { version = "0.4.0", features = ["derive"] }
+av-data = "0.4.2"
 bitflags = "2.4.0"
 cfg-if = "1.0.0"
 libc = "0.2"
 parking_lot = "0.12.2"
 paste = "1.0.14"
 raw-cpuid = "11.0.1"
+static_assertions = "1"
```
kkysen, line 30: Can we keep this like the other dependencies with a full `major.minor.patch`?  We just used what `cargo add` gave us.

lib.rs
```diff
@@ -93,3 +93,846 @@ pub mod src {
...
+        fn from(err: Rav1dError) -> Self {
+            match err {
+                Rav1dError::EAGAIN => Error::Again,
+                Rav1dError::ENOMEM => Error::NotEnoughMemory,
+                Rav1dError::EINVAL => Error::InvalidArgument,
+                Rav1dError::ENOPROTOOPT => Error::UnsupportedBitstream,
+                _ => Error::UnknownError(err),
+            }
+        }
+    }
```
kkysen, line 206: Do we just want to use this as `Rav1dError` in the first place?

lib.rs
```diff
@@ -93,3 +93,846 @@ pub mod src {
...
+        pub fn new() -> Self {
+            unsafe {
+                let mut dav1d_settings = mem::MaybeUninit::uninit();
+
+                dav1d_default_settings(NonNull::new(dav1d_settings.as_mut_ptr()).unwrap());
+
+                Self {
+                    dav1d_settings: dav1d_settings.assume_init(),
+                }
+            }
```
kkysen, line 254: This one is pretty trivial to do safely using the Rust API.

lib.rs
```diff
@@ -93,3 +93,846 @@ pub mod src {
...
+    }
+
+    bitflags::bitflags! {
+        #[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
+        pub struct InloopFilterType: u32 {
+            const DEBLOCK = DAV1D_INLOOPFILTER_DEBLOCK;
+            const CDEF = DAV1D_INLOOPFILTER_CDEF;
+            const RESTORATION = DAV1D_INLOOPFILTER_RESTORATION;
+        }
+    }
```
kkysen, line 346: We'd prefer just making `Rav1dInloopFilterType` this instead of doing a conversion here at the edge of the API.

lib.rs
```diff
@@ -93,3 +93,846 @@ pub mod src {
...
+        fn try_from(value: u32) -> Result<Self, Self::Error> {
+            match value {
+                DAV1D_DECODEFRAMETYPE_ALL => Ok(DecodeFrameType::All),
+                DAV1D_DECODEFRAMETYPE_REFERENCE => Ok(DecodeFrameType::Reference),
+                DAV1D_DECODEFRAMETYPE_INTRA => Ok(DecodeFrameType::Intra),
+                DAV1D_DECODEFRAMETYPE_KEY => Ok(DecodeFrameType::Key),
+                _ => Err(TryFromEnumError(())),
+            }
+        }
+    }
```
kkysen, line 369: Same here.

lib.rs
```diff
@@ -93,3 +93,846 @@ pub mod src {
...
+                DecodeFrameType::Key => DAV1D_DECODEFRAMETYPE_KEY,
+            }
+        }
+    }
+
+    /// The error type returned when a conversion from a C enum fails.
+    #[derive(Debug, Copy, Clone, PartialEq, Eq)]
+    pub struct TryFromEnumError(());
+
+    impl std::fmt::Display for TryFromEnumError {
```
kkysen, line 386: We've preferred fully importing these kinds of items (e.x. `use std::fmt::Display`).

lib.rs
```diff
@@ -93,3 +93,846 @@ pub mod src {
...
+        dec: Dav1dContext,
+        pending_data: Option<Dav1dData>,
+    }
+
+    unsafe extern "C" fn release_wrapped_data<T: AsRef<[u8]>>(
+        _data: *const u8,
+        cookie: Option<SendSyncNonNull<std::ffi::c_void>>,
+    ) {
+        let cookie = cookie.unwrap().as_ptr().as_ptr();
+        let buf = unsafe { Box::from_raw(cookie as *mut T) };
```
kkysen, line 411: We haven't enabled it in CI yet, but we have `#![deny(clippy::undocumented_unsafe_blocks)]` on.  All `unsafe` blocks should be documented with `// SAFETY` comments.

lib.rs
```diff
@@ -93,3 +93,846 @@ pub mod src {
...
+    /// Frame component.
+    #[derive(Eq, PartialEq, Copy, Clone, Debug)]
+    pub enum PlanarImageComponent {
+        /// Y component.
+        Y,
+        /// U component.
+        U,
+        /// V component.
+        V,
+    }
```
kkysen, line 638: These are much better names than we came up with (we don't know much about AV1 or video codecs themselves).  We should try to just use these internally more.

kkysen (changes requested):
Having a public Rust API matching `dav1d-rs` seems like it'd be very nice, as that Rusty API has already been designed for us evidently.  However, this is way too much `unsafe` code that is entirely undocumented/unjustified, and I don't really see why we can't wrap the safe but private Rust API that we have right now (the `rav1d_*` versions of all of the `dav1d_*` functions, which are all safe, but otherwise have very similar APIs besides converting the data types).

emilk:
* Completed in https://github.com/memorysafety/rav1d/pull/1439

PR #1364, opened 2024-10-04
Remove crashy `unwrap()` in `on_error`

emilk:
I got a crash here, and I don't want a crash :)

```
thread 'av1_decoder' panicked at 'called `Option::unwrap()` on a `None` value'
6f12e3f/src/decode.rs:4997
```

It seems like this `unwrap` was introduced in https://github.com/memorysafety/rav1d/pull/663 along with many others.

May I suggest enabling the [`clippy::unwrap_used`](https://rust-lang.github.io/rust-clippy/master/index.html#/unwrap_used) lint? Any `.unwrap()` is a panic waiting to happen :/

**EDIT:** The crash happens during decode of the chunk starting at timestamp `114176` of this video: https://storage.googleapis.com/rerun-test-assets/video/Sintel_1080_10s_av1.mp4

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4994,7 +4994,10 @@ pub fn rav1d_submit_frame(c: &Rav1dContext, state: &mut Rav1dState) -> Rav1dResu
     ) {
         fc.task_thread.error.store(1, Ordering::Relaxed);
         let _ = mem::take(&mut *fc.in_cdf.try_write().unwrap());
-        if f.frame_hdr.as_ref().unwrap().refresh_context != 0 {
+        if f.frame_hdr
+            .as_ref()
+            .is_some_and(|frame_hdr| frame_hdr.refresh_context != 0)
+        {
             let _ = mem::take(&mut f.out_cdf);
         }
         for i in 0..7 {


src/decode.rs
```diff
@@ -4994,7 +4994,10 @@ pub fn rav1d_submit_frame(c: &Rav1dContext, state: &mut Rav1dState) -> Rav1dResu
     ) {
         fc.task_thread.error.store(1, Ordering::Relaxed);
         let _ = mem::take(&mut *fc.in_cdf.try_write().unwrap());
-        if f.frame_hdr.as_ref().unwrap().refresh_context != 0 {
+        if f.frame_hdr
+            .as_ref()
+            .is_some_and(|frame_hdr| frame_hdr.refresh_context != 0)
+        {
```
emilk, line 5000: I'm not sure what the right behavior is in the `None` case, but crashing is certainly not it

kkysen: Places where there is an unwrap should have been unchecked derefs in C, so let me check what C was doing here.

kkysen (commented):
What input were you running this on?

emilk:
I was running this video (the crash happens about half-way through it):

https://storage.googleapis.com/rerun-test-assets/video/Sintel_1080_10s_av1.mp4

src/decode.rs
kkysen: So C is doing the same thing as we are here: 
* https://github.com/memorysafety/rav1d/blob/c7d127e7e31bd3366ac7dc1717dda9782905c605/src/decode.c#L3350
* https://code.videolan.org/videolan/dav1d/-/blob/master/src/decode.c?ref_type=heads#L3722

Can you check if you get an error/crash on `dav1d` here or not?  Thanks!

emilk: Using https://crates.io/crates/dav1d 0.10.3 I do not get a crash, but I get an `Invalid argument` error and some useful info printed:

```
Error parsing frame header
Error parsing OBU data
```

It should be noted though that dereferencing a null pointer is not necessarily a crash in C - it is UB. The compiler is free to assume the pointer is not null, and do whatever crazy shenanigans that implies.

kkysen: Hmm, it's possible UB could be doing something weird here, but I'd still expect a segfault for the most part.  We likely diverged at some point, so we'll have to figure out where and see if that's something we need to fix or if it's okay to crash, just in a different place.

That said, this video should successfully decode with normal `dav1d`, but doesn't?  Is that a bug with `dav1d` as well?

emilk: I don't know if the video should successfully decode, or if it is actually a bad video.

> or if it's okay to crash

I would prefer a `rav1d` that never crashes on any input, but returns errors.

kkysen: > I would prefer a rav1d that never crashes on any input, but returns errors.

Yup, that would definitely be preferable.

kkysen:
Hi @emilk, I have some time to look into this bug now.  What specific command did you use on [`Sintel_1080_10s_av1.mp4`](https://storage.googleapis.com/rerun-test-assets/video/Sintel_1080_10s_av1.mp4) that triggered the bug?  Sorry, I'm not super familiar with all of the CLI options besides what's run in the tests.

thedataking:
@emilk please let us know if you can help us reproduce the issue so we can test your solution.

emilk:
I'm back! Ok, here is a repro :)

The crash requires a single frame context, so `--threads 1` or `--framedelay 1`. With the default thread settings the error takes the multi-threaded path and never reaches the `unwrap`, which is why it did not reproduce for you.

```shell
wget https://storage.googleapis.com/rerun-test-assets/video/Sintel_1080_10s_av1.mp4
ffmpeg -i Sintel_1080_10s_av1.mp4 -c:v copy -f ivf Sintel_1080_10s_av1.ivf
cargo build --release
target/release/dav1d -i Sintel_1080_10s_av1.ivf -o /dev/null --threads 1
```

```
thread 'main' panicked at src/decode.rs:4997:33:
called `Option::unwrap()` on a `None` value
   4: rav1d::src::decode::rav1d_submit_frame::on_error
   5: rav1d::src::decode::rav1d_submit_frame
   6: rav1d::src::lib::gen_picture
   7: _dav1d_send_data
```

**Root cause:** when `c.fc.len() == 1`, `rav1d_submit_frame` calls `rav1d_decode_frame` directly. On failure, `rav1d_decode_frame_exit` does `mem::take(&mut f.frame_hdr)`. `rav1d_submit_frame` then re-locks `f` and calls `on_error`, which unwraps the now-empty `f.frame_hdr`. The multi-frame-context path never calls `on_error` after `rav1d_decode_frame_exit`, so it does not crash. The Sintel file has two frames that fail with `EINVAL` (C dav1d reports the same two errors), which triggers this path.

PR #1368, opened 2024-10-31
Try optimizing `f.mvs` by only allocating when the length changes

kkysen:
I think most frames should be the same size, so hopefully this helps enough without having to have a whole pool allocator, which gets tricky with lifetimes and self-referential structs.

@ivanloz, could you look at this version, too?  Since this one doesn't use any unstable features.


fbossen:
Isn't the memory released anyway at the end of each frame in `rav1d_decode_frame_exit` (see `let _ = mem::take(&mut f.mvs)`) and therefore reallocated in `ravid_submit_frame`?

kkysen:
> Isn't the memory released anyway at the end of each frame in `rav1d_decode_frame_exit` (see `let _ = mem::take(&mut f.mvs)`) and therefore reallocated in `rav1d_submit_frame`?

Hmm, I didn't think of that.

PR #1369, opened 2024-11-10
`struct AlignedVec`: Verify alignment for `T`

HeroicKatora:
Turning the pointer of an aligned chunk into a reference to the overlay element type requires us to verify the element type's alignment invariant on the address. This check need not happen dynamically if we can provide a stronger static assertion.

The use of `AlignedVec` is over-aligning the storage with regards to the elements. This guarantees that all aligned block addresses are at least as aligned as required for elements. With this commit we validate that only vectors with appropriate element types can be instantiated. This verifies the constructor instead of the type parameters due to simplicity. Turning the violation into a compile error may be beneficial during development but should not be strictly necessary. Passing through a constructor is a dominator code path of the creation of a reference from its storage (at least right now).

### DIFF
--- a/src/align.rs
+++ b/src/align.rs
@@ -146,8 +146,14 @@ impl<T: Copy, C: AlignedByteChunk> AlignedVec<T, C> {
         assert!(mem::size_of::<C>() == mem::align_of::<C>());
     }
 
+    const fn check_inner_type_is_aligned() {
+        assert!(mem::align_of::<T>() <= mem::align_of::<C>());
+    }
+
     pub const fn new() -> Self {
         Self::check_byte_chunk_type_is_aligned();
+        Self::check_inner_type_is_aligned();
+
         Self {
             inner: Vec::new(),
             len: 0,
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Note that I have not verified if any element type/chunk type combinations could fail the requirement in practice beyond a cursory review of some instantiations. Among them, `u8` and `RefMvsBlock` and `RefMvsTemporalBlock` all use the highly over-aligned `64` variant and are absolutely sound. As documented it wouldn't be useful, so it's unlikely. This new code just verifies it.

src/align.rs
```diff
@@ -172,13 +178,17 @@ impl<T: Copy, C: AlignedByteChunk> AlignedVec<T, C> {
     pub fn as_slice(&self) -> &[T] {
         // SAFETY: The first `len` elements have been
         // initialized to `T`s in `Self::resize_with`.
+        // SAFETY: the pointer is sufficient aligned, as the chunks are always over-aligned with
+        // respect to `T`.
```
kkysen, line 182: Nit (typo in `sufficiently`):

```suggestion
        // SAFETY: The pointer is sufficiently aligned,
        // as the chunks are always over-aligned with respect to `T`.
```

Can you do this for the other comments, too?

kkysen (approved):
Thanks for finding this! I guess we forgot to check that alignment requirement.  Could you just adjust the comments a bit and fix that typo in `sufficiently`?

kkysen (approved):

Issue #1371, opened 2025-03-11
Backport updates from dav1d 1.5.0


kkysen:

fbossen:
> [@fbossen](https://github.com/fbossen), is this done now? Can you close it if it is done?

This isn't fully done yet. Asm for `x86` and `arm` has been updated though.

PR #1377, opened 2025-03-25
Backport x86 asm changes from dav1d 1.5.0

fbossen:
* Relates to #1371

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -174,7 +174,6 @@ if is_asm_enabled
             'x86/itx_avx512.asm',
             'x86/cdef_avx2.asm',
             'x86/itx_avx2.asm',
-            'x86/looprestoration_avx2.asm',
             'x86/cdef_sse.asm',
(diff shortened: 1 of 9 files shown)


src/meson.build
kkysen, line 1: Do you need to update this in the `build.rs`, too?

kkysen (commented):
The Rust changes look correct so I'm not sure why the tests are failing, but maybe you need to update the `build.rs` in accordance with the `meson.build` change?

src/meson.build
fbossen: Good point

kkysen (approved):
LGTM.  The changes all seem good to me now.  That said, I'm still hesitant since CI is failing, and CI didn't catch the missing changes in `build.rs` before.  So ideally, we'd fix CI before merging this (and the other backport PRs), although this one specifically is probably simple enough that it could be merged early if we want.

PR #1380, opened 2025-04-03
`mod refmvs`: backport changes from dav1d 1.5.0

fbossen:
Backport changes in `refmvs` allocation but keep memory buffers separate for `RefMvsFrame::r` and `RefMvsFrame::rp_proj` as they use different data types.

### DIFF
--- a/src/lib.c
+++ b/src/lib.c
@@ -657,7 +657,7 @@ static COLD void close_internal(Dav1dContext **const c_out, int flush) {
         free(f->lf.level);
         free(f->lf.tx_lpf_right_edge[0]);
         free(f->lf.start_of_tile_row);
-        dav1d_refmvs_clear(&f->rf);
+        dav1d_free_aligned(f->rf.r);
         dav1d_free_aligned(f->lf.cdef_line_buf);
         dav1d_free_aligned(f->lf.lr_line_buf);
     }
--- a/src/refmvs.rs
+++ b/src/refmvs.rs
@@ -120,6 +120,8 @@ pub struct RefMvsFrame {
     pub mfmv_ref2ref: [[i32; 7]; 3],
     pub n_mfmvs: i32,
     pub n_blocks: u32,
+    // TODO: The C code uses a single buffer to store `rp_proj` and `r` to minimize
+    // the number of allocated buffers.
     pub rp_proj: DisjointMut<AlignedVec64<RefMvsTemporalBlock>>,
     pub rp_stride: u32,
     pub r: DisjointMut<AlignedVec64<RefMvsBlock>>,
--- a/src/refmvs.rs
+++ b/src/refmvs.rs
@@ -92,13 +92,12 @@ pub(crate) struct AsmRefMvsFrame<'a> {
     pub mfmv_ref2cur: [i32; 3],
     pub mfmv_ref2ref: [[i32; 7]; 3],
     pub n_mfmvs: i32,
+    pub n_blocks: i32,
     pub rp: *mut RefMvsTemporalBlock,
     pub rp_ref: *const *mut RefMvsTemporalBlock,
     pub rp_proj: *mut RefMvsTemporalBlock,
     pub rp_stride: isize,
     pub r: *mut RefMvsBlock,
-    pub r_stride: isize,
-    pub n_tile_rows: i32,
     pub n_tile_threads: i32,
     pub n_frame_threads: i32,
 }
@@ -120,11 +119,10 @@ pub struct RefMvsFrame {
     pub mfmv_ref2cur: [i32; 3],
     pub mfmv_ref2ref: [[i32; 7]; 3],
     pub n_mfmvs: i32,
+    pub n_blocks: u32,
     pub rp_proj: DisjointMut<AlignedVec64<RefMvsTemporalBlock>>,
     pub rp_stride: u32,
     pub r: DisjointMut<AlignedVec64<RefMvsBlock>>,
-    pub r_stride: u32,
-    pub n_tile_rows: u32,
     pub n_tile_threads: u32,
     pub n_frame_threads: u32,
 }
@@ -207,11 +205,10 @@ impl load_tmvs::Fn {
             mfmv_ref2cur,
             mfmv_ref2ref,
             n_mfmvs,
+            n_blocks,
             ref rp_proj,
             rp_stride,
             ref r,
-            r_stride,
-            n_tile_rows,
             n_tile_threads,
             n_frame_threads,
         } = *rf;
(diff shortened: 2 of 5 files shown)


src/refmvs.rs
```diff
@@ -120,11 +119,10 @@ pub struct RefMvsFrame {
     pub mfmv_ref2cur: [i32; 3],
     pub mfmv_ref2ref: [[i32; 7]; 3],
     pub n_mfmvs: i32,
+    pub n_blocks: i32,
```
kkysen, line 122: I think this can be a `u32`, since I had made `r_stride` and `n_tile_rows` `u32`s already.

kkysen:
```suggestion
    pub n_blocks: u32,
```

src/refmvs.rs
```diff
@@ -239,13 +236,12 @@ impl load_tmvs::Fn {
             mfmv_ref2cur,
             mfmv_ref2ref,
             n_mfmvs,
+            n_blocks,
```
kkysen, line 239:
```suggestion
            n_blocks: n_blocks as _,
```

src/refmvs.rs
```diff
@@ -1582,40 +1578,39 @@ pub(crate) fn rav1d_refmvs_init_frame(
...
+    rf.ih4 = rf.ih8 << 1;
     rf.rp_stride = rp_stride;
-
-    rf.n_tile_rows = n_tile_rows;
     rf.n_tile_threads = n_tile_threads;
     rf.n_frame_threads = n_frame_threads;
+
+    if n_blocks != rf.n_blocks {
+        let uses_2pass = (n_frame_threads > 1) as usize;
+        // `mem::size_of::<refmvs_block>() == 12`,
```
kkysen, line 1600:
```suggestion
        // `mem::size_of::<RefMvsBlock>() == 12`,
```

Noticing this just now.  There's also 2 un-updated uses of `refmvs_block` in docs in `levels.rs` if you wanna fix those, too.

src/lib.c
```diff
@@ -657,7 +657,7 @@ static COLD void close_internal(Dav1dContext **const c_out, int flush) {
         free(f->lf.level);
         free(f->lf.tx_lpf_right_edge[0]);
         free(f->lf.start_of_tile_row);
-        dav1d_refmvs_clear(&f->rf);
+        dav1d_free_aligned(f->rf.r);
```
kkysen, line 660: Do you know why they removed the free of `rf.rp_proj`?  Our logic for this is now quite different, since it's using auto-derived `impl Drop`s, so I want to make sure that stays correct amidst this change.

src/refmvs.rs
```diff
@@ -1582,40 +1578,39 @@ pub(crate) fn rav1d_refmvs_init_frame(
...
+        // but it's accessed using 16-byte loads in asm,
+        // so add `R_PAD` elements to avoid buffer overreads.
+        // TODO fallible allocation
+        rf.r.resize(
+            35 * 2 * n_blocks as usize * (1 + uses_2pass) + R_PAD,
```
kkysen, line 1605: Could you refactor out `r_sz` like C has here now (except for the `sizeof` part, which we don't need)?

src/refmvs.rs
```diff
@@ -1582,40 +1578,39 @@ pub(crate) fn rav1d_refmvs_init_frame(
...
+            FromZeroes::new_zeroed(),
+        );
+        // TODO fallible allocation
+        rf.rp_proj
+            .resize(16 * n_blocks as usize, Default::default());
```
kkysen, line 1610: Same with `rp_proj_sz`.

src/refmvs.rs
```diff
@@ -1582,40 +1578,39 @@ pub(crate) fn rav1d_refmvs_init_frame(
...
+        // `mem::size_of::<refmvs_block>() == 12`,
+        // but it's accessed using 16-byte loads in asm,
+        // so add `R_PAD` elements to avoid buffer overreads.
+        // TODO fallible allocation
+        rf.r.resize(
+            35 * 2 * n_blocks as usize * (1 + uses_2pass) + R_PAD,
+            FromZeroes::new_zeroed(),
+        );
+        // TODO fallible allocation
+        rf.rp_proj
```
kkysen, line 1609: `rp_proj` is now a pointer to the latter section of `r`, so we're gonna need to do that in Rust somehow, too.

src/refmvs.rs
```diff
@@ -1582,40 +1578,39 @@ pub(crate) fn rav1d_refmvs_init_frame(
...
-
-    rf.n_tile_rows = n_tile_rows;
     rf.n_tile_threads = n_tile_threads;
     rf.n_frame_threads = n_frame_threads;
+
+    if n_blocks != rf.n_blocks {
+        let uses_2pass = (n_frame_threads > 1) as usize;
+        // `mem::size_of::<refmvs_block>() == 12`,
+        // but it's accessed using 16-byte loads in asm,
+        // so add `R_PAD` elements to avoid buffer overreads.
```
kkysen, line 1602: Can you add in the slight comment changes from C, too?

kkysen (changes requested):

src/refmvs.rs
kkysen: Oh I see you said in the PR description we'll leave them separate for now.  Could you leave that as a TODO comment on the `rp_proj` field then?  It might be possible to do with `zerocopy` stuff, but we'll have to see if it's worth it.

randomPoison (approved):

src/refmvs.rs
```diff
@@ -120,11 +119,12 @@ pub struct RefMvsFrame {
     pub mfmv_ref2cur: [i32; 3],
     pub mfmv_ref2ref: [[i32; 7]; 3],
     pub n_mfmvs: i32,
+    pub n_blocks: u32,
+    // TODO: The C code uses a single buffer to store `rp_proj` and `r` to minimize
```
kkysen, line 123: Can we leave an open issue for this?

src/refmvs.rs
```diff
@@ -1582,40 +1580,35 @@ pub(crate) fn rav1d_refmvs_init_frame(
...
     rf.rp_stride = rp_stride;
-
-    rf.n_tile_rows = n_tile_rows;
     rf.n_tile_threads = n_tile_threads;
     rf.n_frame_threads = n_frame_threads;
+
+    if n_blocks != rf.n_blocks {
+        // `mem::size_of::<RefMvsBlock>() == 12`,
+        // but it's accessed using 16-byte unaligned loads in save_tmvs() asm,
+        // so add `R_PAD` elements to avoid buffer overreads.
```
kkysen, line 1603:
```suggestion
        // so add `R_PAD` elements to avoid buffer overreads.
        // The C code stores `rp_proj` in the same buffer to reduce allocations,
        // which also allows them to use `rp_proj` as the extra padding.
        // TODO combine `r` and `rp_proj` like C does.
```

kkysen (approved):
LGTM now, except for a little comment thing, and we should open an issue for combining `r` and `rp_proj` if it's worth the performance.

PR #1381, opened 2025-04-03
Backport data size checks from dav1d 1.5.0

fbossen:
* Part of #1371

### DIFF
--- a/src/refmvs.rs
+++ b/src/refmvs.rs
@@ -36,12 +36,16 @@ pub struct RefMvsTemporalBlock {
     pub mv: Mv,
     pub r#ref: i8,
 }
+const _: () = assert!(mem::size_of::<RefMvsTemporalBlock>() == 5);
 
 #[derive(Clone, Copy, PartialEq, Eq, FromZeroes)]
-#[repr(C)]
+// In C, this is packed and is 2 bytes.
+// In Rust, being packed and aligned is tricky
+#[repr(C, align(2))]
 pub struct RefMvsRefPair {
     pub r#ref: [i8; 2],
 }
+const _: () = assert!(mem::size_of::<RefMvsRefPair>() == 2);
 
 impl From<[i8; 2]> for RefMvsRefPair {
     fn from(from: [i8; 2]) -> Self {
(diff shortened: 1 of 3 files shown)


src/refmvs.rs
```diff
@@ -54,6 +56,7 @@ impl From<[i8; 2]> for RefMvsRefPair {
 pub struct RefMvsMvPair {
     pub mv: [Mv; 2],
 }
+const _: () = assert!(mem::size_of::<RefMvsTemporalBlock>() == 5);
```
kkysen, line 59: I think this one was accidentally copied. Should probably be the missing `RefMvsMvPair` check.

src/refmvs.rs
```diff
@@ -54,6 +56,7 @@ impl From<[i8; 2]> for RefMvsRefPair {
...
 
 #[derive(Clone, Copy, FromZeroes)]
 #[repr(C, align(4))]
```
kkysen, line 62: We already have the `RefMvsBlock` size check below, but could you move the comment to the type now.  Previously, I had added the size check in Rust (it wasn't in C yet) since I didn't pack it, since Rust doesn't let you, so we had to manually ensure it was packed.

src/refmvs.rs
```diff
@@ -36,12 +36,14 @@ pub struct RefMvsTemporalBlock {
...
     pub r#ref: i8,
 }
+const _: () = assert!(mem::size_of::<RefMvsTemporalBlock>() == 5);
 
 #[derive(Clone, Copy, PartialEq, Eq, FromZeroes)]
 #[repr(C)]
 pub struct RefMvsRefPair {
     pub r#ref: [i8; 2],
 }
+const _: () = assert!(mem::size_of::<RefMvsRefPair>() == 2);
```
kkysen, line 46: Can you add a similar comment to the one on `RefMvsTemporalBlock` that this is packed, as verified by the size check, but we can't do both packed and aligned in Rust?

src/refmvs.rs
```diff
@@ -36,12 +36,14 @@ pub struct RefMvsTemporalBlock {
     pub mv: Mv,
     pub r#ref: i8,
 }
+const _: () = assert!(mem::size_of::<RefMvsTemporalBlock>() == 5);
 
 #[derive(Clone, Copy, PartialEq, Eq, FromZeroes)]
 #[repr(C)]
```
kkysen, line 42:
```suggestion
#[repr(C, align(2))]
```

kkysen (changes requested):

randomPoison (approved):

kkysen (approved):

PR #1382, opened 2025-04-04
Backport ARM asm from dav1d 1.5.0

fbossen:
* Part of #1371.

Adds some ARM SVE2 assembly.

### DIFF
--- a/src/mc.rs
+++ b/src/mc.rs
@@ -2306,11 +2306,7 @@ impl Rav1dMCDSPContext {
         self.emu_edge = bd_fn!(emu_edge::decl_fn, BD, emu_edge, neon);
 
         #[cfg(all(target_arch = "aarch64", feature = "asm_arm64_dotprod"))]
-        if BD::BITDEPTH == 8 {
-            if !flags.contains(CpuFlags::DOTPROD) {
-                return self;
-            }
-
+        if BD::BITDEPTH == 8 && flags.contains(CpuFlags::DOTPROD) {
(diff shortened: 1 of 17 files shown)


src/mc.rs
```diff
@@ -2369,6 +2369,37 @@ impl Rav1dMCDSPContext {
             });
         }
 
+        #[cfg(all(target_arch = "aarch64", feature = "asm_arm64_sve2"))]
+        if BD::BITDEPTH == 16 {
+            if !flags.contains(CpuFlags::SVE2) {
+                return self;
+            }
```
kkysen, line 2376:
```suggestion
        if BD::BITDEPTH == 16 && flags.contains(CpuFlags::SVE2) {
```

kkysen: Can you do the same changes to `DOTPROD` and `I8MM` so that it matches the `mc.h` logic so we stay synced as similar as possible?

kkysen (changes requested):

src/mc.rs
kkysen: It looks like the C is changed to me: 

```c
    if (flags & DAV1D_ARM_CPU_FLAG_DOTPROD) {
        init_8tap_fns(neon_dotprod);
    }
```

fbossen: Indeed. Must have been looking at the wrong version of the C file.

randomPoison (approved):

kkysen (approved):

PR #1383, opened 2025-04-18
What if we try an older ubuntu image

randomPoison:
couldn't hurt, right?

### DIFF
--- a/.github/workflows/build-and-test-qemu.yml
+++ b/.github/workflows/build-and-test-qemu.yml
@@ -29,7 +29,7 @@ jobs:
             linker: "riscv64-linux-gnu-gcc"
             wrapper: "qemu-riscv64-static"
             packages: "g++-riscv64-linux-gnu libc6-dev-riscv64-cross"
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


randomPoison:
at some point we'll want to find out why ubuntu-latest breaks things, but this should be enough to get us unblocked for now.

.github/workflows/build-and-test-qemu.yml
```diff
@@ -68,4 +68,4 @@ jobs:
         with:
           name: meson-test-logs-${{ matrix.target }}
           path: |
-             ${{ github.workspace }}/build/meson-logs/testlog.txt
\ No newline at end of file
+             ${{ github.workspace }}/build/meson-logs/testlog.txt
```
kkysen, line 71: Was this just a missing trailing newline?

kkysen (approved):
Seems good to unblock things. We can figure out what the actual cause is later.

.github/workflows/build-and-test-qemu.yml
randomPoison: Yeah, looks like I added the trailing newline while I was working on this 🤷

Issue #1372, opened 2025-03-11
Backport updates from dav1d 1.5.1


kkysen:

kkysen:
@fbossen, is this done now? Can you close it if it is done?

fbossen:
> [@fbossen](https://github.com/fbossen), is this done now? Can you close it if it is done?

This isn't fully done yet. Asm was updated for `x86` and `aarch64`. I don't expect to see a noticeable performance difference from the remaining patches on `x86` and `aarch64`.

PR #1385, opened 2025-04-22
Backport x86 asm from dav1d 1.5.1

fbossen:
* Part of #1372.


randomPoison (approved):

src/x86/mc.h
kkysen, line 1: We don't need any corresponding changes for this? This is a non-functional change, right?

kkysen: Should we do the same refactorings?  Not sure (I forget how we set up our macros and stuff).

PR #1392, opened 2025-05-14
ci: ensure every CI workflow caches and explicitly installs a `rustup` toolchain

### LINKED ISSUE #1391: Missing explicit toolchain install in `benchmark-on-self-hosted-x86-64`
We're getting CI failures in `benchmark-on-self-hosted-x86-64`: https://github.com/memorysafety/rav1d/actions/runs/14975453720/job/42066489695#step:4:42 (from #1390).

```shell
error[E0432]: unresolved import `std::ptr::fn_addr_eq`
  --> src/log.rs:10:5
   |
10 | use std::ptr::fn_addr_eq;
   |     ^^^^^^^^^^----------
   |     |         |
   |     |         help: a similar name exists in the module: `addr_eq`
   |     no `fn_addr_eq` in `ptr`

error[E0[43](https://github.com/memorysafety/rav1d/actions/runs/14975453720/job/42066489695#step:4:44)2]: unresolved import `std::ptr::fn_addr_eq`
  --> src/picture.rs:36:5
   |
36 | use std::ptr::fn_addr_eq;
   |     ^^^^^^^^^^----------
   |     |         |
   |     |         help: a similar name exists in the module: `addr_eq`
   |     no `fn_addr_eq` in `ptr`
```

I think this is because the self-hosted benchmarker is still on the old nightly, which doesn't have `std::ptr::fn_addr_eq`, and we don't explicitly install the toolchain specified in `rust-toolchain.toml` since `rustup` used to do that automatically.  We should just need to add an explicit toolchain installation step to this CI workflow, as I think we already added it to the other ones.

kkysen:
* Fixes #1391.

`rustup` no longer automatically installs a toolchain in `rust-toolchain.toml`, so we get errors like https://github.com/memorysafety/rav1d/issues/1391 now.  Moreover, we only use the pinned nightly for targets where CPU feature detection is unstable (`arm`, `riscv64`, `riscv32`).  Elsewhere we use `stable` for stability, so it helps to be more explicit about using the `stable` toolchain there.

This also adds some missing `rustup` toolchain caching steps.

Distinguishing between `stable` and the pinned nightly from `rust-toolchain.toml` is also quite tricky, so this switches to using `dtolnay/rust-toolchain` for that, which is also used by many `rust-lang` repos.

This also switches to using a pinned stable version for benchmarking for performance consistency.

Since I'm updating CI already, I also switched to `rust-cache` instead of our manual caching, as it's a lot smarter and more precise.  For example, in some places (our main CI workflow) we previously had to run `cargo clean` before `cargo build`, defeating the purpose of most of the caching.

`rust-cache` is also used by many `rust-lang` repos.

### DIFF
--- a/.github/workflows/build-and-benchmark-x86.yml
+++ b/.github/workflows/build-and-benchmark-x86.yml
@@ -10,12 +10,18 @@ jobs:
         steps:
             - name: git checkout
               uses: actions/checkout@v4
+            - name: install rust toolchain
+              id: toolchain
+              uses: dtolnay/rust-toolchain@master
+              with:
+                toolchain: 1.87.0 # specific version for consistent perf
+                target: x86_64-unknown-linux-gnu
             - name: release build this PR
-              run:  nice cargo build --release
+              run: nice cargo +${{ steps.toolchain.outputs.name }} build --release
             - name: release build main branch
               run: |
                 git fetch --depth 1 origin main && git checkout origin/main
-                nice cargo +stable build --release --target-dir target.main
+                nice cargo +${{ steps.toolchain.outputs.name }} build --release --target-dir target.main
             - name: benchmark on chimera 8-bit test data
               run: |
                 mkdir -p `dirname $LOCAL_FILE`
--- a/.github/workflows/build-and-test-aarch64-android.yml
+++ b/.github/workflows/build-and-test-aarch64-android.yml
@@ -25,15 +25,7 @@ jobs:
           toolchain: stable
           target: aarch64-linux-android
       - name: cache rust dependencies
-        uses: actions/cache@v4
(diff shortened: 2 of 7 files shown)


djc (approved):
Looks okay to me.

It feels like there's a lot of duplication between the architecture-specific workflows. We should maybe look at trying to reduce that and instead use workflow features to branch on the architecture?

kkysen:
> It feels like there's a lot of duplication between the architecture-specific workflows. We should maybe look at trying to reduce that and instead use workflow features to branch on the architecture?

Yup, I'd definitely prefer that.  Didn't want to do that here in this PR though.

djc:
Sure, makes sense!

PR #1329, opened 2024-07-18
`struct {Rav1dContext,TaskThreadDataDelayedFg}`: Remove TODOed `unsafe impl Send + Sync`s

kkysen:
* Fixes #1177.
* Fixes #1270.

PR #1395, opened 2025-05-17
Run clippy in CI

### LINKED ISSUE #1343: Run `clippy` in CI
Now that we lint against `#![deny(clippy::undocumented_unsafe_blocks)]` and `#![deny(clippy::missing_safety_doc)]`:

* #1305
* #1306
* #1321

, we should run `cargo clippy` in CI to ensure that this is maintained.

djc:
I tried running clippy and got this:

```
error: unsafe impl missing a safety comment
   --> src/internal.rs:308:1
    |
308 | unsafe impl Send for TaskThreadDataDelayedFg {}
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: consider adding a safety comment on the preceding line
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#undocumented_unsafe_blocks
note: the lint level is defined here
   --> lib.rs:9:9
    |
9   | #![deny(clippy::undocumented_unsafe_blocks)]
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: unsafe impl missing a safety comment
   --> src/internal.rs:310:1
    |
310 | unsafe impl Sync for TaskThreadDataDelayedFg {}
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: consider adding a safety comment on the preceding line
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#undocumented_unsafe_blocks

error: unsafe impl missing a safety comment
   --> src/internal.rs:440:1
    |
440 | unsafe impl Send for Rav1dContext {}
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: consider adding a safety comment on the preceding line
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#undocumented_unsafe_blocks

error: unsafe impl missing a safety comment
   --> src/internal.rs:442:1
    |
442 | unsafe impl Sync for Rav1dContext {}
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: consider adding a safety comment on the preceding line
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#undocumented_unsafe_blocks
```

clippy is configured like this:

```
#![allow(clippy::all)]
#![deny(clippy::undocumented_unsafe_blocks)]
#![deny(clippy::missing_safety_doc)]
```

It would be nice if these explicitly enabled clippy lints at least blocked PR CI.

* Closes #1343


kkysen:
#1329 was supposed to fix the warnings, but we didn't have time to fix the issues with #1327 and then it languished.

.github/workflows/cargo-fmt.yml
```diff
@@ -32,3 +32,5 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - name: Check formatting
         run: cargo fmt --check
+      - name: Clippy
+        run: cargo clippy -- -D warnings
```
kkysen, line 36: We should probably rename the workflow or put this after a `cargo check`/`cargo build` as they have a lot of overlap.

djc:
> #1329 was supposed to fix the warnings, but we didn't have time to fix the issues with #1327 and then it languished.

Okay... this seems pretty important. Do you think we can get it fixed?

.github/workflows/cargo-fmt.yml
kkysen: This needs to go after a `cargo build` more than I thought, since it needs to build the code, so it needs things like `nasm`, which aren't necessary for `cargo fmt` and thus not installed in this workflow.

.github/workflows/build-and-test-aarch64-darwin.yml
```diff
@@ -33,6 +33,8 @@ jobs:
         uses: Swatinem/rust-cache@v2
       - name: cargo ${{ matrix.build.name }} build for aarch64-apple-darwin
         run:  cargo +${{ steps.toolchain.outputs.name }} build ${{ matrix.build.cargo_flags }}
+      - name: Clippy
+        run: cargo clippy -- -D warnings
```
kkysen, line 37: Can you run this on all platforms?  It's `cfg`-dependent.

kkysen (approved):

.github/workflows/build-and-benchmark-x86.yml
```diff
@@ -16,12 +16,15 @@ jobs:
               with:
                 toolchain: 1.87.0 # specific version for consistent perf
                 target: x86_64-unknown-linux-gnu
+                components: clippy
```
kkysen, line 19: It seems like this is not getting installed? Idk what's going on here.

djc: It was picking up the `rust-toolchain.toml` instead of the installed toolchain that's forced via the workflow.

djc:
Dropped running clippy in the Android workflow, which seems to fail. It looks like there's not a lot of Android-specific code anyway, so I think it should be fine to land this -- can always expand with more coverage.

kkysen (commented):
> Dropped running clippy in the Android workflow, which seems to fail. It looks like there's not a lot of Android-specific code anyway, so I think it should be fine to land this -- can always expand with more coverage.

Oh, it might be that the android target doesn't have clippy, since it's a tier 2 target without host tools.

PR #1397, opened 2025-05-18
perf: arm: don't init `tmp_buf` before calling padding and filter funcs

ohadravid:
Hi,

Took a look at the `rav1d` vs `dav1d` perf contest, and was able to find a small (~1%) improvement (as measured on my M3 MacBook Air).

This PR includes two changes:
1. In the `cdef_filter_neon_erased` call, replace `Align16([0; TMP_LEN])` with `Align16([MaybeUninit::uninit(); TMP_LEN])`.
2. In `rav1d_cdef_brow`, move up the `let mut lr_bak = Align16([..])` initialization to be outside the loop. (now in #1399 )

I believe these should be safe and correct since the C version doesn't seem to initialize `ALIGN_STK_16`-created data, but this should definitely be double-checked by someone more familiar with the code 🙏 

These changes are the result of comparing runs of `dav1d` and `rav1d`, which showed:

1. `cdef_filter_neon_erased` had a much larger "self" time vs. `cdef_filter_8x8_neon`+`cdef_filter_4x4_neon`, which seems to be due to a `memset`.
3. Similarly, `rav1d::src::cdef_apply::rav1d_cdef_brow` has a slighly larger self time compared to `dav1d_cdef_brow_8bpc`.

### Local hyperfine output

`main` (commit a654c1e82adb2d9a33ae50d2a82a7a747102cbb6):
```
$ hyperfine --warmup 2 "target/release/dav1d -q -i Chimera-AV1-8bit-1920x1080-6736kbps.ivf -o /dev/null --threads 1"
Benchmark 1: target/release/dav1d -q -i Chimera-AV1-8bit-1920x1080-6736kbps.ivf -o /dev/null --threads 1
  Time (mean ± σ):     73.914 s ±  0.151 s    [User: 73.295 s, System: 0.279 s]
  Range (min … max):   73.770 s … 74.132 s    10 runs
```

this branch:
```
$ hyperfine --warmup 2 "target/release/dav1d -q -i Chimera-AV1-8bit-1920x1080-6736kbps.ivf -o /dev/null --threads 1"
Benchmark 1: target/release/dav1d -q -i Chimera-AV1-8bit-1920x1080-6736kbps.ivf -o /dev/null --threads 1
  Time (mean ± σ):     72.892 s ±  0.103 s    [User: 72.314 s, System: 0.269 s]
  Range (min … max):   72.736 s … 73.048 s    10 runs
```

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -510,10 +510,12 @@ fn cdef_find_dir_rust<BD: BitDepth>(
 #[deny(unsafe_op_in_unsafe_fn)]
 #[cfg(all(feature = "asm", any(target_arch = "arm", target_arch = "aarch64")))]
 mod neon {
+    use std::mem::MaybeUninit;
+
     use super::*;
 
     wrap_fn_ptr!(unsafe extern "C" fn padding(
-        tmp: *mut u16,
+        tmp: *mut MaybeUninit<u16>,
         src: *const DynPixel,
         src_stride: ptrdiff_t,
         left: *const [LeftPixelRow2px<DynPixel>; 8],
@@ -526,7 +528,7 @@ mod neon {
     impl padding::Fn {
         fn call<BD: BitDepth>(
             &self,
-            tmp: &mut [u16],
+            tmp: &mut [MaybeUninit<u16>],
             src: *const BD::Pixel,
             src_stride: ptrdiff_t,
             left: *const [LeftPixelRow2px<BD::Pixel>; 8],
@@ -557,7 +559,7 @@ mod neon {
     wrap_fn_ptr!(unsafe extern "C" fn filter(
         dst: *mut DynPixel,
         dst_stride: ptrdiff_t,
-        tmp: *const u16,
+        tmp: *const MaybeUninit<u16>,
         pri_strength: c_int,
         sec_strength: c_int,
         dir: c_int,
@@ -572,7 +574,7 @@ mod neon {
             &self,
             dst: *mut BD::Pixel,
             dst_stride: ptrdiff_t,
-            tmp: &[u16],
+            tmp: &[MaybeUninit<u16>],
             pri_strength: c_int,
             sec_strength: c_int,
             dir: c_int,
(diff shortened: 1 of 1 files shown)


kkysen (commented):
Thanks for finding these improvements!

Could you split the two commits into separate PRs?  The `lr_bak` one is simple and I'll merge that right away, but I want to look through the `fn padding` code more.

kkysen (changes requested):

src/cdef.rs
```diff
@@ -643,9 +645,14 @@ mod neon {
...
         let bd = BD::from_c(bitdepth_max);
 
-        let mut tmp_buf = Align16([0; TMP_LEN]);
+        let mut tmp_buf = Align16([MaybeUninit::uninit(); TMP_LEN]);
         let tmp = &mut tmp_buf.0[2 * TMP_STRIDE + 8..];
         padding::Fn::neon::<BD, W>().call::<BD>(tmp, dst, stride, left, top, bottom, H, edges);
+
+        // SAFETY: see `assume_init_ref` in the `maybe_uninit_slice` unstable feature,
+        // and the `padding` call guarantees that `tmp` is initialized.
+        let tmp = unsafe { &*(tmp as *const _ as *const [_]) };
```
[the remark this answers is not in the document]
ohadravid: Nice catch 😢 I wanted to avoid switching from the slice to a raw pointer but it's probably more resilient to asm changes to do it that way.

Let me know what you think: AFIK passing the raw pointer matches exactly with the semantics of the original C impl.

We can leave it as a `&[MaybeUninit<u16>]` and only convert to a raw pointer inside the `filter` call (adds a little bit of theoretical safety since we get both bound checking and requires an explicit cast before using)

ohadravid:
@kkysen changed the code to always pass around `MaybeUninit` (slice/raw ptr) to avoid the unsoundness, as a separate commit for now, let me know how/if you prefer a different approach 🐛

src/cdef.rs
```diff
@@ -643,9 +645,10 @@ mod neon {
         let bottom = bottom.cast();
         let bd = BD::from_c(bitdepth_max);
 
-        let mut tmp_buf = Align16([0; TMP_LEN]);
+        let mut tmp_buf = Align16([MaybeUninit::uninit(); TMP_LEN]);
         let tmp = &mut tmp_buf.0[2 * TMP_STRIDE + 8..];
         padding::Fn::neon::<BD, W>().call::<BD>(tmp, dst, stride, left, top, bottom, H, edges);
+
```
kkysen, line 651:
```suggestion
```

Nit: Don't really need this newline anymore.

src/cdef.rs
```diff
@@ -643,9 +645,10 @@ mod neon {
         let bottom = bottom.cast();
         let bd = BD::from_c(bitdepth_max);
 
-        let mut tmp_buf = Align16([0; TMP_LEN]);
+        let mut tmp_buf = Align16([MaybeUninit::uninit(); TMP_LEN]);
```
kkysen, line 648:
```suggestion
        // Use `MaybeUninit` here to avoid over-initialization.
        // C doesn't initialize this either and only partially initializes it in `padding`.
        // Since we're just passing this to a few asm calls that are `unsafe` anyways,
        // initializing this in Rust doesn't really add any extra safety.
        let mut tmp_buf = Align16([MaybeUninit::uninit(); TMP_LEN]);
```

kkysen:
> @kkysen changed the code to always pass around `MaybeUninit` (slice/raw ptr) to avoid the unsoundness, as a separate commit for now, let me know how/if you prefer a different approach 🐛

Can you squash those two commits together so we don't have an unsound one?  Thanks!

After adding the docs, too, this should be good to go.

kkysen (approved):

PR #1399, opened 2025-05-18
perf: Only init `lr_bak` once (hoisted from the loop) in `fn rav1d_cdef_brow`

ohadravid:
Split into a separate PR from #1397 @kkysen

### DIFF
--- a/src/cdef_apply.rs
+++ b/src/cdef_apply.rs
@@ -173,6 +173,12 @@ pub(crate) fn rav1d_cdef_brow<BD: BitDepth>(
     let uv_stride: ptrdiff_t = BD::pxstride(f.cur.stride[1]);
 
     let mut bit = false;
+
+    // In C, this is declared uninitialized inside the loop,
+    // so it doesn't matter what it is initialized to in Rust as long as it is initialized,
+    // which means we can hoist it out of the loop to safely avoid re-initializing it.
+    let mut lr_bak = Align16([[[[0.into(); 2 /* x */]; 8 /* y */]; 3 /* plane */ ]; 2 /* idx */]);
+
     for by in (by_start..by_end).step_by(2) {
         let tf = tc.top_pre_cdef_toggle != 0;
         let by_idx = (by & 30) >> 1;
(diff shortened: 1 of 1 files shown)


src/cdef_apply.rs
```diff
@@ -173,6 +173,9 @@ pub(crate) fn rav1d_cdef_brow<BD: BitDepth>(
     let uv_stride: ptrdiff_t = BD::pxstride(f.cur.stride[1]);
 
     let mut bit = false;
+
+    let mut lr_bak = Align16([[[[0.into(); 2 /* x */]; 8 /* y */]; 3 /* plane */ ]; 2 /* idx */]);
```
kkysen, line 177:
```suggestion
    // In C, this is declared uninitialized inside the loop,
    // so it doesn't matter what it is initialized in Rust as long as it is initialized,
    // which means we can hoist it out of the loop to safely avoid re-initializing it.
    let mut lr_bak = Align16([[[[0.into(); 2 /* x */]; 8 /* y */]; 3 /* plane */ ]; 2 /* idx */]);
```

kkysen (approved):
The code itself LGTM, although can you add that comment explaining it a bit?  Thanks again for finding this!

src/cdef_apply.rs
```diff
@@ -173,6 +173,12 @@ pub(crate) fn rav1d_cdef_brow<BD: BitDepth>(
     let uv_stride: ptrdiff_t = BD::pxstride(f.cur.stride[1]);
 
     let mut bit = false;
+
+    // In C, this is declared uninitialized inside the loop,
+    // so it doesn't matter what it is initialized in Rust as long as it is initialized,
```
kkysen, line 178:
```suggestion
    // so it doesn't matter what it is initialized to in Rust as long as it is initialized,
```

Sorry, I made a typo.

kkysen (changes requested):

src/cdef_apply.rs
ohadravid: Fixed 😄

kkysen (approved):

PR #1404, opened 2025-05-20
Move crate root to idiomatic location

djc:


meson.build
```diff
@@ -46,7 +46,7 @@ cdata = configuration_data()
 cdata_asm = configuration_data()
 
 # Include directories
-dav1d_inc_dirs = include_directories(['.', 'include/dav1d', 'include'])
+dav1d_inc_dirs = include_directories(['.', 'src/include/dav1d', 'src/include'])
```
kkysen, line 49: We do need to make sure that `git cherry-pick` works on `dav1d` commits that touch C files.  I'm not sure if git can handle this rename (it might, not sure, though).  If it can't, we should leave the C files where they are and just move the Rust files.

djc:
```
djc-2021 reroot rav1d $ git cherry-pick 7d4b789f55389dad1820d6caf6a650038dad06e2
Auto-merging src/include/common/attributes.h
[reroot 36236a45] Mark C globals with small code model
 Author: Pranav Kant <prka@google.com>
 Date: Fri Feb 21 15:55:00 2025 +0000
 2 files changed, 13 insertions(+)
```

7d4b789 is the latest commit from dav1d/master that touches `include/`. Seems okay?

src/lib.rs
```diff
@@ -1,5 +1,187 @@
...
+pub(crate) mod levels;
+mod lf_apply;
+mod lf_mask;
+pub(crate) mod log;
+mod loopfilter;
+mod looprestoration;
+mod lr_apply;
+mod mc;
+#[path = "mem.rs"]
+mod memmod;
```
kkysen, line 159: What's this for?

src/lib.rs
```diff
@@ -1,5 +1,187 @@
...
+mod qm;
+mod recon;
+mod refmvs;
+mod scan;
+mod tables;
+mod thread_task;
+mod warpmv;
+mod wedge;
+
+pub use error::Dav1dResult;
```
kkysen, line 173:
```suggestion
pub use crate::error::Dav1dResult;
```

We've been trying to be consistent about using `crate::` so there's less ambiguity.

kkysen: Although I think `lib.rs` previously lacked a `crate::` here, too.

kkysen (approved):
This mostly looks good I think.  Seems like `git cherry-pick`s from `dav1d` should continue to work.  Just a few small comments left.

Actually, let me know when you merge it, just because there may be a bunch of conflicts.

src/lib.rs
djc: I don't understand the ambiguity issue. It's not like there is an `error` crate, nor are there multiple `error` modules? It seems more idiomatic to me import directly from the module that is declared in the same scope.

src/lib.rs
kkysen: Ahh, I see.  That's annoying.

Could we just rename it to `pool.rs` then?  It just contains `MemPool`.

src/lib.rs
kkysen: I don't mean `error` specifically is ambiguous, but for example, `mem` is, and referring to the module directly without going through `crate::` means that it'd change if you moved it to another module.

src/lib.rs
djc: Yup, good idea. Done.

lib.rs
kkysen, line 1: I think this might've gotten added back when you rebased.

djc:
IIRC we discussed this (presumably in Discord) and @kkysen preferred keeping them together? I don't have a strong preference where the C source resides although I do like having the `.c` and `.rs` next to each other. It's unfortunate that the `checkasm` target isn't exercised in CI...

PR #1407, opened 2025-05-21
Be stricter about sequence and frame header presence

iximeow:
this is a rather broad change, but the overall intention is to move `f.seq_hdr.as_ref().unwrap()` and `f.frame_hdr.as_ref().unwrap()` as far up the call tree as the highest unconditional execution of that statement. in most cases these unwraps are truly dead due to their equivalent execution by some caller several frames up, and in some functions (like decode_b and decode_sb) these are redundantly checked many times.

there doesn't seem to be a substantial effect on runtime or instruction count here, but this does delete many unreachable `unwrap()`s and might help legibility some. this also reduces a release build's size by about 2.6kb.

i didn't see any point that branches the `Some`-ness of these `Arc`s, which is particularly interesting. looks like that's all intended to be decided somewhere higher up when scheduling work? it's not exactly clear to me what a principled boundary would look like - "functions in `thread_task.rs` may not rely on header presence, called functions in the rest of `rav1d` should be able to"?

this is a pretty rough patch as-is, mostly wondering what your interest level is here! there's a somewhat more meaningful change to be had by grouping `{Rav1dContext, Rav1dTaskContext, Rav1dFrameData(WithHeaders)}` together into a single parameter, because the combination of number of parameters, number of calls, and inner calls not being tail calls, does add up to a mild (like.. ~0.5%) amount of overall overhead.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -68,6 +68,7 @@ use crate::src::internal::Rav1dFrameContext;
 use crate::src::internal::Rav1dFrameContextFrameThread;
 use crate::src::internal::Rav1dFrameContextLf;
 use crate::src::internal::Rav1dFrameData;
+use crate::src::internal::Rav1dFrameDataWithHeaders;
 use crate::src::internal::Rav1dState;
 use crate::src::internal::Rav1dTaskContext;
 use crate::src::internal::Rav1dTileState;
@@ -279,7 +280,7 @@ fn read_mv_residual(ts_c: &mut Rav1dTileStateContext, ref_mv: &mut Mv, mv_prec: 
 
 fn read_tx_tree(
     t: &mut Rav1dTaskContext,
-    f: &Rav1dFrameData,
+    f: &Rav1dFrameDataWithHeaders,
     ts_c: &mut Rav1dTileStateContext,
     from: TxfmSize,
     depth: c_int,
@@ -417,7 +418,7 @@ fn neg_deinterleave(diff: u8, r#ref: SegmentId, max: u8) -> u8 {
 }
 
 fn find_matching_ref(
-    f: &Rav1dFrameData,
+    f: &Rav1dFrameDataWithHeaders,
     t: &Rav1dTaskContext,
     intra_edge_flags: EdgeFlags,
     bw4: c_int,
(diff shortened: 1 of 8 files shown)


src/internal.rs
```diff
@@ -840,20 +840,108 @@ pub(crate) struct Rav1dFrameData {
...
+        unsafe {
+            std::mem::transmute(self)
+        }
     }
+}
+
+#[repr(C)]
+pub(crate) struct Rav1dFrameDataWithHeaders {
+    pub seq_hdr: Arc<DRav1d<Rav1dSequenceHeader, Dav1dSequenceHeader>>,
+    pub frame_hdr: Arc<DRav1d<Rav1dFrameHeader, Dav1dSequenceHeader>>,
```
iximeow, line 883: an obvious first improvement to this patch would be to extract all fields *other than* `seq_hdr` and `frame_hdr` into a struct on its own and share that between both forms of `Rav1dFrameData`

src/lr_apply.rs
```diff
@@ -154,7 +154,7 @@ fn lr_sbrow<BD: BitDepth>(
     let chroma = (plane != 0) as c_int;
     let ss_ver = chroma & (f.sr_cur.p.p.layout == Rav1dPixelLayout::I420) as c_int;
     let ss_hor = chroma & (f.sr_cur.p.p.layout != Rav1dPixelLayout::I444) as c_int;
-    let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
+    let frame_hdr = &f.frame_hdr;
```
iximeow, line 157: ... and an obviousish second step to improve this patch would be to clean up this kinda thing. emphasis on "do the broad strokes seem useful" :)

djc:
I think it's a great idea to statically clarify whether these values are expected to be there via the type system. I don't really love the elaborate `Rav1dDataFrameWithHeaders` name -- and maybe the "burden of ugly name" should be the other way?

src/decode.rs
```diff
@@ -4830,48 +4901,48 @@ pub(crate) fn rav1d_decode_frame_exit(
...
     }
 
-    let _ = mem::take(&mut f.cur_segmap);
-    let _ = mem::take(&mut f.prev_segmap);
-    let _ = mem::take(&mut f.mvs);
+    let _ = mem::take(&mut f.content.cur_segmap);
+    let _ = mem::take(&mut f.content.prev_segmap);
+    let _ = mem::take(&mut f.content.mvs);
     let _ = mem::take(&mut f.seq_hdr);
     let _ = mem::take(&mut f.frame_hdr);
```
iximeow, line 4944: any time the `seq_hdr`/`frame_hdr` fields are `take`n from like this, it would be absolutely critical to reconstruct the `Arc<T>` if these were `Some` so as to not leak the header refcount. seems pretty subtle and easy to mistake. might be less bad with a `Rav1dFrameDataMaybeData::drop_state()` or something, but ouch.

PR #1409, opened 2025-05-22
Use Box instead of Arc for itut_t35

jplatte:
Remove an unnecessary unwrap.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -381,7 +381,7 @@ pub struct Rav1dState {
     pub(crate) frame_hdr: Option<Arc<DRav1d<Rav1dFrameHeader, Dav1dFrameHeader>>>, // TODO(kkysen) Previously pooled.
     pub(crate) content_light: Option<Arc<Rav1dContentLightLevel>>,
     pub(crate) mastering_display: Option<Arc<Rav1dMasteringDisplay>>,
-    pub(crate) itut_t35: Arc<Mutex<Vec<Rav1dITUTT35>>>,
+    pub(crate) itut_t35: Box<Mutex<Vec<Rav1dITUTT35>>>,
 
     // decoded output picture queue
     pub(crate) in_0: Rav1dData,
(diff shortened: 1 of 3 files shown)


djc (approved):

src/internal.rs
```diff
@@ -381,7 +381,7 @@ pub struct Rav1dState {
     pub(crate) frame_hdr: Option<Arc<DRav1d<Rav1dFrameHeader, Dav1dFrameHeader>>>, // TODO(kkysen) Previously pooled.
     pub(crate) content_light: Option<Arc<Rav1dContentLightLevel>>,
     pub(crate) mastering_display: Option<Arc<Rav1dMasteringDisplay>>,
-    pub(crate) itut_t35: Arc<Mutex<Vec<Rav1dITUTT35>>>,
+    pub(crate) itut_t35: Box<Mutex<Vec<Rav1dITUTT35>>>,
```
kkysen, line 384:
```suggestion
    
    /// This is ref-counted in C, but since we first accumulate the `Vec<Rav1dITUTT35>`
    /// during  [`parse_obus`] and then convert to an [`Arc`] in [`Rav1dITUTT35::to_immut`],
    /// we don't need an [`Arc`] here.
    pub(crate) itut_t35: Box<Mutex<Vec<Rav1dITUTT35>>>,
```

kkysen, line 384: If `Arc` isn't needed (good find!), then we also don't need the `Mutex` or `Box` either, so it can just be a `Vec`.

kkysen (changes requested):

jplatte:
Addressed the review feedback. I have a question: Is `parse_obus` called with the same `Rav1dState` a lot? In that case, I'd have a small perf improvement PR I could follow up with.

kkysen (approved):
Thanks!

kkysen:
> Addressed the review feedback. I have a question: Is `parse_obus` called with the same `Rav1dState` a lot? In that case, I'd have a small perf improvement PR I could follow up with.

I think it's the same `Rav1dState` per `Rav1dContext`, which rarely changes, but `parse_obus` is only called once per picture/frame, so it's not super frequent (@fbossen, do you know if this is this right?).  You can open the PR, though, and I can take a look.  It might still help marginally or just make the code cleaner.

PR #1413, opened 2025-05-25
perf: Use drain + collect over mem::take + into_boxed_slice

jplatte:
This makes `to_immut` always allocate unlike the previous implementation, which (re)allocated whenever the capacity of `state.itut_t35` didn't match its length. However, this new implementation means that the `state.itut_t35` keeps hold of the previous `Vec` allocation, so subsequent calls to `parse_obus` with the same state don't have to allocate when pushing to `state.itut_t35` (up to the capacity previously allocated).

Here's the optimization I was talking about in #1409. I hope the explanation above makes sense. Unless `to_immut` is usually / always called with `Vec`s that are fully filled, this should reduce the number of calls to the allocator, and thus improve performance. Unfortunately, runtime variance for the benchmarks is too high on my notebook for me to be able to detect any perf changes with them (plus this might also be small enough not to register on less noisy systems). Let me know what you think :)

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -5135,9 +5135,6 @@ pub fn rav1d_submit_frame(c: &Rav1dContext, state: &mut Rav1dState) -> Rav1dResu
 
     // allocate frame
 
-    // We must take itut_t35 out of the context before the call so borrowck can
-    // see we mutably borrow `c.itut_t35` disjointly from the task thread lock.
-    let itut_t35 = mem::take(&mut state.itut_t35);
     let res = rav1d_thread_picture_alloc(
         &c.fc,
         &c.logger,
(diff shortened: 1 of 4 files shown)


src/decode.rs
```diff
@@ -5135,9 +5135,6 @@ pub fn rav1d_submit_frame(c: &Rav1dContext, state: &mut Rav1dState) -> Rav1dResu
 
     // allocate frame
 
-    // We must take itut_t35 out of the context before the call so borrowck can
-    // see we mutably borrow `c.itut_t35` disjointly from the task thread lock.
-    let itut_t35 = mem::take(&mut state.itut_t35);
```
jplatte, line 5140: AFAICT this comment was stale, since it compiles just fine to borrow `state.itut_t35` for the following function call rather than `mem::take`ing its value. I also don't see what "the task thread lock" could refer to here.

kkysen (commented):
@fbossen, correct me if I'm wrong, but from my understanding, ITU-T T.35s are uncommon, and there are rarely more than one of them, because the `dav1d` code for this used an `Option` for a while.  Most of the tests, for example, have 0, a few have 1, and one has 2.

So I'm not sure it's worth bothering to change the allocation behavior here since the numbers are so small.  It's probably better to use something like `tinyvec`, since there are usually just 0 or 1 itut t35s.

PR #1414, opened 2025-05-25
Replace most uses of `let _ = mem::take(_)`

jplatte:
This is a pretty weird way of writing the `Default` value to a place.

Should improve perf (probably just marginally) because it removes a few `.clone()` calls.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -4428,8 +4428,8 @@ pub(crate) fn rav1d_decode_frame_init(c: &Rav1dContext, fc: &Rav1dFrameContext) 
                 .pal_idx
                 .resize(pal_idx_sz as usize * 128 * 128 / 8, Default::default());
         } else if !f.frame_thread.pal.is_empty() {
-            let _ = mem::take(&mut f.frame_thread.pal);
-            let _ = mem::take(&mut f.frame_thread.pal_idx);
+            f.frame_thread.pal = Default::default();
+            f.frame_thread.pal_idx = Default::default();
         }
     }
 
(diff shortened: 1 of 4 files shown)


djc (approved):
Nice! I don't love `Default::default()` -- using the actual type name provides a clearer idea of what's going on without an increase in tokens.

jplatte:
The exact type is sometimes a bit complex like `[SomeRav1dType; 7]` in one case IIRC. I'm happy to still do it of course. Should I elide any array lenghts? ^^

jplatte:
I force-pushed to get rid of that one clone call I missed. Anything else I should do before this can merge?

kkysen (commented):
Sorry I missed reviewing this more earlier.

> This is a pretty weird way of writing the `Default` value to a place.

That's not what most of these are meant to do.  They're essentially `drop`s, but not yet refactored to be more Rusty so that they can actually be a `drop`.

kkysen (changes requested):
I think removing the clones is helpful and should marginally improve performance, but I think we should leave the `mem::take`s.  They'll be the same performance wise, and `mem::take` much more accurately describes the intent of `drop`ping and should be simpler to refactor into real `drop`s in the future.

djc:
> The exact type is sometimes a bit complex like `[SomeRav1dType; 7]` in one case IIRC. I'm happy to still do it of course. Should I elide any array lenghts? ^^

Yes, eliding the array lengths is fine. Still think just `Default` is a bit too abstract.

> I think removing the clones is helpful and should marginally improve performance, but I think we should leave the `mem::take`s. They'll be the same performance wise, and `mem::take` much more accurately describes the intent of `drop`ping and should be simpler to refactor into real `drop`s in the future.

What do you mean by "refactor into real `drop`s"? IMO the `let _ = mem::take(_)` way of writing things is definitely unidiomatic: if you don't need the value, there's no need to take it. I would agree that the dropping of the former value isn't super explicit in the new spelling, but at least it looks more idiomatic.

kkysen:
> IMO the `let _ = mem::take(_)` way of writing things is definitely unidiomatic: if you don't need the value, there's no need to take it. I would agree that the dropping of the former value isn't super explicit in the new spelling, but at least it looks more idiomatic.

Would `drop(mem::take(_))` be clearer?  We're taking away the value and dropping it.

> What do you mean by "refactor into real `drop`s"? 

For example, in `fn rav1d_decode_frame_exit`, most fields of `f` are dropped.  With some refactoring, we may be able to just drop the entire `f`.

jplatte:
Feel free to cherry-pick any changes from this PR. I don't think I'm going to rebase it.

Issue #1394, opened 2025-05-16
prepare_intra_edges is slower in rav1d than dav1d


jrmuizel:
Specifically this loop:

```rust
            for i in 0..px_have {
                left[sz - 1 - i] = *(dst + (i as isize * stride - 1)).index::<BD>();
            }
```
```C
            for (int i = 0; i < px_have; i++)
                left[sz - 1 - i] = dst[PXSTRIDE(stride) * i - 1];
```

Rust version

```
57      lea rcx, qword [rax + r13 * 1]
108     cmp rcx, rdx
        jnb 0xfdb8d
15      lea rcx, qword [r10 + rbx * 1]
59      cmp rcx, rsi
        jnb 0xfdba0
56      movzx ecx, byte [rbp + rax * 1]
502     mov byte [r11 + rbx * 1], cl
161     dec rbx
26      add rax, r15
48      cmp r9, rbx
        jnz 0xfd500
```

C version

```
9       movzx eax, byte [rdi + r11 * 1]
316     lea ebp, dword [rcx + 0x1]
18      movsxd rbp, ebp
17      mov byte [r15 + rbp * 1], al
124     movzx eax, byte [rdi + rsi * 1]
159     movsxd rcx, ecx
24      mov byte [r15 + rcx * 1], al
47      add rdi, rbx
37      add ecx, -0x2
23      add rdx, -0x2
        jnz 0x143e20
```

The Rust version has 1032 cycles vs C's 774

The C version has been unrolled once and doesn't have the bounds checks but I don't have any data as to why it's faster than the Rust version.

kkysen:
Thanks for finding this!  I agree this is somewhere we could definitely improve performance, although the use of `DisjointMut` and having to ensure we don't have overlapping `&mut`s makes this a lot more difficult, and I suspect is the main blocker to being optimized better here (though I haven't looked in detail).  If you can find a perf fix here that remains sound, that'd be great!

ERobsham:
Took a swing at a minor optimization for these loops: https://github.com/memorysafety/rav1d/pull/1415

Essentially, rather than taking the `DisjointImmutGuard` on every iteration of the loop, we're taking it just once on the range we're iterating over.

And due to the implicit bounds checking by taking the `slice()`, we can also explicitly opt out of any bounds checking safely.

PR #1415, opened 2025-05-25
perf: `fn rav1d_prepare_intra_edges`: pre-slice

ERobsham:
* Part of #1394.

rework `rav1d_prepare_intra_edges` to precalculate and take a slice of the data to be iterated over so we only incur the overhead of taking the `DisjointImmutGuard` once, which also allows us to safely use the unsafe `unsafe fn get_unchecked(..)` within the loop avoiding the overhead of any runtime bounds checks, since the `slice()` opertaion already implcitly bounds checked the slice we're asking for.

inital hyperfine results show a overall improvement (albeit fairly minisucle):
```
Benchmark 1: ./target/release/dav1d -q -i test-data/Chimera-AV1-10bit-1920x1080-6191kbps.ivf -o /dev/null
  Time (mean ± σ):      9.384 s ±  0.057 s    [User: 108.379 s, System: 6.500 s]
  Range (min … max):    9.342 s …  9.516 s    10 runs

Benchmark 1: ./target.main/release/dav1d -q -i test-data/Chimera-AV1-10bit-1920x1080-6191kbps.ivf -o /dev/null
  Time (mean ± σ):      9.490 s ±  0.076 s    [User: 110.062 s, System: 6.521 s]
  Range (min … max):    9.383 s …  9.606 s    10 runs
```
(changed version took ~98.88% of the original runtime)

### DIFF
--- a/src/ipred_prepare.rs
+++ b/src/ipred_prepare.rs
@@ -187,7 +187,7 @@ pub fn rav1d_prepare_intra_edges<BD: BitDepth>(
     let bitdepth = bd.bitdepth();
     let stride = dst.pixel_stride::<BD>();
     let is_neg_stride = 0 > stride;
-    let abs_stride = if is_neg_stride { (stride * -1) as usize } else { stride as usize };
+    let abs_stride = stride.unsigned_abs();
 
     match mode {
         VERT_PRED..=VERT_LEFT_PRED => {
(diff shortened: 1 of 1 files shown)


src/ipred_prepare.rs
```diff
@@ -245,8 +247,28 @@ pub fn rav1d_prepare_intra_edges<BD: BitDepth>(
...
-            for i in 0..px_have {
-                left[sz - 1 - i] = *(dst + (i as isize * stride - 1)).index::<BD>();
+            {
+                let slice_offset = if is_neg_stride {
+                    (px_have - 1) as isize * stride
+                } else {
+                    0
+                };
+                let dst_slice =
+                    &*(dst + slice_offset - 1isize).slice::<BD>((px_have - 1) * abs_stride + 1);
```
kkysen, line 257: The slice here overlaps ones from other threads, as the individual indices accessed are interleaved.  That's why `.index::<BD>()` works, but not a single `.slice::<BD>()`.  It's unfortunately a very tricky problem.

PR #1418, opened 2025-05-29
perf: `fn rav1d_cdef_brow`: reduce stack usage

ohadravid:
* Supersedes #1402.

## Summary

There seems to be a slowdown in some of the Arm asm functions due to increased stack usage in `rav1d_cdef_brow` (`def_filter4_pri_edged_8bpc_neon` is ~20% slower when called from rav1d).

By keeping `WithOffset`s "internal" to the function, LLVM is able to optimize away most of the `alloca`-s (first commit).

After doing that, it seemed that `backup2x8` was still slower when calling `.stride()` (vs. a version which doesn't have the Rust fallback args _at all_), which I think is related to how LLVM optimizes the entire `rav1d_cdef_brow` (second commit).

main:

(`$ cargo asm -p rav1d-cli --bin dav1d --llvm rav1d_cdef_brow 0`)

```llvm
; rav1d::src::cdef_apply::rav1d_cdef_brow
; Function Attrs: nounwind
define internal fastcc void @rav1d::src::cdef_apply::rav1d_cdef_brow(..) {
  %10 = alloca [16 x i8], align 8
  %11 = alloca [16 x i8], align 8
  %12 = alloca [16 x i8], align 8
  %13 = alloca [16 x i8], align 8
  %14 = alloca [16 x i8], align 8
  %15 = alloca [16 x i8], align 8
  %16 = alloca [16 x i8], align 8
  %17 = alloca [24 x i8], align 8
  %18 = alloca [24 x i8], align 8
  %19 = alloca [4 x i8], align 4
  %20 = alloca [96 x i8], align 16
  %21 = icmp sgt i32 %5, 0
  %22 = select i1 %21, i32 12, i32 8
  %23 = load ptr, ptr %3, align 8
  ..
```

This branch:
```llvm
define internal fastcc void @rav1d::src::cdef_apply::rav1d_cdef_brow(..) {
  %10 = alloca [16 x i8], align 8
  %11 = alloca [24 x i8], align 8
  %12 = alloca [24 x i8], align 8
  %13 = alloca [4 x i8], align 4
  %14 = alloca [96 x i8], align 16
```

~This branch:~
(This is the most "optimized" version in this sense, but couldn't get the same with the current Rust/Asm argument arrangement requirements:)

```llvm
; rav1d::src::cdef_apply::rav1d_cdef_brow
; Function Attrs: nounwind
define internal fastcc void @rav1d::src::cdef_apply::rav1d_cdef_brow(..) {
  %10 = alloca [16 x i8], align 8
  %11 = alloca [4 x i8], align 4
  %12 = alloca [96 x i8], align 16
  %13 = icmp sgt i32 %5, 0
  %14 = select i1 %13, i32 12, i32 8
  %15 = load ptr, ptr %3, align 8
  ..
```

## Full details

When comparing to `dav1d`, the `cdef_filter4_pri_edged_8bpc_neon` asm function seems to be ~20% slower when called from `rav1d`.

Looking at the per-instruction sample count, there's one with a big, consistent diff:

dav1d:
dav1d

rav1d:
rav1d

From this, I tried to look at the callers. The problem _seems_ to be that when the src buffer in `x13` is placed "far enough back" in the stack, the load stalls.

This fix is inspired by what I saw in https://github.com/rust-lang/rust/issues/141649 - since `WithOffset`s  seemed to match the `%10,..,%18` `alloca`-s in the IR above, I tried to force LLVM to do the right thing and optimize them away (I tried a few other things, but this is the most effective one).

With this fix, the diff in the sample count is gone, with a nice speedup:

``` bash
rav1d-pr % hyperfine --warmup 3 --runs 15 --parameter-list profile target/release/dav1d,target/release-baseline/dav1d  '{profile} -q -i ~/workspace/video_files_for_rav1d/Chimera-AV1-8bit-1920x1080-6736kb
ps.ivf -o /dev/null --threads 1'

Benchmark 1: target/release/dav1d -q -i Chimera-AV1-8bit-1920x1080-6736kbps.ivf -o /dev/null --threads 1
  Time (mean ± σ):     71.918 s ±  0.066 s    [User: 71.595 s, System: 0.233 s]
  Range (min … max):   71.812 s … 72.032 s    15 runs

Benchmark 2: target/release-baseline/dav1d -q -i Chimera-AV1-8bit-1920x1080-6736kbps.ivf -o /dev/null --threads 1
  Time (mean ± σ):     72.294 s ±  0.078 s    [User: 71.945 s, System: 0.235 s]
  Range (min … max):   72.183 s … 72.438 s    15 runs

release-baseline = 1be76ea9cbe70ae7cb4ffc9cb4a05971f82669ac
```

I'm not sure how much better this will be in x86, but it should still be faster.

### DIFF
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -19,6 +19,7 @@ use crate::include::common::bitdepth::LeftPixelRow2px;
 use crate::include::common::bitdepth::BPC;
 use crate::include::common::intops::apply_sign;
 use crate::include::common::intops::iclip;
+use crate::include::dav1d::picture::Rav1dPictureDataComponent;
 use crate::include::dav1d::picture::Rav1dPictureDataComponentOffset;
 use crate::pic_or_buf::PicOrBuf;
 use crate::strided::Strided as _;
@@ -55,9 +56,9 @@ wrap_fn_ptr!(pub unsafe extern "C" fn cdef(
     damping: c_int,
     edges: CdefEdgeFlags,
     bitdepth_max: c_int,
-    _dst: *const FFISafe<Rav1dPictureDataComponentOffset>,
-    _top: *const FFISafe<CdefTop>,
-    _bottom: *const FFISafe<CdefBottom>,
+    _dst: WithOffset<*const FFISafe<Rav1dPictureDataComponent>>,
+    _top: WithOffset<*const FFISafe<DisjointMut<AlignedVec64<u8>>>>,
+    _bottom: WithOffset<*const FFISafe<PicOrBuf<'_, AlignedVec64<u8>>>>,
 ) -> ());
 
 pub type CdefTop<'a> = WithOffset<&'a DisjointMut<AlignedVec64<u8>>>;
@@ -87,12 +88,23 @@ impl cdef::Fn {
         let left = ptr::from_ref(left).cast();
         let top_ptr = top.as_ptr::<BD>().cast();
         let bottom_ptr = bottom.wrapping_as_ptr::<BD>().cast();
-        let top = FFISafe::new(&top);
-        let bottom = FFISafe::new(&bottom);
         let sec_strength = sec_strength as c_int;
         let damping = damping as c_int;
         let bd = bd.into_c();
-        let dst = FFISafe::new(&dst);
+
+        let dst = WithOffset {
+            data: FFISafe::new(dst.data),
+            offset: dst.offset,
+        };
+        let top = WithOffset {
+            data: FFISafe::new(top.data),
+            offset: top.offset,
+        };
+        let bottom = WithOffset {
+            data: FFISafe::new(&bottom.data),
+            offset: bottom.offset,
+        };
+
         // SAFETY: Rust fallback is safe, asm is assumed to do the same.
         unsafe {
             self.get()(
@@ -385,18 +397,31 @@ unsafe extern "C" fn cdef_filter_block_c_erased<BD: BitDepth, const W: usize, co
     damping: c_int,
     edges: CdefEdgeFlags,
     bitdepth_max: c_int,
-    dst: *const FFISafe<Rav1dPictureDataComponentOffset>,
-    top: *const FFISafe<CdefTop>,
-    bottom: *const FFISafe<CdefBottom>,
+    dst: WithOffset<*const FFISafe<Rav1dPictureDataComponent>>,
+    top: WithOffset<*const FFISafe<DisjointMut<AlignedVec64<u8>>>>,
+    bottom: WithOffset<*const FFISafe<PicOrBuf<'_, AlignedVec64<u8>>>>,
 ) {
-    // SAFETY: Was passed as `FFISafe::new(_)` in `cdef_dir::Fn::call`.
-    let dst = *unsafe { FFISafe::get(dst) };
+    let dst = WithOffset {
+        // SAFETY: Was passed as `FFISafe::new(_)` in `cdef::Fn::call`.
+        data: unsafe { FFISafe::get(dst.data) },
+        offset: dst.offset,
+    };
+
     // SAFETY: Reverse of cast in `cdef::Fn::call`.
     let left = unsafe { &*left.cast() };
-    // SAFETY: Was passed as `FFISafe::new(_)` in `cdef::Fn::call`.
-    let top = *unsafe { FFISafe::get(top) };
-    // SAFETY: Was passed as `FFISafe::new(_)` in `cdef::Fn::call`.
-    let bottom = *unsafe { FFISafe::get(bottom) };
+
+    let top = WithOffset {
+        // SAFETY: Was passed as `FFISafe::new(_)` in `cdef::Fn::call`.
+        data: unsafe { FFISafe::get(top.data) },
+        offset: top.offset,
+    };
+
+    let bottom = WithOffset {
+        // SAFETY: Was passed as `FFISafe::new(_)` in `cdef::Fn::call`.
+        data: *unsafe { FFISafe::get(bottom.data) },
+        offset: bottom.offset,
+    };
+
     let bd = BD::from_c(bitdepth_max);
     cdef_filter_block_rust(
         dst,
@@ -632,9 +657,9 @@ mod neon {
         damping: c_int,
         edges: CdefEdgeFlags,
         bitdepth_max: c_int,
-        _dst: *const FFISafe<Rav1dPictureDataComponentOffset>,
-        _top: *const FFISafe<CdefTop>,
-        _bottom: *const FFISafe<CdefBottom>,
+        _dst: WithOffset<*const FFISafe<Rav1dPictureDataComponent>>,
+        _top: WithOffset<*const FFISafe<DisjointMut<AlignedVec64<u8>>>>,
+        _bottom: WithOffset<*const FFISafe<PicOrBuf<'_, AlignedVec64<u8>>>>,
     ) {
         use crate::align::Align16;
 
--- a/src/with_offset.rs
+++ b/src/with_offset.rs
@@ -7,6 +7,7 @@ use std::ops::Sub;
 use std::ops::SubAssign;
 
 #[derive(Clone, Copy)]
+#[repr(C)]
 pub struct WithOffset<T> {
     pub data: T,
     pub offset: usize,


src/cdef.rs
```diff
@@ -44,20 +46,20 @@ bitflags! {
...
     pri_strength: c_int,
     sec_strength: c_int,
     dir: c_int,
     damping: c_int,
     edges: CdefEdgeFlags,
     bitdepth_max: c_int,
-    _dst: *const FFISafe<Rav1dPictureDataComponentOffset>,
-    _top: *const FFISafe<CdefTop>,
-    _bottom: *const FFISafe<CdefBottom>,
 ) -> ());
```
kkysen, line 63: We can't change this signature because it's used by asm.  I guess it's not used by arm asm yet, but that may change, and it is used by x86 asm.  So all of the changed arguments (the offsets) need to go at the end, or nowhere at all.  I'm pretty sure this is why CI is failing.

kkysen (changes requested):

ohadravid:
Ok @kkysen - I think everything should be fixed now 😄

src/cdef_apply.rs
```diff
@@ -89,11 +89,12 @@ fn backup2x8<BD: BitDepth>(
     flag: Backup2x8Flags,
 ) {
     let x_off = x_off as isize;
+    let stride_0 = src[0].pixel_stride::<BD>();
```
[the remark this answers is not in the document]
kkysen:
```suggestion
    let strides = src.map(|src| src.pixel_stride::<BD>());
```

Does this work, too?  Or is it important to LLVM to only precompute `stride[0]` first?

src/cdef.rs
```diff
@@ -71,28 +76,31 @@ impl cdef::Fn {
     /// so in order to get access to pre-filter top pixels, use `top`.
     pub fn call<BD: BitDepth>(
```
kkysen, line 77: So LLVM is bad at optimizing `WithOffset` even for a normal Rust function, not just the `extern "C"` one with extra args at the end, including the weird `FFISafe` pointers?

ohadravid: In my experiments LLVM doesn't always optimize them (or other small Copy structs) away even if there are only normal Rust functions involved, but never when they cross FFI boundaries (for example, if you set bottom = ptr:null() here, the extra stack usage will be optimized away, but if you pass in references to WithOffsets, used them, then call FFI with the new deconstructed parameters, they still won't be optimized away.

*I tried dozens of things so it's possible I missed something or confused some of the results so please take this with a grain of salt 🙏

src/cdef_apply.rs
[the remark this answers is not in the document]
ohadravid: Seems like it's important to split `stride[0]` the keep them `strides` as a simple array: setting `let strides = src.map(..)` once at the start is a lot slower (by about 100ms in my testing), and even only using `src.map` in the second part is slightly slower.

src/cdef.rs
kkysen: 😭 That's very unfortunate.  `WithOffset` is such a simple, thin wrapper over two args.  I could see `*const FFISafe<WithOffset<T>>` not being optimized well to `*const FFISafe<T>` and `usize`, but the much simpler `WithOffset<T>` not being optimized the same as `T` and `usize` is really atrocious.

src/cdef.rs
```diff
@@ -71,28 +76,31 @@ impl cdef::Fn {
...
-        dst: Rav1dPictureDataComponentOffset,
+        dst_data: &Rav1dPictureDataComponent,
+        dst_offset: usize,
         left: &[LeftPixelRow2px<BD::Pixel>; 8],
-        top: CdefTop,
-        bottom: CdefBottom,
+        top_data: &DisjointMut<AlignedVec64<u8>>,
+        top_offset: usize,
+        bottom_data: &PicOrBuf<'_, AlignedVec64<u8>>,
+        bottom_offset: usize,
```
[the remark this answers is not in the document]
kkysen: And I think the difference happening here is that `&bot.data` in `fn rav1d_cdef_brow` is higher on the stack than the `&bottom.data` in `fn Cdef::call`.

Can you undo the changes for `dst` and `top` since we know those are okay now?

src/cdef_apply.rs
```diff
@@ -89,11 +89,13 @@ fn backup2x8<BD: BitDepth>(
     flag: Backup2x8Flags,
 ) {
     let x_off = x_off as isize;
+    // Precomputing optimizes better.
```
kkysen, line 92:
```suggestion
    // Precomputing (only `stride[0]`) optimizes better.
```

ohadravid:
@kkysen -

1. ✅ moved the `backup2x8` to a separate PR 
2. ❓ w.r.t only changing `bot` vs all 3 variables: in my tests, only changing all of them gives the perf boost - https://github.com/memorysafety/rav1d/pull/1418#discussion_r2136944989

src/cdef.rs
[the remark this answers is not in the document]
kkysen: Hmm, that's odd.  With `dest` and `top` using `WithOffset`, I see the same asm, but with `bottom` using `WithOffset`, I see this asm diff:

```diff
-       sub sp, sp, #976
-       .cfi_def_cfa_offset 1072
+       sub sp, sp, #992
+       .cfi_def_cfa_offset 1088
```

aarch64, too, so I'm not sure what the difference between us is here.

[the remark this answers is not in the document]
kkysen: This is the diff I used to test `dest` and `top`:

```diff
diff --git a/src/cdef.rs b/src/cdef.rs
index 02839407..bd156faa 100644
--- a/src/cdef.rs
+++ b/src/cdef.rs
@@ -76,11 +76,9 @@ impl cdef::Fn {
     /// so in order to get access to pre-filter top pixels, use `top`.
     pub fn call<BD: BitDepth>(
         &self,
-        dst_data: &Rav1dPictureDataComponent,
-        dst_offset: usize,
+        dst: Rav1dPictureDataComponentOffset,
         left: &[LeftPixelRow2px<BD::Pixel>; 8],
-        top_data: &DisjointMut<AlignedVec64<u8>>,
-        top_offset: usize,
+        top: CdefTop,
         bottom_data: &PicOrBuf<'_, AlignedVec64<u8>>,
         bottom_offset: usize,
         pri_strength: c_int,
@@ -90,16 +88,16 @@ impl cdef::Fn {
         edges: CdefEdgeFlags,
         bd: BD,
     ) {
-        let dst_ptr = dst_data.as_mut_ptr_at::<BD>(dst_offset).cast();
-        let stride = dst_data.stride();
+        let dst_ptr = dst.data.as_mut_ptr_at::<BD>(dst.offset).cast();
+        let stride = dst.data.stride();
         let left = ptr::from_ref(left).cast();
-        let top_ptr = top_data.as_ptr_at::<BD>(top_offset).cast();
+        let top_ptr = top.data.as_ptr_at::<BD>(top.offset).cast();
         let bottom_ptr = bottom_data.wrapping_as_ptr_at::<BD>(bottom_offset).cast();
         let sec_strength = sec_strength as c_int;
         let damping = damping as c_int;
         let bd = bd.into_c();
-        let dst = FFISafe::new(dst_data);
-        let top = FFISafe::new(top_data);
+        // let dst = FFISafe::new(dst.data);
+        // let top = FFISafe::new(top.data);
         let bottom = FFISafe::new(bottom_data);
         // SAFETY: Rust fallback is safe, asm is assumed to do the same.
         unsafe {
@@ -115,10 +113,10 @@ impl cdef::Fn {
                 damping,
                 edges,
                 bd,
-                dst,
-                dst_offset,
-                top,
-                top_offset,
+                FFISafe::new(dst.data),
+                dst.offset,
+                FFISafe::new(top.data),
+                top.offset,
                 bottom,
                 bottom_offset,
             )
diff --git a/src/cdef_apply.rs b/src/cdef_apply.rs
index 283e0354..76739dc0 100644
--- a/src/cdef_apply.rs
+++ b/src/cdef_apply.rs
@@ -344,11 +344,9 @@ pub(crate) fn rav1d_cdef_brow<BD: BitDepth>(
                             let adj_y_pri_lvl = adjust_strength(y_pri_lvl, variance);
                             if adj_y_pri_lvl != 0 || y_sec_lvl != 0 {
                                 f.dsp.cdef.fb[0].call::<BD>(
-                                    bptrs[0].data,
-                                    bptrs[0].offset,
+                                    bptrs[0],
                                     &lr_bak[bit as usize][0],
-                                    top.data,
-                                    top.offset,
+                                    top,
                                     &bot.data,
                                     bot.offset,
                                     adj_y_pri_lvl,
@@ -361,11 +359,9 @@ pub(crate) fn rav1d_cdef_brow<BD: BitDepth>(
                             }
                         } else if y_sec_lvl != 0 {
                             f.dsp.cdef.fb[0].call::<BD>(
-                                bptrs[0].data,
-                                bptrs[0].offset,
+                                bptrs[0],
                                 &lr_bak[bit as usize][0],
-                                top.data,
-                                top.offset,
+                                top,
                                 &bot.data,
                                 bot.offset,
                                 0,
@@ -450,11 +446,9 @@ pub(crate) fn rav1d_cdef_brow<BD: BitDepth>(
                                 });

                                 f.dsp.cdef.fb[uv_idx as usize].call::<BD>(
-                                    bptrs[pl].data,
-                                    bptrs[pl].offset,
+                                    bptrs[pl],
                                     &lr_bak[bit as usize][pl],
-                                    top.data,
-                                    top.offset,
+                                    top,
                                     &bot.data,
                                     bot.offset,
                                     uv_pri_lvl.into(),
```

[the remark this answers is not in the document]
ohadravid: Oh got it now - keep the Offsets away from the FFI boundary. Will update the impl 👍

ohadravid:
OK @kkysen - all done 🦀

src/cdef.rs
```diff
@@ -55,9 +56,12 @@ wrap_fn_ptr!(pub unsafe extern "C" fn cdef(
...
     bitdepth_max: c_int,
-    _dst: *const FFISafe<Rav1dPictureDataComponentOffset>,
-    _top: *const FFISafe<CdefTop>,
-    _bottom: *const FFISafe<CdefBottom>,
+    dst: *const FFISafe<Rav1dPictureDataComponent>,
+    dst_offset: usize,
+    top: *const FFISafe<DisjointMut<AlignedVec64<u8>>>,
+    top_offset: usize,
+    bottom: *const FFISafe<PicOrBuf<'_, AlignedVec64<u8>>>,
+    bottom_offset: usize,
```
kkysen, line 64:
```suggestion
    _dst: *const FFISafe<Rav1dPictureDataComponent>,
    _dst_offset: usize,
    _top: *const FFISafe<DisjointMut<AlignedVec64<u8>>>,
    _top_offset: usize,
    _bottom: *const FFISafe<PicOrBuf<'_, AlignedVec64<u8>>>,
    _bottom_offset: usize,
```

As a convention for readability, I've been prefixing the extra args not used by asm with a `_`.

src/cdef.rs
```diff
@@ -55,9 +56,12 @@ wrap_fn_ptr!(pub unsafe extern "C" fn cdef(
     damping: c_int,
     edges: CdefEdgeFlags,
     bitdepth_max: c_int,
-    _dst: *const FFISafe<Rav1dPictureDataComponentOffset>,
-    _top: *const FFISafe<CdefTop>,
-    _bottom: *const FFISafe<CdefBottom>,
+    dst: *const FFISafe<Rav1dPictureDataComponent>,
+    dst_offset: usize,
```
kkysen, line 60: I'm wondering now, does `WithOffset<*const FFISafe<Rav1dPictureDataComponent>` optimize okay?  I think, as it was for `bottom`, that the pointer/reference indirection is throwing off the optimization, and I would hope that `WithOffset<T>` and `T, usize` can still be optimized the same.

ohadravid: 🤯 yep that seems to work - changed to that and set `repr(C)` in `WithOffset` and got the nice stack usage.

src/cdef.rs
```diff
@@ -87,12 +88,23 @@ impl cdef::Fn {
...
-        let bottom = FFISafe::new(&bottom);
         let sec_strength = sec_strength as c_int;
         let damping = damping as c_int;
         let bd = bd.into_c();
-        let dst = FFISafe::new(&dst);
+
+        let dst = WithOffset {
+            data: FFISafe::new(dst.data),
+            offset: dst.offset,
+        };
```
kkysen, line 98: Do you think it's worth it to add to `WithOffset` a `fn map` method that maps the `data` type while keeping the same `offset`?  It'd make these briefer.

Also, now that we've mostly figured out what's being misoptimized with `WithOffset`, have you checked any other places where we might be using `WithOffset` like this, too?

ohadravid: Added a few utility functions to make this nicer 👍 

> have you checked any other places where we might be using WithOffset like this, too?

Followed up on this and checked everywhere else, updated all `FFISafe<WithOffset<..>>` to `WithOffset<FFISafe<..>>`, in this commit:
https://github.com/ohadravid/rav1d/commit/8c9d4333f37fd4fbb4ed65f2d294d6be8a501a31

Examples where this appears to help:

(taken from `sub sp, sp, #XXX`):
rav1d_recon_b_intra: 688 bytes -> 672 bytes
rav1d_recon_b_inter: 576 bytes -> 544 bytes
read_coef_tree: 256 bytes -> 224

I'll try to measure if this impacts real world perf - maybe overall `rav1d_recon_b_intra` improvement is ~150ms, I think `rav1d_recon_b_inter` and `read_coef_tree` are the same.

Let me know if you want this commit here / in a seperate PR / only for some functions 😄

ohadravid:
@kkysen - added the utlity functions - let me know if you like them (can update the names / docs / etc).

Also tried to apply the same trick everywhere https://github.com/memorysafety/rav1d/pull/1418#discussion_r2145087617 .

src/cdef.rs
kkysen: > Let me know if you want this commit here / in a seperate PR / only for some functions

Hmm, the new stuff looks great.  It might be easier to split them into separate PRs, though.  The original PR here is all good now and I can merge it.

ohadravid:
@kkysen do you want me to split [680eb25](https://github.com/memorysafety/rav1d/pull/1418/commits/680eb250c3ba5cf5e6e983730a97532d0323acf5) and the followup update to the other callsites to a new PR?

PR #1420, opened 2025-05-30
Start fixing clippy lints

djc:
This is mostly pretty boring.


src/align.rs
```diff
@@ -132,6 +132,7 @@ def_align!(64, Align64);
 /// A [`Vec`] that uses [`mem::size_of`]`::<C>()` aligned allocations.
 ///
 /// Only works with [`Copy`] types so that we don't have to handle drop logic.
+#[expect(clippy::len_without_is_empty)]
```
kkysen, line 135: Can we add `reason`s for any `expect`s?

src/cdef.rs
```diff
@@ -181,9 +181,9 @@ fn padding<BD: BitDepth>(
     let stride = src.pixel_stride::<BD>();
 
     // Fill extended input buffer.
-    let mut x_start = 2 - 2;
+    let mut x_start = 0;
```
kkysen, line 184: It was `2 - 2` for a reason.  This shouldn't be simplified.  Can you leave it as is?

src/ipred.rs
```diff
@@ -791,7 +791,7 @@ fn get_filter_strength(wh: c_int, angle: c_int, is_sm: bool) -> c_int {
             (..=16, ..) => 0,
             (..=24, 4..) => 3,
             (..=24, ..) => 0,
-            (.., _) => 3,
+            (..) => 3,
```
kkysen, line 794: How does this work?

djc: I think this is `..` as in `{ .. }` rather than `..` as in `0..len`?

src/align.rs
kkysen: Oh perhaps, just a simple reason is fine then.

include/dav1d/picture.rs
```diff
@@ -520,7 +519,7 @@ impl From<Rav1dPicture> for Dav1dPicture {
             seq_hdr_ref: seq_hdr.map(RawArc::from_arc),
             content_light_ref: content_light.map(RawArc::from_arc),
             mastering_display_ref: mastering_display.map(RawArc::from_arc),
-            itut_t35_ref: Some(itut_t35).map(RawArc::from_arc),
+            itut_t35_ref: Some(RawArc::from_arc(itut_t35)),
```
kkysen, line 522: All of the other fields here are `.map(RawArc::from_arc)`, so I'd prefer to keep that consistency.

src/decode.rs
```diff
@@ -3465,12 +3457,6 @@ fn decode_sb(
     let mut by8 = 0;
     let ctx = match pass {
         FrameThreadPassState::First(ts_c) => {
-            if false && bl == BlockLevel::Bl64x64 {
-                println!(
-                    "poc={},y={},x={},bl={:?},r={}",
-                    frame_hdr.frame_offset, t.b.y, t.b.x, bl, ts_c.msac.rng,
-                );
-            }
```
kkysen, line 3473: Please don't delete this.  It's currently dead code, but just because it's been disabled.  It shouldn't be deleted.

src/fg_apply.rs
```diff
@@ -133,7 +133,7 @@ pub(crate) fn rav1d_apply_grain_row<BD: BitDepth>(
     let seq_hdr = &***out.seq_hdr.as_ref().unwrap();
     let frame_hdr = &***out.frame_hdr.as_ref().unwrap();
     let data = &frame_hdr.film_grain.data;
-    let data_c = &data.clone().into();
+    let data_c = &data.clone();
```
kkysen, line 136: These (`data` and `data_c`) were previously different `Rav1d*` and `Dav1d*` types, but they've now been unified and we must've forgotten to update this.  I think we can elide the `.clone()` here, too, and just have `let data_c = data`.

```suggestion
    let data_c = data;
```

src/fg_apply.rs
```diff
@@ -226,7 +226,7 @@ pub(crate) fn rav1d_apply_grain<BD: BitDepth>(
     r#in: &Rav1dPicture,
 ) {
     let mut grain = Default::default();
-    let rows = (out.p.h as usize + FG_BLOCK_SIZE - 1) / FG_BLOCK_SIZE;
+    let rows = (out.p.h as usize).div_ceil(FG_BLOCK_SIZE);
```
kkysen, line 229: Same here about perf.

src/internal.rs
```diff
@@ -274,6 +274,7 @@ impl<BD: BitDepth> Default for GrainBD<BD> {
     }
 }
 
+#[allow(clippy::large_enum_variant)]
```
kkysen, line 277: Can you make this an `expect` with a `reason`?

src/error.rs
```diff
@@ -2,6 +2,7 @@ use std::ffi::c_int;
 use std::ffi::c_uint;
 use strum::FromRepr;
 
+#[allow(clippy::upper_case_acronyms)]
```
kkysen, line 5: Can you make this an `expect` with a `reason` (the reason being these are the same names as the C errors)?

src/msac.rs
```diff
@@ -208,6 +208,10 @@ pub struct MsacAsmContext {
     ) -> c_uint,
 }
 
+#[cfg_attr(
+    not(all(feature = "asm", target_arch = "x86_64")),
+    expect(clippy::derivable_impls)
```
kkysen, line 213:
```suggestion
    expect(clippy::derivable_impls, reason = "Doesn't apply on `x86_64`")
```

src/msac.rs
```diff
@@ -555,20 +558,19 @@ pub fn rav1d_msac_decode_symbol_adapt4(s: &mut MsacContext, cdf: &mut [u16], n_s
...
-            ret = unsafe {
+            let ret = unsafe {
                 dav1d_msac_decode_symbol_adapt8_neon(&mut s.asm, cdf.as_mut_ptr(), n_symbols as usize)
             };
         } else {
-            ret = rav1d_msac_decode_symbol_adapt_rust(s, cdf, n_symbols);
+            let ret = rav1d_msac_decode_symbol_adapt_rust(s, cdf, n_symbols);
         }
     }
     debug_assert!(ret < 8);
```
kkysen, line 576: I did not know this worked with `cfg_if!`.  Cool that it does.

src/obu.rs
```diff
@@ -385,13 +385,10 @@ fn parse_seq_hdr(
...
-    let monochrome;
-    if profile != Rav1dProfile::High {
-        monochrome = gb.get_bit() as u8;
-    } else {
-        // Default initialization.
-        monochrome = Default::default();
-    }
+    let monochrome = match profile {
+        Rav1dProfile::High => gb.get_bit() as u8,
+        _ => 0,
```
kkysen, line 390:
```suggestion
        _ => Default::default(), // Default initialization.
```

src/obu.rs
```diff
@@ -483,13 +480,10 @@ fn parse_seq_hdr(
...
-    let separate_uv_delta_q;
-    if monochrome == 0 {
-        separate_uv_delta_q = gb.get_bit() as u8;
-    } else {
-        // Default initialization.
-        separate_uv_delta_q = Default::default();
-    }
+    let separate_uv_delta_q = match monochrome {
+        0 => gb.get_bit() as u8,
+        _ => 0,
```
kkysen, line 485:
```suggestion
        _ => Default::default(), // Default initialization.
```

src/obu.rs
```diff
@@ -1862,13 +1852,10 @@ fn parse_frame_hdr(
...
-    let frame_id;
-    if seqhdr.frame_id_numbers_present != 0 {
-        frame_id = gb.get_bits(seqhdr.frame_id_n_bits.into()) as u32;
-    } else {
-        // Default initialization.
-        frame_id = Default::default();
-    }
+    let frame_id = match seqhdr.frame_id_numbers_present {
+        0 => gb.get_bits(seqhdr.frame_id_n_bits.into()),
+        _ => 0,
```
kkysen, line 1857:
```suggestion
        _ => Default::default(), // Default initialization.
```

src/recon.rs
```diff
@@ -543,6 +543,7 @@ fn decode_coefs<BD: BitDepth>(
     let frame_hdr = &***f.frame_hdr.as_ref().unwrap();
     let lossless = frame_hdr.segmentation.lossless[b.seg_id.get()];
     let t_dim = &dav1d_txfm_dimensions[tx as usize];
+    #[allow(clippy::overly_complex_bool_expr)]
```
kkysen, line 546:
```suggestion
    #[expect(clippy::overly_complex_bool_expr)]
```

src/obu.rs
djc: Why do you think this is better? I'm personally not a fan of `Default::default()` because it relies on contextual knowledge, so it is harder to review in a diff.

src/thread_task.rs
```diff
@@ -710,19 +707,18 @@ fn delayed_fg_task<'l, 'ttd: 'l>(
...
         if (row + 1) < progmax {
             ttd.cond.notify_one();
         } else if row + 1 >= progmax {
-            *task_thread_lock = Some(ttd.lock.lock());
+            let _locked = ttd.lock.lock();
             ttd.delayed_fg_exec.set(0);
             if row >= progmax {
                 break;
             }
-            let _ = task_thread_lock.take();
```
kkysen, line 725: This doesn't look the same?  Is it?

Can you just do `drop(task_thread_lock.take())` at the end?

kkysen: Was this an automatic `cargo clippy --fix` change?

djc: I don't see the point of storing into `task_thread_lock` here?

kkysen: The former doesn't drop the lock `if row >= progmax`, but this does, right?

djc: Ahh, that's true.

src/obu.rs
kkysen: Because in C, this variable is uninitialized in this case, and I think `Default::default()` (with the comment, too) that this is the default initialization, whatever that may be, and it doesn't matter what value it is.

src/thread_task.rs
kkysen: We have to be especially careful with the threading code here, because errors just make the program slower and won't cause any test failures.

src/lf_mask.rs
```diff
@@ -702,7 +702,7 @@ pub(crate) fn rav1d_calc_lf_values(
         return;
     }
 
-    let mr_deltas = hdr.loopfilter.mode_ref_deltas.clone().into();
+    let mr_deltas = hdr.loopfilter.mode_ref_deltas.clone();
```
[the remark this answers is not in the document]
kkysen: I moved this to #1422 since it may affect perf a bit.

Issue #1345, opened 2024-09-09
Replace `assume` with `assert_unchecked` (Rust 1.81)


kkysen:
Rust 1.81 stabilized [`std::hint::assert_unchecked`](https://doc.rust-lang.org/std/hint/fn.assert_unchecked.html), which is what we want our `assume` to be.  Our `assume` is a slightly different implementation (that just calls `unreachable_unchecked` in the `false` case), and `assert_unchecked` is implemented more directly with regards to LLVM IR generation.  They are largely the same, but `assert_unchecked` may be optimized slightly better, and Rust is standardizing around this new interface, so we should switch to it.

PR #1421, opened 2025-05-31
Replace `assume` with `assert_unchecked`

danielbyiringiro:
Rust 1.81 stabilized `std::hint::assert_unchecked`, which is what we want our assume to be. Our assume is a slightly different implementation (that just calls `unreachable_unchecked` in the `false` case), and `assert_unchecked` is implemented more directly with regards to LLVM IR generation. They are largely the same, but `assert_unchecked` may be optimized slightly better, and Rust is standardizing around this new interface, so we should switch to it.

Deleted the `assume` function since it is no longer needed.

* Resolves: #1345


kkysen (approved):
LGTM, just can you run `cargo fmt`?

PR #1429, opened 2025-06-12
`rustfmt.toml`: add one to enforce import rules that were mostly followed

kkysen:
I had been mostly following these rules already through `rust-analyzer`'s settings, but apparently you can enforce them more easily with `rustfmt`.


djc (commented):
I don't see an actual change to `rustfmt.toml`. Is that not tracked as part of the repo?

Also somewhat independent, but IMO the one line per import standard is pretty unidiomatic and makes navigating the codebase harder. Why was this chosen over a more standard approach of grouping either on a crate or module basis?

kkysen:
> I don't see an actual change to `rustfmt.toml`. Is that not tracked as part of the repo?

Oh whoops that must've not gotten checked in.

> Also somewhat independent, but IMO the one line per import standard is pretty unidiomatic and makes navigating the codebase harder. Why was this chosen over a more standard approach of grouping either on a crate or module basis?

It was causing a lot of big diffs to unrelated changes.

PR #1430, opened 2025-06-13
Use more idiomatic import style

djc:
Using module-level grouping saves 800 lines of code, avoiding having people to scroll past all the imports. Using import groups helps external crate imports stand out. I consider both to be idiomatic. Personally I think with the current rate of change, changes to imports aren't going to be a huge burden.


rustfmt.toml
```diff
@@ -1,2 +1,2 @@
-imports_granularity = "Item"
-group_imports = "One"
+imports_granularity = "Module"
+group_imports = "StdExternalCrate"
```
kkysen, line 2: Grouping imports like this I agree is good, and I was thinking of possibly doing the same.

rustfmt.toml
```diff
@@ -1,2 +1,2 @@
-imports_granularity = "Item"
-group_imports = "One"
+imports_granularity = "Module"
```
kkysen, line 1: I'm still not sure I agree with this.

> Using module-level grouping saves 800 lines of code, avoiding having people to scroll past all the imports.

I don't think just having fewer lines of code (with the same information) is really that helpful.  It's quite easy to collapse the imports in your editor and not think about this.

> Personally I think with the current rate of change, changes to imports aren't going to be a huge burden.

It's probably won't cause as many conflicts as before, but it still does, and we can just avoid that.

One-per-line imports also have a bunch of other benefits, besides diffing and merge conflicts.  They're easier to consistently search and grep for and easier to move around; they're context independent.

djc: > > Using module-level grouping saves 800 lines of code, avoiding having people to scroll past all the imports.
> 
> I don't think just having fewer lines of code (with the same information) is really that helpful. It's quite easy to collapse the imports in your editor and not think about this.

If we don't want to think about imports, spending less space on them in the canonical representation of the source code seems like a better option for most contributors.

> One-per-line imports also have a bunch of other benefits, besides diffing and merge conflicts. They're easier to consistently search and grep for and easier to move around; they're context independent.

By your own argument of using the editor, I use the editor to look for references to symbols. Normal search also mostly works fine on grouped imports.

In the end, IMO making rav1d more like a "normal" Rust library would be a good thing, and I don't think import style is worth spending "difference tokens" on.

kkysen (approved):
Oh sorry, meant to approve this after we discussed it.

PR #1432, opened 2025-06-16
perf: Further optimise the `PartialEq` implementations for `Mv`, `RefMvsRefPair` and `RefMvsMvPair`

daxtens:
In #1400, we replaced the derived `PartialEq` implementations for `Mv`, `RefMvsRefPair` and `RefMvsMvPair` with implementations that used `.as_bytes()`. In the case of `Mv`, this was an improvement because it the derived implementation compared field at a time, which involved two separate branches.

However, while staring at the assembly for other reasons, I realised that the implementation has other problems - it does two loads (`y`, `x`), then it saves those values back to the stack in two stores, then it reads the value on the stack back as a single 32-bit value and finally compares _that_. 

The best generated code I've found so far comes from expressing the contents as arrays and comparing the arrays. The generated code still doesn't usually reduce to a single load, but it does manage avoid the round-trip via the stack.

I don't have figures from Linux and `perf` yet, but I did compare with `hyperfine` on a Windows system with an AMD Ryzen 5 2600 6-core processor. I ran with 6 threads (1 per core) in order to reduce system noise. Under that configuration:
 - Chimera 10 bit: ~1.9% overall improvement 
 - Chimera 8 bit: ~0.5% overall improvement

(I'm not sure why there would be such a difference between 10bit and 8bit; I will rerun the benchmarks later and confirm. I am confident that it's a measurable improvement in both cases.)

cc: @ohadravid

### DIFF
--- a/src/levels.rs
+++ b/src/levels.rs
@@ -395,9 +395,43 @@ impl PartialEq for Mv {
     fn eq(&self, other: &Self) -> bool {
         // `#[derive(PartialEq)]` compares per-field with `&&`,
         // which isn't optimized well and isn't coalesced into wider loads.
-        // Comparing all of the bytes at once optimizes better with wider loads.
         // See <https://github.com/rust-lang/rust/issues/140167>.
-        self.as_bytes() == other.as_bytes()
+
+        // A very common compare is the is_invalid().
+        // With the default implementation, this becomes two branches, one for each field.
+        //
+        // Comparing all of the bytes at once optimizes better:
+        // self.as_bytes() == other.as_bytes()
+        //
+        // However, on nightly-2025-05-01-aarch64-apple-darwin, while it does avoid a second compare,
+        // it involves two reads from memory of the two array elements,
+        // then it _stores_ them to the stack, then loads them again as a 32 bit,
+        // and the compares that with the constant Mv::INVALID:
+        //
+        // mov w9, #-2147450880 ; load [MIN, MIN] combined invalid mv
+        // ldrh w10, [x3]       ; load mv.y
+        // ldrh w11, [x3, #2]   ; load mv.x
+        // strh w10, [sp, #8]   ; save mv.y to the stack
+        // strh w11, [sp, #10]  ; save mv.x to the stack
+        // ldr w8, [sp, #8]     ; load combined mv from stack
+        // cmp w8, w9           ; compare mv with [MIN, MIN]
+        //
+        // x86_64 code appears similar. This is disappointingly not zero-copy.
+        //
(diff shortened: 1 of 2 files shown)


daxtens:
What happens if you look at the first few instructions of
add_spatial_candidate or add_temporal_candidate? I think the inlining is
likely to be making a difference.

daxtens:
So with better access to my x86 machines, I poked around again.

I do like what this does with `add_spatial_candidate`:

Before:
```
rav1d::refmvs::add_spatial_candidate:
        push rbp
        push r14
        push rbx
        sub rsp, 16
        movzx ebp, word ptr [rcx]
        movzx ebx, word ptr [rcx + 2]
        mov word ptr [rsp + 8], bp
        mov word ptr [rsp + 10], bx
        cmp dword ptr [rsp + 8], -2147450880
        je .LBB337_29
```

After:
```
rav1d::refmvs::add_spatial_candidate:
        push rbp
        push r15
        push r14
        push rbx
        push rax
        movzx ebp, word ptr [rcx]
        movzx ebx, word ptr [rcx + 2]
        mov eax, ebx
        shl eax, 16
        or eax, ebp
        cmp eax, -2147450880
        jne .LBB337_1
```

Before, we went from heap -> registers -> stack -> register -> compare, after we go from heap -> registers -> shift & or -> compare.

But, with further benchmarking (running 20 times with the Chimera files instead of 10), while the first patch does definitely seem to be a consistent win, something in the subsequent two seems to be a regression. I think there's a win to be had here, but evidently it's not quite ready for prime time just yet. I will put this PR into draft and do some more work. I'll mark it as ready for review and comment when I think it's good to go.

daxtens:
Another thing I've realised in investigating this: It looks like sometimes when the read an mv in two halves in `add_spatial_candidate`, it's so we can reuse the constituent parts later in the function without having to do a separate memory access. Whether or not that's better or worse I'm still working out.

kkysen (commented):
> I've taken one of the things I learned - that alignment is a key factor here - and have done #1433, which I think is good to go.
> 
> I'm then going to repeat some of the things I tried in this PR on top of #1433. If there's anything of value I will force-push it to this PR, otherwise I will just close this PR.

Sounds good.  🤞 you can find some more on top of #1433, too.  Sorry I didn't get to review this PR much yet, but it sounds like you found a great, simple win in #1433.  I'll hold off on reviewing stuff here until you force push something new.

kkysen:
@daxtens, is there anything left here, or can I close this?

kkysen:
Feel free to re-open if you have something new.

PR #1433, opened 2025-06-19
perf: `struct Mv`: make `align(4)`

daxtens:
(This takes some of the things I learned from #1432, without a bunch of the noise and missteps. Unlike #1432, I think this is good in its current state.)

Making `Mv` `align(4)` matches up with how it works in the C code, where it's a union with a u32.

It requires a bit of finesse around the `RefMvsTemporalBlock`, but nothing we can't handle.

It improves the code in `add_spatial_candidate` a lot. On nightly on macOS, we see for example this:

```asm
ldrh  w11, [x3]
ldrh  w10, [x3, #2]
strh  w11, [sp, #8]
strh  w10, [sp, #10]
ldr   w8, [sp, #8]
```

becomes this:
```asm
ldr   w10, [x3]
```

And later on we see some branches replaced with conditional moves. We also see better code on x86.

`perf` on x86 shows a significant improvement in time spent in `add_spatial_candidate`. 

I see an overall improvement around 1% on the Chimera files on a Windows system with an AMD Ryzen 5 2600 6-core processor. (I tested with `hyperfine`, with 6 threads (1 per core) in order to reduce system noise, but the figures are still very rough and noisy.)


src/levels.rs
```diff
@@ -401,6 +403,21 @@ impl PartialEq for Mv {
...
+            y: value[0],
+            x: value[1],
+        }
+    }
+}
+
+impl From<Mv> for [i16; 2] {
+    fn from(value: Mv) -> Self {
+        [value.y, value.x]
+    }
```
kkysen, line 418:
```suggestion
    fn from(Mv { y, x }: Mv) -> Self {
        [y, x]
    }
```

src/refmvs.rs
```diff
@@ -34,7 +34,9 @@ use zerocopy::FromZeroes;
 #[derive(Clone, Copy, Default, PartialEq, Eq)]
 #[repr(C, packed)]
 pub struct RefMvsTemporalBlock {
-    pub mv: Mv,
+    /// This is an [`Mv`]. However, because the type [`Mv`] is `align(4)`,
+    /// we cannot include it in this packed type. Include the underlying array instead.
```
kkysen, line 38:
```suggestion
    /// This is a [`Mv`]. However, because [`Mv`] is `align(4)`,
    /// we cannot include it in this packed `struct`.
    /// So we use the underlying array instead.
```

src/refmvs.rs
```diff
@@ -34,7 +34,9 @@ use zerocopy::FromZeroes;
...
+    pub mv: [i16; 2],
```
kkysen, line 39: Could we use a dedicated `UnalignedMv` type instead?  One that doesn't implement `PartialEq`, `fn is_invalid`, etc?  And then have methods like `fn align(self: UnalignedMv) -> Mv` and `fn unalign(self: Mv) -> UnalignedMv` instead of the more generic `from` and `into` so it's clearer what exactly we're doing.

daxtens: Sure, I'll give that a go.

kkysen (changes requested):
Just a few minor changes left.

kkysen:
https://github.com/memorysafety/rav1d/actions/runs/15770411757/job/44454210077#step:6:44

Isn't this showing performance regressed?

```
     0.97%     +0.85%  dav1d                 [.] rav1d::refmvs::add_spatial_candidate
     2.77%     -0.37%  dav1d                 [.] rav1d::refmvs::rav1d_refmvs_find
     1.44%     +0.19%  dav1d                 [.] rav1d::refmvs::add_temporal_candidate
```

kkysen:
> https://github.com/memorysafety/rav1d/actions/runs/15770411757/job/44454210077#step:6:44
> 
> Isn't this showing performance regressed?
> 
> ```
>      0.97%     +0.85%  dav1d                 [.] rav1d::refmvs::add_spatial_candidate
>      2.77%     -0.37%  dav1d                 [.] rav1d::refmvs::rav1d_refmvs_find
>      1.44%     +0.19%  dav1d                 [.] rav1d::refmvs::add_temporal_candidate
> ```

Oh wait, I think our CI has the args backwards and this is an improvement:

```sh
perf diff perf.data.$input perf.data.$input.main | head -n 100
```

#1436 should fix this.

src/levels.rs
```diff
@@ -397,6 +402,34 @@ impl PartialEq for Mv {
...
+}
+
+impl UnalignedMv {
+    /// Create an aligned [`Mv`] from this `UnalignedMv`
+    pub fn to_aligned(self) -> Mv {
+        Mv {
+            y: self.y,
+            x: self.x,
+        }
+    }
```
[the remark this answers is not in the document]
daxtens: Yeah, sorry, most of my experience is with C so I'm still learning some of the more subtle rust conventions! Will fix that and the clippy complaints later tonight.

daxtens:
Thanks, I've applied those and rebased it on main.

daxtens:
Done! :smile:

kkysen (approved):

PR #1438, opened 2025-06-30
`mod itx`: backport changes from dav1d 1.5.0

fbossen:
Main item is "restrict number of columns iterated over based on EOB" which applies to the generic code for transforms (not asm).

### DIFF
--- a/src/in_range.rs
+++ b/src/in_range.rs
@@ -23,7 +23,7 @@ where
 
 impl<T, const MIN: u128, const MAX: u128> InRange<T, MIN, MAX>
 where
-    T: TryFrom<u128, Error: Debug> + PartialEq + Eq + PartialOrd + Ord,
+    T: TryFrom<u128, Error: Debug> + PartialEq + Eq + PartialOrd + Ord + Copy,
 {
     fn in_bounds(&self) -> bool {
         *self >= Self::min() && *self <= Self::max()
@@ -41,6 +41,10 @@ where
     pub fn get(self) -> T {
         // SAFETY: Checked in `Self::new`.
         unsafe { assert_unchecked(self.in_bounds()) };
+        self.0
+    }
+
+    pub const fn const_get(&'static self) -> T {
         self.0
     }
 }
--- a/src/itx.rs
+++ b/src/itx.rs
@@ -32,7 +32,9 @@ use crate::levels::{
     FLIPADST_ADST, FLIPADST_DCT, FLIPADST_FLIPADST, H_ADST, H_DCT, H_FLIPADST, IDTX,
     N_TX_TYPES_PLUS_LL, V_ADST, V_DCT, V_FLIPADST, WHT_WHT,
 };
+use crate::scan::DAV1D_LAST_NONZERO_COL_FROM_EOB;
 use crate::strided::Strided as _;
+use crate::tables::DAV1D_TXFM_DIMENSIONS;
 use crate::wrap_fn_ptr::wrap_fn_ptr;
 
 pub type Itx1dFn = fn(c: &mut [i32], stride: NonZeroUsize, min: i32, max: i32);
@@ -42,15 +44,17 @@ fn inv_txfm_add<BD: BitDepth>(
     dst: Rav1dPictureDataComponentOffset,
     coeff: &mut [BD::Coef],
     eob: i32,
-    w: usize,
-    h: usize,
+    tx: TxfmSize,
     shift: u8,
-    first_1d_fn: Itx1dFn,
-    second_1d_fn: Itx1dFn,
-    has_dc_only: bool,
+    txtp: TxfmType,
     bd: BD,
 ) {
     let bitdepth_max = bd.bitdepth_max().as_::<i32>();
+
+    let t_dim = &DAV1D_TXFM_DIMENSIONS[tx as usize];
+    let w = 4 * t_dim.w as usize;
+    let h = 4 * t_dim.h as usize;
+    let has_dc_only = txtp == DCT_DCT;
 
     assert!(w >= 4 && w <= 64);
     assert!(h >= 4 && h <= 64);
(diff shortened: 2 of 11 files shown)


src/in_range.rs
```diff
@@ -43,6 +43,10 @@ where
         unsafe { assert_unchecked(self.in_bounds()) };
         self.0
     }
+
+    pub const fn const_get(&'static self) -> T {
+        self.0
+    }
```
kkysen, line 49:
```suggestion
    /// This doesn't [`assert_unchecked`] that the value is in bounds (for optimization)
    /// because [`Self::in_bounds`] is not a `const fn` due to its usage of `trait`s.
    /// However, because this is meant to be called in a `const` context,
    /// everything should be known already, and the [`assert_unchecked`]
    /// should be unnecessary for optimization.
    pub const fn const_get(self) -> T {
        self.0
    }
```

src/itx.rs
```diff
@@ -78,6 +82,63 @@ fn inv_txfm_add<BD: BitDepth>(
...
+            (Dct, 8) => rav1d_inv_dct8_1d_c,
+            (Dct, 16) => rav1d_inv_dct16_1d_c,
+            (Dct, 32) => rav1d_inv_dct32_1d_c,
+            (Dct, 64) => rav1d_inv_dct64_1d_c,
+            (Adst, 4) => rav1d_inv_adst4_1d_c,
+            (Adst, 8) => rav1d_inv_adst8_1d_c,
+            (Adst, 16) => rav1d_inv_adst16_1d_c,
+            (FlipAdst, 4) => rav1d_inv_flipadst4_1d_c,
+            (FlipAdst, 8) => rav1d_inv_flipadst8_1d_c,
+            (FlipAdst, 16) => rav1d_inv_flipadst16_1d_c,
```
kkysen, line 134: Can you remove the `rav1d_*` prefixes?  I think `dav1d` removed its `dav1d_*` prefixes for all of these.

src/scan.rs
```diff
@@ -228,3 +229,54 @@ pub static DAV1D_SCANS: [&'static [Scan]; TxfmSize::COUNT] = [
...
+    let mut max_col: u8 = 0;
+    const_for!(n in 0..S => {
+        let rc = scan[n].const_get();
+        let rcx = (rc & (h - 1) as u16) as u8;
+        max_col = if rcx > max_col { rcx } else {max_col };
+        last_nonzero_col_from_eob[n] = max_col;
+    });
+
+    last_nonzero_col_from_eob
+}
```
kkysen, line 245:
```suggestion
const fn init_tbl<const N: usize>(scan: &'static [Scan; N], h: u16) -> [u8; N] {
    let mut last_nonzero_col_from_eob = [0; N];

    let mut max_col = 0;
    const_for!(n in 0..N => {
        let rc = scan[n].const_get();
        let rcx = (rc & (h - 1)) as u8;
        max_col = const_max!(max_col, rcx);
        last_nonzero_col_from_eob[n] = max_col;
    });

    last_nonzero_col_from_eob
}
```

with

```rust
macro_rules! const_max {
    ($a:expr, $b:expr) => {{
        let a = $a;
        let b = $b;
        if a < b {
            a
        } else {
            b
        }
    }};
}

/// [`std::cmp::max`] is not `const` since it would need `const` `trait` `fn`s,
/// so this implements it with a macro instead so it can remain `const`.
pub(crate) use const_max;
```

src/scan.rs
```diff
@@ -228,3 +229,54 @@ pub static DAV1D_SCANS: [&'static [Scan]; TxfmSize::COUNT] = [
...
+    &LAST_NONZERO_COL_FROM_EOB_32X16,
+    &LAST_NONZERO_COL_FROM_EOB_32X32,
+    &LAST_NONZERO_COL_FROM_EOB_32X32,
+    &LAST_NONZERO_COL_FROM_EOB_4X16,
+    &LAST_NONZERO_COL_FROM_EOB_16X4,
+    &LAST_NONZERO_COL_FROM_EOB_8X32,
+    &LAST_NONZERO_COL_FROM_EOB_32X8,
+    &LAST_NONZERO_COL_FROM_EOB_16X32,
+    &LAST_NONZERO_COL_FROM_EOB_32X16,
+];
```
kkysen, line 282: It's better to use `enum_map!` here so that the order isn't implicit.

```suggestion
pub static DAV1D_LAST_NONZERO_COL_FROM_EOB: enum_map_ty!(TxfmSize, &'static [u8]) = enum_map!(TxfmSize => &'static [u8]; match key {
    S4x4 => LAST_NONZERO_COL_FROM_EOB_4X4.as_slice(),
    S8x8 => LAST_NONZERO_COL_FROM_EOB_8X8.as_slice(),
    S16x16 => LAST_NONZERO_COL_FROM_EOB_16X16.as_slice(),
    S32x32 => LAST_NONZERO_COL_FROM_EOB_32X32.as_slice(),
    S64x64 => LAST_NONZERO_COL_FROM_EOB_32X32.as_slice(),
    R4x8 => LAST_NONZERO_COL_FROM_EOB_4X8.as_slice(),
    R8x4 => LAST_NONZERO_COL_FROM_EOB_8X4.as_slice(),
    R8x16 => LAST_NONZERO_COL_FROM_EOB_8X16.as_slice(),
    R16x8 => LAST_NONZERO_COL_FROM_EOB_16X8.as_slice(),
    R16x32 => LAST_NONZERO_COL_FROM_EOB_16X32.as_slice(),
    R32x16 => LAST_NONZERO_COL_FROM_EOB_32X16.as_slice(),
    R32x64 => LAST_NONZERO_COL_FROM_EOB_32X32.as_slice(),
    R64x32 => LAST_NONZERO_COL_FROM_EOB_32X32.as_slice(),
    R4x16 => LAST_NONZERO_COL_FROM_EOB_4X16.as_slice(),
    R16x4 => LAST_NONZERO_COL_FROM_EOB_16X4.as_slice(),
    R8x32 => LAST_NONZERO_COL_FROM_EOB_8X32.as_slice(),
    R32x8 => LAST_NONZERO_COL_FROM_EOB_32X8.as_slice(),
    R16x64 => LAST_NONZERO_COL_FROM_EOB_16X32.as_slice(),
    R64x16 => LAST_NONZERO_COL_FROM_EOB_32X16.as_slice(),
});
```

This requires a couple extra small `impl`s, so I can add those if you want.

kkysen (changes requested):

PR #1439, opened 2025-07-06
Implement `dav1d-rs`'s Rust API

### LINKED ISSUE #1252: Provide Rust API

leo030303:
I've copy pasted the PR from #1362 and updated it with some of the suggestions made on that pull request.  The main changes are:
* Have all the public-facing Rust API code in its own file.
* Used some `enum`s from `rav1d` instead of redefining new ones, and added in doc comments from [the original `dav1d-rs` library](https://github.com/rust-av/dav1d-rs/blob/master/src/lib.rs) to a few items.
* Removed as much `unsafe` code as I could and replaced it with the Rust methods from `rav1d` as much as possible.

It currently works as a drop-in replacement for `dav1d-rs`; adding in `use rav1d as dav1d;` to my fork of image makes everything work fine.

The only functional changes I made are I removed the `unsafe impl`s of `Send` and `Sync` for `InnerPicture` so `Picture` is no longer `Sync` or `Send`. I looked through the code and I don't believe `DisjointMut<Rav1dPictureDataComponentInner>`, which is a field of one of its children, is thread safe, though I'm open to correction there; I'm pretty unfamiliar with `unsafe` Rust.

I also don't have safety comments on the two `unsafe` blocks in `rust_api.rs`; I'm unsure what these would look like, so open to suggestions there. These are mostly taken verbatim from the old pull request.

* Closes #1252.


src/rust_api.rs
```diff
@@ -0,0 +1,762 @@
...
+    any(target_arch = "riscv32", target_arch = "riscv64"),
+    feature(stdarch_riscv_feature_detection)
+)]
+#![deny(unsafe_op_in_unsafe_fn)]
+#![allow(clippy::all)]
+#![deny(clippy::undocumented_unsafe_blocks)]
+#![deny(clippy::missing_safety_doc)]
+
+#[cfg(not(any(feature = "bitdepth_8", feature = "bitdepth_16")))]
+compile_error!("No bitdepths enabled. Enable one or more of the following features: `bitdepth_8`, `bitdepth_16`");
```
kkysen, line 13: How come these things need to be repeated from `lib.rs`?

src/include/dav1d/headers.rs
```diff
@@ -378,12 +378,17 @@ impl From<Rav1dWarpedMotionParams> for Dav1dWarpedMotionParams {
...
+    /// Monochrome.
     #[default]
     I400 = 0,
+    /// 4:2:0 planar.
     I420 = 1,
+    /// 4:2:2 planar.
     I422 = 2,
+    /// 4:4:4 planar.
     I444 = 3,
 }
```
kkysen, line 393: These kinds of docs are generally very useful, so if you'd like to pull them out into a separate PR, I can review and merge that right away.

src/rust_api.rs
```diff
@@ -0,0 +1,762 @@
...
+            None
+        } else {
+            Some(NonNull::new(ptr).unwrap())
+        }
+    }
+
+    /// Error enum return by various `dav1d` operations.
+    #[derive(Debug, Copy, Clone, PartialEq, Eq)]
+    #[non_exhaustive]
+    pub enum Error {
```
kkysen, line 90: Does this need to exactly match `dav1d-rs`?  Otherwise, changes like this with more Rust-idiomatic names could be made directly to `Rav1dError` (though for all of the variants).

src/rust_api.rs
```diff
@@ -0,0 +1,762 @@
...
+                Error::UnknownError(err) => write!(fmt, "Unknown error {err:?}"),
+            }
+        }
+    }
+
+    impl std::error::Error for Error {}
+
+    /// Settings for creating a new [`Decoder`] instance.
+    /// See documentation for native `Dav1dSettings` struct.
+    pub struct Settings {
```
kkysen, line 150:
```suggestion
    #[derive(Default)]
    pub struct Settings {
```

src/rust_api.rs
```diff
@@ -0,0 +1,762 @@
...
+        }
+    }
+
+    impl Settings {
+        /// Creates a new [`Settings`] instance with default settings.
+        pub fn new() -> Self {
+            Self {
+                dav1d_settings: Rav1dSettings::default(),
+            }
+        }
```
kkysen, line 168:
```suggestion
    impl Settings {
        /// Creates a new [`Settings`] instance with default settings.
        pub fn new() -> Self {
            Self::default()
        }
```

src/rust_api.rs
```diff
@@ -0,0 +1,762 @@
...
+    }
+
+    impl std::error::Error for Error {}
+
+    /// Settings for creating a new [`Decoder`] instance.
+    /// See documentation for native `Dav1dSettings` struct.
+    pub struct Settings {
+        pub(crate) dav1d_settings: Rav1dSettings,
```
kkysen, line 151:
```suggestion
        pub(crate) rav1d_settings: Rav1dSettings,
```

src/rust_api.rs
```diff
@@ -0,0 +1,762 @@
...
+
+        pub fn set_apply_grain(&mut self, apply_grain: bool) {
+            self.dav1d_settings.apply_grain = apply_grain;
+        }
+
+        pub fn get_apply_grain(&self) -> bool {
+            self.dav1d_settings.apply_grain
+        }
+
+        pub fn set_operating_point(&mut self, operating_point: u8) {
```
kkysen, line 194: I think `operating_point` is a `u32` in `dav1d-rs`.

src/rust_api.rs
```diff
@@ -0,0 +1,762 @@
...
+            self.dav1d_settings.n_threads as u32
+        }
+
+        pub fn set_max_frame_delay(&mut self, max_frame_delay: u32) {
+            self.dav1d_settings.max_frame_delay = max_frame_delay as i32;
+        }
+
+        pub fn get_max_frame_delay(&self) -> u32 {
+            self.dav1d_settings.max_frame_delay as u32
+        }
```
kkysen, line 184: You can make `n_threads` and `max_frame_delay` `u32`s directly in `Rav1dSettings`.  Those are still left a `c_int`s and we haven't fixed their types yet.  The range is checked to be in `0..=256`, so `u32` should be fine.

kkysen (changes requested):
Thanks for the much safer implementation!  I'm still taking a closer look at the `unsafe`s, as the few remaining are still quite critical, but I wanted to give some initial feedback in general first.

src/rust_api.rs
leo030303: Overzealous copy pasting lol, pulled out now

src/rust_api.rs
leo030303: Makes sense, didn't want to change these initially so it'd be a more self contained PR, added the changes

src/rust_api.rs
leo030303: It is indeed, must've been a quirk copied over from the other PR

src/rust_api.rs
kkysen: Yeah, that makes sense.  Could you pull this change out into its own PR then?  I have some comments, but it'll be easier if I put them on its own PR.  Thanks!

src/rust_api.rs
leo030303: Updated

leo030303: New PR is here https://github.com/memorysafety/rav1d/pull/1443

src/rust_api.rs
kkysen: That's fine, too.  I just wasn't sure if it was supposed to exactly match `dav1d-rs` or not.

src/rust_api.rs
```diff
@@ -0,0 +1,679 @@
...
+    pub struct BitsPerComponent(pub usize);
+
+    impl Picture {
+        /// Stride in pixels of the `component` for the decoded frame.
+        pub fn stride(&self, component: PlanarImageComponent) -> u32 {
+            let s = match component {
+                PlanarImageComponent::Y => 0,
+                _ => 1,
+            };
+            self.inner.pic.stride[s] as u32
```
kkysen, line 435: Stride can also be negative.  So you might want to panic here or take the absolute value, or account for the negative strides somehow.

src/rust_api.rs
```diff
@@ -0,0 +1,679 @@
...
+        type Target = [u8];
+
+        fn deref(&self) -> &Self::Target {
+            self.as_ref()
+        }
+    }
+
+    /// Number of bits per component.
+    #[derive(Copy, Clone, Debug, PartialEq, Eq)]
+    pub struct BitsPerComponent(pub usize);
```
kkysen, line 426:
```suggestion
    pub struct BitsPerComponent(pub u8);
```

src/rust_api.rs
```diff
@@ -0,0 +1,679 @@
...
+        V,
+    }
+
+    impl From<usize> for PlanarImageComponent {
+        fn from(index: usize) -> Self {
+            match index {
+                0 => PlanarImageComponent::Y,
+                1 => PlanarImageComponent::U,
+                2 => PlanarImageComponent::V,
+                _ => panic!("Invalid YUV index: {}", index),
```
kkysen, line 383: This should probably be a `TryFrom` instead of panicking.

src/rust_api.rs
```diff
@@ -0,0 +1,679 @@
...
+
+    /// Frame component.
+    #[derive(Eq, PartialEq, Copy, Clone, Debug)]
+    pub enum PlanarImageComponent {
+        /// Y component.
+        Y,
+        /// U component.
+        U,
+        /// V component.
+        V,
```
kkysen, line 374: Specifying which are luminance and chrominance could also be useful.

src/rust_api.rs
```diff
@@ -0,0 +1,679 @@
...
+    ///
+    /// This can be used like a `&[u8]`.
+    #[derive(Clone)]
+    pub struct Plane(Picture, PlanarImageComponent);
+
+    impl AsRef<[u8]> for Plane {
+        fn as_ref(&self) -> &[u8] {
+            let (stride, height) = self.0.plane_data_geometry(self.1);
+            unsafe {
+                std::slice::from_raw_parts(
```
kkysen, line 408:
```suggestion
                slice::from_raw_parts(
```

src/rust_api.rs
```diff
@@ -0,0 +1,679 @@
...
+                    self.0.plane_data_ptr(self.1) as *const u8,
+                    (stride * height) as usize,
+                )
+            }
```
kkysen, line 412: This is unsound if the stride is negative.

kkysen (changes requested):

src/rust_api.rs
```diff
@@ -0,0 +1,679 @@
...
+    pub use crate::include::dav1d::dav1d::Rav1dInloopFilterType as InloopFilterType;
+    pub use crate::include::dav1d::headers::{
+        Rav1dContentLightLevel as ContentLightLevel, Rav1dMasteringDisplay as MasteringDisplay,
+        Rav1dPixelLayout as PixelLayout,
+    };
+    pub use crate::include::dav1d::picture::RAV1D_PICTURE_ALIGNMENT as PICTURE_ALIGNMENT;
+    use crate::internal::Rav1dContext;
+    use crate::pixels::Pixels;
+    use crate::{
+        self as rav1d, c_arc, c_box, dav1d_get_frame_delay, rav1d_close, rav1d_flush,
```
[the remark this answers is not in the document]
leo030303: Ya I removed that since it wasn't actually needed

src/rust_api.rs
```diff
@@ -0,0 +1,679 @@
...
+        fn from(x: Infallible) -> TryFromEnumError {
+            match x {}
+        }
+    }
+
+    impl std::error::Error for TryFromEnumError {}
+
+    /// A `dav1d` decoder instance.
+    pub struct Decoder {
+        dec: Arc<Rav1dContext>,
```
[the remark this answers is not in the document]
leo030303: Most of these are just leftovers from the old pull request so your guess is as good as mine for some of the names, `ctx` makes more sense

src/rust_api.rs
```diff
@@ -0,0 +1,679 @@
...
+        /// Flush the decoder.
+        ///
+        /// This flushes all delayed frames in the decoder and clears the internal decoder state.
+        ///
+        /// All currently pending frames are available afterwards via [`Decoder::get_picture`].
+        pub fn flush(&mut self) {
+            rav1d_flush(&self.dec);
+            if let Some(mut pending_data) = self.pending_data.take() {
+                let _ = mem::take(&mut pending_data);
+            }
```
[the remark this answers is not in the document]
leo030303: Ah actually no `mem::take` is necessary, pending_data here is a `Rav1dData` so I'm setting that to its default value, so I want `self.pending_data = Some(Rav1dData::default())` instead of `None`, since thats whats happening in the `dav1d_data_unref` calls this replaced

src/rust_api.rs
leo030303: I'm not sure what the stride value is so I'm open to guidance on what the best approach is here

src/rust_api.rs
leo030303: Both `stride` and `height` are u32 so should be fine
```rust
        pub fn plane_data_geometry(&self, component: PlanarImageComponent) -> (u32, u32) {
```

src/rust_api.rs
leo030303: For now I'll replace it with a `try_into().unwrap()`

src/rust_api.rs
kkysen: But the stride from that method may originally be negative. If you cast it to `u32` and then use it as part of the offset, it'll be incorrect, and that's UB.

kkysen: Yeah, this should be enough to avoid UB.  It's still not fully correct, but it'll work for positive strides and won't cause UB.

kkysen (commented):
@leo030303, could you rebase this on `main`?  There are a lot of other commits in here now, so it's harder to review.

leo030303:
Ah sorry yes thought I'd done that already, should be good to go now!

src/include/dav1d/headers.rs
```diff
@@ -565,6 +567,34 @@ impl Rav1dColorPrimaries {
...
+            Rav1dColorPrimaries::BT601 => pixel::ColorPrimaries::BT470BG,
+            Rav1dColorPrimaries::SMPTE240 => pixel::ColorPrimaries::ST240M,
+            Rav1dColorPrimaries::FILM => pixel::ColorPrimaries::Film,
+            Rav1dColorPrimaries::BT2020 => pixel::ColorPrimaries::BT2020,
+            Rav1dColorPrimaries::XYZ => pixel::ColorPrimaries::ST428,
+            Rav1dColorPrimaries::SMPTE431 => pixel::ColorPrimaries::P3DCI,
+            Rav1dColorPrimaries::SMPTE432 => pixel::ColorPrimaries::P3Display,
+            Rav1dColorPrimaries::EBU3213 => pixel::ColorPrimaries::Tech3213,
+            Rav1dColorPrimaries(x) => {
+                if (23..=DAV1D_COLOR_PRI_RESERVED).contains(&(x as u32)) {
```
kkysen, line 588: What's the meaning of 19 here?  Perhaps you could name it?

src/include/dav1d/headers.rs
```diff
@@ -646,6 +676,45 @@ impl Rav1dTransferCharacteristics {
...
+            Rav1dTransferCharacteristics::BT2020_12BIT => {
+                pixel::TransferCharacteristic::BT2020Twelve
+            }
+            Rav1dTransferCharacteristics::SMPTE2084 => {
+                pixel::TransferCharacteristic::PerceptualQuantizer
+            }
+            Rav1dTransferCharacteristics::SMPTE428 => pixel::TransferCharacteristic::ST428,
+            Rav1dTransferCharacteristics::HLG => pixel::TransferCharacteristic::HybridLogGamma,
+            Rav1dTransferCharacteristics(x) => {
+                if (19..=DAV1D_TRC_RESERVED).contains(&(x as u32)) {
```
kkysen, line 708: Same for the 19 here.

src/include/dav1d/headers.rs
```diff
@@ -709,6 +778,44 @@ impl Rav1dMatrixCoefficients {
...
+            Rav1dMatrixCoefficients::SMPTE2085 => pixel::MatrixCoefficients::ST2085,
+            Rav1dMatrixCoefficients::CHROMAT_NCL => {
+                pixel::MatrixCoefficients::ChromaticityDerivedNonConstantLuminance
+            }
+            Rav1dMatrixCoefficients::CHROMAT_CL => {
+                pixel::MatrixCoefficients::ChromaticityDerivedConstantLuminance
+            }
+            Rav1dMatrixCoefficients::ICTCP => pixel::MatrixCoefficients::ICtCp,
+            Rav1dMatrixCoefficients(x) => {
+                if (15..=DAV1D_MC_RESERVED).contains(&(x as u32)) {
```
kkysen, line 809: And the 15 here.

src/rust_api.rs
```diff
@@ -0,0 +1,563 @@
...
+        /// Get the decoder delay.
+        pub fn get_frame_delay(&self) -> u32 {
+            // The only fields this actually needs from Rav1dSettings are n_threads and max_frame_delay so we just pass these in directly
+
+            rav1d_get_frame_delay(&Rav1dSettings {
+                n_threads: self.n_threads,
+                max_frame_delay: self.max_frame_delay,
+                ..Default::default()
+            })
+            .unwrap() as u32
```
kkysen, line 313: You can change `rav1d_get_frame_delay` to return the `usize` directly since it's always `Ok`.

src/rust_api.rs
```diff
@@ -0,0 +1,563 @@
+/// Public Rust API.
+///
+/// This is more or less the same API as <https://crates.io/crates/dav1d>,
+/// and is indeed a fork of that work.
+pub mod dav1d {
```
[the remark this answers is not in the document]
leo030303: Ya its unneeded, will remove

src/include/dav1d/headers.rs
leo030303: `Rav1dColorPrimaries::EBU3213` is the highest value variant at 22, so its just to cover anything above that. It does look pretty arbitrary being defined there though, so I'll add an `is_unspecified` method to each of the three functions

kkysen: Let me think (might be a simpler way to make it a bit clearer).

kkysen: Actually there are holes, just explicitly named. I forget exactly why.

leo030303: Ya this makes sense, (this section was pulled verbatim from the old pull request) it does look like you can just return `Unspecified` for each of these

kkysen: So, according to the spec, `2` is `CP_UNSPECIFIED` (https://aomediacodec.github.io/av1-spec/av1-spec.pdf#page=129).  We should probably rename that to `UNSPECIFIED` instead of `UNKNOWN`, as the spec should take precedence here.  And it's the same for transfer characteristics and matrix coefficients.

There are reserved (not unspecified) values in dav1d's tests, although I think coercing them to `Unspecified` does work.  Can you do that for now?  I'll look into seeing if we can fix this more properly.

leo030303:
I think the core of this PR, the rust API, seems to be in a  good place, for the work aligning the implementation closer to the spec I think it makes sense to do those under separate PR's, but overall I think the rust API should be good to merge if it all looks fine from your end

src/include/dav1d/headers.rs
kkysen: Sorry I took a while.  #1453 is what I meant, if that looks good (can you review it?).

leo030303:
I think this was closed accidentally?

kkysen:
> I think this was closed accidentally? @kkysen

Oh, whoops, I didn't realize "fix" expressions for a PR comment close them.  I thought it only worked on issues.

Shnatsel:
Is there anything blocking this? I'd love to see this merged and published so we could drop the last remaining C library, `dav1d`, from the `image` crate.

leo030303:
Is there anything stopping this from being merged? Happy to make any requested changes @kkysen

src/rust_api.rs
```diff
@@ -0,0 +1,551 @@
...
+    }
+
+    /// Plane geometry of the `component` for the decoded frame.
+    ///
+    /// This returns the stride and height.
+    pub fn plane_data_geometry(&self, component: PlanarImageComponent) -> (u32, u32) {
+        let height = match component {
+            PlanarImageComponent::Y => self.height(),
+            _ => match self.pixel_layout() {
+                PixelLayout::I420 => (self.height() + 1) / 2,
```
197g:
```suggestion
                PixelLayout::I420 => self.height().div_ceil(2),
```

Handles the case where `height = u32::MAX` correctly, too.

src/rust_api.rs
```diff
@@ -0,0 +1,551 @@
...
+/// A single plane of a decoded frame.
+///
+/// This can be used like a `&[u8]`.
+#[derive(Clone)]
+pub struct Plane(Picture, PlanarImageComponent);
+
+impl AsRef<[u8]> for Plane {
+    fn as_ref(&self) -> &[u8] {
+        let (stride, height) = self.0.plane_data_geometry(self.1);
+        // SAFETY: both stride and height can't be negative, the `stride` and `height` methods panic if they are so there's no undefined behaviour
```
197g: The wording here misses the mark and ends in a non-sequitur. It should justify why the extent is correct for the pointer that's used but only justifies how the inputs are somehow bounded in `[0; u32::MAX)` (repeating the type explicitly would also help readability).

narodnik:
```rust
    pub fn send_data<T: AsRef<[u8]> + Send + Sync + 'static>(
        &mut self,
        buf: T,
        offset: Option<i64>,
        timestamp: Option<i64>,
        duration: Option<i64>,
    ) -> Result<(), Rav1dError> {
        assert!(
            self.pending_data.is_none(),
            "Have pending data that needs to be handled first"
        );

        let buf = buf.as_ref().to_vec().into_boxed_slice();
```

Why is buf marked as `'static`? It doesnt need it since you copy it immediately with `.to_vec()`. I have a non-static `&[u8]` and have to copy it with `.to_vec()` just to pass it in and then you immediately copy it again.

kkysen (commented):
> Is there anything blocking this? I'd love to see this merged and published so we could drop the last remaining C library, `dav1d`, from the `image` crate.

> Is there anything stopping this from being merged? Happy to make any requested changes @kkysen

Sorry, I haven't had time to work on `rav1d`.  I can finish reviewing this now.  I'll take another look over it again now.  Thanks for all of your patience.

src/lib.rs
```diff
@@ -139,6 +139,8 @@ mod pool;
 mod qm;
 mod recon;
 mod refmvs;
+/// This API copies that of `dav1d-rs`, and should function as a drop-in replacement, aside from `dav1d::Error` being renamed to `Rav1dError`
+pub mod rust_api;
```
kkysen: I'm not sure if this will show up as doc comments.  It may have to be `//!` doc comments in `rust_api.rs`.  If I'm wrong about how that works, though, it's fine here.

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+
+    pub fn set_apply_grain(&mut self, apply_grain: bool) {
+        self.inner.apply_grain = apply_grain;
+    }
+
+    pub fn get_apply_grain(&self) -> bool {
+        self.inner.apply_grain
+    }
+
+    pub fn set_operating_point(&mut self, operating_point: u8) {
```
kkysen: Do you want this to match `dav1d-rs`, which uses `u32` for `operating_point`: https://docs.rs/dav1d/latest/dav1d/struct.Settings.html#method.set_operating_point?

src/rust_api.rs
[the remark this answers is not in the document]
kkysen: So the `self.pending_data.take()` here replaces `self.pending_data = None` and then moves the `Rav1dData` to the local variable `pending_data`.  That local variable is dropped no matter what, so the `mem::take(&mut pending_data)` should be redundant.

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+    pub fn flush(&mut self) {
+        rav1d_flush(&self.ctx);
+        if let Some(mut pending_data) = self.pending_data.take() {
+            let _ = mem::take(&mut pending_data);
+        }
+    }
+
+    /// Send new AV1 data to the decoder.
+    ///
+    /// After this returned `Ok(())` or `Err([Error::Again])` there might be decoded frames
```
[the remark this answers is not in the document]
kkysen: Actually, what's `Error::Again`?

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+    pub fn send_data<T: AsRef<[u8]> + Send + Sync + 'static>(
+        &mut self,
+        buf: T,
+        offset: Option<i64>,
+        timestamp: Option<i64>,
+        duration: Option<i64>,
+    ) -> Result<(), Rav1dError> {
+        assert!(
+            self.pending_data.is_none(),
+            "Have pending data that needs to be handled first"
```
kkysen:
```suggestion
            "There's pending data that needs to be handled first."
```

src/rust_api.rs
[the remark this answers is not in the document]
kkysen: Is this supposed to be `Rav1dError::TryAgain`?

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+            if matches!(err, Rav1dError::TryAgain) {
+                self.pending_data = Some(data);
+            } else {
+                let _ = mem::take(&mut data);
+            }
+
+            return Err(err);
+        }
+
+        if data.data.as_ref().is_some_and(|d| d.len() > 0) {
```
kkysen: nit:

```suggestion
        if data.data.as_ref().is_some_and(|d| !d.is_empty()) {
```

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+            }
+            Some(data) => data,
+        };
+
+        if let Err(err) = rav1d_send_data(&self.ctx, &mut data) {
+            if matches!(err, Rav1dError::TryAgain) {
+                self.pending_data = Some(data);
+            } else {
+                let _ = mem::take(&mut data);
+            }
```
kkysen:
```suggestion
            }
```

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+
+            return Err(err);
+        }
+
+        if data.data.as_ref().is_some_and(|d| d.len() > 0) {
```
kkysen:
```suggestion
        if data.data.as_ref().is_some_and(|d| !d.is_empty()) {
```

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+            Ok(Picture {
+                inner: Arc::new(pic),
+            })
+        }
+    }
+
+    /// Get the decoder delay.
+    pub fn get_frame_delay(&self) -> u32 {
+        // The only fields this actually needs from Rav1dSettings are n_threads and max_frame_delay so we just pass these in directly
+
```
kkysen: nit:

```suggestion
        // The only fields this actually needs from `Rav1dSettings`
        // are `n_threads` and `max_frame_delay`, so we just pass these in directly.
```

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+            ..Default::default()
+        }) as u32
+    }
+}
+
+impl Drop for Decoder {
+    fn drop(&mut self) {
+        if let Some(mut pending_data) = self.pending_data.take() {
+            let _ = mem::take(&mut pending_data);
+        }
```
kkysen: Should be unnecessary, as `self.pending_data` will be dropped anyways.

```suggestion
```

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+pub struct Plane(Picture, PlanarImageComponent);
+
+impl AsRef<[u8]> for Plane {
+    fn as_ref(&self) -> &[u8] {
+        let (stride, height) = self.0.plane_data_geometry(self.1);
+        let data = self.0.plane_data_ptr(self.1) as *const u8;
+        if stride == 0 || data.is_null() {
+            return &[];
+        }
+        // SAFETY: Copied checks from david-rs added in this pull request https://github.com/rust-av/dav1d-rs/pull/121
```
kkysen: I think this needs its own `SAFETY` comment.  There's none in https://github.com/rust-av/dav1d-rs/pull/121.

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+pub struct BitsPerComponent(pub u8);
+
+impl TryFrom<u8> for BitsPerComponent {
+    type Error = Rav1dError;
+
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            0 => Ok(BitsPerComponent(8)),
+            1 => Ok(BitsPerComponent(10)),
+            2 => Ok(BitsPerComponent(12)),
```
kkysen: What's this for exactly?  It's a non-trivial conversion, so I also wouldn't put it under `TryFrom<u8>` vs. its own named function.

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+    }
+
+    /// Plane geometry of the `component` for the decoded frame.
+    ///
+    /// This returns the stride and height.
+    pub fn plane_data_geometry(&self, component: PlanarImageComponent) -> (u32, u32) {
+        let height = match component {
+            PlanarImageComponent::Y => self.height(),
+            _ => match self.pixel_layout() {
+                PixelLayout::I420 => (self.height() + 1) / 2,
```
kkysen:
```suggestion
                PixelLayout::I420 => self.height().div_ceil(2),
```

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+
+impl AsRef<[u8]> for Plane {
+    fn as_ref(&self) -> &[u8] {
+        let (stride, height) = self.0.plane_data_geometry(self.1);
+        let data = self.0.plane_data_ptr(self.1) as *const u8;
+        if stride == 0 || data.is_null() {
+            return &[];
+        }
+        // SAFETY: Copied checks from david-rs added in this pull request https://github.com/rust-av/dav1d-rs/pull/121
+        unsafe { slice::from_raw_parts(data, stride as usize * height as usize) }
```
kkysen: I would do a `.checked_mul` here for `stride * height`.  Since they're both `u32`s, this shouldn't overflow ever when `usize = u64`, but since that's not always the case, I think it's better to be safe, and it'll optimize out in the common case.

kkysen (changes requested):
Sorry again for such a long wait.  I left some comments, and could you also rebase to remove the merge commits?  It's also important that we keep each commit buildable/testable so that we can do benchmarking on each commit, so if any commits aren't, could you rebase them such that they are?  Thanks!!

tats-u:
@Shnatsel How about "Squash and merge" instead of "Create a merge commit" or "Rebase and merge"? It will automatically do it.

src/lib.rs
leo030303: Actually nevermind, cargo doc complained about the fix, the original was correct

src/rust_api.rs
leo030303: Actually renamed it to `from_hbd` since its for use with the high bit depth flag, the conversion copies the implementation from dav1d seen here https://github.com/videolan/dav1d/blob/de4ce4f32d97aee77790941972011e23a620e56b/include/dav1d/headers.h#L222

leo030303:
@kkysen Thanks for the review, resolved most of the comments, and happy to see all the outside interest in the PR!

src/rust_api.rs
```diff
@@ -0,0 +1,550 @@
...
+        assert!(
+            self.pending_data.is_none(),
+            "Have pending data that needs to be handled first"
+        );
+
+        let buf = buf.as_ref().to_vec().into_boxed_slice();
+        let slice = &*buf;
+        let len = slice.len();
+
+        let mut data = Rav1dData::create(len).unwrap();
```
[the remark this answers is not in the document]
kkysen: `Rav1dData::wrap` doesn't do an extra allocation and is what `dav1d-rs` is doing, too.

src/rust_api.rs
kkysen: > Actually renamed it to `from_hbd`

Yeah, that's a lot clearer.

Shnatsel (commented):
Here's a safety comment and a potential soundness fix.

Are there any other outstanding work items or is this good to go?

src/rust_api.rs
```diff
@@ -0,0 +1,495 @@
+// This whole module was originally copied from https://github.com/rust-av/dav1d-rs/
+// (specifically https://github.com/rust-av/dav1d-rs/blob/94b1deaa1e25bf29c77bb5cc8a08ddaf7663eede/src/lib.rs)
+// with modifications.
+// `dav1d-rs` is under the MIT license, replicated here:
+
```
kkysen:
```suggestion
//
```

Otherwise auto imports get put here.

Shnatsel:
> If we restrict the API to just Box<[u8]>, it's simple, but allowing AsRef<[u8]>, like Vec<u8> and &[u8], is more complex.

`Box<[u8]>` doesn't really gel with the APIs e.g. `image` exposes. It'll work but require an extra copy of each batch of the input plus an allocation for each batch; when using this with an `impl Read` this can introduce noticeable overhead.

I like approach 2 in the short term and approach 3 in the long term.

kkysen (commented):
> `Box<[u8]>` doesn't really gel with the APIs e.g. `image` exposes. It'll work but require an extra copy of each batch of the input plus an allocation for each batch; when using this with an `impl Read` this can introduce noticeable overhead.

What API is `image`, etc. exposing?  Being able to take that into account would help designing the right API here.

> I like approach 2 in the short term and approach 3 in the long term.

👍 

> Speaking of API improvements, I don't think `T: AsRef<[u8]>` is even a good idea. It just causes more monomophisation

Hmm, you're right about the monomorphization.  That would be quite annoying.

> doesn't even help ergonomics that much compared to a `&[u8]`.

It does let use other types like `[u8; N]`, `Box<[u8]>`, `Vec<u8>`, `Arc<[u8]>`, etc.  I'm not sure how useful any of those are from the caller's perspective, as I don't have much familiarity with that.

> But the real trick is getting rid of the `'static` requirement without copying or infecting much of the surrounding code with lifetimes.

The tricky part here is the `pending_data`.  If that's moved to another intermediary type wrapping `Rav1dContext`, we could limit the lifetime to just `CBox`, `CArc`, `Rav1dData`, the new wrapper type, and maybe a few others.  I also experimented with adding a `R: AsRef<T>`/`R: AsRef<[u8]>` bound on these types (without the `pending_data` separation), but without a `+ 'static` bound, and it seems to compile fine.

kkysen (commented):
@Shnatsel, if you change the `.to_vec()` (https://github.com/image-rs/image/blob/5d0418d0eff747c829d52738c092581312f775c9/src/codecs/avif/decoder.rs#L86) to `Box::<[u8]>::from(...)` and change the `dav1d-rs` API to just accept `Box<[u8]>` instead of the `impl AsRef<[u8]>`, we should be able to avoid an extra copy or two here.  We'll still need the one, the `Box::<[u8]>::from(...)`, but we won't be able to fix that one until we fix the `'static` lifetime, which is much more tricky.  #1470 sets things up to work well with `Box<[u8]>`.

src/rust_api.rs
```diff
@@ -0,0 +1,496 @@
...
+            self.pending_data.is_none(),
+            "Have pending data that needs to be handled first"
+        );
+
+        let buf = buf.as_ref().to_vec().into_boxed_slice();
+        let slice = &*buf;
+        let len = slice.len();
+
+        let mut data = Rav1dData::create(len).unwrap();
+        data.data = Some(CArc::wrap(CBox::from_box(buf)).unwrap());
```
kkysen:
```suggestion
    pub fn send_data(
        &mut self,
        buf: Box<[u8]>,
        offset: Option<i64>,
        timestamp: Option<i64>,
        duration: Option<i64>,
    ) -> Result<(), Rav1dError> {
        assert!(
            self.pending_data.is_none(),
            "Have pending data that needs to be handled first"
        );

        let mut data = Rav1dData::wrap(CRef::Box(buf))?;
```

With the new API, this should minimize copies to a single `&'a [u8]` to `Box<[u8]>` done by the caller.

src/rust_api.rs
```diff
@@ -0,0 +1,495 @@
...
+
+        //  Get a raw pointer to the plane data of the `component` for the decoded frame.
+        let index: usize = component.into();
+        let raw_plane_data_pointer =
+            self.inner.data.as_ref().unwrap().data[index].as_byte_mut_ptr() as *const u8;
+
+        if stride == 0 || raw_plane_data_pointer.is_null() {
+            return &[];
+        }
+        // SAFETY: Copied checks from dav1d-rs added in this pull request https://github.com/rust-av/dav1d-rs/pull/121
```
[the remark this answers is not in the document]
leo030303: Sweet, I'll commit it for the time being

kkysen (commented):
Oh, I added `Rc` to `CRef`, which isn't `Send` or `Sync`. I can remove that.

kkysen (commented):
> Since CRef isn't Send or Sync, if its trivial to add that back in then we should

I've added it back in #1472 now.  Once that merges you can rebase/merge.

leo030303:
Think that's most of the outstanding issues wrapped up!

leo030303:
@kkysen if fixing the API for that to never be null isn't straightforward then it could probably be left for a follow up pr since I think the core of this pr should be good to go

src/rust_api.rs
```diff
@@ -0,0 +1,506 @@
...
+        //    alive for the duration of the borrow.
+        // 6. No mutable aliases: The Picture is only returned after decoding is complete,
+        //    so the decoder no longer writes to this buffer. The public API only exposes
+        //    shared (&[u8]) access, and no &mut references to this data can exist once
+        //    the Picture is handed to the user.
+        //
+        // Past dav1d-rs PRs relevant to this line:
+        // https://github.com/rust-av/dav1d-rs/pull/121
+        // https://github.com/rust-av/dav1d-rs/pull/123
+        unsafe { slice::from_raw_parts(raw_plane_data_pointer, data_length) }
```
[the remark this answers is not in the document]
kkysen: Opened #1479 for that.

kkysen (approved):
I think it all LGTM now.  I added a couple of small changes, and I left one last (🤞) about needing `impl TryFrom<usize> for PlanarImageComponent`.  The safety all looks good now.  Thanks for working on this for so long!  Once #1479 and #1480 merge, we should be able to merge this finally!

kkysen (approved):
Yay, it's finally all done!! Thanks!

leo030303:
Thanks again for all the help reviewing!

emilk:
Thanks for this ❤️

Shnatsel:
Thank you!

For this to be usable in image-rs we'll also need a new release to crates.io, would you please publish one?

src/rust_api.rs
```diff
@@ -0,0 +1,467 @@
...
+    }
+
+    /// YUV color range.
+    pub fn color_range(&self) -> pixel::YUVRange {
+        self.inner.seq_hdr.as_ref().unwrap().color_range.into()
+    }
+
+    /// Sample position for subsampled chroma.
+    pub fn chroma_location(&self) -> pixel::ChromaLocation {
+        self.inner.seq_hdr.as_ref().unwrap().chr.try_into().unwrap()
```
Shnatsel, line 446: These `.unwrap()` calls can panic in production; the fuzzer in #1483 quickly unearthed inputs that can trigger this panic.

I believe these APIs have to be changed to return `Result`

Shnatsel: I've opened https://github.com/memorysafety/rav1d/pull/1484 to return `Option` here instead of panicking

kkysen:
> For this to be usable in image-rs we'll also need a new release to crates.io, would you please publish one?

Yup, sure!  Let's fix the fuzzing bugs you've found first, but then we definitely can.

Shnatsel:
Sounds good! I'll run the fuzzer some more and make an updated PoC that wires up rav1d to image-rs, and once that's all good you can actually ship it to crates.io.

rectalogic:
I noticed some incompatibilities with the dav1d API.

[Picture::plane()](https://docs.rs/dav1d/latest/dav1d/struct.Picture.html#method.plane) is named `Picture::plane_data()` in rav1d

Also in dav1d, Picture is [`Send`](https://docs.rs/dav1d/latest/dav1d/struct.Picture.html#impl-Send-for-Picture%3CA%3E) and `Sync` but is neither in rav1d.

Shnatsel:
I've opened https://github.com/memorysafety/rav1d/pull/1485 to undo the needless renaming.

PR #1442, opened 2025-07-07
`enum Rav1dError`: rename fields to more Rusty versions and add docs

leo030303:
Pulled out the `Rav1dError` changes from #1439 into a separate PR.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -32,7 +32,7 @@ pub enum Rav1dError {
     ///
     /// If this is returned by [`Decoder::get_picture`] then no decoded frames are pending
     /// currently and more data needs to be sent to the decoder.
-    Again = libc::EAGAIN as u8,
+    TryAgain = libc::EAGAIN as u8,
 
     /// Out of memory.
     ///
@@ -58,7 +58,7 @@ pub enum Rav1dError {
 impl Rav1dError {
     pub const fn as_str(&self) -> &'static str {
         match self {
-            Rav1dError::Again => "Try again",
+            Rav1dError::TryAgain => "Try again",
             Rav1dError::InvalidArgument => "Invalid argument",
             Rav1dError::OutOfMemory => "Not enough memory available",
             Rav1dError::UnsupportedBitstream => "Unsupported bitstream",
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,13 +2,13 @@ use std::ffi::{c_int, c_uint};
 
 use strum::FromRepr;
 
+/// Error enum return by various `rav1d` operations.
 #[derive(Clone, Copy, PartialEq, Eq, FromRepr, Debug)]
 #[repr(u8)]
 #[non_exhaustive]
 pub enum Rav1dError {
     /// This represents a generic `rav1d` error.
     /// It has nothing to do with the other `errno`-based ones
-    /// (and that's why it's not all caps like the other ones).
     ///
     /// Normally `EPERM = 1`, but `dav1d` never uses `EPERM`,
     /// but does use `-1`, as opposed to the normal `DAV1D_ERR(E*)`.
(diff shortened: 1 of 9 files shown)


src/error.rs
```diff
@@ -17,15 +17,59 @@ pub enum Rav1dError {
...
+    /// Frame size is larger than the limit
+    OutOfRange = libc::ERANGE as u8,
+    /// Unsupported bitstream.
+    ///
+    /// The provided bitstream is not supported by `rav1d`.
+    UnsupportedBitstream = libc::ENOPROTOOPT as u8,
 }
 
+impl std::fmt::Display for Rav1dError {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
```
kkysen, line 57: Could you import items like these directly (and `fmt::Result` since that clashes)?

src/error.rs
```diff
@@ -17,15 +17,59 @@ pub enum Rav1dError {
...
+        match self {
+            Rav1dError::Again => write!(fmt, "Try again"),
+            Rav1dError::InvalidArgument => write!(fmt, "Invalid argument"),
+            Rav1dError::NotEnoughMemory => write!(fmt, "Not enough memory available"),
+            Rav1dError::UnsupportedBitstream => write!(fmt, "Unsupported bitstream"),
+            Rav1dError::EGeneric => write!(fmt, "Generic error"),
+            Rav1dError::InvalidBuffer => write!(fmt, "Invalid buffer"),
+            Rav1dError::CantOpenFile => write!(fmt, "Can't open file: IO error"),
+            Rav1dError::OutOfRange => write!(fmt, "Out of range"),
```
kkysen, line 66: Could you pull these out into a `pub const fn as_str(&self) -> &'static str` method?

src/error.rs
```diff
@@ -17,15 +18,70 @@ pub enum Rav1dError {
...
-    EIO = libc::EIO as u8,
-    EAGAIN = libc::EAGAIN as u8,
-    ENOMEM = libc::ENOMEM as u8,
-    EINVAL = libc::EINVAL as u8,
-    ERANGE = libc::ERANGE as u8,
-    ENOPROTOOPT = libc::ENOPROTOOPT as u8,
+    /// Invalid buffer.
+    ///
+    /// No Sequence Header OBUs were found in the buffer.
+    InvalidBuffer = libc::ENOENT as u8,
```
kkysen, line 24:
```suggestion
    /// No entity.
    ///
    /// No Sequence Header OBUs were found in the frame.
    NoEntity = libc::ENOENT as u8,
```

I think `NoEntity` is more appropriate for `ENOENT` given that it's much closer to `ENOENT`s meaning (no entity/entry) and fits well with its usage here, no seq hdr OBU.

kkysen (changes requested):

src/error.rs
```diff
@@ -17,15 +18,70 @@ pub enum Rav1dError {
     /// which is more optimal since `0` is no error for [`Dav1dResult`].
     EGeneric = 1,
```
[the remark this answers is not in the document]
leo030303: I'd agree with `Other`

src/error.rs
leo030303: The only place its used is `rav1d_parse_sequence_header` and the C version of that function has the "no sequence header..." description for that error, so I'd agree something like  NoSequenceHeader makes sense

src/error.rs
```diff
@@ -17,15 +18,70 @@ pub enum Rav1dError {
...
+    /// before processing any further pending data.
+    ///
+    /// If this is returned by [`Decoder::get_picture`] then no decoded frames are pending
+    /// currently and more data needs to be sent to the decoder.
+    Again = libc::EAGAIN as u8,
+
+    /// Not enough memory.
+    ///
+    /// Not enough memory is currently available for performing this operation.
+    NotEnoughMemory = libc::ENOMEM as u8,
```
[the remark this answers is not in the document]
leo030303: Ya I like that better

src/error.rs
```diff
@@ -17,15 +18,70 @@ pub enum Rav1dError {
...
+    CantOpenFile = libc::EIO as u8,
+
+    /// Try again.
+    ///
+    /// If this is returned by [`Decoder::send_data`] or [`Decoder::send_pending_data`] then there
+    /// are decoded frames pending that first have to be retrieved via [`Decoder::get_picture`]
+    /// before processing any further pending data.
+    ///
+    /// If this is returned by [`Decoder::get_picture`] then no decoded frames are pending
+    /// currently and more data needs to be sent to the decoder.
```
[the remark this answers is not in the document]
leo030303: Sorry I'm not sure what you mean here?

src/error.rs
kkysen: Hmm, possibly, although dav1d may add another usage in the future and we do want to keep a one to one mapping. Not sure though.

src/error.rs
```diff
@@ -1,31 +1,82 @@
...
-    ENOMEM = libc::ENOMEM as u8,
-    EINVAL = libc::EINVAL as u8,
-    ERANGE = libc::ERANGE as u8,
-    ENOPROTOOPT = libc::ENOPROTOOPT as u8,
+    Other = 1,
+
+    /// No sequence header.
+    ///
+    /// No Sequence Header OBUs were found in the buffer.
+    NoSequenceHeader = libc::ENOENT as u8,
```
kkysen, line 25: I'd rather this go back to just `NoEntity`. Since this will now be part of the public API, if `dav1d` changes to return some other `ENOENT` error somewhere, we'll have to rename this and cause a breaking change.

We can keep the more specific docs, though, like you're doing for `OutOfRange` and the frame size.

src/error.rs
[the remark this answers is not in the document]
kkysen: `Decoder` doesn't exist in this PR yet, so these docs don't yet make sense. But they should be able to refer to equivalent C API functions (e.g. `dav1d_get_picture`), and then #1439 can also add the references to `Decoder`.

kkysen (changes requested):
Sorry I took a while to get back to finish reviewing this.  It's almost done now, I think.  And then one other thing, for suggested changes that you accept, could you give the commit a descriptive message instead of the default `Update file.rs` message?  Or if it's small, you can squash it into another commit as well.  Thanks!

src/error.rs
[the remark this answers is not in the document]
leo030303: Cool I get you, I changed it to reference the rav1d functions as the C dav1d functions don't actually use this error variant

src/thread_task.rs
```diff
@@ -999,10 +999,7 @@ pub fn rav1d_worker_task(task_thread: Arc<Rav1dTaskContextTaskThread>) {
                             abort_frame(
                                 c,
                                 fc,
-                                match res {
-                                    Err(err) => Err(err),
-                                    Ok(_) => Err(Rav1dError::InvalidArgument),
-                                },
+                                res.and_then(|_| Err(Rav1dError::InvalidArgument))
```
kkysen, line 1002: I think you need to rebase and format this.

src/error.rs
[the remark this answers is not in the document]
kkysen: Oh yeah, that's better.

src/thread_task.rs
leo030303: Fixed!

PR #1446, opened 2025-07-18
Fix warnings from newer toolchain versions

NexusXe:
This resolves a couple of new warnings that appear when building with newer nightly toolchains.

These warnings don't show up with the toolchain version specified in `rust-toolchain.toml`, but the fixes are fully compatible with it. These changes are purely cosmetic and result in an identical compiled binary.

## Fix mismatched_lifetime_syntaxes warning

The first change addresses the `mismatched_lifetime_syntaxes` lint that became enabled by default in nightly [1.89](https://github.com/rust-lang/rust/pull/138677) - this lint suggests making elided lifetimes in return types explicit to avoid ambiguity.

## Fix function pointer comparison warning

The second change silences a warning about deriving `PartialEq` and `Eq` for the function wrapper `Fn(FnPtr)`, as the address of a given function is not guaranteed to be unique across codegen units.

This was caused by `Fn(FnPtr)` deriving `PartialEq` and `Eq` in the wrap_fn_ptr! macro. Since these traits don't appear to be used anywhere in the project, removing them resolves the warning. If the Actions tests prove this change wrong I'll revert it.

### DIFF
--- a/src/cdf.rs
+++ b/src/cdf.rs
@@ -145,7 +145,7 @@ impl CdfThreadContext {
         }
     }
 
-    pub fn cdf_write(&self) -> RwLockWriteGuard<CdfContext> {
+    pub fn cdf_write(&self) -> RwLockWriteGuard<'_, CdfContext> {
         match self {
             Self::QCat(_) => panic!("Expected a Cdf"),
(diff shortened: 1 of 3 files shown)


djc (approved):

djc:
Thanks, LGTM!

PR #1447, opened 2025-07-22
Simplify and slightly optimise how tasks are stored and selected

daxtens:
I had hoped to be able to optimise multithreaded behaviour. Unfortunately, I found it almost impossible to make heads or tails of what was going on. This PR is an attempt to unpick a couple of the complicated parts, replacing them with simpler and more documented implementations. As a happy side-effect, it's also a bit faster.

There are a set of  'frame contexts' for frames that we are decoding, and each frame context holds a dynamic set of tasks required to decode the frame. Some of these tasks can be executed in parallel. The tasks have dependencies, so not every task being tracked for a frame context is necessarily executable at any given point in time: tasks, and indeed entire frame contexts, can be 'blocked'.

Previously, these tasks were stored in an array in order of insertion. To execute them in priority order, each task stored the index of the next highest priority task, and the structure holding the tasks kept the index of the highest 'live' task. Remove this linked-list implementation, and instead implement `PartialOrd` and `Ord` for `Rav1dTask`, and use Rust's `Vec`'s search and sort methods to keep the tasks in order. Fascinatingly, despite seeming like more work (especially in terms of copying tasks around), this seems to go faster. (Possibly because we called `index` a lot? `index` takes a read lock on the vector each time.)

As checking to see if a task is blocked is not free, we also keep track of completely blocked frame contexts and the most recent frame context to have had progress. Because a frame can depend on information from a previous frame, if a frame context becomes unblocked, that and all subsequent frame contexts should be checked for executable tasks too. This process was tracked by the `first`, `cur`, and `reset_cur_index` variables and mostly managed by the `reset_task_cur` and `reset_task_cur_async` functions. These functions were complicated and their documentation was not particularly easy to follow. Replace them with simpler ones. The original ones also had a number of attempted optimisations, not all of which seem to be worth carrying over.

Finally, we make various passing optimisations:
 - it's not necessary to repeatedly mark the same frame context as having new tasks when we're merging pending tasks
 - if we can avoid repeatedly taking a read-lock—even when we're using `try_read()`—so much the better.

I am in the process of doing a final set of performance measurements on my latest tweaks, and I will update this when I have final results. Previous testing at 10, 6 and 2 threads on my 6 core AMD Ryzen box with Windows 11 indicates improvements somewhere less than 1% for the 10 bit Chimera file. For the 8 bit file, the change was more in the noise, but at least not slower. A fair amount of variability is also present. I haven't tested single-threaded performance; I assume it will be unchanged.

I would mostly advocate for these changes in terms of their simplicity: it's nice that they're faster but even if they made no performance change at all I think the simplicity is worth the churn.

Other notes:
 - I'm still not 100% clear on how I should be formatting comments so my apologies if they need work.
 - I've structured this to be read commit-by-commit.
 - There is definitely more that could be done here: for example, there are definitely possible optimisations to the way that delayed film grain is handled. But I need to take a break for a while and I want to at least try to land something!

### DIFF
--- a/src/thread_task.rs
+++ b/src/thread_task.rs
@@ -90,8 +90,8 @@ impl Rav1dTasks {
     fn merge_pending_frame(&self, c: &Rav1dContext) -> bool {
         let merge = self.pending_tasks_merge.swap(false, Ordering::SeqCst);
         if merge {
-            let new_start = self.len();
             let mut tasks = self.tasks.try_write().unwrap();
+            let new_start = tasks.len();
 
             {
                 let mut pending_tasks = self.pending_tasks.lock();
@@ -133,11 +133,6 @@ impl Rav1dTasks {
         }
 
         merge
-    }
-
-    /// How many (non-pending) tasks do we have in the task list?
-    pub fn len(&self) -> usize {
-        self.tasks.try_read().unwrap().len()
     }
 
     /// Are there non-pending tasks in the task list?
@@ -582,7 +577,7 @@ pub fn rav1d_worker_task(task_thread: Arc<Rav1dTaskContextTaskThread>) {
             continue 'outer;
         }
 
-        let (fc, t_idx) = 'found: {
+        let (fc, t_idx, new_t) = 'found: {
             if c.fc.len() > 1 {
                 // run init tasks second
                 'init_tasks: for fc in
@@ -599,7 +594,7 @@ pub fn rav1d_worker_task(task_thread: Arc<Rav1dTaskContextTaskThread>) {
 
                     let t = tasks.index(t_idx);
                     if t.type_0 == TaskType::Init {
-                        break 'found (fc, t_idx);
+                        break 'found (fc, t_idx, None);
                     }
                     if t.type_0 == TaskType::InitCdf {
                         // XXX This can be a simple else, if adding tasks of both
@@ -619,7 +614,7 @@ pub fn rav1d_worker_task(task_thread: Arc<Rav1dTaskContextTaskThread>) {
                             fc.task_thread
                                 .error
                                 .fetch_or((p1 == TILE_ERROR) as c_int, Ordering::SeqCst);
-                            break 'found (fc, t_idx);
+                            break 'found (fc, t_idx, None);
                         }
                     }
                 }
@@ -630,8 +625,7 @@ pub fn rav1d_worker_task(task_thread: Arc<Rav1dTaskContextTaskThread>) {
                 let fc = &c.fc[(first + ttd.cur.get()) as usize % c.fc.len()];
                 let tasks = &fc.task_thread.tasks;
                 tasks.merge_pending_frame(c);
-                'tasks: for t_idx in 0..tasks.len() {
-                    let t = tasks.index(t_idx);
+                'tasks: for (t_idx, t) in tasks.tasks.try_read().unwrap().iter().enumerate() {
                     if t.type_0 == TaskType::InitCdf {
                         continue 'tasks;
                     }
@@ -649,7 +643,7 @@ pub fn rav1d_worker_task(task_thread: Arc<Rav1dTaskContextTaskThread>) {
                         // if not bottom sbrow of tile, this task will be re-added
                         // after it's finished
                         if check_tile(&f, &fc.task_thread, &t, (c.fc.len() > 1) as c_int) == 0 {
-                            break 'found (fc, t_idx);
+                            break 'found (fc, t_idx, None);
                         }
                     } else if t.recon_progress != 0 {
                         // We need to block here because we are seeing rare
@@ -693,16 +687,15 @@ pub fn rav1d_worker_task(task_thread: Arc<Rav1dTaskContextTaskThread>) {
                             if next_t.sby == start {
                                 f.frame_thread.next_tile_row[p as usize].set(ntr);
                             }
-                            drop(t);
-                            fc.task_thread.insert_task(c, next_t, 0);
+                            break 'found (fc, t_idx, Some(next_t));
                         }
-                        break 'found (fc, t_idx);
+                        break 'found (fc, t_idx, None);
                     } else if t.type_0 == TaskType::Cdef {
                         let p1_1 = fc.frame_thread_progress.copy_lpf.try_read().unwrap()
                             [(t.sby - 1 >> 5) as usize]
                             .load(Ordering::SeqCst);
                         if p1_1 as c_uint & (1 as c_uint) << (t.sby - 1 & 31) != 0 {
-                            break 'found (fc, t_idx);
+                            break 'found (fc, t_idx, None);
                         }
                     } else {
                         if t.deblock_progress == 0 {
@@ -713,7 +706,7 @@ pub fn rav1d_worker_task(task_thread: Arc<Rav1dTaskContextTaskThread>) {
                             fc.task_thread
                                 .error
                                 .fetch_or((p1_2 == TILE_ERROR) as c_int, Ordering::SeqCst);
-                            break 'found (fc, t_idx);
+                            break 'found (fc, t_idx, None);
                         }
                     }
                 }
@@ -731,6 +724,11 @@ pub fn rav1d_worker_task(task_thread: Arc<Rav1dTaskContextTaskThread>) {
         // found:
         // remove t from list
         let mut t = fc.task_thread.tasks.remove(t_idx);
+
+        // insert new task into list if one was given
+        if let Some(new_t) = new_t {
+            fc.task_thread.insert_task(c, new_t, 0);
+        }
 
         if t.type_0 > TaskType::InitCdf && fc.task_thread.tasks.is_empty() {
             ttd.cur.update(|cur| cur + 1);


src/internal.rs
```diff
@@ -421,28 +419,66 @@ impl Rav1dTask {
...
+}
 
-    pub fn next(&self) -> Rav1dTaskIndex {
-        self.next.load(Ordering::SeqCst)
-    }
-
-    pub fn set_next(&self, next: Rav1dTaskIndex) {
-        self.next.store(next, Ordering::SeqCst)
+impl PartialOrd for Rav1dTask {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
```
kkysen, line 637: Can you import `std::cmp`?  Or if `std::sync::atomic::Ordering` is no longer imported, just import `std::cmp::Ordering`?

src/internal.rs
```diff
@@ -421,28 +419,66 @@ impl Rav1dTask {
...
-            deblock_progress: self.deblock_progress,
-            deps_skip: self.deps_skip.clone(),
-            next: Default::default(),
+impl Ord for Rav1dTask {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // We want `sort` to put tasks in priority order.
+        // To do that, we return:
+        //  - `Less`    if `self` is of higher priority than `other`
+        //  - `Greater` if `self` is of lower priority than `other`
+        // This requires that there are no Init, InitCdf or film grain tasks.
```
kkysen, line 648: Could this be a doc comment on the function?

src/internal.rs
```diff
@@ -421,28 +419,66 @@ impl Rav1dTask {
...
-            recon_progress: self.recon_progress,
-            deblock_progress: self.deblock_progress,
-            deps_skip: self.deps_skip.clone(),
-            next: Default::default(),
+impl Ord for Rav1dTask {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // We want `sort` to put tasks in priority order.
+        // To do that, we return:
+        //  - `Less`    if `self` is of higher priority than `other`
+        //  - `Greater` if `self` is of lower priority than `other`
```
kkysen, line 647: What about `Equal`?  `a.cmp(&a)` should be `Equal`, right?

src/thread_task.rs
```diff
@@ -143,152 +139,49 @@ pub struct Rav1dTasks {
...
     pending_tasks_merge: AtomicBool,
 
-    pub head: Atomic<Rav1dTaskIndex>,
-    // Points to the task directly before the cur pointer in the queue.
-    // This cur pointer is theoretical here, we actually keep track of the
-    // "prev_t" variable. This is needed to not loose the tasks in
-    // [head;cur-1] when picking one for execution.
-    pub cur_prev: RelaxedAtomic<Rav1dTaskIndex>,
+    // Have we looked at any tasks in this set of tasks?
+    pub looked: AtomicBool,
```
kkysen: If all of the loads/stores are relaxed, can this be a `RelaxedAtomic<bool>`?

src/internal.rs
```diff
@@ -272,6 +272,43 @@ pub(crate) struct TaskThreadData {
     pub delayed_fg: RwLock<TaskThreadDataDelayedFg>,
 }
 
+impl TaskThreadData {
+    /// When we finish a frame we advance `first`, and we must also adjust `cur`,
+    /// because `cur` is an offset of `first`.
+    /// We may also need to adjust `reset_task_cur`.
+    ///
+    /// The task thread lock must be held and there must be more than 1 frame context.
```
kkysen, line 303: Can you link to the task thread lock field directly?

src/internal.rs
```diff
@@ -272,6 +272,43 @@ pub(crate) struct TaskThreadData {
...
+    pub(crate) fn advance_first(&self, n_fc: usize) {
+        // This read-modify-write doesn't need to be an atomic CAS:
+        // modifications are protected by the lock that we're holding.
```
kkysen, line 306: If we already have a lock for this, why is `SeqCst` needed?

kkysen: Especially because the `.load` and `.fetch_add(1` aren't atomic together.

kkysen: Don't need to fix this right now, but I'm wondering.

src/thread_task.rs
```diff
@@ -197,15 +197,19 @@ impl Rav1dTasks {
             let new_start = self.len();
             {
                 let mut pending_tasks = self.pending_tasks.lock();
+                if pending_tasks.len() == 0 {
+                    // A spurious merge.
```
kkysen, line 99:
```suggestion
                    // A spurious, empty merge.
```

src/thread_task.rs
```diff
@@ -197,15 +197,19 @@ impl Rav1dTasks {
...
+                    // We don't synchronise the two parts of `add_pending` so this is always possible.
```
kkysen, line 100: I'm not sure what this means, although I haven't looked at this code as much as you have lately.

src/thread_task.rs
```diff
@@ -197,15 +197,19 @@ impl Rav1dTasks {
...
                 tasks.extend(pending_tasks.drain(..));
             }
             {
                 // take a read lock because reset_task_cur takes a read lock also
                 let tasks = self.tasks.try_read().unwrap();
-                for t in tasks[new_start..].iter() {
-                    self.insert_tasks(c, t.frame_idx, 0);
-                }
+                let frame_idx = tasks[new_start].frame_idx;
+                self.insert_tasks(c, frame_idx, 0);
```
kkysen: What's the reason for why we can do this?

kkysen (commented):
Still reviewing ce64d3c and b4e4bb9, but here's some initial feedback.  It looks very nice so far.

src/internal.rs
daxtens: I'll have a look!

src/internal.rs
daxtens: 👍  will do

src/internal.rs
daxtens: It's a good question. At the moment it's a direct move of the existing code. I did have a commit which I didn't include which made it Relaxed, but I didn't include it because it starts to get quite complex to tease out what is and isn't an access which might require a more rigorous ordering.

I'll have another look at including that commit in the series; the other thing I wanted to compare was putting `cur`, `reset_task_cur` and `first` in a lock all of their own so that there was just one single acquire/release pair of atomic operations, and it would simplify a lot of the code.

src/thread_task.rs
daxtens: I'll include some more explanation.

src/thread_task.rs
kkysen: Ah, okay, nevermind then.

daxtens:
Here's some performance figures from `hyperfine`, Windows 11, AMD Ryzen something or other with 6 cores, 12 hyperthreads. I ran with 4 warmup rounds. I ran the tests multiple times, some had outliers, others were just a bit 'weird', so I've just supplied 1 figure that I feel is at least somewhat representative of the common case. As such, please do take these with a definite grain of salt.

| File | Threads | main | daxtens/tasks | Improvement |
|-|-|-|-|-|
| 8 bit | 1 thread | 149.6 s ± 0.169 s | 148.5 s ± 0.219 s | 0.7% |
| 8 bit | 2 threads | 85.711 s ± 0.572 s | 85.249 s ± 0.556 s | 0.5% |
| 8 bit | 6 threads | 32.325 s ± 0.101 s | 32.195 s ± 0.140 s | 0.4% |
| 8 bit | 10 threads | 28.225 s ± 0.080 s | 28.123 s ± 0.067 s | 0.4% |
| 10 bit | 1 thread | 196.0 s ± 0.330 s | 194.8 ± 0.224 s | 0.6% |
| 10 bit | 2 threads | 114.747 s ± 0.584 s | 114.160 s ± 0.516 s | 0.5% |
| 10 bit | 6 threads | 47.555 s ± 0.069 s | 47.123 s ± 0.104 s | 0.9% |
| 10 bit | 10 threads | 43.752 s ± 0.097 s | 43.648 s ± 0.163 s | 0.2% |


I found it very odd that the single-thread case also went faster, but I haven't actually looked into how the single-thread case works; if it's the same in terms of finding and running tasks then it does make sense that these improvements would help there too.

src/thread_task.rs
kkysen: Okay, that makes sense.

src/internal.rs
daxtens: We still use `std::sync::atomic::Ordering`; I'm going to import `std::sync::atomic` and `std::cmp` so that we always qualify `Ordering` but never need to fully qualify it with `std::`.

src/thread_task.rs
daxtens: See new ce1f0a42cfa51065194292dda8a7d85ba9d448d0

src/thread_task.rs
daxtens: See new ce1f0a42cfa51065194292dda8a7d85ba9d448d0

src/internal.rs
daxtens: Folded into the new https://github.com/memorysafety/rav1d/pull/1447/commits/3a02af0d171821917612e20705d80ee8bbad6623

src/internal.rs
daxtens: Folded into the new https://github.com/memorysafety/rav1d/pull/1447/commits/3a02af0d171821917612e20705d80ee8bbad6623.

daxtens:
Looks like your x86 runner needs an updated `perf`.

src/thread_task.rs
```diff
@@ -301,68 +90,60 @@ impl Rav1dTasks {
...
+                    // We now have an empty vector: A has drained both the original contents,
+                    // and the task added by thread B - but `pending_tasks_merge` is true,
+                    // because it was set by B.
+                    //
+                    // So, if `merge_pending_frame` is called before another pending task is added,
+                    // we will get this spurious merge.
+                    return false;
                 }
-                let start = tasks.len() as u32;
                 tasks.extend(pending_tasks.drain(..));
```
kkysen, line 118: `tasks.append(pending_tasks)` should be more efficient.

src/thread_task.rs
```diff
@@ -301,68 +90,60 @@ impl Rav1dTasks {
...
-                let mut tasks = self.tasks.try_write().unwrap();
-                if self.head.load(Ordering::Relaxed).is_none() {
-                    tasks.clear();
+            let mut tasks = self.tasks.try_write().unwrap();
+            let new_start = tasks.len();
+
+            {
+                let mut pending_tasks = self.pending_tasks.lock();
+                if pending_tasks.len() == 0 {
+                    // This is a 'spurious' merge: that is, `merge` is set but the vector is empty.
```
kkysen, line 99: Is the optimization from skipping the actual merge (`.extend` and `.drain`, or more efficiently, `.append), or from returning `false`?

src/thread_task.rs
```diff
@@ -301,68 +90,60 @@ impl Rav1dTasks {
...
-    pub fn raw_index(self) -> Option<u32> {
-        self.0.map(|i| i.get() - 1)
-    }
 
-    pub fn is_none(self) -> bool {
-        self.0.is_none()
-    }
+            // Unlike in the C code, we only do a single `insert_tasks`. There are two reasons:
+            //
+            //  - in the C code, the `insert_tasks` is what creates the sorted linked list.
```
kkysen, line 123: nit:

```suggestion
            //  - In the C code, the `insert_tasks` is what creates the sorted linked list.
```

src/thread_task.rs
```diff
@@ -301,68 +90,60 @@ impl Rav1dTasks {
...
+            //    In the Rust code, the sorting is done via sorting the `Vec` explictly.
+            //
+            //  - The key remaining function of is updating our position in the frame contexts,
+            //    by updating the `reset_task_cur` and `cur` variables. Here, as in C,
+            //    every task we're merging is for the same frame context, so the arguments are always the same,
+            //    and nothing is changed by trying to repeatedly do the same update.
+            let frame_idx = tasks[new_start].frame_idx;
+            self.insert_tasks(c, frame_idx, 0);
+            tasks.sort();
```
kkysen, line 132: I'm not sure, but is a stable sort needed here?  I would think it isn't, as `fn cmp` doesn't even ever return `Equal`.

src/internal.rs
```diff
@@ -421,28 +633,69 @@ impl Rav1dTask {
...
+    /// We want `sort` to put tasks in priority order.
+    /// To do that, we return:
+    ///  - `Less`    if `self` is of higher priority than `other`
+    ///  - `Greater` if `self` is of lower priority than `other`
+    ///
+    /// This is farily straightforwardly translated from `insert_tasks` in `task_thread.c`,
+    /// and as such inherits the limitations of that function. Specifically,
+    /// it requires that there are no Init, InitCdf or film grain tasks,
+    /// and that there are no duplicate tasks.
+    fn cmp(&self, other: &Self) -> cmp::Ordering {
```
kkysen, line 654: You can do this in a separate PR (if I'm right), but I think this could be greatly simplified to this:

```rust
    fn cmp(&self, other: &Self) -> cmp::Ordering {
        (self.type_0, self.sby, self.tile_idx).cmp(&(other.type_0, other.sby, other.tile_idx))
    }
```

Where `TaskType` is also changed to delete the manual values (they don't seem to be used) and `TileEntropy` is moved to be first.

I haven't thought this through super closely to make sure I'm not messing anything up, but it would greatly simplify this and should make it easier to understand.

Can also be written as

```rust
impl Rav1dTask {
    fn cmp_fields(&self) -> impl Ord {
        (self.type_0, self.sby, self.tile_idx)
    }
}

impl Ord for Rav1dTask {
    fn cmp(&self, other: &Self) -> cmp::Ordering {
        self.cmp_fields().cmp(&other.cmp_fields())
    }
}
```

src/thread_task.rs
```diff
@@ -30,265 +28,56 @@ use crate::internal::{
...
-            prev_t = t;
-            t = self.index(t).next();
+    fn push(&self, c: &Rav1dContext, task: Rav1dTask, cond_signal: c_int) {
+        let frame_idx = task.frame_idx;
+        {
+            let mut tasks = self.tasks.try_write().unwrap();
+            let idx = match tasks.binary_search(&task) {
+                Ok(_) => panic!("Tried to add an identical task!"),
+                Err(idx) => idx,
+            };
```
kkysen, line 62: The code didn't use to do this, right?  Maybe just make it a `debug_assertions`-only check?

src/internal.rs
kkysen: I guess we do do `t.type_0 > TaskType::InitCdf`, which this order would change.  But we can probably get around that and still have the much simpler `fn cmp`.

kkysen: Wait, if we're already assuming that there aren't any `Init` or `InitCdf`s, then `TileEntropy` is the first anyways.  So no ordering changes needed.

daxtens:
I'm back at work this week so it may be a little while before I am able to look at this more.

I also need to think a bit more about some of my analysis of the atomic ordering around `first` and `reset_task_cur`; I think the code may actually be safer than I thought with SeqCst; I think some of the stuff around `wrapping_sub` might actually not be needed unless we make some things Relaxed. I'll let you know what I manage to work out.

daxtens:
Huh, yeah, I think the code actually cannot have `first` and `reset_task_cur` fall out of sync. IOW, it's always the case that `reset_task_cur >= first` except for a brief moment when `mark_new_task_for_index` and `advance_first` are running concurrently. Because `reset_cur_index` is always run under the same lock as `advance_first`, it's not possible for it to observe any case where `reset_task_cur` is less than `first`.

If we make `first` `Relaxed`, it's suddenly very easy for that to happen on a weakly ordered architecture.

I'll update this PR to remove the `wrapping_sub` from the existing commits - if I do get a relaxed solution I'm happy with, I'll put it in a separate commit that adds the wrapping sub back and adds the requisite explanation. I currently think that given the way the old code worked, we would be OK with the slight loss of synchronisation that results from using Relaxed, but given how tricky this code is I'd want to think more carefully. Using two variables with seqcst like this is not a pattern I'm used to and so it's tricky for me to reason about.

kkysen (commented):
> Looks like your x86 runner needs an updated `perf`.

@thedataking, can you update the kernel headers `perf` needs?

See the error here: https://github.com/memorysafety/rav1d/actions/runs/16540023760/job/46779890408.

Issue #1417, opened 2025-05-27
[Linux] calls to assembly routines have excessive indirection


iximeow:
this is about a 1.5% difference with 32 threads, 0.5% difference with 2 threads, and no difference for single-threaded measurements. this might not be an issue on macOS targets, though i can't really check that. but this definitely applies for Linux targets, probably BSDs. maybe Windows too, but i doubt it? it's also probably more of a Rust issue but i'm not yet sure how to describe what exactly is happening. so, some notes here in case it inspires some tinkering :)

all call sites to functions provided by `librav1dasm.a` are indirected through the GOT rather than direct calls like to any other internal function. which is to say, they're `ff 15 <addr>` calls like `call *addr(%rip)` rather than `e8 <offset>` calls like `call $+<offset>`. [this gist](https://gist.github.com/iximeow/4e83c245f507d251d4e6feabbebd2bbe) is a standalone demonstration of the issue. here, though, i'm going to use the particularly hot loop in `decode_coefs` as a more relevant example. the Rust code is, roughly, [here](https://github.com/memorysafety/rav1d/blob/3150db5/src/recon.rs#L931-L943):
```rust
            let level = &mut levels[level_off..];
            ctx = get_lo_ctx(level, tx_class, &mut mag, lo_ctx_offsets, x, y, stride);
            if tx_class == TxClass::TwoD {
                y |= x;
            }
            tok = rav1d_msac_decode_symbol_adapt4(&mut ts_c.msac, &mut lo_cdf[ctx as usize], 3);
            if dbg {
                println!(
                    "Post-lo_tok[{}][{}][{}][{}={}={}]: r={}",
                    t_dim.ctx, chroma, ctx, i, rc_i, tok, ts_c.msac.rng,
                );
            }
            if tok == 3 {
```

("roughly" because the assembly that follows is from [a change i was measuring and ought to PR](https://github.com/iximeow/rav1d/blob/715e31cc16dd8c1dc9f62241c10eb63351d4d829/src/recon.rs#L947-L965), but for the region here it's about the same with more branches on `main`.)

`perf`-annotated assembly from build via `cargo +nightly build --release` transcoding `Chimera-AV1-8bit-1920x1080-6736kbps.ivf` looks something like this:

```
0.63 │        add    0xc0(%rsp),%rsi
0.64 │        shr    $0x7,%ecx
0.67 │        cmp    $0x201,%eax
0.63 │        movzbl %dl,%eax
[...]

PR #1448, opened 2025-07-23
experiment: use `#[rustc_std_internal_symbol]` to make `#[no_mangle]` tables have hidden visibility

kkysen:
Based on @ahomescu's investigation:
* https://github.com/memorysafety/rav1d/issues/1417#issuecomment-3100873696
* https://github.com/rust-lang/rfcs/pull/3834
* https://github.com/rust-lang/rust/issues/143422

I tried using `#[rustc_std_internal_symbol]` on `rav1d`'s `#[no_mangle]` symbols (some of the tables in `tables.rs`).  Per https://github.com/rust-lang/rust/issues/143422, `#[rustc_std_internal_symbol]` is unstable and not meant to be used for this, and in fact, no longer works on `#[no_mangle]` symbols (so I had to move our pinned nightly back 1 day), but this is just for experimenting.  This is able to successfully build as a `cdylib` with `cargo build`, not with `cargo build --release`, where I get `error: undefined symbol` for all of these symbols.  So I don't think this works even for the hidden visibility issue (note: also fixed in #1367 through small changes to the assembly).

As for the performance issues raised in #1417, I don't think this addresses them.  #1417 is about Rust calls to asm functions having unnecessary indirection, not about any symbols that Rust defines having any issues.  And it can be fixed by using `RUSTFLAGS="-Z plt=yes"`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ repository = "https://github.com/memorysafety/rav1d"
 exclude = ["*.c", "*.h", "*.in", "meson*", "tests/", "doc/", "package/"]
 
 [lib]
-crate-type = ["staticlib", "rlib"]
+crate-type = ["staticlib", "rlib", "cdylib"]
 
 [dependencies]
 assert_matches = "1.5.0"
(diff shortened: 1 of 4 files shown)


kkysen:
Closing since this doesn't actually work and is very unstable so we wouldn't use it anyways.

PR #1451, opened 2025-08-16
`fn dav1d_inv_wht4_1d_c`: backport trimming

kkysen:
Splitting out of the large #1438.


kkysen (commented):
LGTM, but since I opened the PR to split it out of #1438, I can't approve it myself. @fbossen, can you?

I also simplified how the `#[cfg]` was done, as there was an import error before and I saw there was a simpler way to write it.

fbossen (approved):

PR #1453, opened 2025-08-16
`struct Rav1d{ColorPrimaries,TransferCharacteristics,MatrixCoefficients}`: align with spec

kkysen:
This switches these 3 `enum`s to use `UNSPECIFIED`/`Unspecified` like the [spec](https://aomediacodec.github.io/av1-spec/av1-spec.pdf#page=129) does.  It also makes these proper Rust `enum`s instead of `enum`-like `u8` newtype wrappers, coercing unknown values to `Unspecified`, which the spec allows for.  This should fix https://github.com/memorysafety/rav1d/pull/1439#discussion_r2229448001.


leo030303 (approved):
Looks good to me

src/include/dav1d/headers.rs
```diff
@@ -544,25 +544,25 @@ pub const DAV1D_COLOR_PRI_EBU3213: Dav1dColorPrimaries = Rav1dColorPrimaries::EB
...
-    pub const FILM: Self = Self(8);
-    pub const BT2020: Self = Self(9);
-    pub const XYZ: Self = Self(10);
-    pub const SMPTE431: Self = Self(11);
-    pub const SMPTE432: Self = Self(12);
-    pub const EBU3213: Self = Self(22);
-
     const fn to_dav1d(self) -> Dav1dColorPrimaries {
-        self.0 as Dav1dColorPrimaries
+        self as _
```
djc, line 566: I think we can just get rid of this method now?

src/obu.rs
```diff
@@ -355,11 +355,14 @@ fn parse_seq_hdr(
     let trc;
     let mtrx;
     if color_description_present != 0 {
-        pri = Rav1dColorPrimaries(gb.get_bits(8) as u8);
+        pri = gb
```
djc: Nit: I'd spell this as

```rust
pri = match Rav1dColorPrimaries::try_from(gb.get_bits(8)) {
     Ok(pri) => pri,
     _ => Rav1dColorPrimaries::Unspecified,
};
```

src/include/dav1d/headers.rs
```diff
@@ -618,32 +618,32 @@ pub const DAV1D_TRC_HLG: Dav1dTransferCharacteristics =
...
-    pub const SRGB: Self = Self(13);
-    pub const BT2020_10BIT: Self = Self(14);
-    pub const BT2020_12BIT: Self = Self(15);
-    pub const SMPTE2084: Self = Self(16);
-    pub const SMPTE428: Self = Self(17);
-    pub const HLG: Self = Self(18);
-
     const fn to_dav1d(self) -> Dav1dTransferCharacteristics {
-        self.0 as Dav1dTransferCharacteristics
+        self as _
```
djc, line 648: As above.

src/obu.rs
```diff
@@ -359,11 +359,14 @@ fn parse_seq_hdr(
             .get_bits(8)
             .try_into()
             .unwrap_or(Rav1dColorPrimaries::Unspecified);
-        trc = Rav1dTransferCharacteristics(gb.get_bits(8) as u8);
+        trc = gb
```
djc: As above.

src/include/dav1d/headers.rs
```diff
@@ -662,52 +662,53 @@ impl TryFrom<Dav1dTransferCharacteristics> for Rav1dTransferCharacteristics {
...
-    pub const BT2020_NCL: Self = Self(9);
-    pub const BT2020_CL: Self = Self(10);
-    pub const SMPTE2085: Self = Self(11);
-    pub const CHROMAT_NCL: Self = Self(12);
-    pub const CHROMAT_CL: Self = Self(13);
-    pub const ICTCP: Self = Self(14);
-
     const fn to_dav1d(self) -> Dav1dMatrixCoefficients {
-        self.0 as Dav1dMatrixCoefficients
+        self as _
```
djc, line 714: As above.

src/obu.rs
```diff
@@ -363,11 +363,14 @@ fn parse_seq_hdr(
             .get_bits(8)
             .try_into()
             .unwrap_or(Rav1dTransferCharacteristics::Unspecified);
-        mtrx = Rav1dMatrixCoefficients(gb.get_bits(8) as u8)
+        mtrx = gb
```
djc: As above.

src/include/dav1d/headers.rs
```diff
@@ -544,9 +544,10 @@ pub const DAV1D_COLOR_PRI_EBU3213: Dav1dColorPrimaries = Rav1dColorPrimaries::EB
 // this symbol is defined by dav1d, but not part of the spec
 pub const DAV1D_COLOR_PRI_RESERVED: Dav1dColorPrimaries = 255;
 
-#[derive(Clone, Copy, PartialEq, Eq, FromRepr)]
+#[derive(Clone, Copy, PartialEq, Eq, FromRepr, Default)]
 pub enum Rav1dColorPrimaries {
     BT709 = 1,
+    #[default]
```
djc, line 550: I'm not such a fan of this -- `Unspecified` doesn't seem like an obvious `#[default]`.

kkysen: The way we're using it is as a default though.

djc: IMO the litmus test for a `Default` implementation is whether something is an obvious, logical default. I don't think that's the case here, and replacing the single fallback with `_or_default()` doesn't seem like a material enough improvement.

src/obu.rs
kkysen: Why though? That just seems more wordy and non left-to-right logic that's harder to follow.  `.unwrap_or` is pretty idiomatic.

src/include/dav1d/headers.rs
leo030303: I'd agree it is a logical default, the default for an `Option` is usually `None` and `Unspecified` is the equivalent of `None` here as far as I understand

djc (approved):

src/obu.rs
djc: I think it is easier to follow. IMO your version also relies on non left-to-right logic in that you have to look at the `unwrap_or()` clause to get the inferred type for `try_into`.

PR #1454, opened 2025-08-27
Misc fixes to build dav1d C sources cleanly

thedataking:


kkysen (approved):
LGTM.

> Would be nice to add something to CI to avoid this regressing again.

When I add the benchmarking script, I might do this.  I imagine it'd come pretty naturally in doing that.

PR #1455, opened 2025-08-29
`const DAV1D_MAX_{THREADS,FRAME_DELAY}`: add missing `const`s that weren't transpiled

kkysen:


.gitignore
```diff
@@ -20,3 +20,5 @@ target/
 
 .gdb_history
 .gdbinit
+
+.cache/clangd/index
```
djc, line 24: Should we just ignore all of `.cache`?

djc (approved):

.gitignore
kkysen: I'm not sure.  I just know `clangd` keeps making this file so I'm ignoring it.

PR #1468, opened 2026-02-06
`fn rav1d_get_frame_delay`: simplify

kkysen:
* Pulled from #1439.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -166,6 +166,7 @@ use crate::in_range::InRange;
 use crate::include::common::bitdepth::BitDepth16;
 #[cfg(feature = "bitdepth_8")]
 use crate::include::common::bitdepth::BitDepth8;
+use crate::include::common::intops::clip;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -259,31 +260,32 @@ pub unsafe extern "C" fn dav1d_default_settings(s: NonNull<Dav1dSettings>) {
...
+        s.n_threads.get() // TODO propagate `InRange`
     } else {
-        rav1d_num_logical_processors()
-            .get()
-            .clamp(1, RAV1D_MAX_THREADS)
+        clip(
+            rav1d_num_logical_processors().get(),
+            1,
+            const { RAV1D_MAX_THREADS as u16 },
+        )
```
randomPoison: What did this change achieve? It seems like it's doing the exact same thing as before, but now you've also got this `const` expr in there?

randomPoison (approved):

Issue #1344, opened 2024-09-09
Replace unnecessary `usize::try_from`s with `as` casts


kkysen:
Replace `usize::try_from`s that were there to avoid overflow for lengths with unchecked `as` casts,
as we now bounds check, and overflow here is fine.

PR #1473, opened 2026-03-01
perf: replace unnecessary `usize::try_from` with `as` casts

Nicolas0315:
Replace `usize::try_from(n).unwrap()` with `n as usize` in three locations where the checked conversion is unnecessary:

- **`fn read_pal_indices`** (decode.rs:643): Block dimensions (w4, h4, bw4, bh4) are always non-negative per the AV1 spec.

- **Segmentation map update** (decode.rs:3062-3063): The original comment about needing checked casts for `from_raw_parts_mut` is stale — the code now uses bounds-checked `DisjointMut` indexing, so overflow in the cast would cause a bounds-check panic rather than UB. The outdated comment is removed.

- **`fn cfl_ac_rust`** (ipred.rs:1310): w_pad and h_pad are non-negative and immediately asserted to be less than width/height.

This removes the overhead of the checked conversion (comparison + unwrap/panic path) on every call.

* Addresses #1344.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -640,7 +640,7 @@ fn read_pal_indices(
     bw4: c_int,
     bh4: c_int,
 ) {
-    let [w4, h4, bw4, bh4] = [w4, h4, bw4, bh4].map(|n| usize::try_from(n).unwrap());
+    let [w4, h4, bw4, bh4] = [w4, h4, bw4, bh4].map(|n| n as usize);
     let pli = pl as usize;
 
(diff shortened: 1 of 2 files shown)


kkysen (changes requested):
Could you leave inline comments saying that they're non-negative like you have in the PR description?

Also, we should probably update these variables/arguments/fields to be actually unsigned (e.g., `u32`), but not sure if we want to do that right here (feel free to do that if you'd like, though).

PR #1474, opened 2026-03-01
perf: document zero-initialized scratch buffers in Rust fallback functions

Nicolas0315:
## Summary

In the Rust fallback functions (those ending in `_rust`), scratch buffers like `let mut mid = [[0i16; MID_STRIDE]; 135]` are zero-initialized but don't need to be — the C version (dav1d's `mc_tmpl.c`) declares these as uninitialized stack variables.

The `MaybeUninit` pattern was already applied successfully in `cdef.rs` (PR #1397) where buffers are passed to ASM calls. However, in `mc.rs`, `looprestoration.rs`, and `itx.rs`, the Rust fallback buffers are used with safe Rust indexing, making `MaybeUninit` more invasive for limited benefit since these functions are only used when ASM is unavailable.

## What this PR does

- Adds `TODO(perf)` comments documenting each buffer that matches an uninitialized C counterpart
- No buffers were found inside loops that could be hoisted outside to avoid repeated initialization
- All buffers are standalone allocations at function or branch scope level

## Files annotated

- **`src/mc.rs`**: 10 buffers across `put_8tap_rust`, `prep_8tap_rust`, `put_8tap_scaled_rust`, `prep_8tap_scaled_rust`, `put_bilin_rust`, `prep_bilin_rust`, `put_bilin_scaled_rust`, `prep_bilin_scaled_rust`, `warp_affine_8x8_rust`, `warp_affine_8x8t_rust`
- **`src/looprestoration.rs`**: 5 function-level annotations covering `wiener_rust`, `selfguided_filter`, `sgr_5x5_rust`, `sgr_3x3_rust`, `sgr_mix_rust`
- **`src/itx.rs`**: 2 buffers in `inv_txfm_add` and `inv_txfm_add_wht_wht_4x4_rust`

## Why not apply MaybeUninit here?

Unlike `cdef.rs` where the buffers are passed directly to ASM calls (already `unsafe`), these buffers are used with safe Rust indexing. Using `MaybeUninit` would require either:
- Wrapping all indexing in `unsafe` blocks with `assume_init` calls
- Or creating a safe wrapper abstraction

Both approaches add complexity for functions that are only exercised when platform-specific ASM implementations are unavailable. The comments serve as documentation for future optimization if profiling shows these Rust fallbacks are performance-critical.

## Related PRs

- #1397: Applied `MaybeUninit` pattern in `cdef.rs` for ASM-passed buffers
- #1399: Related scratch buffer optimization

### DIFF
--- a/src/itx.rs
+++ b/src/itx.rs
@@ -95,6 +95,10 @@ fn inv_txfm_add<BD: BitDepth>(
     let row_clip_max = !row_clip_min;
     let col_clip_max = !col_clip_min;
 
+    // TODO(perf): This buffer is zero-initialized but only needs to be written
+    // before read, matching the uninitialized C counterpart in dav1d's itx_tmpl.c.
+    // Could use MaybeUninit to avoid zeroing, but this is a Rust fallback function
+    // (only used when ASM is unavailable). See PR #1397 and #1399.
     let mut tmp = [0; 64 * 64];
     let mut c = &mut tmp[..];
     for y in 0..sh {
@@ -304,6 +308,10 @@ fn inv_txfm_add_wht_wht_4x4_rust<BD: BitDepth>(
 
     let coeff = &mut coeff[..W * H];
 
+    // TODO(perf): This buffer is zero-initialized but only needs to be written
+    // before read, matching the uninitialized C counterpart in dav1d's itx_tmpl.c.
+    // Could use MaybeUninit to avoid zeroing, but this is a Rust fallback function
+    // (only used when ASM is unavailable). See PR #1397 and #1399.
     let mut tmp = [0; W * H];
     let mut c = &mut tmp[..];
     for y in 0..H {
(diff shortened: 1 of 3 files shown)


src/itx.rs
```diff
@@ -95,6 +95,10 @@ fn inv_txfm_add<BD: BitDepth>(
     let row_clip_max = !row_clip_min;
     let col_clip_max = !col_clip_min;
 
+    // TODO(perf): This buffer is zero-initialized but only needs to be written
+    // before read, matching the uninitialized C counterpart in dav1d's itx_tmpl.c.
+    // Could use MaybeUninit to avoid zeroing, but this is a Rust fallback function
+    // (only used when ASM is unavailable). See PR #1397 and #1399.
```
kkysen:
```suggestion
    // before read, matching the uninitialized C counterpart in dav1d's `itx_tmpl.c`.
    // Could use `MaybeUninit` to avoid zeroing, but this is a Rust fallback function
    // (only used when asm is unavailable). See PR #1397 and #1399.
```

Could you fix these for the others, too?

kkysen (changes requested):
I'm not sure how important fixing these are considering they're in the Rust fallback functions, where safety is probably more important than performance.  But leaving comments is good.  Just had a couple tweaks for them.

PR #1475, opened 2026-03-14
Fuzzing via cargo-fuzz

Shnatsel:
Stacked on top of https://github.com/memorysafety/rav1d/pull/1439 which provides the necessary safe Rust API


Shnatsel:
Now that the Rust API is merged, I'll rebase this on top of main in a new PR.

PR #1479, opened 2026-03-18
`fn DisjointImmutGuard::get_unchecked`: add `unsafe fn` to bypass `DisjointMut`

kkysen:
For use in https://github.com/memorysafety/rav1d/pull/1439#discussion_r2945310882.

### DIFF
--- a/src/disjoint_mut.rs
+++ b/src/disjoint_mut.rs
@@ -208,6 +208,18 @@ impl<'a, T: ?Sized + AsMutPtr, V: ?Sized> Deref for DisjointImmutGuard<'a, T, V>
     }
 }
 
+impl<'a, T: ?Sized + AsMutPtr, V: ?Sized> DisjointImmutGuard<'a, T, V> {
+    // # Safety
+    //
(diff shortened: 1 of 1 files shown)


src/disjoint_mut.rs
```diff
@@ -208,6 +208,15 @@ impl<'a, T: ?Sized + AsMutPtr, V: ?Sized> Deref for DisjointImmutGuard<'a, T, V>
     }
 }
 
+impl<'a, T: ?Sized + AsMutPtr, V: ?Sized> DisjointImmutGuard<'a, T, V> {
+    // # Safety
+    //
+    // This is no longer checked at runtime to be legitimately disjoint mut.
+    pub unsafe fn unchecked_disjoint_mut(&self) -> &'a V {
```
randomPoison: Why's this called `unchecked_disjoint_mut` when it returns an immutable reference? Wouldn't something like `slice_unchecked` make more sense? Am I missing something about how this is meant to be used?

kkysen: Done now.

randomPoison (approved):

Issue #1476, opened 2026-03-14
Panic on malformed input in getbits.rs: range end index 18446744073709551615 out of range for slice of length 3


Shnatsel:
Found by cargo-fuzz using the fuzzing harness provided in https://github.com/memorysafety/rav1d/pull/1475

The minimal input triggering this issue is `[47, 47, 5, 4, 128, 0, 0, 0]`, here it is as a file: [minimized-from-1634880c334c62483cbc61860eb94e36b3739664.zip](https://github.com/user-attachments/files/25999914/minimized-from-1634880c334c62483cbc61860eb94e36b3739664.zip)

PR #1483, opened 2026-04-04
Fuzzing via cargo-fuzz, take 2

Shnatsel:
1. Add a fuzzing harness using `cargo fuzz`
2. Expose a Rust API to disable logging, which slows down fuzzing significantly

This is essentially #1475 but rebased on top of main. It has already discovered an issue: #1476


Shnatsel:
No issues other than #1484 discovered after 75 million iterations

fuzz/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+
+[dependencies.rav1d]
+path = ".."
+default-features = false # Disables ASM; comment out to enable
```
thedataking, line 15: Does fuzzing not work with asm enabled? or is this to get coverage on the non-assembly versions?

In practice, assembly is always enabled as speed is otherwise unacceptable and I imagine it would speed up fuzzing too.

If this has to stay off, could you document the reason?

fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+    Ok(())
+}
+
+fn decode(data: Box<[u8]>) -> std::io::Result<()> {
+    let mut settings = rav1d::rust_api::Settings::new();
+    settings.set_logger_enabled(false);
+    settings.set_frame_size_limit(256 * 256); // in pixels, 256x256 image max
+    let mut dec = rav1d::rust_api::Decoder::with_settings(&settings).expect("failed to create decoder instance");
```
thedataking, line 36: Should we bound the worker count? I think you get one worker per logical CPU.
My worry is that this could cause a slowdown if someone runs multiple parallel fuzz jobs.

fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,66 @@
...
+                }
+            }
+            Err(e) => {
+                return Err(std::io::Error::other(format!("Error sending pending data to the decoder: {}", e)));
+            }
+            _ => (),
+        }
+
+        // Handle all pending pictures before sending the next data.
+        handle_pending_pictures(&mut dec, false)?;
```
thedataking, line 63: Could we add an end-of-input drain after this call? `send_data()` accepting the buffer doesn’t mean decoding is finished, and a single `get_picture()` can return `TryAgain` while threaded decoding is still pending, or return only the first of several pictures.

This could let the fuzzer cover more of the decoder logic.

thedataking (commented):
This looks like a good PR for initial fuzzing support. Thanks!

I think this can land once a few minor items have been considered.

nit: could we rename `fuzz/fuzz_targets/fuzz_target_1.rs` to something like `fuzz/fuzz_targets/decoder.rs`?

PR #1484, opened 2026-04-04
Don't force panics in the Rust API

Shnatsel:
`chroma_location()` panic has been actually triggered by #1483, the other two changes are preemptive.

### DIFF
--- a/src/rust_api.rs
+++ b/src/rust_api.rs
@@ -416,12 +416,12 @@ impl Picture {
 
     /// Chromaticity coordinates of the source colour primaries.
     pub fn color_primaries(&self) -> pixel::ColorPrimaries {
-        self.inner.seq_hdr.as_ref().unwrap().pri.try_into().unwrap()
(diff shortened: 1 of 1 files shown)


src/rust_api.rs
```diff
@@ -372,13 +372,13 @@ impl Picture {
     }
 
     /// Width of the frame.
-    pub fn width(&self) -> u32 {
-        self.inner.p.w.try_into().unwrap()
+    pub fn width(&self) -> i32 {
```
kkysen: I thought these were always unsigned, even if the type used isn't yet.

Shnatsel: I've reverted the u32->i32 change. Thanks!

Now there is only one functional change in this PR and I am certain it is needed.

kkysen (approved):
LGTM!  Thanks!  I also rebased to delete the commit that was reverted instead of keeping the revert commit for it.

Shnatsel:
Thanks! I'll resume running the fuzzer once this is merged.

PR #1487, opened 2026-05-06
Support stable Rust on 32-bit ARM

laurens-pilot:
## Summary

This removes the nightly-only 32-bit ARM `stdarch_arm_feature_detection` gate by avoiding runtime ARM feature detection on `target_arch = "arm"`.

On 32-bit ARM, `std::arch::is_arm_feature_detected!` still requires nightly. This means crates depending on `rav1d` cannot build for 32-bit ARM targets with stable Rust.

Instead, this patch uses the stable `#[cfg(target_feature = "neon")]` compile-time fallback for 32-bit ARM and leaves the existing runtime detection paths for other architectures unchanged.

## Details

- Removes `#![cfg_attr(target_arch = "arm", feature(stdarch_arm_feature_detection))]`
- Replaces 32-bit ARM runtime detection for `neon` with `#[cfg(target_feature ="neon")]`
- Does not change the `aarch64` runtime detection path

The tradeoff is that on 32-bit ARM this only enables NEON when it is known at compile time. That seems preferable to requiring nightly for all stable 32-bit ARM consumers.

## Testing

Tested locally with stable Rust:

```sh
cargo +stable check --lib --no-default-features --features bitdepth_8,bitdepth_16
cargo +stable check --lib --target armv7-linux-androideabi --no-default-features
--features bitdepth_8,bitdepth_16

### DIFF
--- a/scripts/pure-rust-smoke-test.sh
+++ b/scripts/pure-rust-smoke-test.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+usage() {
+    cat <<'EOF'
+Usage: scripts/pure-rust-smoke-test.sh [--release] [--target TARGET] [--skip-check]
+
+Builds the rav1d CLI with assembly disabled and verifies a small, representative
+set of AV1 bitstreams against the upstream test-data MD5 checksums.
+
+Options:
+  --release       Build and run the release binary.
+  --target TARGET Run cargo check for TARGET with the same pure-Rust features.
+                  The decode smoke vectors are run on the host binary.
+  --skip-check    Skip the cargo check step.
+EOF
+}
+
+profile=debug
+cargo_profile_args=()
+target=
+skip_check=0
+
+while (($#)); do
+    case "$1" in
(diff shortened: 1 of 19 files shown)


thedataking (changes requested):
Thanks for your PR!

> The tradeoff is that on 32-bit ARM this only enables NEON when it is known at compile time. That seems preferable to requiring nightly for all stable 32-bit ARM consumers.

I disagree.  It does not seem preferable to do runtime feature detection on some targets (arm32) and compile-time detection on others.  Consistency between the different targets and consistency with dav1d makes troubleshooting, maintenance, and performance analysis simpler; we should keep it that way!

At a more fundamental level, I think we need feature detection to happen at runtime. Rav1d is intended to be compiled on one machine and distributed to many others. Choosing features at compile time would either exclude arm32 targets that do not have neon or forego neon acceleration on targets that do have the feature. Neither option is acceptable.

If there is no way to support runtime feature-detection on stable Rust, I'm afraid this PR must block on stabilization or we need to look for an acceptable workaround.

PR #1498, opened 2026-08-14
Impl `Debug` for `SendSyncNonNull` regardless of pointee

Jules-Bertholet:
This avoids relying on the `Debug` impl for `c_void`, which is dubious; see https://github.com/rust-lang/rust/pull/160115 for details.

### DIFF
--- a/src/send_sync_non_null.rs
+++ b/src/send_sync_non_null.rs
@@ -1,3 +1,4 @@
+use std::fmt;
 use std::ptr::NonNull;
 
 /// A [`NonNull`] that is [`Send`]` + `[`Sync`].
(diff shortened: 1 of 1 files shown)


thedataking:
Try rebasing this, CI should be green again now, in part thanks to your fix in #1499.

thedataking:
This change should wait on your upstream PR https://github.com/rust-lang/rust/pull/160115

Jules-Bertholet:
No, it's the opposite; the Rust project doesn't like to break the ecosystem, so that PR (if it goes in at all) needs to wait on this one.

thedataking:
> so that PR (if it goes in at all) needs to wait on this one.

Ack. In that case, let's leave this PR open until you have higher confidence the Rust PR is headed for approval.

I don't love that you're implementing `fmt::Debug` without calling out why you're not doing the obvious thing via a comment. Did you consider whether we can just get rid of the `[derive(Debug)]` attribute and leave a comment explaining why? What would be less code for us to understand and maintain.

Jules-Bertholet:
> Did you consider whether we can just get rid of the `[derive(Debug)]` attribute

Would mean removing `derive(Debug)` for `c_box::Free` also. If you are OK with that, I can do it

PR #1499, opened 2026-08-14
Appease Clippy

Jules-Bertholet:
Make CI green again


thedataking:
~~Could you be more specific about what this PR is fixing and why it needs fixing?~~

~~It looks like CI failed due to a problem with `nasm`, unrelated to your change.~~

I pulled your commit into #1500 which solves an unrelated problem. Both changes seem to depend on each other.

thedataking:
Thanks for your contribution; will land as part of #1500.

PR #1502, opened 2026-09-11
ipred: fix oversized z1 top_out scratch buffer

ssablin-meta:
Fix a transcribed `*` in TOP_OUT_SIZE, reducing the zero-initialized stack allocation from 10,192 to 302 elements. This reduces instructions retired by 1.275% on the all-intra benchmark (1 thread, median of 5).


kkysen (approved):
Good catch!

ssablin-meta:
do I need to do anything else before anyone can merge this? @kkysen

